PR #154, opened 2014-06-22
Speedup for TangengBinormalGenerator

abies:
Parallel version of generate can speed up generation according to number of cores in best case.
processTriangle now uses temp vars instead of allocation, which gives around 30% speedup in Sponza testcase I was running. Entire methods seems modified, but this is because of indent change - actual modification is just few lines at the start and tmp.release at the end.


empirephoenix:
Any news on this? What is missing? does it require a different testcase to further evaluate the usefullness? Should TangentGenerator speed even be important? It shouldnt be done in a deployed application anyway or?

abies:
Won't people want to use it possibly for some of dynamically generated geometry? Plus, if speedup is observed in general case, not only things I was checking, it might be a reasonable speedup even for SDK.
I'll try to provide proper testcase by end of the week - if I miss the deadline, feel free to close it.

Nehon:
there is no dead line. IMO this is an improvement, even if TBN is usually used at design time, it an't hurt to have it go faster.
I tried to make a test case back then but couldn't notice any improvement. The thing is, now That I think of it, is that I used the stock sponza.j3o we have, and that's one big geometry. I guess that the parallisation is really relevant if there are more than one geom.

normen:
I do like the idea behind this, its a perfect example for parallelization really.

empirephoenix:
So if there is no problem with it, merge?

Nehon:
Well...still want a test case :p

abies:
Working on that right now

abies:
This is very 'artificial' test, but I think it shows the speedup clearly. 


import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;

import com.jme3.scene.Node;
import com.jme3.scene.shape.Sphere;
import com.jme3.util.TangentBinormalGenerator;

public class TangentTest {

    static ExecutorService executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());

        for (int i = 0; i < 10; i++) {
            {
                Node root = new Node("Root");
                for (int count = 0; count < 10; count++) {
                    for (int samples = 4; samples < 50; samples++) {
                        Geometry g = new Geometry();
                        g.setMesh(new Sphere(samples, samples, 1.0f));
                        root.attachChild(g);
                    }
                }

                long start = System.currentTimeMillis();
                TangentBinormalGenerator.generate(root);
                System.out.println("Serial " + (System.currentTimeMillis() - start));
            }

            {
                Node root = new Node("Root");
                for (int count = 0; count < 10; count++) {
                    for (int samples = 4; samples < 50; samples++) {
                        Geometry g = new Geometry();
                        g.setMesh(new Sphere(samples, samples, 1.0f));
                        root.attachChild(g);
                    }
                }

                long start = System.currentTimeMillis();
                TangentBinormalGenerator.generateParallel(root, executor);
                System.out.println("Parallel " + (System.currentTimeMillis() - start));
            }

        }

    }

}

pspeed42:
Note: as far as this: "Won't people want to use it possibly for some of
dynamically generated geometry?"

...no, I think maybe not.  If one is dynamically generating geometry then
they are already in the best place to make tangents as they go for free.

I agree that a speed-up is nice, though.

Nehon:
4 times faster with 4 cores. 
gonna merge

PR #160, opened 2014-07-28
Utility methods for getting arbitrary properties for triangles from mesh

abies:


empirephoenix:
This sounds pretty dead to me? close?

empirephoenix:
Yep it's dead, killing it.

PR #170, opened 2014-08-21
added a pre-order traversal of the scene graph.

shamanDevel:
Adds a new method called depthLastTraversal() to Spatial and subclasses as an extension to depthFirstTraversal() and breathFirstTraversal().

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -1493,15 +1493,22 @@ public abstract class Spatial implements Savable, Cloneable, Collidable, Cloneab
 
     /**
      * Visit each scene graph element ordered by DFS, also called post order traversal.
+     * Delegates to {@link #depthFirstTraversal(com.jme3.scene.SceneGraphVisitor, boolean) }
+     * with {@code preOrder} set to {@code false}.
      * @param visitor
      */
-    public abstract void depthFirstTraversal(SceneGraphVisitor visitor);
+    public void depthFirstTraversal(SceneGraphVisitor visitor) {
+        depthFirstTraversal(visitor, false);
+    }
     
     /**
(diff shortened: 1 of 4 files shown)


empirephoenix:
So any answer here? This is kinda long stalled as well, but look sfine for me, merge?

empirephoenix:
@shadowislord  @pspeed42 Any rejections? Else I will merge this next time I review the open pull requests.

Shredder121:
Is it maybe a good idea to **add** a method rather than replacing it?
Backward compatible patches are more likely to be applied quickly.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -1492,10 +1492,23 @@ public Matrix4f getLocalToWorldMatrix(Matrix4f store) {
...
     /**
-     * Visit each scene graph element ordered by DFS
+     * Visit each scene graph element ordered by DFS, also called post order traversal.
+     * Delegates to {@link #depthFirstTraversal(com.jme3.scene.SceneGraphVisitor, boolean) }
+     * with {@code preOrder} set to {@code false}.
      * @param visitor
      */
-    public abstract void depthFirstTraversal(SceneGraphVisitor visitor);
+    public void depthFirstTraversal(SceneGraphVisitor visitor) {
+        depthFirstTraversal(visitor, false);
```
Shredder121: I didnt see this one, my comment can be ignored.

empirephoenix:
Closing due to merge conflict, however I would like to see this again in a form that just adds methods, as that can be easily applies without braking anything. I personally find different traversal methods usefull.

shamanDevel:
Oops, I totally forgot that this was still hanging around.
I'll start with a clean pull from the head of the jme3 branch again. Then there should be no merge conflicts and I'll include all the suggestions. You'll see a new pull request soon.

PR #196, opened 2014-12-05
Collide with optimization

Bebul:
`Node.collideWith` always traverses the whole scene through `child.collideWith(...)` recursion. 

This optimization tries to avoid recursion when possible using boundingVolume test first.


Bebul:
Are there any additional issues making this Push Request unacceptable?
  Any response is welcome.
  Thank you.

shadowislord:
@Bebul 
Why are you not using TempVars.collisionResults but instead creating your own public singleton class?

Bebul:
@shadowislord 
> Why are you not using TempVars.collisionResults but instead creating your own public singleton class?

The `TempVars.collisionResults` can be used, of course. Using `DummyCollisionResults` avoids collecting of colliding geometries at all, so it is slightly more effective. It may also help reader to understand the collisionResults are of no interest in this particular case.

But we can use `TempVars.collisionResults`, no problem. Just say.

shadowislord:
@Bebul 
> Using DummyCollisionResults avoids collecting of colliding geometries at all, so it is slightly more effective.

Yes but the CollisionResult object still has to be allocated if there is a valid collision, the collection step is negligible in this case (but either way we are talking peanuts). The only solution there is to do what @pspeed42 said and having an optimized intersection check that does not allocate anything. 
So I still stand by my suggestion to use TempVars.collisionResults as that's exactly what it was intended for. If you wish to take the more advanced route @pspeed42 suggested that works too.

PR #211, opened 2015-01-30
Optimize RenderShadow to use scene hierarchy for culling

Bebul:
Shadow rendering was using the flattened list of all scene geometries to check culling. This optimization traverses the scene hierarchy recursively from `rootNode` and tries to cull the whole `Node` whenever possible.
When the scene is constructed carefully to take advantage of hierarchical culling the optimization is significant.

The optimization tries to maintain the jME `public interface`. Some shadow related `RenderManager` or `RenderQueue` members were marked `@Deprecated`, mainly the flattened lists of shadow occluders and receivers, which are no longer used.

Two deprecated renderers, `BasicShadowRenderer` and `PssmShadowRenderer` were updated too, as these would not work with flattened occluders and receivers missing.

We are not aware of any other dependencies to be addressed.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
@@ -658,8 +658,6 @@ public class RenderManager {
     public void renderScene(Spatial scene, ViewPort vp) {
         //reset of the camera plane state for proper culling (must be 0 for the first note of the scene to be rendered)
         vp.getCamera().setPlaneState(0);
-        //remember the scene for possible later use
-        vp.getQueue().setRootScene(scene);
         //rendering the scene
         renderSubScene(scene, vp);
     }
--- a/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
@@ -591,6 +591,7 @@ public class RenderManager {
      * is still rendered in the shadow frustum (shadow casting queue)
      * through this recursive method.
      */
+    @Deprecated
     private void renderShadow(Spatial s, RenderQueue rq) {
         if (s instanceof Node) {
             Node n = (Node) s;
@@ -681,6 +682,8 @@ public class RenderManager {
     public void renderScene(Spatial scene, ViewPort vp) {
         //reset of the camera plane state for proper culling (must be 0 for the first note of the scene to be rendered)
         vp.getCamera().setPlaneState(0);
+        //remember the scene for possible later use
+        vp.getQueue().setRootScene(scene);
         //rendering the scene
         renderSubScene(scene, vp);
     }
@@ -690,10 +693,6 @@ public class RenderManager {
 
         // check culling first.
         if (!scene.checkCulling(vp.getCamera())) {
-            // move on to shadow-only render
-            if ((scene.getShadowMode() != RenderQueue.ShadowMode.Off || scene instanceof Node) && scene.getCullHint() != Spatial.CullHint.Always) {
-                renderShadow(scene, vp.getQueue());
-            }
             return;
         }
 
@@ -717,12 +716,6 @@ public class RenderManager {
             }
 
             vp.getQueue().addToQueue(gm, scene.getQueueBucket());
-
-            // add to shadow queue if needed
-            RenderQueue.ShadowMode shadowMode = scene.getShadowMode();
-            if (shadowMode != RenderQueue.ShadowMode.Off) {
-                vp.getQueue().addToShadowQueue(gm, shadowMode);
-            }
         }
     }
 
--- a/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
@@ -587,30 +587,6 @@ public class RenderManager {
     }
 
     /**
-     * If a spatial is not inside the eye frustum, it
-     * is still rendered in the shadow frustum (shadow casting queue)
-     * through this recursive method.
-     */
-    @Deprecated
-    private void renderShadow(Spatial s, RenderQueue rq) {
-        if (s instanceof Node) {
-            Node n = (Node) s;
-            List<Spatial> children = n.getChildren();
-            for (int i = 0; i < children.size(); i++) {
-                renderShadow(children.get(i), rq);
-            }
-        } else if (s instanceof Geometry) {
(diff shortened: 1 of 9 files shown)


Nehon:
Thanks.
it looks good, I'll look into it this WE and keep you updated

Nehon:
Ok I looked into it, and all of our shadow tests runs fine. That's already nice :p
Could be nice to also have the exact test case where it's faster than before.
From my understanding it's will now be a lot faster to cull chunks scene if it's properly organized in sub graphs, am I right?

About the code itself : IMO all methods and attributes you marked as deprecated can be safely removed. they are very internal stuffs, and even the public methods should not be used by users, unless they did their own renderer. But the deprecation won't really help out these people much neither, so I'm all for making some clean up.

One thing I dislike though it the ShadowUtil.rootScene = rq.getRootScene(). Not sure why you did that. I suspect it was to keep the ShadowUtil API unchanged.
This is bad because you're making a static Util class stateful. And even if it's handled properly here, it might not in user code, because they don't know how things work under the hood.
IMO the rootScene should be passed as a parameter in ShadowUtil.updateShadowCamera() and not be a static class attribute..
Also if the occluders parameter is not needed anymore in this method, just remove it.
So short story : don't be afraid to change the ShadowUtil api. 3.1 will have a lot of breaking changes, this one is very internal and you go unnoticed.

Nice work.
I'd be grateful you update the PR taking my remarks into consideration.

Bebul:
Thanks for fast and detailed response.

> From my understanding it’s will now be a lot faster to cull chunks scene if it’s properly organized in sub graphs, am I right?

 Yes, the scene must be constructed in subgraphs based on position. We have used trivial simple scene organized as quad tree.

>One thing I dislike though it the `ShadowUtil.rootScene = rq.getRootScene()`. Not sure why you did that. I suspect it was to keep the ShadowUtil API unchanged.

You are right again, I was afraid of changing the ` ShadowUtil.updateShadowCamera()` signature.

> So short story : don't be afraid to change the ShadowUtil api.

OK, nice to hear it. Will do. Thank you.

Bebul:
The PR updated to take Nehon's remarks into consideration.
 * multiple relict render shadow functionality removed
 * ShadowUtil.rootScene removedpassed as a parameter in `ShadowUtil.updateShadowCamera()` instead

shadowislord:
I see two issues:
1) This breaks the shadow processors (but not shadow filters). This is because you removed the part that adds geometries to the shadow receiver bucket. The shadow processors only render items in the shadow receiver bucket but with this change it will always be empty.
2) If there's more than one scene attached to the viewport, it will only render the casters from the last attached scene. This is because of the RenderQueue.setRootNode() which is not necessary at all, because viewPort.getScenes() should be used instead.

Bebul:
Considering the issues mentioned:
1) the shadow processors - you mean renderers, like Directional+Spot+Point+Pssm+BasicLightShadowRenderer? All of these were tested and we believe they work. Moreover, shadow filters use their related renderer as member variable, so the optimization affects them too, also tested. We implemented the occluders recursive collection process first, but then we have realized the same holds for receivers too, as there is no need to collect them all first and select only these inside camera frustum later. The reason why there is still `RenderQueue.shadowRecv` left is to possibly reuse the collected receivers, as having the list of receivers in the camera frustum can become handy. So it is not necessarily empty! It is also used it in our next to come push request, which builds over this one, so we must wait whether this one is accepted or not.
2) the optimization is inside `RenderManager.renderSubScene` which is called from `RenderManager.renderScene` where the ` RenderQueue.setRootScene()` is set. But the renderScene is called for all scenes attached to viewport:
```
        List<Spatial> scenes = vp.getScenes();
        for (int i = scenes.size() - 1; i >= 0; i--) {           
            renderScene(scenes.get(i), vp);
        }
```

Bebul:
Are there any additional issues making this Push Request unacceptable?
Any response is welcome.
Thank you.

shadowislord:
@Bebul 
> the same holds for receivers too, as there is no need to collect them all first and select only these inside camera frustum later

All receivers are already guaranteed to the in the camera frustum since they are rendered from the eye camera. There's no need to render them again in the shadow renderer. The patch should only fix the issues with the shadow casters. The shadow receivers already work fine and should be left alone.

> the renderScene is called for all scenes attached to viewport

Yes it is, but RenderQueue.getRootNode() is used at the very end after all scenes have been rendered, in AbstractShadowRenderer.postQueue(), so for multiple scenes, only the last scene will have proper shadow casters set.

Bebul:
> All receivers are already guaranteed to the in the camera frustum since they are rendered from the eye camera.

  Yes, it was guaranteed, but such huge list of receivers need shadows rendered only for the `DirectionalLight`.
  Different and thinner receivers list can be used for other lights.
  `PointLightShadowRenderer.getReceivers` and `SpotLightShadowRenderer.getReceivers` use `ShadowUtil.getLitGeometriesInViewPort` now, which is the intersection of two camera frustums.
  When cones of lights intersects only part of eye view frustum, it is more effective.

> There's no need to render them again in the shadow renderer.

  Are they rendered again? I believe they are not as it would make shadows twice darker!
  To prove it I have forced to render receivers twice for `SpotLightShadowRenderer` and it was immediately darker. The code was:
  ```
    GeometryList getReceivers(GeometryList sceneReceivers, GeometryList lightReceivers) {
        ...
        ShadowUtil.getLitGeometriesInViewPort(viewPort.getQueue().getRootScene(), viewPort.getCamera(), cameras, RenderQueue.ShadowMode.Receive, lightReceivers);
        if (proveIt) ShadowUtil.getLitGeometriesInViewPort(viewPort.getQueue().getRootScene(), viewPort.getCamera(), cameras, RenderQueue.ShadowMode.Receive, lightReceivers);
        return lightReceivers;
    }  
  ```
  In our tests we were able to switch between old and new implementation, so we were sure the shadows are the same, not darker or so.

> ... so for multiple scenes, only the last scene will have proper shadow casters set.
  
You are right at this point. We will fix it.

Bebul:
@shadowislord 
> If there's more than one scene attached to the viewport, it will only render the casters from the last attached scene. This is because of the RenderQueue.setRootNode() which is not necessary at all, because viewPort.getScenes() should be used instead.

The fix ed369135fa471495b9d6190e31a9d1094cc628d7 handles the issue reported. `setRootScene()` is not used anymore. Multiple scenes attached to viewport were tested, both renderers and filters.

shadowislord:
Looks good to me.
@Nehon please merge if you think its fine.

PR #215, opened 2015-02-25
Allow stats to be displayed without counting stats themselves

OndrejSpanel:
Currently the stats shown on pressing F5 include even the primitives used to render the stats themselves. While this is not incorrect, sometimes it can be confusing.

F5 now cycles between three states: off, full, post not included.

In the last mode the stats shown are captured when the stats object starts rendering, which means the stats do not include the stats object itself.

Depending on what are other viewports present in postFrame it may also skip some other primitives.

If you think changing the F5 for all applications is too drastic, I may adjust it so that it is done only when the application opts for it explicitely.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/StatsAppState.java
+++ b/jme3-core/src/main/java/com/jme3/app/StatsAppState.java
@@ -57,6 +57,7 @@ public class StatsAppState extends AbstractAppState {
     protected boolean showSettings = true;
     private boolean showFps = true;
     private boolean showStats = true;
+    private boolean includePostStats = true;
     private boolean darkenBehind = true;
     
     protected Node guiNode;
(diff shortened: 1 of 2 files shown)


empirephoenix:
Is this only about the primitives or also about shaders ect?

An alternative solutions with less downsides would be to calculate the primitives under the stats Node and just substract them in the second mode. This would not affect any other Viewport.

Nehon:
@empirephoenix the problem with your approach is that you have to count the number of primitives  of the stat view on each frame as it may have changed (for example is fps goes from 99 to 100 there are 2 more vertices)

As for the original PR, I'm gonna test what it implies for other viewports.

empirephoenix:
Any news on this one?

Nehon:
Nope, sorry. I'll check it out

empirephoenix:
@Nehon meow? ^^

Nehon:
Just been bumped up at the top of the stack ;)

Nehon:
ok...so the major downside here is that for example this enturely skips stats from the shadow post pass (that is done in the postFrame). We may need to find a better solution to actually not account for the stat view geometries in the stat.
Maybe a more in depth rework of the stat gathering.

shadowislord:
If the StatsAppState is responsible for rendering the StatsView, it can just set some flag on the Statistics class that says all calls should be ignored while it is rendering the StatsView.

pspeed42:
And when exactly does it do that since the stats are just Geometry rendered like everything else?

Nehon:
maybe the stat view should be rendered in the render mthod of the statAppstate, without being added to the gui node.

pspeed42:
I guess it just depends on how far we really want to go to remove 0.1% of the vertexes from the typical stats display.

pspeed42:
Probably easier to just do the math and subtract it.

PR #219, opened 2015-02-28
A very small change that creates two getters in VehicleWheel,,

empirephoenix:
These are necessary if jme's bullet wrapper is used in a non Scenegraph enviroment,
for example a ES based server. As without this there is no clean way to get the values of these two variables.
-> wheelspat being null in this case, normally the rotation and position is set to the Wheel Spatial.

In the case of a Scenegraphless Server it makes no sense to have a dummy spatial just for this, especially if all logic necessary already exist, and only the getters are missing.

Signed-off-by: Kai Börnert <kai-boernert@visiongamestudios.de>

### DIFF
--- a/jme3-bullet/src/main/java/com/jme3/bullet/objects/VehicleWheel.java
+++ b/jme3-bullet/src/main/java/com/jme3/bullet/objects/VehicleWheel.java
@@ -420,4 +420,22 @@ public class VehicleWheel implements Savable {
         this.applyLocal = applyLocal;
     }
 
+    /**
+    * write the content of the wheelWorldRotation into the store
+    * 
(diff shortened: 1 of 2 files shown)


shadowislord:
This change should be made in JBullet as well, I guess.

empirephoenix:
Added jbullet as well.
Dont ask me why the build failed this time, according to it its due to some sdk stuff, wich I didn't even touch.

empirephoenix:
Yep just checked out the master and tried to build it, current master is failing!

empirephoenix:
Ok, works with current master now, and builds fine :)
Can we merge this?

PR #220, opened 2015-02-28
Geometry shader and Tessellation shader support

zzuegg:
At current state everything beside the ShaderNode test works. I added two simple tests to show that geometry shaders as well as the tessellation shader stage works.

Would be good if someone looks over the changes, to give me an imput, what i should/could change. 

I would also like to mark a few Vertex+Fragment shader specific methods as deprecated, and in the the same time promote the generic versions like 
"getShaderProgramLanguage(ShaderType.Vertex) instead of getVertexShaderLanguage()"

So all in all this is still a 99% working WIP, but i would like to have some inputs before working completely in the wrong direction


Nehon:
Nice!
gonna look into it
Thanks

empirephoenix:
As you can see at  commit 99d69d1 I also looked at every single change and I could not find anything that should make any problems.

shadowislord:
Looks good I guess .. Could use a bit of cleanup around ShaderKey. Anyway, we can look at that later..

Nehon:
Is that me or Tesselation control shader and Tesselation Evaluation shader are inverted?

from this example https://developer.nvidia.com/content/opengl-sdk-simple-tessellation-shader
TCS looks like that https://developer.nvidia.com/sites/default/files/akamai/gamedev/images/code_snippet_2.jpg
and yours like that
https://github.com/jMonkeyEngine/jmonkeyengine/blob/5b58bda23ffa1861952f302d7205d2fb67117175/jme3-testdata/src/main/resources/Materials/Tess/SimpleTess.tsctrl

TES looks like 
https://developer.nvidia.com/sites/default/files/akamai/gamedev/images/code_snippet_3.jpg
And yours
https://github.com/jMonkeyEngine/jmonkeyengine/blob/5b58bda23ffa1861952f302d7205d2fb67117175/jme3-testdata/src/main/resources/Materials/Tess/SimpleTess.tseval

It looks like it's the other way around.

zzuegg:
going to check out in the evenig, the only plce which comes into my mind are the bit vlues in the GL4 interface. maybe i switched them..

zzuegg:
@Nehon: you are right, the bit values for Control and Evaluation Shaders are switched.. Going to fix this, and the tests today evening.

PR #224, opened 2015-03-03
Tessellation shader fix

zzuegg:
Fixed the swapped bits for the shader flags and renamed the shaders for the test


Nehon:
Thanks will try that tonight ;)

Nehon:
Alright works fine.

PR #227, opened 2015-03-04
build: quick fix to run javadoc with jdk < 1.8

davidB:


normen:
Better use if (JavaVersion.current().isJava8Compatible())

davidB:
advice applied

PR #228, opened 2015-03-05
fix a NPE and a ClassCastExcpetion on Cinematic.clear()

davidB:


Nehon:
Thanks David

davidB:
np, I forgot it.
I push it when I work on xbuf. 2 notifications about xbuf in the same day, it's a sign :-P

PR #245, opened 2015-03-12
Update OBJLoader.java

simon-heinen:
Improved the parsing performance (important mainly for Android). Example model with 32k vertices took initially 71 seconds and loads now in around 15 seconds. I tried to keep as much as possible of the original structure. Main problem on Android is that the Scanner class is horribly slow and the Pattern class as well so I got rid of them and added my own implementation of the split method.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/scene/plugins/OBJLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/scene/plugins/OBJLoader.java
@@ -73,571 +73,571 @@ import com.jme3.util.IntMap;
  */
 public final class OBJLoader implements AssetLoader {
 
-	private static final Logger logger = Logger.getLogger(OBJLoader.class
-			.getName());
-
-	private static final char SPLIT_CHAR_SPACE = ' ';
-
-	private static final char SPLIT_CHAR_SLASH = '/';
-
-	protected final ArrayList<Face> faces = new ArrayList<Face>();
-	protected final HashMap<String, ArrayList<Face>> matFaces = new HashMap<String, ArrayList<Face>>();
-
-	protected String currentMatName;
-	protected String currentObjectName;
-
-	protected final HashMap<Vertex, Integer> vertIndexMap = new HashMap<Vertex, Integer>(
-			100);
-	protected final IntMap<Vertex> indexVertMap = new IntMap<Vertex>(100);
-	protected int curIndex = 0;
-	protected int objectIndex = 0;
-	protected int geomIndex = 0;
-
-	protected ModelKey key;
-	protected AssetManager assetManager;
-	protected MaterialList matList;
-
-	protected String objName;
-	protected Node objNode;
-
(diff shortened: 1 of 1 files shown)


shadowislord:
Hi,
Thanks for this. It looks like the formatting has changed though, it is using tabs which makes it hard to see the changes. Is there any way you can use the previous formatting so the changes can be seen?

simon-heinen:
I will check if I can edit the default Eclipse preferences for the auto formatting

simon-heinen:
well looks like its a common problem http://stackoverflow.com/questions/407929/how-do-i-change-eclipse-to-use-spaces-instead-of-tabs ;) I will push an updated version

NetzwergX:
> Is there any way you can use the previous formatting so the changes can be seen? 

You can always add `?w=0` to the url to get a diff that ignores whitespace changes:
https://github.com/simon-heinen/jmonkeyengine/commit/ee801717473936f934a5b34c75bd94744c21cfcf?w=0

It's not the same as retaining the original format, but good enough to take a quick look at the actual changes ;)

gouessej:
Sorry but please can you minimize the changes? You seem to use a smaller value for the maximum line size. Dropping the scanner is a good idea as I suggested several months ago: http://hub.jmonkeyengine.org/t/obj-importer-does-not-support-backslash-as-line-continuation/31056/2?u=gouessej

Why have you removed the public constructor from the internal class "Face"? Why have you dropped the commented source code in ObjectGroup? Why do you throw an exception instead of adding a message into the logs? Is your "warning" method really necessary? Why have you stopped using the enhanced for loop in constructMesh()?

To sum up, please concentrate on the necessary changes (especially in splitByChar()) and drop the unnecessary changes so that we can really understand what you have tried to modify.

gouessej:
In my humble opinion, you should move your cosmetic changes (braces, variables renaming, ...) into a separate commit.

simon-heinen:
@gouessej Yes you are right about the separation into individual commits, I initially did not plan to commit my changes at all so I only have this final version with all the cosmetic changes by the auto-formatting and the actual refactoring process is not understandable anymore. I will check if I have later some time to go through your questions at least and answer them.

Btw is there a jme-formatting-rules file for Eclipse I can download?

shadowislord:
@simon-heinen 
jME3 uses NetBeans IDE default formatting. I am not sure if such selection is available on Eclipse.

empirephoenix:
If you are under linux, my suggestions would be to use meld
checkout stock jme and your own stuff into seperate folders,
meld foldera folderb.
It then has a nice compare ui where you can quickly hop trough all formating changes.
BTW there are a ton of classes in jme not according to the netbeans default formating, maybee we should do a format all in netbeans and commit that at some time?

shadowislord:
@simon-heinen
can you remove changes unassociated with the parsing performance improvement?

empirephoenix:
3 month idle, author seemed to moved on to android development.

PR #248, opened 2015-03-22
UUIDSerializer

NemesisMate:
Just another serializer, this one for UUIDs

### DIFF
--- a/jme3-networking/src/main/java/com/jme3/network/serializing/Serializer.java
+++ b/jme3-networking/src/main/java/com/jme3/network/serializing/Serializer.java
@@ -103,6 +103,7 @@ public abstract class Serializer {
         registerClass(Float.class,     new FloatSerializer());
         registerClass(Double.class,    new DoubleSerializer());
         registerClass(String.class,    new StringSerializer());
+        registerClass(UUID.class,    new UUIDSerializer());
 
         registerClass(Vector3f.class,  new Vector3Serializer());
 
--- a/jme3-networking/src/main/java/com/jme3/network/serializing/serializers/UUIDSerializer.java
+++ b/jme3-networking/src/main/java/com/jme3/network/serializing/serializers/UUIDSerializer.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2009-2012 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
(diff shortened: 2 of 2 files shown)


NemesisMate:
Can the UUID serialization be added too on the [BinaryOutputCapsule] (https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryOutputCapsule.java)?:

pspeed42:
I don't really think this should be part of core... and definitely not included by default in the serializers.  It's a fairly rare edge case that would require this since normally more efficient IDs can be used... and it seems trivial to have the message that includes the UUID to use string or something else.

public void setId( UUID id ) {
    this.id = id.toString();
}

public UUID getId() {
    return UUID.fromString(id);
}

By including this by default, you make everyone using this force upgrade their clients when otherwise it wouldn't have been necessary.  It also sends a bad message that using UUIDs is actually a good way to do networking when in a high-performance networking situation, it's almost always better to used shorter localized IDs.

NemesisMate:
I understand your point of view. I didn't think on the need of upgrade the clients for have this working and the "high-performance networking situation" issue as for my case, actually, UUIDs are the best way. However, adding it to the BinaryOutputCapsule would be possible?

pspeed42:
I don't think that's a great idea, either.  Nothing prevents your from writing/reading UUIDs from a capsule.  The point of these storage mechanisms is not so "support everything that might be needed" but to support a "minimal set of primitives" necessary for reading/writing.  It's trivial to read/write a UUID to a capsule as it is.  Just like it is trivial to do so for networking.

In your networking case, it is really surprising that UUIDs are the best way... as that implies that you have 2^128 objects to keep track of... which would never fit in the memory of any desktop computer.  :)

NemesisMate:
You are right, it's not the most efficient way, but my "best" is based on some other thoughs. Maybe, it would be better to use a simple double for it and store on disk the last generated ID but this may cause some other troubles and complexities when possibly doing a balanced server for having this ID synchronized. I'm not experienced on creating game servers so I think is easier and can save me time and trouble if I use UUIDs rather than longs (my networking traffic and disk storage is not really critical by the moment).

About adding this, I proposed just because I didn't really know your "minimal set of primitives" support philosophy ;).

pspeed42:
That's more a topic for the forum... but you can have IDs on the server that are different than on the client.  But if efficiency is not really needed then it's also not a big deal to just send the UUIDs as strings.

empirephoenix:
Reading the comments and activity I would say kinda rejected as not neccessary.

PR #250, opened 2015-03-27
Added sweepTest to jme3-bullet-native

rainmantsr:


empirephoenix:
This lookes fine for me, @normen what do you think about this?

shadowislord:
I can confirm TestSweepTest works, however, it seems the sweep test returns the object being sweeped as well (meaning, the capsule shape). This behavior doesn't seem to occur with JBullet. Currently investigating ...

shadowislord:
I changed TestSweepTest so it ignores the shape being sweeped. This is a behavior difference between JBullet and native bullet that needs to be accounted for.

normen:
Its okay the way it is but you copied the ugly way the other hit results are copied to java (addResult/addSweepResult), which was done as a contribution.. I wanted to change how this is handled, guess its my fault for keeping this bad example in  ^^

PR #251, opened 2015-03-30
Added: GhostObject sweep- and raytest

rainmantsr:


empirephoenix:
Hi, thanks for your work,

can you shortly explain what this does and why it is necessary? 
Also what the difference to the normal ray and sweep is. I have not used Ghostobjects before and don't know what they are usefull for :)

rainmantsr:
Hmm - actually I am not 100% sure about it. I just wrapped it;) I think it reuses the results form the broadphase to make a faster sweep/raytest. Bullet uses it internal for the KinematicCharacterController. In my tests it has been faster the the sweepTest against the dynamicsWorld, but not as fast as I hoped;) I am currently working on a small physx wrapper prototype to check if this gives me the performance i need...

empirephoenix:
I think, that since they apparently did not give a great performance boost, and are not documented in the bullet library at all, this could create quite some issues. So if no one objects I'm closing this next time i review the pull requests.

PR #272, opened 2015-05-11
Added support for 16 bit GL_LUMINANCE Textures

zzuegg:
Not a new feature at all, but added missing loader for 16 bit grayscale textures. 3.1 loaded the files as RGB8, while 3.0 was loading a Luminance16 

Not sure why this feature was removed from 3.1. I am probably overseeing some big downside of this format. But i have not yet found any other way to load a 16 bit heightmap in 3.1 due to the missing formats. 

I am open to suggestions


shadowislord:
The reason those formats were removed is because in OpenGL 3 all GL_LUMINANCE formats were removed. See here:
https://stackoverflow.com/questions/20076814/replacement-for-gl-luminance-gl-luminance-alpha

So if you're targeting OpenGL 3+ you will not be able to use those formats. The replacement is GL_RED and GL_RG. The issue is, those formats are not available in OpenGL 2.

There's no single channel format that is available in both OpenGL 2 and 3 - not sure how that happened, but that's how it is. 

Maybe we need a new format, like X8 and X16, which essentially means, whatever single channel format the GPU supports.

zzuegg:
Oh, seems it is a dealbreaker for me, wonder why nobody else notices that yet, basically every heightmap should not work anymore in 3.1 (beside the 8 bit one).

Just as a note, also the ImagePainter needs to support the X8, X16 format, as it is used by the ImageBasedHeightMap to generate the CollisionShapes

shadowislord:
Its probably best to bring those formats back .. It will cause too much damage to models that have embedded luminance textures, for example. 

If Core Profile is used, luminance textures can be uploaded by using the texture swizzle extension, that's probably a better approach. That extension became core in 3.3, which should be supported by most GL3+ level hardware.

PR #274, opened 2015-05-18
SDK SceneComposer : some refactor

Dokthar:
So smalls changes : now the DeleteShortcut cannot be activated at the same time as and other shortcutTool (prevent form miss deleting). And cleaned some code to be more functional (reduce side effects). Get involved here : http://hub.jmonkeyengine.org/t/improving-scene-composer-tools/32230/29

ps : don't care about the two last commits, wrong branch commits ^^


Dokthar:
I close this request, i'll make a new branch to avoid further conflict (with the reverted commit)

normen:
That actually wasn't the issue, I was in Iceland so I couldn't merge this :)

Dokthar:
ok, but i just was afraid of having trouble while trying to merge a branch on which i doesn't want the reverted commit

PR #276, opened 2015-05-22
Fixes the initialisation of the default SpotLight

zzuegg:


shadowislord:
I will merge this, but since the default for point lights is to have infinite range, I am a bit confused as to why it is finite for spot lights. 
@Nehon Do spot lights support infinite range?

PR #277, opened 2015-05-24
Added GL_MAX_VERTEX_UNIFORM_COMPONENTS

zzuegg:
Required to know how many uniforms can be used in the vertex shader.

Probably also required by the hardware skinning?

The code was already in 3.0 but was commented out in 3.1, you probably had a reason for this i guess. 

Following some stackoverflow talk, on some ATI cards the returned number has to be divided by 4. Not implemented that because i have not access to testing machines. 

Returns 4096 on my geforce 780


shadowislord:
Looks OK, but it won't work on OpenGL ES 2.0 - i.e. Android and iOS devices.
Can you make it use GL_MAX_VERTEX_UNIFORM_VECTORS if using OpenGL ES?

zzuegg:
No problem. Adding GL_MAX_VERTEX_UNIFORM_VECTORS and keeping, or removing GL_MAX_VERTEX_UNIFORM_COMPONENTS? 

Also since i am dooing it should i add the queries for the other shader stages too?

There should also be a note in the doc what that number actually says. 
like 
```
uniform int someNumber //takes 1 full vector or 4 components
uniform int someArray[10] //takes surprisingly 10 vectors or 40 components on nvidia while i assumed it would align and take 3 vectors or 12 components.
```

shadowislord:
I don't think desktop OpenGL 2.0 supports GL_MAX_VERTEX_UNIFORM_COMPONENTS, and the constants have different values, so I guess you will need to do things differently depending on whether OpenGL ES 2.0 or OpenGL 2.0 is used.
The docs should match the OpenGL specification. Its OK to add for other shaders stages if you like.

Hardware skinning doesn't use this, it just tries to compile the hardware skinning shader and if that fails, will use software skinning.

zzuegg:
So i just add them all, and its up to a higher layer to decide if COMPONENTS or VECTORS has to be used? 

Alternative is to make our own limit, like MAX_SHADER_VARIABLES and decide in the init step if COMPONENTS or VECTORS is used. 

I personally would prefer first, since it is in line with the gl specs. But more error prone since the user has to check himself.

shadowislord:
I would just expose the VECTORS and compute it for desktop OpenGL based on COMPONENTS. 
Its just a division by 4 isn't it?

zzuegg:
All right, going to update it into that direction.
It should be division by 4. On most cards..

shadowislord:
@zzuegg any update on this? It should be fairly simple to do ...

zzuegg:
oh damn, totally forgot about this. Think i have already implemented it locally. Just have to cherry pick the commits and test quickly..

zzuegg:
Ouch, sorry about the autoformatting. That key combination is burned in :(

shadowislord:
Looks good, but removing those spaces wasn't really necessary ...

zzuegg:
@shadowislord I know, thats why i started the thread about the auto-reformat tool. I really tried, but i was not able to re-add those spaces. Tried reverting to a previous commit and copy/paste from github.

PR #279, opened 2015-05-27
Fixed parallax in PBRLighting

riccardobl:
I've added the missing code taken from Lighting.*


Nehon:
Sorry but, I can't integrate that PR.
I use World space to compute lighting in PBR as opposed to viewSpace in Lighting.
Parallax needs to be computed in world space.
Could you add an Issue? Or if you feel like it, modify this PR.

riccardobl:
The pbr's fragment uses the vViewDir that is missing, so I thought that maybe it was accidentally removed as result of a refactoring and the result of this fix looks fine to me so I erroneously thought it was ok...
I think I will leave the task to rewrite the parallax code to someone who knows it better.

Nehon:
It was removed on purpose actually, but yeah it broke PBR.
It works fine as you did, but there are much unnecessary operations going on.
Thanks anyway, I'll look into it.

riccardobl:
Ok, I've added an issue as you asked https://github.com/jMonkeyEngine/jmonkeyengine/issues/280

PR #282, opened 2015-05-28
SDK SceneComposer : smalls changes and cleaner code

Dokthar:
smalls changes : now the DeleteShortcut cannot be activated at the same time as and other shortcutTool (prevent form miss deleting). And cleaned some code to be more functional (reduce side effects). Get involved here : http://hub.jmonkeyengine.org/t/improving-scene-composer-tools/32230/29


Nehon:
looks good thanks

PR #287, opened 2015-06-18
Fix for ZipLocator and path with the root slash

riccardobl:
While for the FileLocator "/path/x.png" and "path/x.png" are the same location, the ZipLocator (since they are passed directly to ZipFile.getEntry) treats them as two different location, as it should be on a zip archive, but since there is no need of this for the asset manager I think it's reasonable to make both FileLocator and ZipLocator work in the same way.

I noticed this problem after I load a blend model from a zip file and its textures were missed because all path were starting with a slash. I'm not sure if the HttpZipLoader is affected too.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/asset/plugins/ZipLocator.java
+++ b/jme3-core/src/plugins/java/com/jme3/asset/plugins/ZipLocator.java
@@ -82,6 +82,7 @@ public class ZipLocator implements AssetLocator {
 
     public AssetInfo locate(AssetManager manager, AssetKey key) {
         String name = key.getName();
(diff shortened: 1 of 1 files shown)


shadowislord:
Asset paths starting with "/" are pretty unusual .. Perhaps the change should be made in the Blender loader instead?

riccardobl:
Initially i thought to do in this way, but then i realized that editing the ZipLoader has more sense, since '/' should logically represent the root of the filesystem, and all the locators should have the same behavior for developers in order to be easily interchangeable.

shadowislord:
Yeah this makes sense ... Compatibility is probably more important here.

PR #288, opened 2015-06-18
add GL3 support for Unshaded and MRT

davidB:


shadowislord:
I am hoping maybe we can have some compatibility layer that can be used to create GLSL 1.1 and GLSL 1.5 shaders simultaneously (without duplicating code). 
For example: 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/resources/Common/ShaderLib/GLSL150Compat.glsllib

Nehon:
*cough* shader nodes *cough*

shadowislord:
Yeah but I think we should still support writing traditional shaders, even if its within the shader node system.

Nehon:
Shader node will generate a glsl 1.5 shader if the hardware supports it, else it will generate a glsl1.1 shader. Without the need of thinking about it in 90% of the case

Nehon:
Meaning except for the change in the GLRenderer, this change's use case could be worked out with  the UnshadedNode.j3md that is already in the engine.

davidB:
Today, I tried/explored several ways to use GLSL150Compat (without success). 
My experiments:
1. in GLRenderer.updateShaderSourceData(...) where "#version" is inserted, I insert "#import GLSL150Compat" if version < 150 && gl3 != null.
    failed because #import already processed.
2. in GLRenderer.updateShaderSourceData(...), try to insert content of GLSL150Compat, but GLRenderer should not dependents of AssetManager, so it's a bad idea
3. in J3MLoader try to create Techniques for GLSL150, but only defaultTechnique can have multiple definitions, and we can't convert shader, because target OpenGL version is unknown.
4. try to create a basic .vert and .frag with 2 imports : GLSL150Compat and the .vert/.frag for version 110.
  failed because GLSLLoader only accept .glsllib as import, and it also require to create the duplicate techniques in j3md (only for default techniques)

What I didn't experiement is to create a ShaderUtils.convertToGLSL150(Shader) that do replace like convertTOGLSL130 (so don't use the .glsllib) and call it into GLRenderer.updateShaderSourceData(...).
I guess the last should work.

davidB:
@Nehon I have to add Unshaded15.{vert|frag} to be able to display Fps/Stats (I also used it to display light point). I saw UnshadedNode.j3md but I didn't know how to use it (for thoses cases).

davidB:
Sorry for the diff noise, with trailing whitespace removal (it's the conf in my company)

shadowislord:
@davidB 
The purpose of including that GLSL lib is to use the same vert / frag files for both GLSL 1.1 / 1.5. It automatically converts a GLSL 1.1 shader to 1.5. It doesn't do anything unless GLSL 1.5+ is the version being targeted.
This means you can declare two techniques in a material that specifies the same shader, but uses a different GLSL version. 

I am not sure if switching to shader nodes was actually needed here, perhaps @Nehon can chime in if that's the intended usage.

davidB:
@shadowislord 
It's why I guess, but it requires to duplicate the technique in j3md, and will not work for named technique (one technique per name, one Shader version per technique)

davidB:
@shadowislord , @Nehon what should I change in the PR ?
[diff without space, append ?w=1](https://github.com/jMonkeyEngine/jmonkeyengine/pull/288/files?w=1)

shadowislord:
I believe @Nehon  should be the one to review it since it is using shader nodes now.

As an aside: I really don't understand what's the benefit of using shader nodes here, the technique issues can probably be fixed in the core itself... I.e. allow multiple named techniques and choose one automatically based on caps. Also there's some runtime overhead with using shader nodes which now would be imposed on everything due to these changes.

Nehon:
I'd be surprised you can mesure that overhead.
IMO here the benefit of shader nodes is precisely that you don't need to care about glsl version.
Appart from that, I admit the benefit is quite limited.
I find it a bit dumb to need an unshaded15.j3md to be honest.

davidB:
@shadowislord ,

GLSL150Compat is not enough. on an other branch I rewrite the PR to use GLSL150Compat instead of UnshadedNode but, it failed at runtime.

```
Jul 05, 2015 11:42:12 AM com.jme3.renderer.opengl.GLRenderer updateShaderSourceData
WARNING: Bad compile of:
1	#version 150 core
2	#define HAS_COLOR 1
3	// -- begin import Common/ShaderLib/GLSL150Compat.glsllib --
4	#if _VERSION_ >= 150
5	out vec4 outFragColor;
6	#  define texture1D texture
7	#  define texture2D texture
8	#  define texture3D texture
9	#  define texture2DLod texture
10	#  if defined VERTEX_SHADER
11	#    define varying out
12	#    define attribute in
13	#  elif defined FRAGMENT_SHADER
14	#    define varying in
15	#    define gl_FragColor outFragColor
16	#  endif
17	#endif
18	// -- end import Common/ShaderLib/GLSL150Compat.glsllib --
19	
20	#if defined(HAS_GLOWMAP) || defined(HAS_COLORMAP) || (defined(HAS_LIGHTMAP) && !defined(SEPARATE_TEXCOORD))
21	    #define NEED_TEXCOORD1
22	#endif
23	
24	#if defined(DISCARD_ALPHA)
25	    uniform float m_AlphaDiscardThreshold;
26	#endif
27	
28	uniform vec4 m_Color;
29	uniform sampler2D m_ColorMap;
30	uniform sampler2D m_LightMap;
31	
32	varying vec2 texCoord1;
33	varying vec2 texCoord2;
34	
35	varying vec4 vertColor;
36	
37	void main(){
38	    vec4 color = vec4(1.0);
39	
40	    #ifdef HAS_COLORMAP
41	        color *= texture2D(m_ColorMap, texCoord1);     
42	    #endif
43	
44	    #ifdef HAS_VERTEXCOLOR
45	        color *= vertColor;
46	    #endif
47	
48	    #ifdef HAS_COLOR
49	        color *= m_Color;
50	    #endif
51	
52	    #ifdef HAS_LIGHTMAP
53	        #ifdef SEPARATE_TEXCOORD
54	            color.rgb *= texture2D(m_LightMap, texCoord2).rgb;
55	        #else
56	            color.rgb *= texture2D(m_LightMap, texCoord1).rgb;
57	        #endif
58	    #endif
59	
60	    #if defined(DISCARD_ALPHA)
61	        if(color.a < m_AlphaDiscardThreshold){
62	           discard;
63	        }
64	    #endif
65	
66	    gl_FragColor = color;
67	}

Jul 05, 2015 11:42:12 AM com.jme3.app.Application handleError
com.jme3.renderer.RendererException: compile error in: ShaderSource[name=Common/MatDefs/Misc/Unshaded.frag, defines, type=Fragment, language=GLSL150]
ERROR: 0:4: '' : syntax error: incorrect preprocessor directive
ERROR: 0:4: '' : syntax error: unexpected tokens following the preprocessor directive - expected a newline
ERROR: 0:32: 'varying' : syntax error: syntax error

	at com.jme3.renderer.opengl.GLRenderer.updateShaderSourceData(GLRenderer.java:1120)
	at com.jme3.renderer.opengl.GLRenderer.updateShaderData(GLRenderer.java:1147)
	at com.jme3.renderer.opengl.GLRenderer.setShader(GLRenderer.java:1211)
	at com.jme3.material.Material.render(Material.java:1215)
	at com.jme3.renderer.RenderManager.renderGeometry(RenderManager.java:568)
```

I also tried with
```
#if __VERSION__ >= 150
```

as I use `__VERSION__ == 120` on some shader (I copied/adapt from G3D).

UnshadedNodes: +1 point ;-)

davidB:
replaced by #294

PR #290, opened 2015-06-22
Improve compressed serialization

AlrikG:
The GZIP and ZIP serializer copying the whole temporary buffer instead of only the part where the data was written. This can be easily fixed.

Forum thread: http://hub.jmonkeyengine.org/t/gzipserializer-creates-too-big-buffers/32925


shadowislord:
Nobody should be using those classes.. whatever.

PR #292, opened 2015-06-29
Reduce uniforms

AlrikG:
I've found an issue with the uniforms. It seams that they are sent to the GPU even if the value hasn't changed.
In my game I was able to reduce the uniforms from ~10.000 to ~2000 which results in up to 15% of performance increase. For me it was about 10 fps more than before.

forum link: http://hub.jmonkeyengine.org/t/performance-issue-too-much-uniform-updates/33016


Nehon:
Hello, I'm really surprised by this issue, but I can't figure what you changed because the formatting is all wrong.
Could you please try to keep the right formatting?

AlrikG:
Hi,
OK, I've tried to use the default formatting. Sry for the typo in the branch name and in the last commit. Is it possible to rename my branch/commit without creating a new one?

shadowislord:
Not exactly sure what this patch does? There's already a check to verify the value being set is not equal to the current value, why not just utilize this preexisting check? Also you should not use FloatBuffer.compareTo() as it will compare the contents of the float buffers which will significantly slowdown hardware skinning (which changes the matrix array per geometry per frame)

AlrikG:
Because you cannot compare two variables of the same reference. This patch prevents that the reference of a variable is written directly into the uniform variable "value", so without that comparing is not possible. I also have to check if matrices has changed. Have you tested hardware skinning in a complex scene with this patch applied?

shadowislord:
There's no need to test hardware skinning because its quite clear what the result would be based on the changes. First you have FloatBuffer.compareTo() which does a member-wise comparison between every float in the buffer, then later you have a buffer copy operation. This has to happen for each matrix and array uniform, per geometry, per frame. Let's say you have 100 bones in each model and 50 models on screen, that's (100\*16)\*50\*60 = **4,800,000** float comparisons plus 4,800,000 * 4 = **18 MB** of data copied per frame. Given that matrices and arrays change every frame in the engine (since the camera moves and the model animates), there's really no point in having those extra checks and copies.

AlrikG:
"There's no need to test hardware skinning because its quite clear what the result would be based on the changes"

I don't think so. Look at the screen shots on you own:

javaw 2015-07-01 10-30-06-43
javaw 2015-07-01 10-29-22-54

empirephoenix:
I would guess, that the float buffers itself are no the gain, but the other uniforms used by the models can be reused efficiently.

shadowislord:
@empirephoenix 
After testing this it seems the main gains are the per-camera matrices that don't change per model, e.g. ViewProjectionMatrix, ProjectionMatrix, and ViewMatrix. Those save the majority of uniforms it seems.

empirephoenix:
Hm, so if the camera moves, the gain is gone, right? Aka the test case might be flawed.

shadowislord:
No, because the material is shared among all geometries, the camera uniforms are set on the material once at the beginning of the frame and then the check passes so it does not set them again.

empirephoenix:
So, it actually does work then and help gain performance.

shadowislord:
Yes it does, but it has a very severe worst case performance: if you have a large matrix array which doesn't change, the compare has to go through the entire native buffer. If the comparison was changed only to happen for Matrix3 / Matrix4 type uniforms as opposed to arrays you would get the performance benefits without the worst case performance decrease.

empirephoenix:
Couldn't the part setting the uniforms for the camera matrices be made frame aware? So it only sets the camera stuff once per frame? Woudl reduce all comparison, while still giving the upside?

zzuegg:
It would have to be frame and camera aware to allow rendering to multiple cameras.. And that would probably be nearly the same as comparing the matices directly. 

Might be an option to cache the hashvalue however so that the check is basically only and int==int

shadowislord:
I definitely want to see something that would optimize away View/ProjectionMatrix since those never change for the entire frame - it seems to save a lot of glUniformMatrix calls especially for the lighting shader which loves to use those matrices ... For all other cases, its either already implemented or not necessary to be implemented (like for skinning).

empirephoenix:
Transformed to issue for the outstanding part:
Reduce uniform changes by comparing them before frame #351

PR #294, opened 2015-07-05
add GL3 support for Unshaded and MRT (v2)

davidB:
I remade the PR #288 to have a cleaner diff and history (commit log).

The PR include a fix for  GLSL150Compat  (the version + auto-injection of the define XXXX_SHADER).


shadowislord:
Looks good.

PR #301, opened 2015-07-14
Support for setting minification and magnification filters on a texture in the j3m material file

### LINKED ISSUE #295: Allow to set min filter and mag filter for a texture in the j3m file
Textures can have diffrent convolution algorithm when they are minified or magnified on screen.
There is a way to change this algorithm by code, throught the Texture API.

Though when you make a material though a j3m file, and want different filtering, you have to load your material by code and change the filtering in the code. It woudl be nice to be able to set the filtering directly in the j3m file the same we specify Flip or Repeat.

The keyworkd could be any value from the MinFilter and MagFilter enum foun din the Texture class 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/texture/Texture.java

But each keyword should be prefixed with Min or Mag for example : 
MinTrilinear MagBilinear.

DannyJo:
Added support for setting minification and magnification filters on a texture in the j3m material file. This also adds support for double and single quoted paths as well as being able to set WrapMode for a specific WrapAxis.

This PR also contains a slight change to a couple of build scripts to include the other modules as dependencies rather than have their sources referenced directly. This was causing an issue with IntelliJ IDEA which does not allow source paths residing outside a modules root.

This pull request resolves #295


DannyJo:
It says the build fails when building the SDK. This was happening in the clean updated code before I created this branch as well. Not sure if anyone else can shed some light on this, is there something I've missed or is/was master in a bad state?

DannyJo:
I've reverted some build script changes and the build now appears to pass again. Should be good to go now.

empirephoenix:
Hi, looks good, one question tho, why did you comment out the bullet project in the gradle file?

DannyJo:
That was never meant to be there in the first place, I should have deleted that row altogether really.

Nehon:
Do you have a test case and a file? I could help to validate the PR.

DannyJo:
I used the TestPostFilters test case and just commented out the filter and made some temporary changes to BrickWall.j3m and Sign Post.j3m while testing the new code. Is that good enough or do you actually want something else? I could write a unit test for the J3MLoader which is probably the right thing to do really ;)

DannyJo:
I have now added a unit test for the J3MLoader which tests that the old and new way of parsing texture parameters work as expected. The class could do with some refactoring as a whole though but that's a different issue altogether.

empirephoenix:
I would say lets merge it, and see what happens, after all git head is not supposed to be stable. If there are still any issues, they can be found faster, and fixed.

Nehon:
I can't test it before 2 weeks. (answering from my phone). So yes,  go ahead merge this. But please pull it before and at least try to load stock materials.

kwando:
:+1: for unit tests :smiley:

DannyJo:
Any chance we could get this PR merged in soon @Nehon as I've foolishly merged this into my own master branch before I started a new feature branch for #314 and I'd rather not create a PR for that before this is merged in.

Nehon:
Yes! I'll check it out tonight

Nehon:
Hello, looks pretty good.
But why the changes in the gradle build?

DannyJo:
There are some changes in the build script to add some test dependencies and enable unit testing as a part of the build. Does that look wrong for some reason?

DannyJo:
I've merged in master into this issue branch now so the pull request can be merged without any conflicts.

Nehon:
Tested it, it seems to work fine.
We need to have the SDK support those changes though.
Since it's backward compatible I merge it and I'll make the change on the material editor later.

Issue #297, opened 2015-07-13
Add new constructors for lights


shadowislord:
Add the following convenience constructors:
`AmbientLight(ColorRGBA color)`
`DirectionalLight(ColorRGBA color, Vector3f direction)`
`PointLight(ColorRGBA color, Vector3f position)`
`SpotLight(ColorRGBA color, Vector3f position, Vector3f direction, float innerAngle, float outerAngle)`

Dokthar:
Btw adding a cloneForSpatial(Spatial spatial) into the LightControl would be very appreciate into the scene editor.

Nehon:
Please, this is harldy related to the issue.
Create an other issue, and describe it thouroughly.
Thanks

pspeed42:
Not to mention that LightControl already has cloneForSpatial()... so.... yeah, if you do mention this on a forum or a different issue then include exactly what you are talking about at the same time.

zzuegg:
any news on this, there are two PR open for this issue: 306 and 302

DannyJo:
Pull Request #306 resolves this issue.

PR #302, opened 2015-07-15
Constructors for the lights. Issue #297

zzuegg:


Nehon:
Thanks for this.
But you need to add the empty constructors too.

zzuegg:
Added the missing default empty constructors

DannyJo:
Nice stuff. Might not be a big deal but in the second constructor in SpotLight.java you get double calls to computeAngleParameters().

Nehon:
Yup, if you remove this this PR is ready to be merged.

jme3-core/src/main/java/com/jme3/light/SpotLight.java
```diff
@@ -74,6 +75,39 @@ public SpotLight() {
...
+     * @param position Position of the light
+     * @param direction Direction of the light
+     * @param spotInnerAngle Inner angle of the light
+     * @param spotOuterAngle Outer angle of the light
+     * @param spotRange Range of the light
+     */
+    public SpotLight(ColorRGBA lightColor,Vector3f position, Vector3f direction, float spotInnerAngle, float spotOuterAngle,float spotRange) {
+        this(lightColor,position,direction,spotInnerAngle,spotOuterAngle);
+        setSpotRange(spotRange);
+        computeAngleParameters();
```
Nehon: This call to computeAngleParameters(); can be removed as it's been called in the prior thi(...) call

jme3-core/src/main/java/com/jme3/light/SpotLight.java
```diff
@@ -58,19 +59,50 @@
...
+    protected float spotOuterAngle;
+    protected float spotRange;
+    protected float invSpotRange;
+    protected float packedAngleCos;
     
     protected float outerAngleCosSqr, outerAngleSinSqr;
     protected float outerAngleSinRcp, outerAngleSin, outerAngleCos;
     
     public SpotLight() {
+        this(new ColorRGBA(1f, 1f, 1f, 1f), new Vector3f(), new Vector3f(0, -1, 0), FastMath.QUARTER_PI / 8, FastMath.QUARTER_PI / 6);
```
Shredder121: Can't these calculations be put in a static final field or something?

empirephoenix: Why? that would bloat the class, and the calculations need no real time at all. It is not that you are creating 1 million lights per frame. And if you want, you can simply write yourself a factory class using the second constructor, that has precalculated values in the factory.

Shredder121: Okay then, just wanted to add my 2 cents.

zzuegg: Actually, if you submit your own color/direction, you are saving 2 object creations compared with the previous version. 
+The divisions should be resolved during compile time.

Nehon:
Ok back to this.
I'm a bit annoyed because there are 2 PRs that does the same thing see #306.
We've been bad at talking to each others I guess. That's just too bad that the work has been done twice.
So since I'm going to integrate one or the other I have to choose and I'm gonna choose the one that will give me the less work because I'm a lazy ass.
I'm gonna choose PR #306 because it's a tiny bit cleaner to my taste, and it even fixes my lousy spelling in the java doc.
However it doesn't have the constructor with the radius for PointLight and the spot angles for the SpotLight...so mixed feelings, I'll add the 2 constructors.
So @zzuegg I'm really sorry to have wasted your time. Tell me how I can make it up to you ;)

PR #303, opened 2015-07-16
Bullet RigidBody : added support to vector3f based AngularFactor & LinearFactor

Dokthar:
This modification allow to "lock" a degree of freedom of one rigidBody, without joint. This ability is only available with native bullet. 
This can be useful for using bullet for 2D game (maybe not the best option), this can be done like so : 
```java
body.setAngularFactor(new Vector3f(0,0,1)); // <-- only rotate on Z axis
body.setLinearFactor(new Vector3f(1,1,0)); // <-- only move on a 2D plane (X-Y)
```
You can also set a value between [0;1] to "reduce" the movement / rotation. 
Using angular/linear factor do not preserve the conservation of energy. (the factor only scale the "forces")

### DIFF
--- a/jme3-bullet-native/src/native/cpp/com_jme3_bullet_objects_PhysicsRigidBody.cpp
+++ b/jme3-bullet-native/src/native/cpp/com_jme3_bullet_objects_PhysicsRigidBody.cpp
@@ -829,20 +829,20 @@ extern "C" {
      * Method:    setAngularFactor
      * Signature: (JF)V
      */
-    JNIEXPORT void JNICALL Java_com_jme3_bullet_objects_PhysicsRigidBody_setAngularFactor
-    (JNIEnv *env, jobject object, jlong bodyId, jfloat value) {
-        btRigidBody* body = reinterpret_cast<btRigidBody*>(bodyId);
-        if (body == NULL) {
-            jclass newExc = env->FindClass("java/lang/NullPointerException");
-            env->ThrowNew(newExc, "The native object does not exist.");
-            return;
-        }
-        btVector3 vec1 = btVector3();
-        vec1.setX(value);
-        vec1.setY(value);
-        vec1.setZ(value);
-        body->setAngularFactor(vec1);
-    }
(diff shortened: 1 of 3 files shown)


normen:
Sorry, this commit would break compatibility of the jbullet and bullet implementations. bullet may have additional features but not break code written for jbullet

empirephoenix:
Why would this break code written for jbullet? It sure would not allow code written with these two features to run on jbullet, but aside from that I see no problems.

afonsolage:
Because the changes on cpp file, as I can see, it changes many functions:

-    private native float getAngularFactor(long objectId);
+    private native void getAngularFactor(long objectId, Vector3f vec);

shadowislord:
First of all it changes the API so old code would not compile anymore. Second it changes the serialization format so exported models would stop loading correctly. If those functions are to be added, they have to be added alongside the existing ones, not replacing them.

empirephoenix:
Ah ok, I did miss that, it sounded to me as only 2 new methods were added, and their values put into the capsule. (The capsule itself can handle non used values, and defaults, so no probem there if done properly or?)

Dokthar:
You are all right. I'll make some modification to not break the API and the serialization.

normen:
Now the vector form isn't serialized at all or am I wrong?

Dokthar:
You are wright the vector isn't serialized.  So i could put a try/catch for the "read" and maybe i can write a float instead of a vector if x,y and z are equal.

normen:
read won't fail if the variable name doesn't exist, it will simply return the default value you enter (second argument)

normen:
That works better but simply use different names for the vector and float variable, then it should be safest.

shadowislord:
@normen I think its OK in this case, since it just uses overloading

normen:
Okay, if you say so I'll merge ^^

Dokthar:
but like this is okay to go from jbullet -> bullet-native, the other way will break if you used theses features.

normen:
I already merged it but no, why would it? If you wouldn't get the vector value you'd look for the float one, just like you do now but with different names. Contrary with jbullet you'd do it the other way around. If you already use the vector version your code isn't compatible anyway.

PR #305, opened 2015-07-24
fix ArrayOutOfBoundException in SkeletonControl when maxWeightsPerVer…

davidB:
…t is less than 4


shadowislord:
This is incorrect, in the case where `weights[0] != 0`, it will always increment by 4. There are two parts where `idxWeights` is incremented, in the for each weight in vertex loop, and then afterwards where its incremented by `fourMinusMaxWeights` - thus it is always incremented by 4.
The crash is likely due to the importer providing garbage data. The weights buffer should have 4 entries per vertex regardless of `maxWeightsPerVert` or `weights[0]`.

davidB:
So the weight buffer should be always  4 x vertex ?

PR #306, opened 2015-07-26
light constructors

mbenson:
per discussion at http://hub.jmonkeyengine.org/t/new-contributor-wants-to-submit-small-usability-tweaks/33150. The new constructors have been added in such a way as to preserve existing functionality whereby most direction and light data are copied during set* operations.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/AmbientLight.java
+++ b/jme3-core/src/main/java/com/jme3/light/AmbientLight.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2012, 2015 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


DannyJo:
Is this not the same as PR #302 ?

mbenson:
I wasn't aware of this PR. They are very similar. Differences that strike me:

* My version makes the lights' colors optional by providing additional constructors. The constructors that include color put it last, the idea being that when you're constructing a `DirectionalLight`, its direction is more important to you than its color.
* I included a `Light(ColorRGBA)` constructor which is called by the color-specifying constructors.
* I did not include constructors that specified the mathematical properties of a `SpotLight` as I feared that was becoming "just too much." If one were going to create such a constructor, I would consider embedding these parameters into some kind of enclosing struct-like object for tidiness' sake.
* It still appears to me that the `SpotLight` constructors have the potential to make multiple calls to `#computeAngleParameters()`; I took pains to ensure my modifications did not do this, although admittedly I may have missed something.

PR #309, opened 2015-07-29
Make 'Q' and 'Z' move flyCam along its up vector

g-rocket:
In FlyByCamera, rather than hardcoding them to be along the Y axis, 'Q' and 'Z' now move with and against respectively whatever direction has been set as its up vector (using setUpVector).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
+++ b/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
@@ -356,7 +356,7 @@ public class FlyByCamera implements AnalogListener, ActionListener {
     }
 
     protected void riseCamera(float value){
-        Vector3f vel = new Vector3f(0, value * moveSpeed, 0);
+        Vector3f vel = initialUpVec.mult(value * moveSpeed);
         Vector3f pos = cam.getLocation().clone();
 
         if (motionAllowed != null)


afonsolage:
And this will break all existing flyCam. At least add a boolean to enable or disable it. In my project, I use flyCam as it is and this change will lead me to reverse it on custom code.

DannyJo:
I think this sounds like a bug fix to be honest and it shouldn't break code unless you've set the initial up vector to something other than Y and then use Q and Z to move along the Y axis anyway which to be honest sounds a bit odd to be doing. I'm guessing @g-rocket actually uses a different up vector for the fly cam and that's why this has come about.

This pull request has my vote unless a core dev can spot an error in that one line of code that has changed.

empirephoenix:
I think @afonsolage miinterpreted your comment, as using the up vector of the current view direction. Which would change behaviour.

shadowislord:
This is a behavior change, yes. I am not sure if its significant enough to warrant avoiding fixing this, since it is not expected that it would always use Y when you change the up vector.

pspeed42:
Though if we are going to add bloat to a class that shouldn't be used in real games... why not have made it a boolean flag to turn on/off?

empirephoenix:
not relevant/class not intended for productive use

PR #320, opened 2015-08-21
Fixed default window titlebar to say 3.1 instead of 3.0

8Keep:


DannyJo:
Good spot, although with the recent version changes from @shadowislord one might think the default could come from https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/system/JmeVersion.java

Issue #313, opened 2015-08-03
Not calling Mesh.updateCounts() causes negative statistics


shadowislord:
If the user forgets to call `Mesh.updateCounts()`, the vertices / triangles value will be -1, thus causing the renderer statistics to be subtracted and eventually become negative. If the stats value is -1 the mesh should be considered invalid and crash with exception.

shadowislord:
Also it seems that an empty Mesh (with no buffers) can also cause problems

Issue #64, opened 2014-03-24
VertexBuffer.updateData() does not update vertex/triangle count (undefined OGL behavior may result)


ghost:
Reference on googlecode: https://code.google.com/p/jmonkeyengine/issues/detail?id=487

PR #334, opened 2015-09-06
Fixes Issues #313 & #64 : Calculation Errors for custom Meshes & Deprecation of VB.updateData()

MeFisto94:
Quite a simple commit:
- computeNumElements could return -2 depending on the Mode (TRIANGLE_STIRP) and now it always returns zero as it should.

- The Statistics-Render will now crash when there is an erroneous Mesh (without proper tri/vert count).
Note: This also means that a clean ```Geometry g = new Geometry("Geo", new Mesh());```will lead to a crash as the plain constructor doesn't set the tri/vert count.

Also note that when you use BasicGameTemplate the vert-count goes up by 4 when adding an "empty" Mesh, this is because the object-count goes up from 9 to 10 and thus requiring another digit.

Edit: Sorry, I was unable to open two separate pull requests since Github always wants to use the newest version of my fork :/ What would've been the right way? One branch for each new feature? (Like each issue?)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/Statistics.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/Statistics.java
@@ -128,6 +128,9 @@ public class Statistics {
         if( !enabled )
             return;
             
+        if (mesh.getTriangleCount(lod) < 0 || mesh.getVertexCount() < 0)
+            throw new IllegalStateException("Invalid triangle or vertex count on this mesh. Call .updateCounts() before rendering");
+        
         numObjects += 1;
         numTriangles += mesh.getTriangleCount(lod) * count;
         numVertices += mesh.getVertexCount() * count;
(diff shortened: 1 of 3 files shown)


shadowislord:
I was actually thinking of changing it so that computing counts is not required at all.. E.g. the Renderer and / or Statistics class can simply use the limit / componentLength of the first buffer as the vertex count, whereas the primitive count is based on the index buffer limit / mesh mode size.

This way the error prone updateCounts() method can be removed.

MeFisto94:
But wouldn't that lead to a (really small) performance overhead? 
If you call limit each frame and do some divisions (and/or) +/- instead of like a few Bytes of Memory usage?

With the current way you could even make the updateCounts() private since the only way to update Buffers is using updateBuffer. 

I guess it wouldn't make a real difference though?

empirephoenix:
I had the same underlying problem as well, having negative recorded rendered stuff, so I am for a fix to this, no matter the underlying way it works (as the stats are for debug either small ram or memory overhead are ignorable, tho I would tend to increased cpu useage, as the memory is also used if the stats are not used at all).  However I am closing this one, as it cannot be merged anymore.

Issue #314, opened 2015-08-04
LWJGL 3.x support


shadowislord:
Need to add support for LWJGL 3.x.

There shouldn't be any issues, except the loss of LwjglCanvas which may break the SDK.
AWT panels need to be stable and working before we can fully make the switch.

DannyJo:
I'm currently looking at this and the conclusion is to create a new module (jme3-lwjgl3) which can be dropped in place of the jme3-lwjgl module in order to use the lwjgl 3 renderer.

So far everything is going OK, I've got the GLFW window running and keyboard and mouse input are working as expected. Joysticks still need some work. Audio seems to be fine as well.

I also need to check that the offscreen rendering is working and test the SDK with it to ensure the canvas rendering is OK.

normen:
The SDK already uses AWT panels by default, canvas is optional.

DannyJo:
This work has now been merged into master as of pull request #340

PR #340, opened 2015-09-13
Added LWJGL 3.x module and implementation in regards to issue #314

DannyJo:
I realise this is a huge pull request but most of the real changes are in the new module so new files altogether.

Some other changes will have been made, in particular to how lwjgl native libraries were registered, this now happens in the two modules jme3-lwjgl and jme3-lwjgl3.

I also took the liberty to change the info log statement of the audio renderer to better match the one of the gl renderer.

Take note that the jme3-lwjgl3 module currently depends on LWJGL 3.0.0b which is a snapshot version. It will probably stay like this until we get a proper 3.0.0 version.

Support for keyboard, mouse and joystick is all working with a new joystick mapping added as well.

Also note that currently there will be some duplicate code between the jme3-lwjgl module and the new jme3-lwjgl3 module.


jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -0,0 +1,484 @@
...
+        glfwWindowHint(GLFW_STEREO, settings.useStereo3D() ? GL_TRUE : GL_FALSE);
+
+        int frameRateCap = settings.getFrameRate();
+
+        if (!autoFlush) {
+            frameRateCap = 20;
+        }
+
+        if (frameRateCap > 0) {
+            glfwWindowHint(GLFW_REFRESH_RATE, frameRateCap);
```
shadowislord: This is not really correct. `GLFW_REFRESH_RATE` is the display refresh rate for fullscreen mode, whereas `settings.getFrameRate()` is a software-enforced limit of framerate. `GLFW_REFRESH_RATE` should be set to `settings.getFrequency()` instead.

Probably should stick to the LWJGL2 implementation for framerate.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglTimer.java
```diff
@@ -0,0 +1,139 @@
...
+ * <code>Timer</code> handles the system's time related functionality. This
+ * allows the calculation of the framerate. To keep the framerate calculation
+ * accurate, a call to update each frame is required. <code>Timer</code> is a
+ * singleton object and must be created via the <code>getTimer</code> method.
+ *
+ * @author Mark Powell
+ * @version $Id: LWJGLTimer.java,v 1.21 2007/09/22 16:46:35 irrisor Exp $
+ */
+public class LwjglTimer extends Timer {
+
```
shadowislord: Probably doesn't make much sense to implement `Timer`. We should be using `NanoTimer` by default for everything anyway...

common.gradle
```diff
@@ -16,6 +16,9 @@ repositories {
     maven {
         url "http://nifty-gui.sourceforge.net/nifty-maven-repo"
     }
+    maven {
```
shadowislord: Perhaps this should go into the jme3-lwjgl3 project instead, since no code outside of it should be using LWJGL 3 anyway.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
```diff
@@ -125,23 +124,6 @@ public static void registerNativeLibrary(String name, Platform platform,
...
-        
-        // OpenAL
-        // For OSX: Need to add lib prefix when extracting
-        registerNativeLibrary("openal", Platform.Windows32, "native/windows/OpenAL32.dll");
-        registerNativeLibrary("openal", Platform.Windows64, "native/windows/OpenAL64.dll");
-        registerNativeLibrary("openal", Platform.Linux32,   "native/linux/libopenal.so");
-        registerNativeLibrary("openal", Platform.Linux64,   "native/linux/libopenal64.so");
-        registerNativeLibrary("openal", Platform.MacOSX32,  "native/macosx/openal.dylib", "libopenal.dylib");
-        registerNativeLibrary("openal", Platform.MacOSX64,  "native/macosx/openal.dylib", "libopenal.dylib");
-
```
shadowislord: Removing this can cause issues. The SDK is relying on being able to extract natives even though the context is not initialized. Also I strongly suggest using a different library name instead of `"lwjgl"` for lwjgl3, probably should just use `"lwjgl3"` to avoid conflicts. For openal you can use `"lwjgl3-openal"` instead. Once LWJGL 3.0 becomes official / stable we can then get rid of LWJGL 2 and replace the lwjgl and openal libraries with their LWJGL 3.0 equivalents.

jme3-lwjgl3/build.gradle
```diff
@@ -0,0 +1,12 @@
+if (!hasProperty('mainClass')) {
+    ext.mainClass = ''
+}
+
+dependencies {
+    compile project(':jme3-core')
+    compile project(':jme3-desktop')
+    compile 'org.lwjgl:lwjgl:3.0.0b-SNAPSHOT'
```
shadowislord: Any specific reason why we are using the 3.0.0b snapshot instead of 3.0.0a?

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
DannyJo: That's fine, however the LWJGL2 implementation uses Display.sync() which no longer exists. Should we just write our own implementation of that using some kind of wait mechanism?

common.gradle
DannyJo: Fair enough, I will change that.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
shadowislord: Yeah, we'll have to do that I am afraid.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglTimer.java
DannyJo: Sure thing, I suppose we don't really need GLFW for timering.

jme3-lwjgl3/build.gradle
DannyJo: Basically the 3.0.0b has some api changes in it to do with creating the GL capabilities and since this is more of an experiment module for now I thought having the 3.0.0b snapshot would be OK in order to get those api changes in place.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -0,0 +1,369 @@
...
+    private boolean runningFirstTime = true;
+    private boolean mouseWasGrabbed = false;
+    
+    private boolean mouseWasCreated = false;
+    private boolean keyboardWasCreated = false;
+
+    private long window;
+
+    private class GLCanvas extends Canvas {
+        @Override
```
shadowislord: I don't think it is possible to implement canvas using GLFW. It only supports standalone windows or fullscreen. 
Should probably just remove this class then.

zzuegg:
For pure offscreen rendering you have to create a windowed window, and set glfwWindowHint(GLFW_VISIBLE,false) before glfwCreateWindow

It does not show up in the taskbar or anywhere.. but creates a hidden context..

shadowislord:
It looks good with the latest changes. If other members are OK we can merge this.

DannyJo:
Excellent. It shouldn't be very intrusive with these changes now, basically just a new module which can be excluded for most people. I will of course be testing it myself in my project to see how it goes and I'm sure @phr00t will be interested in this one way or another.

phr00t:
I look forward to it being merged & tested. I'd like to also pull it into my repository!

phr00t:
@DannyJo  can you create a pull request of this into my branch, please?

https://github.com/phr00t/jmonkeyengine

empirephoenix:
I also propose to merge this. It wont break anyones game who is not explicitly using this, and master is unstable anyway. But it allows to check the newer lwjgl functionalities, and reduce some bugs on linux platforms.

shadowislord:
Because there are no core modifications, there should be no regressions with this. 
I am looking forward to getting rid of the LWJGL 2.x backend and replacing with this one.
Merging..

PR #344, opened 2015-09-18
use normal for irradiance, not reflvec

RYDB3RG:


Nehon:
Hello. Could you please explain.
Using normal here is definitely wrong and will produce wrong reflection.
But I'm willing to hear your arguments.

RYDB3RG:
Hi,
wsNormal seems to be used in most other PBR shaders ([cinder experiments](https://github.com/simongeilfus/Cinder-Experiments/blob/master/PBR%20Image%20Base%20Lighting%202/assets/PBR.frag#L217), [cmftstudio](https://github.com/dariomanesku/cmftStudio/blob/master/src/shaders/fs_mesh.shdr#L225), [pbrviewer](http://www.alexandre-pestana.com/physically-based-rendering-viewer/))

Nehon:
Ho that's only for the diffuse IBL, not reflection...
Mhh ok, will test this.
Did you see an improvement?

RYDB3RG:
Its barely noticable. Using a testcubemap (which is red on -X) and outputting indirectDiffuse only, it seems to be a bit more red'ish with [normal](http://i.imgur.com/a6yU9aG.png) than [refl](http://i.imgur.com/ARblE5C.png)

Nehon:
Maybe it's more noticeable with hi frequency normal map...
I'll test that.

shadowislord:
@Nehon 
Seems like this change is already applied, can we close this then?

Nehon:
Yes it has been added to the shader

PR #346, opened 2015-09-22
Eliminates the duplicated code

alexVengrovsk:
Shortening code using Java7's catch block, that can handle exceptions of the different types.


normen:
The change is not really necessary and ups the baseline for the plugin to Java7, I think we can pass on this.

PR #349, opened 2015-09-23
Optimise Strings concatenation

alexVengrovsk:
In the ```toString()``` method concatenation will be performed during the runtime, because not all parameters are ```String``` type literals.
That is why it is better to use ```StringBuilder``` here.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/util/NativeObject.java
+++ b/jme3-core/src/main/java/com/jme3/util/NativeObject.java
@@ -145,7 +145,9 @@ public abstract class NativeObject implements Cloneable {
 
     @Override
     public String toString(){
-        return "Native" + getClass().getSimpleName() + " " + id;
+        return new StringBuilder("Native")
+                .append(getClass().getSimpleName()).append(" ").append(id)
+                .toString();
     }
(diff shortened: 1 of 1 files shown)


Nehon:
mhhh this is the typical example where the compiler will do it for us...
Since java 6 String concat are optimised into StringBuilder by the compiler.
So, for readability's sake, I'd rather keep the String concat syntax.

Also not really sure this method is a critical performance bottleneck, that would require this kind of micro optimisation, if the compiler wasn't doing it for free.

pspeed42:
Agree that this is an unoptimization because it makes the code harder to read but generates basically the same byte code.

To the OP, + is only bad when it's not part of the same expression, as in:
String myValue = string1;
myValue += string2 + string3;
...and so on.
Because it's possible that each separate line gets its own StringBuilder.

alexVengrovsk:
@pspeed42 @Nehon You are right. Sorry for bothering you.

Nehon:
No problem really, you don't bother us.
It may not sound like that in our answers, but we appreciate the effort of people willing to optimise the engine ;)

PR #350, opened 2015-09-23
Delete close() of ByteArrayOutputStream type objecs

alexVengrovsk:
According to the Oracle's [docummentation](http://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html#close()): "Closing a ```ByteArrayOutputStream``` has no effect. The methods in this class can be called after the stream has been closed without generating an ```IOException```."


empirephoenix:
While you are technically right, does this offer any other improvement?

alexVengrovsk:
Not realy, just let's to make code shorter.

Issue #362, opened 2015-10-07
DefaultLightFilter does not support BoundingSphere


shadowislord:
If the user is using bounding spheres in their scene graph, the following error will appear:
`UnsupportedOperationException: Only AABB supported for now`

jME3 lights need to implement an intersectSphere method in order to support light v. sphere intersection.

Dokthar:
(yeah you see two commit here because i did a mess with the first one, tried to resolve conflict ... ouch )

PR #364, opened 2015-10-07
Lights : added light v. sphere intersection

Dokthar:
- implementations of intersectsSphere() for Ambient / Directional / Point / Spot  Light
(see #362)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/AmbientLight.java
+++ b/jme3-core/src/main/java/com/jme3/light/AmbientLight.java
@@ -32,6 +32,7 @@
 package com.jme3.light;
 
 import com.jme3.bounding.BoundingBox;
+import com.jme3.bounding.BoundingSphere;
 import com.jme3.math.ColorRGBA;
 import com.jme3.math.Vector3f;
 import com.jme3.renderer.Camera;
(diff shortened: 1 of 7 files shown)


shadowislord:
We have some unit tests for the light culling here:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/test/java/com/jme3/light/LightFilterTest.java

Could you include a test case for a model which uses a bounding sphere?

Dokthar:
Yes, i'll add this when i have time.

Dokthar:
I could also move the code for collision of sphere v. sphere (and v. cone) into Intersection.java , what do you think ?

shadowislord:
@Dokthar If you think it makes sense to share that code between bounding volumes and lights then go ahead.

Dokthar:
@shadowislord I agree that's feel strange to put the sphere v. cone collision code inside Intersection, as it's need a lot of parameters (already computed inside the cone light). 
Also I found that the collision code used for sphere v. sphere was already wrote (inside Intersection), so i use this one.

Dokthar:
Don't merge this yet, one test case is wrong, i am trying to fix this.

Dokthar:
@shadowislord : I don't have more changes comming so you can review this.

PR #372, opened 2015-10-30
Fast, precise mult with Vector3f

TehLeo:
Fast, precise implementation of multLocal(Vector3f), mult(Vector3f, Vector3f) 

From forum thread: http://hub.jmonkeyengine.org/t/4x-faster-quaternion-vector-multiplication/32618/25

### DIFF
--- a/jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
+++ b/jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
@@ -62,8 +62,8 @@ public class QuatMultTest {
             
             randVect(vect);
             
-            quat.mult(vect, store60);
-            mult15V2(quat, vect, store15);
+            quat.mult(vect, store15);
+            mult60(quat, vect, store60);
     
             //eps stands for the epsillon in the approximateEquals method
             //set your desired eps
@@ -113,21 +113,21 @@ public class QuatMultTest {
         v.set(rand.nextFloat(), rand.nextFloat(), rand.nextFloat());
         v.multLocal(10000f).subtractLocal(5000f, 5000f, 5000f);
     }
-    public static void mult15V2(Quaternion q, Vector3f v, Vector3f s) {
-        //15 mult, 15 add
+    public static void mult60(Quaternion q, Vector3f v, Vector3f s) {
         float x = q.getX();
         float y = q.getY();
         float z = q.getZ();
         float w = q.getW();
         
-        //v + 2*q.xyz cross (q.xyz cross v + w*v )
-        //q.xyz x v.xyz + w*v
-        float vx = y*v.z - z*v.y + w*v.x;
-        float vy = z*v.x - x*v.z + w*v.y;
-        float vz = x*v.y - y*v.x + w*v.z;
-        vx += vx; vy += vy; vz += vz;
-        s.x = v.x + y*vz - z*vy;
-        s.y = v.y + z*vx - x*vz;
-        s.z = v.z + x*vy - y*vx;
+        float vx = v.x, vy = v.y, vz = v.z;
+        s.x = w * w * vx + 2 * y * w * vz - 2 * z * w * vy + x * x
+                    * vx + 2 * y * x * vy + 2 * z * x * vz - z * z * vx - y
+                    * y * vx;
+        s.y = 2 * x * y * vx + y * y * vy + 2 * z * y * vz + 2 * w
+                    * z * vx - z * z * vy + w * w * vy - 2 * x * w * vz - x
+                    * x * vy;
+        s.z = 2 * x * z * vx + 2 * y * z * vy + z * z * vz - 2 * w
+                    * y * vx - y * y * vz + 2 * w * x * vy - x * x * vz + w
+                    * w * vz;
     }
 }
--- a/jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
+++ b/jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2009-2015 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.jme3.math;
+
+import java.util.Random;
+import org.junit.*;
+import static org.junit.Assert.*;
+import static com.jme3.math.FastMath.*;
+
+public class QuatMultTest {
+    Random rand;
+    @Before
+    public void setUp() {
+        rand = new Random(0);
+    }
+    
+    @After
+    public void tearDown() {
+    }
+   
+    @Test
+    public void test() {
+        Quaternion quat = new Quaternion();
+        Vector3f vect = new Vector3f();
+        Vector3f store60 = new Vector3f(); //stores result of the current implementation
+        Vector3f store15 = new Vector3f(); //stores result of the proposed change
+
+//        final int reps = 100000;
+        final int reps = 10;
+         
+        for(int i = 0; i < reps; i++) {
+            randQuat(quat);
+            
+            randVect(vect);
+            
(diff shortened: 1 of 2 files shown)


shadowislord:
Recommend creating some unit tests comparing this improved method vs the old one to make sure there's no major differences. You can use [approximateEquals](https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/math/FastMath.java#L922) for this.

TehLeo:
UnitTest: http://pastebin.com/00bFr50P

I have wrote a test and used approximateEquals. With epsilon 0.00001f approximateEquals will often return false. The test will pass the fist 100000 iterations with epsilon = 0.029f.

Example output:
Running with 100k iterations with epsilon 0.025f, the test will stop at iteration: 67821. (output included below) Looking at the results we see that 0.004163742(original) is not equal to 0.004043579(improved) using the epsilon 0.025f(it is with epsilon 0.029f). The bigger the epsilon needs to be to pass the test, the bigger the difference between the two methods is. This does not answer which method is more accurate. We can, however, calculate the quaternion vector product using a mathematical software. In this case SageMath outputs the result: (3806.79915617854, 0.00397302788272214, 182.414682736066) and 0.00397302788272214 is closer to 0.004043579(improved) than to 0.004163742(original).

------Output with eps = 0.025f
Iteration: 67821
quat (-0.08545344, 0.2544827, -0.93980426, -0.2114336), vect (-3361.9124, -1759.1035, 357.90723)
store60 (3806.7993, 0.004163742, 182.41466), store15 (3806.7993, 0.004043579, 182.4147)
diff (0.0, 1.20162964E-4, -4.5776367E-5)
(true,false,true)

shadowislord:
Can you put it into `jme3-core/src/test/java/com/jme3/math/`?

TehLeo:
The test is now inside the directory.

jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
```diff
@@ -0,0 +1,133 @@
...
+    public static void mult15V2(Quaternion q, Vector3f v, Vector3f s) {
+        //15 mult, 15 add
+        float x = q.getX();
+        float y = q.getY();
+        float z = q.getZ();
+        float w = q.getW();
+        
+        //v + 2*q.xyz cross (q.xyz cross v + w*v )
+        //q.xyz x v.xyz + w*v
+        float vx = y*v.z - z*v.y + w*v.x;
```
shadowislord: Isn't this the exact algorithm that was implemented above? Shouldn't this compare against the old algorithm instead?

jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
```diff
@@ -0,0 +1,133 @@
...
+            
+            quat.mult(vect, store60);
+            mult15V2(quat, vect, store15);
+    
+            //eps stands for the epsillon in the approximateEquals method
+            //set your desired eps
+            
+            //final float eps = 0.00001f; 
+            //final float eps = 0.001f; 
+            final float eps = 0.08f;
```
shadowislord: Is it intended to use such a high epsilon value? Is it failing on the default value in `FastMath.approximateEquals`?

jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
TehLeo: Yes, sorry for this. First I made a test and uploaded it to pastebin. Then I forgot I have to change this since the new build will use the new method. I'll fix it now.

jme3-core/src/test/java/com/jme3/math/QuatMultTest.java
TehLeo: Yes, rotating a vector by quaternion requires such "high" epsilon value. 
As I wrote above, eg. if we want to rotate vector (-3361.9124, -1759.1035, 357.90723) by a quaternion (-0.08545344, 0.2544827, -0.93980426, -0.2114336):

We get a results:
New method      : (3806.7993, 0.004043579, 182.4147)
Jme old method: (3806.7993, 0.004163742, 182.41466), 
SageMath:          (3806.79915617854, 0.00397302788272214, 182.414682736066)

From these three results(mainly the y-coordinate in this example) and taking SageMath output as a reference, the new method would need quite a high epsilon to be considered equal. In this example eps = 0.046f for the old method and eps = 0.018f for the new.

pspeed42:
My concern in all of this was less whether we were getting the same values as the existing Quaternion and more about whether we were potentially introducing jitter that wasn't there before.  Quaternions are accumulated and used to position every object in the scene.  Any fluctuations out reasonably far from origin will likely be noticed as a lack of smoothness.

So, as a test, I used the following methodology...

First, I created a double-based straight cos/sin based series of radials.  3600 radials from 0 to 2 PI at a radius of 1000.  I measured the distance between each successive point.  For perfect math, this distance should always be the same.  Indeed, the 'pure' version varies almost not at all... I calculate the max and min distances and take the difference: 2.0237145292867353E-12.

Then I ran the same test using JME's quaternion.  Here we already have a small disadvantage because we're using float.  I do the distance calculation in double, though.  The variation in this case (max - min) is: 8.137615737373594E-4

Worse, of course, but at a thousand meters from origin it is probably not noticeable.

Then I ran with the fast-mult approach.  The variation in this case is definitely worse:
0.0010742516680835035

...it no longer even needs to render in scientific notation.  This was enough that I was going to sit on it overnight again and think about it.  Then I decided to graph the datasets to see where the jitter actually is.  After all, if the variations are at least consistent locally then it doesn't matter... no jitter in that case.

I loaded up each set of deltas in open office calc and made some charts.
http://i.imgur.com/YiimsbP.png

When the scale is made the same between each chart, the 'pure' version shows no variation at all.  (Before that, it's small variations followed the same rough pattern of the other charts.)

This tells a slightly better story as the last two charts look very similar.  The third chart has some spikey outliers that are a little troubling.  I'm not sure it matters though as ultimately even the JME version is pretty chaotic at this scale.

I have a double version of Quaternion in my SimMath package that does the same calculation as JME.  I may convert it to the faster mult and do some additional jitter tests.  I use that for my physics engine so I need more accuracy there but I'd also like the speed-up.

So far I'm kind of optimistic about applying this patch but I thought I'd present my evidence in case it bothers someone else more than it bothers me.

I'll include my test just in case someone spots a flaw.
```
import com.jme3.math.*;

public class TestQuaternion {

    public static Vector3f fastMult( Quaternion q, Vector3f v ) {
        float x = q.getX();
        float y = q.getY();
        float z = q.getZ();
        float w = q.getW();
         
        float vx = y * v.z - z * v.y;
        float vy = z * v.x - x * v.z;
        float vz = x * v.y - y * v.x;
        vx += vx; vy += vy; vz += vz;
        
        Vector3f store = new Vector3f();
        store.x = v.x + w * vx + y * vz - z * vy;
        store.y = v.y + w * vy + z * vx - x * vz;
        store.z = v.z + w * vz + x * vy - y * vx;
        
        return store;
    }            

    public static void main( String... args ) {
        
        int radialCount = 3600;
        double maxRadians = Math.PI * 2;
        double radius = 1000;
        double[][] baseline = new double[radialCount][2];
 
        System.out.println("Baseline test with double-base trig:");
        double min = 100000;
        double max = 0;
        for( int i = 0; i < radialCount; i++ ) {
            double rads = i * maxRadians / radialCount;
            baseline[i][0] = Math.cos(rads) * radius; 
            baseline[i][1] = Math.sin(rads) * radius;
            
            if( i > 0 ) {
                double xd = baseline[i][0] - baseline[i-1][0];   
                double yd = baseline[i][1] - baseline[i-1][1];
                double d = Math.sqrt(xd * xd + yd * yd);
                //System.out.println("[" + baseline[i][0] + ", " + baseline[i][1] + "] dist:" + d);
                //System.out.println(d);
                max = Math.max(d, max);
                min = Math.min(d, min);
            } 
        }
        System.out.println("Variation:" + (max - min) + "  Max delta:" + max + "  min delta:" + min); 
 
        System.out.println("Testing JME quaternion:");
        Quaternion quat = new Quaternion();       
        double[][] jme = new double[radialCount][2];
        min = 100000;
        max = 0;
        for( int i = 0; i < radialCount; i++ ) {
            float rads = (float)(i * maxRadians / radialCount);
            Vector3f v = new Vector3f((float)radius, 0, 0);
            quat.fromAngles(0, 0, rads);
            v = quat.mult(v);
            jme[i][0] = v.x;
            jme[i][1] = v.y;
            
            if( i > 0 ) {
                double xd = jme[i][0] - jme[i-1][0];   
                double yd = jme[i][1] - jme[i-1][1];
                double d = Math.sqrt(xd * xd + yd * yd);
                //System.out.println("[" + jme[i][0] + ", " + jme[i][1] + "] dist:" + d);
                //System.out.println(d);
                max = Math.max(d, max);
                min = Math.min(d, min);
            } 
        }
        System.out.println("Variation:" + (max - min) + "  Max delta:" + max + "  min delta:" + min); 

        System.out.println("Testing fast quaternion:");
        double[][] fast = new double[radialCount][2]; 
        min = 100000;
        max = 0;
        for( int i = 0; i < radialCount; i++ ) {
            float rads = (float)(i * maxRadians / radialCount);
            Vector3f v = new Vector3f((float)radius, 0, 0);
            quat.fromAngles(0, 0, rads);
            v = fastMult(quat, v);
            fast[i][0] = v.x;
            fast[i][1] = v.y;
            
            if( i > 0 ) {
                double xd = fast[i][0] - fast[i-1][0];   
                double yd = fast[i][1] - fast[i-1][1];
                double d = Math.sqrt(xd * xd + yd * yd);
                //System.out.println("[" + fast[i][0] + ", " + fast[i][1] + "] dist:" + d);
                //System.out.println(d);
                max = Math.max(d, max);
                min = Math.min(d, min);
            } 
        }
        System.out.println("Variation:" + (max - min) + "  Max delta:" + max + "  min delta:" + min); 
           
    }
}
```

pspeed42:
Just adding more information.  I did further testing with my double-based quaternions and the good news (at least for me) is that the math is still pretty accurate in that case.  Jitter errors are at a similar number of decimal places out (11 or 12).  It doesn't really help the other case but it's maybe useful information.

As an experiment, I also tried doing the float-in/out based calculations with doubles internally.  ie: promote the floats to double, calculate, cast doubles back to float.  In a microbenchmark (on desktop at least) it's still faster than JME's built in approach.  The bad news is that the jitter is just as bad as the full float version.  So it's not really an issue with the calculations (further confirmed by the double test) but a combination of float-errors in the source data + float errors in the answers being exaggerated by the new math.

As an aside, I added loop timings to the above test using nano timer.  In many ways, this is a more accurate timing test than a more traditional targeted microbenchmark as it includes more than just the quat calculation (which in general is a teeny tiny part of a JME update loop).  If I up the iteration count to 36,000 radials then the timing differences are almost at the level of measurement error.  The largest gain I've seen is from JME:10.743824 ms to Fast:10.275049 ms.  And again, that's for 36,000 quaternion multiply calculations.  Every now and then through random JVM fluctuations that plagued microbenchmarks everywhere, the numbers even reverse.  (For example, my last run even has a JME:12.402135 ms to Fast:12.605513 ms)

The jitter does worry me a bit and I don't think it's entirely unfounded.  For example, imagine you had a blue cube slowly orbiting a planet at 1000 meters and the camera is positioned right next to the cube so that the 1 meter cube fills the screen.  Error rate of the vertex transformations would be at the third decimal place.  On a 1900x1200 screen, that's slightly better than a pixel's worth of jitter that wouldn't have been there with the JME-stock way.

Will it be noticeable?  Maybe.  Is the chance that it is worth the small gains?  I dunno.  I'll let others decide.  Maybe my methodology is flawed.

empirephoenix:
I would say 1km distances are sure within jme's scope without using any intelligent system. Eg many map based shooters simply have their map loaded, at that size a 1px jitter at the end of the map would be problamatic, as the camera would jitter.

empirephoenix:
Closed due to unclear state of useability, and no further involvement by the creator or someone else

PR #375, opened 2015-11-07
jogl: some fixes about mouse handler

davidB:
* Y hotspot is top in jogl, and bottom in jme
* in jme when mouse is invisible it should be confined into the window (on mac it could go out of the window)
* when window doesn't has the focus the mouse should not behave like when has the focus like 
  * trying to re-center
  * stay invisible


davidB:
@gouessej can you review this PR ?

gouessej:
Hi. I'm currently studying your changes...

gouessej:
It's ok for me.

davidB:
Merci

PR #379, opened 2015-11-11
jogl: use pixel unit for window's dimension (fix issue on MaxOS X)

davidB:


gouessej:
Shouldn't it be done for AWT too?

davidB:
do you mean for mousePressed and mouseReleased ? or for AwtMouseInput
(that as no jogl dependency) ?

davidB:
With this last change, I've got a working version of JME3-JFX with jogl. The last point, is that I need to access the GLWindow to setup Display. To do is I create my class in the com.jme3.jogl package. WDYT to provide a public accessor in JoglNewtAbstractDisplay to canvas ?

gouessej:
JoglAbstractDisplay still uses an AWT GLCanvas, it is probably affected by the bug. The AWT input should be overridden to use the same fix.

davidB:
but the fix use method 'getSurfaceXxxx' that is not available in the
AwtMouseInput (it uses an AWT Component regardless of AWT GLCanvas)

gouessej:
You can override it locally here or pass an instance of a class extending AwtMouseInput:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-jogl/src/main/java/com/jme3/system/jogl/JoglAbstractDisplay.java#L162

davidB:
For joglAbstractDisplay, I thouh about 3 alternatives:
1. use a class extending AwtMouse, lot of copy/paste (3 method of several lines only to change `component.getHeight` to `((GLCanvas)component).getSurfaceHeight`)
2. use a class extending GLCanvas and override `getHeight` and `getWidth` (by `getSurfaceHeight`,...), I'm not fan to change the behavior, less code than 1.
3. Maybe configure canvas with
```
canvas.setSurfaceScale(new float[]{ScalableSurface.IDENTITY_PIXELSCALE, ScalableSurface.IDENTITY_PIXELSCALE});
```
(like I read on jogl forum).

Last point, how can I test it (for AWT) ?

gouessej:
You can use JoglDisplay instead of JoglNewtDisplay to test it. I think that it requires a deeper code redesign to move the code common to the current AWT input system and the JOGL AWT specific input system into some abstract classes to avoid code duplication. Both concrete implementations would have to override getSurfaceHeight and getSurfaceWidth, the plain AWT one would just call getHeight and getWidth. Do you see what I mean? It mustn't impact other backends using AWT but without JOGL.

davidB:
By using only  jME 3.1 API (without fork), we can't use AWT Display. So why keeping both Display  (NewT and AWT) ?
eg. to use JoglDisplay I have to change JmeDesktopSystem.newContextJogl(...)

PS: I'll not do the redesign (out-of my skill and goals).

gouessej:
Ok a decision must be made about that, you're right. I can accept your pull request but please can you reproduce the bug with AWT? I received a Mac Mini but it doesn't work exactly as I expected :s

davidB:
I'm not on mac today, I'll create a simple sample to test (I got the issue
when I adapt JME3-JFX).

gouessej:
I'll have to add a warning into the AWT code path or to drop it later. Thanks.

PR #381, opened 2015-11-14
Interpolate particle positions

TripleSnail:
Currently there is a problem with ParticleEmitter. It doesn't take into account the the distance it has travelled between updates. Or at least this is my interpretation - I didn't invent this solution. methusalah did. Please see the following threads for more information:

http://hub.jmonkeyengine.org/t/interpolation-of-particle-spawning-point/30385/9
http://hub.jmonkeyengine.org/t/interpolation-of-particle-spawning-point-when-emitter-has-moved/30400

The following images are from my game. FPS is 60 and the emitter emits 2000 particles per second. Projectile's linear speed is 200 world units / s (it uses RigidBodyControl).

Without interpolation
Without interpolation

With interpolation
With interpolation

### DIFF
--- a/jme3-core/src/main/java/com/jme3/effect/ParticleEmitter.java
+++ b/jme3-core/src/main/java/com/jme3/effect/ParticleEmitter.java
@@ -1021,7 +1021,7 @@ public class ParticleEmitter extends Geometry {
             Particle p = emitParticle(min, max);
             if (p != null){
                 p.life -= tpf;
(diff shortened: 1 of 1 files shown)


Nehon:
That's interesting. I totally missed the post on the forum.
Looks a lot better indeed.

Nehon:
Semms good to me @shadowislord do we merge it?

PR #385, opened 2015-11-21
Removed useless Quaternion from AbstractPhysicsControl

NemesisMate:
The private **tmp_inverseWorldRotation** quaternion is only used to make calculations that are completely lost and, thus, unnecessary.


shadowislord:
Its not true that its unused, because in the last line:
```tmp_inverseWorldRotation.set(spatial.getParent().getWorldRotation()).inverseLocal().mult(localRotationQuat, localRotationQuat);```
The `mult()` call actually writes the result to `localRotationQuat`, which is used.

NemesisMate:
It's true -O-, I can't believe I missed it.

PR #391, opened 2015-12-18
Fixed BetterCharacterControl "flickering" when not moving.

NemesisMate:
Sometimes, when the BetterCharacterControl is not moving, it flickers due it physics being always active. The reason for it being always active is that on the prePhysicsTick method it always set the rigidBody's linear velocity (and the method PhysicsRigidBody.setLinearVelocity(Vector3f) reactivates the physics each time is called).

The fix consist on just comparing if the current velocity and the setting one aren't the same and, thus, not setting it (using ZERO_TOLERANCE instead 0 to best results).


NemesisMate:
Link to the post talking about it: http://hub.jmonkeyengine.org/t/fixed-bettercharactercontrol-flickering-when-not-moving/34650/2?u=nemesismate

PR #398, opened 2016-01-02
Branch v3.1 alpha1

lightbringer:
As the master head failed to run for me, I branched of the 3.1.alpha1 tag. Sorry for the clutter in the diff, but I ran the classes I edited through my code cleaner/formatter and many indentations, alphabetical/qualifier method ordering and missing final tags were changed. 

This PR fixes two things:
1. Fixing issue 394, when specifying the Serializer class through the @Serializable annotation. The edit took place in  public static SerializerRegistration registerClass( Class<?> cls, boolean failOnMiss ) in Serializer.java
and now correctly uses reflection to instantiates the Serializer rather than looking up the Serializer for the SerializerClass.
2. It adds a new feature: I needed to send Savables larger than 32K over a message,the relevant changes are in the send/broadcast methods in DefaultClient and DefaultServer. I also added a new method to their respective interface to configure the buffer size dynamically.

This also fixes an issue that the SavableSerializer had, the ByteBuffer..Streams did not implement the Output/InputBufferStream contract correctly. This only showed when either of them is wrapped in a Buffered..Stream and buffers larger than 32K are read/written. 

If you decide to discard the >32K buffers feature for DefaultClient/Server, I recommend to use the fixed SavableSerializer anyhow.

### DIFF
--- a/jme3-networking/src/main/java/com/jme3/network/Client.java
+++ b/jme3-networking/src/main/java/com/jme3/network/Client.java
@@ -31,8 +31,9 @@
  */
 package com.jme3.network;
 
-import com.jme3.network.service.ClientServiceManager;
+import java.nio.BufferOverflowException;
 
+import com.jme3.network.service.ClientServiceManager;
 
 /**
  *  Represents a remote connection to a server that can be used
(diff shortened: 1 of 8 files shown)


pspeed42:
For sending data larger than 32k you should break your message up into
smaller chunks.  Since Serializer has to work with ByteBuffers there is no
real way to make them self-expanding so a fixed size has to be chosen.
Already 32k is pretty larger considering every outbound message has to
create a full size buffer.  Anyway, we won't be applying that patch for
sure.

We can't apply the rest of them either because of all of the reformatting.
So you will have to keep using your branch until I get around to fixing
them myself I guess.

pspeed42:
Also, "master head doesn't run for me" isn't a particularly good reason to submit an incompatible patch... since it runs for a large number of other people then it must be a fixable problem.  Thank you for the effort but it's frustrating that all of this work went into something unusable. :-/

lightbringer:
Well, it's up to you. I however wanted to make the network layer as transparent as possible, not dealing with breaking up messages in my app code, and the methods accepting other ByteBuffer were already there. I'll create a a new Pull request for the master branch with just the fixes then.

Issue #394, opened 2015-12-29
Specifying Serializer through @Serializable


lightbringer:
Trying to specify the Serializer for a class annotated with `@Serializable` though the "serializer " attribute does not work:
The class is correctly fetched in Serializer.java:224 from the annotation, but the succeeding call to `getSerializer` in line 228 is forwarded through line 316 to `getSerializerRegistration`, effectively looking for a Serializer for the specified Serializer class. Since that class isn't registered, the method falls through to the end (`failOnerror` is set to false by `getSerializer`) and returns `fieldSerializer`, which is actually returned back to the original `registerClass` call for the annotated class.

Workaround:
specifying an instance of the desired Serializer upon registering the class:

```
@Serializable( serializer = MySerializer.class )
class Foo{
}
```

Does not work:
```
Serializer.registerClass( Foo.class );
```
Works:
```
Serializer.registerClass( Foo.class, new MySerializer() );
```

pspeed42:
The logic here is weird.  It's line 228 that seems bizarre to me.  Why
would anyone want to look up the serializer for the serializer instead of
just instantiating it.  Very strange.

PR #399, opened 2016-01-02
- Fix for issue #394. 

lightbringer:
This is basically a re-submit of PR #398 with just the fixes for the master branch. It fixes issue #394 by instantiating class supplied the by @Serializable for a Serializer instead of looking for a SerializerRegistration for the Serializer class in registerClass. 

The second file, SavableSerializer, had a minor thing in SavableSerializer.ByteBufferInputStream: it did not implement the OutputStream contract contract completely correct. Hence it corrupts the stream when wrapped into a BufferedInputstream. This doesn't affect jMonkey's current master branch but broke things I changed in my local branch.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Node.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Node.java
@@ -687,6 +687,7 @@ public class Node extends Spatial {
 //            childClone.parent = nodeClone;
 //            nodeClone.children.add(childClone);
 //        }
+        nodeClone.invalidateUpdateList();
         return nodeClone;
     }
 
--- a/jme3-networking/src/main/java/com/jme3/network/serializing/Serializer.java
+++ b/jme3-networking/src/main/java/com/jme3/network/serializing/Serializer.java
@@ -218,29 +218,47 @@ public abstract class Serializer {
      */
     @SuppressWarnings("unchecked")
     public static SerializerRegistration registerClass(Class cls, boolean failOnMiss) {
-        if (cls.isAnnotationPresent(Serializable.class)) {
-            Serializable serializable = (Serializable)cls.getAnnotation(Serializable.class);
+        if (cls.isAnnotationPresent( Serializable.class )) {
+            final Serializable serializable = (Serializable) cls.getAnnotation( Serializable.class );
 
-            Class serializerClass = serializable.serializer();
+            final Class<? extends Serializer> serializerClass = serializable.serializer();
             short classId = serializable.id();
-            if (classId == 0) classId = nextId();
+            if (classId == 0) {
+                classId = nextId();
+            }
+            Serializer serializer = null;
 
-            Serializer serializer = getSerializer(serializerClass, false);
+            try {
+                serializer = serializerClass.newInstance();
+            }
+            catch (InstantiationException | IllegalAccessException e) {
+                if (failOnMiss) {
+                    throw new IllegalStateException( e );
+                }
+                else {
+                    log.log( Level.WARNING, "Failed to instantiate serializer (" + serializerClass.getName() + ") for class: " + cls.getName() + " with error "
+                                    + e.getMessage() + ". Falling back to FieldSerializer" );
+                }
+            }
 
-            if (serializer == null) serializer = fieldSerializer;
+            if (serializer == null) {
+                serializer = fieldSerializer;
+            }
 
-            SerializerRegistration existingReg = getExactSerializerRegistration(cls);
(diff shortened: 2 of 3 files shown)


pspeed42:
Thanks for redoing it.  I'll take a look at this later tonight and try to apply it.

shadowislord:
@pspeed42 I think you have this issue fixed on master already, right? Can we close this then?

pspeed42:
I guess the original PR was strangely submitted directly against the alpha
branch instead of master.  If I applied it then maybe it was applied there?

I don't know.  I can't really apply this PR as written because it includes
three different unrelated commits... one of which is to Node.java which
isn't even in the same package.

pspeed42:
Also, it's 1000x easier to apply a patch if you don't reformat unrelated
lines of code.

lightbringer:
Something weird happened here. The update in Node.java was done for another issue, which to my knowledge has been applied elsewhere (PR #403 ). No idea why this pops in this thread. I blame git for being super confusing to me.

Anyway, I havent't touched the original issue here for a while. But this PR should't have any lines besides a few. The original PR was a bit messy, true.

I remember there also being another problem with the SavableSerializer I found after submitting this: you can't put more than one Savable in a ByteBuffer, because the Importer will suck the Buffer dry no matter if it belongs to object it's importing. Has that been addressed recently or is that by design?

pspeed42:
I think mostly people don't send the scene graph over the wire and so it
never comes up.

As to the other, even for the one class I'm willing to apply, there were
about a dozen lines changes just because you applied a different
formatting.  It makes it harder to see what the actual change was, personal
preference aside.  Seemed to me it could have been done in one if-block and
I could have just clicked "accept" with little to no extra interpretation
on my part.

lightbringer:
The actual fix is replacing line 228 with a try/catch block (Though my IDE seems to have ignore my wish to make this Java 6 compliant and converted it to a multi-catch). 

Line 236 removes an unnecessary object instantiation.

The rest is just auto-formatting screwup. Is their a style file for Eclipse or something similar somewhere in the repo? Haven't found anything.

The changes SavableSerializer.java aren't solving the whole problem, as I commented on above.

Comment on sending SceneGraphs over the wire: I think the ability to clone the scene graph is one of the key advantages of JMonkey when developing AI methods that use a forward-model, i.e. agents that want to predict the outcome of an action by cloning the graph, invoking the action on the copy, and observing the resulting world state, without implementing a custom parallel scene graph. (I additionally send these to number crunching cluster nodes). However, this makes e.g.Unity unusable for me. Who thought "making the SceneGraph a singleton" is a brilliant idea?

empirephoenix:
You should not use the scenegraph for logic, only for visuals, you are actually doing an antipattern here. (be it in unity or any other engine)

JME just does not force you do tdo stuff some specific way, so you can do this, if you are sure this is the best way.

pspeed42:
re:"The rest is just auto-formatting screwup."

Step 1: turn off auto-formatting.
Step 2: profit

...at least when working on an open source project.  Else it's like
rearranging everyone's kitchen every time you walk through.

lightbringer:
>You should not use the scenegraph for logic, only for visuals, you are actually doing an antipattern here. 

Not entirely sure how to read your comment, and what you mean by "logic" or how this is an anti pattern. Besides, my application has no visuals - it runs completely headless. I'm not writing a game here ;-)
The point of a forward-model is to predict the outcome of an action that you don't know, and you want to replicate the actual environment as closely as possible. Of course, I could implement something that looks like the scene graph but isn't, but I don't see the point of that. See https://en.wikipedia.org/wiki/Internal_model_%28motor_control%29

>...at least when working on an open source project. 

Yeah, discussing pro&con of automatic code styles is like entering the rabbit hole. At one point in my life, I worked on industry projects with 200+ participants on the code base, and every time there was some poorly formatted code checked in, the person to blame had to throw a round. Because that If-Statement there was supposed to cover the whole block. What block? Oh ...
But hey, it's your kitchen ;-)

pspeed42:
If you code on enough open source projects, eventually you learn to easily
type in a lot of different styles.  Else your patches never get applied.

pspeed42:
So, just so we're crystal clear on why this PR hasn't been applied... it is
because it is unacceptable in its current form.  Not because of the
formatting but because it has erroneous changes in it. (I'm not even
allowed to apply it because it has merge conflicts.)  Fixing the formatting
will just lessen the time I have to waste on it next time.  Recommend
trimming it to just the serializer changes.  Do the other savable stuff as
a different PR if you like.

PR #406, opened 2016-02-07
Update Quaternion.java individual angle constructor

Triangle345:
added constructor to simplify creation of Quaternion by using individual angle parameters.


pspeed42:
Use new Quaternion().fromAngles().  A three-float-arg constructor is too ambiguous on its own.

PR #407, opened 2016-02-09
Changed TGALoader

JavaSaBr:
Added support of overriding and added support the resize image on loading.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/texture/plugins/TGALoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/texture/plugins/TGALoader.java
@@ -77,6 +77,16 @@ public class TGALoader implements AssetLoader {
     // 11 - run-length encoded, black and white image
     public static final int TYPE_BLACKANDWHITE_RLE = 11;
 
+    protected static TGALoader loader;
+
+    public static void setLoader(TGALoader loader) {
+        TGALoader.loader = loader;
+    }
+
+    public TGALoader() {
+        setLoader(this);
+    }
+
     public Object load(AssetInfo info) throws IOException {
 
         if (!(info.getKey() instanceof TextureKey)) {
@@ -87,7 +97,7 @@ public class TGALoader implements AssetLoader {
         InputStream in = null;
         try {
             in = info.openStream();
-            Image img = load(in, flip);
+            Image img = loadImage(in, flip);
             return img;
         } finally {
             if (in != null) {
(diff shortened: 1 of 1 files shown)


DannyJo:
Slightly confused by this implementation, but is it right that in order to use this new feature you have to extend the TGALoader and override the getNewWidth() and getNewHeight() methods in order to resize the texture? If so, you are essentially resizing every texture that's being loaded by this loader regardless if it's smaller or larger.

On another note I thought the Graphics api was notoriously bad at resizing images when resizing by a large amount of pixels (quality wise I mean).

What is your business case for this if I may ask? Why could you not just resize the texture before shipping it with your application in the first place?

JavaSaBr:
I did it to be able to implement centralized configuration of quality of textures in my game.

DannyJo:
So this means you will be shipping larger textures and then downscaling in your code at runtime/load time?

Does that not also mean you will be shipping several more megabytes of data for the client to download than what your application will be using?

Or are you simply saying that you ship large textures with the application and then the users can select what quality they want in the application?

JavaSaBr:
I'm using it in offline client. It gives me the opportunity not to have storage options of different textures. I use Scalar in my loader to scale, which is able to use different algorithms.

jme3-core/src/plugins/java/com/jme3/texture/plugins/TGALoader.java
```diff
@@ -38,6 +38,12 @@
 import com.jme3.texture.Image;
 import com.jme3.texture.Image.Format;
 import com.jme3.util.BufferUtils;
+
+import java.awt.*;
+import java.awt.image.BufferedImage;
+import java.awt.image.DataBuffer;
+import java.awt.image.DataBufferByte;
+import java.awt.image.Raster;
```
shadowislord: `jme3-core` should not reference the `java.awt` package, because it is not available on Android. 
`java.awt` is allowed to be referenced in desktop only projects, like `jme3-desktop` and `jme3-lwjgl`.

jme3-core/src/plugins/java/com/jme3/texture/plugins/TGALoader.java
```diff
@@ -71,7 +77,18 @@
     // 11 - run-length encoded, black and white image
     public static final int TYPE_BLACKANDWHITE_RLE = 11;
 
+    protected static TGALoader loader;
```
shadowislord: Loaders are typically managed via `AssetManager`, there's no need to have a static reference in any `AssetLoader`.

jme3-core/src/plugins/java/com/jme3/texture/plugins/TGALoader.java
```diff
@@ -463,29 +502,139 @@ public static Image load(InputStream in, boolean flip) throws IOException {
...
+     * Change this image if need.
+     *
+     * @param width current width of image
+     * @param height current height of image
+     * @param pixelDepth pixel depth of image
+     * @param format format of image.
+     * @param rawData data of image.
+     * @return result image.
+     */
+    protected Image resizeImage(int width, int height, final int pixelDepth, final Format format, byte[] rawData) {
```
shadowislord: Although resizing images automatically could be useful, I don't think this functionality should only be available in certain loaders. This is much better to be implemented as a `TextureProcessor` feature instead.

shadowislord: Also, I'll mention that personally I don't think resizing images at run-time is a good idea, because it slows down loading for people who want lower texture quality. Further, resizing is not possible to do for compressed formats like DXT or ETC. A better option might be to use a mipmap level offset when specifying the texture in OGL, so that it ignores mipmap level 0 for example, effectively halving the texture resolution.

empirephoenix:
Rejecting this due to usage of awt classes.

A JME wide logic for resizing images at load is something I would like to see however, as shipping all textures with half resolution would add a heavy +40% in download size. I think some global implementation based on the mippmaps might be interesting here.

2lost4u:
In my opinion TGALoader should take care of resize of the image, at the algorithm not as a extension, that is if i understood well the problem

PR #408, opened 2016-02-09
Added support of overriding

JavaSaBr:
Added support of overriding for class EnvironmentCamera

### DIFF
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentCamera.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentCamera.java
@@ -31,10 +31,10 @@
  */
 package com.jme3.environment;
 
-import com.jme3.environment.generation.JobProgressListener;
-import com.jme3.environment.util.EnvMapUtils;
 import com.jme3.app.Application;
 import com.jme3.app.state.BaseAppState;
+import com.jme3.environment.generation.JobProgressListener;
+import com.jme3.environment.util.EnvMapUtils;
 import com.jme3.light.LightProbe;
 import com.jme3.math.ColorRGBA;
 import com.jme3.math.Vector3f;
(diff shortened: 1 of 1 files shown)


Nehon:
There is a lot of noise in this change because I guess of your formatting scheme.
Several questions : 
Why do you need to override this class? What feature do you miss?
Why the early returns instead of the more classic "if"

JavaSaBr:
> Why do you need to override this class? What feature do you miss?
For my case:
```
@Override
    protected Camera createOffCamera(final int mapSize, final Vector3f worldPos, final Vector3f axisX, final Vector3f axisY, final Vector3f axisZ) {
        final Camera offCamera = super.createOffCamera(mapSize, worldPos, axisX, axisY, axisZ);
        offCamera.setFrustumPerspective(90f, 1f, 1, GraphicUtils.SKY_SPHERE_RADIUS);
        return offCamera;
    }
```

> Why the early returns instead of the more classic "if"
Anti practice for:
```
if(cond) {
			if(cond) {
				if (cond) {
					if (cond) {
						
					}
				}
			}
		}
```

Nehon:
1. createOffCamera is already protected. You can do that without the change. If you need any attribute from the class I'd rather create accessors than making them protected.
2. Not relevant in this case, there is only one condition... your change doesn't make it more readable...

JavaSaBr:
1. protected FINAL...

Nehon:
OH! duh. You should have started with that.
Ok I'll push a change in this regard.

pspeed42:
Early returns are a billions times more readable.  It's funny because people who are against early returns seem to have no problem with using "break" in a loop and it's exactly the same thing.  In C, there was a point to this structured dogma but in any language with a finally {} clause there is no point.

gouessej:
@pspeed42 It's a matter of taste but I like neither early returns nor break (I prefer picking a right loop condition) & continue. Using early returns can enhance the readability but abusing of returns in the whole body of a method drives it harder to understand (in my humble opinion). If you have only small methods, having several points of exit isn't a source of trouble but it's not the case if you have large methods.

I find structured programming helpful to make the proof of a program, not only with languages with no RAII (Resource Acquisition Is Initialization).

PR #412, opened 2016-02-14
Added SimpleAppState class to make the use of AppStates less cumbersome when accessing SimpleApplication's resources

john01dav:
This pull request includes a new class called SimpleAppState. It implements a feature that I have felt was lacking in this library since I learned how to use AppStates -- it allows for seamless interaction with all resources that your main class may have through an AppState by providing delegates for SimpleApplication's and Application's methods such as getAssetManager(). It also provides improved lifecycle management such as what is available already in BaseAppState to not force users to choose between the convenient method delegates and BaseAppState's lifecycle management. Lastly, it provides a method to enqueue() a Runnable rather than a Callable to allow those who are implementing games in Java 8 use simple enqueue lambda expressions such as `enqueue(() -> doSomething())` rather than `enqueue(() -> {doSomething(); return null;});`

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/state/SimpleAppState.java
+++ b/jme3-core/src/main/java/com/jme3/app/state/SimpleAppState.java
@@ -1,0 +1,349 @@
+/*
+ * Copyright (c) 2009-2012 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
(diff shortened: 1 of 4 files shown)


pspeed42:
This is fine for a utility you use yourself but this will not be included in the engine.  I appreciate the work you put into it but it has several problems that prevent it from being included.
1) it completely goes against BaseAppState which will eventually be THE base abstract app state.
2) in the process, it messes up the 'state' of the app state since an app state can be disabled and not yet initialized and those are two distinct states.  (For example, an app state that you want to attach but enable later.)
3) It does redundant processing in update().  The app state update() already won't be called until it's attached and enabled.
4) It confuses getState() that means two different things depending on the argument passed.
5) it has a ton of nearly useless methods that merely save a method call... and it's likely that half of these methods become totally broken when SimpleApplication goes away someday.  (setDisplayFps(), etc... versus even in your own class getApplication().setDisplayFps().)  These kinds of methods are 100% clutter.
6) It includes a runnable wrapper thing that would have been better as its own utility class.  (Surprised Java8 doesn't include it already for the reason you describe.)  (In fact, since the underlying could could easily handle both I've often wanted to add an enqueue(Runnable) right on application.)

In general, this class takes an "everything and the kitchen sink" approach which is the exact opposite of the direction we are trying to head.

john01dav:
Would it be better if I did the following:

- Fix the state management (ie. add more possible states)

- Rename one of the getState() methods

- Remove SimpleApplication delegations

- Move the RunnableWrapper class to SimpleApplication (probably in a separate pull request, in fact I'm going to go do this now)

As for BaseAppState, I'm not sure how I can fully integrate this with it (not that I think that is a bad idea) due to the generic type it provides and the enum to keep track of the state rather than set of booleans.

pspeed42:
The enum is just a bad idea to track two completely different boolean
states.  I cannot think of any changes that would make this palatable for
core that wouldn't end up making it look just like BaseAppState.

john01dav:
The main thing I want to contribute is something to provide easy access to the application class via generics. I don't think it is wise to add a generic type to an existing api class. Perhaps extending BaseAppState with GenericAppState to simply add the generic feature would be wise?

pspeed42:
But in this case it's providing nothing except some small code sugar.  It
doesn't even really provide type safety since none of the other classes
need this.  Plus, in the future when SimpleApplication is deprecated it
will be completely unncessary to cast to any special application class
except in the very most specific cases.

It seems like a lot of trouble just to avoid ((MyAppClass)getApplication()).

pspeed42:
Typo: "since none of the other classes need this" "since none of the other
classes use this"

PR #413, opened 2016-02-14
Added enqueue runnable

john01dav:
This commit adds a new enqueue() method to Application that accepts a Runnable object rather than a callable to allow for the following cleaner syntax when using Java 8 lambda expressions.

callable-only enqueue:
```
enqueue(() -> {
    doSomething();
    return null;
});
```
runnable enqueue:
`enqueue(() -> doSomething());`

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/Application.java
+++ b/jme3-core/src/main/java/com/jme3/app/Application.java
@@ -650,11 +650,27 @@ public class Application implements SystemListener {
      * Callables are executed right at the beginning of the main loop.
      * They are executed even if the application is currently paused
      * or out of focus.
+     * 
+     * @param callable The callable to run in the main jME3 thread
      */
     public <V> Future<V> enqueue(Callable<V> callable) {
         AppTask<V> task = new AppTask<V>(callable);
         taskQueue.add(task);
         return task;
+    }
+    
+    /**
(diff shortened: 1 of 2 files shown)


empirephoenix:
Hm yes, I'm for this :) would greatly reduce the bloat i have.

pspeed42:
Almost ready.  General convention is to have inner classes at the bottom of the class.  If you can make that change, apply it, and flatten the PR then I'll apply it.

Thanks. I've had this on my to-do list for a long time.  (Though my plans were more invasive because it bugs me a little that Runnables incur all of the Callable+Future overhead where Callables could easily have been wrapped in a Runnable and just have a Runnable queue..  It's a small thing in the big picture..)

empirephoenix:
Yep I kinda hat the callables as in around 100% of cases I do a return null.

pspeed42:
But once we have an enqueue(Runnable) then we can flip it over internally later.

The while( task = getTask ) { !canceled ) loop could easily just be a loop over runnables.  The runnable wrapping Callable could be smart enough not to run it if canceled.  Done.  :)

pspeed42:
By the way, that's why I LIKE that RunnableWrapper is internal to this class... easier to change later.

john01dav:
I have made the requested changes.

pspeed42:
Thanks!

PR #414, opened 2016-02-16
Fixed indirect light extraction in Material

BorealFeast:


Nehon:
... how did I miss this?
....how did it work???

Thanks for the fix

PR #415, opened 2016-02-17
Fixes #360 : Material now serializes name

### LINKED ISSUE #360: Material does not serialize the "name" property
The `Material` class should serialize the name property into the `Savable`.

MeFisto94:
Yet another simple fix, but hey, atleast it's a fix :P


MeFisto94:
Oh and I don't know if we should increase SAVABLE_VERSION?
I think the usual behavior would be to increase it and only read the name if it's >= 3, though since we have specified "null" as default value, we have the same behavior, no matter if the patch is inside or not

shadowislord:
No need to change `SAVABLE_VERSION`, name will simply be set as null as before if it was not written.

Issue #342, opened 2015-09-17
Make AudioNode constructors less confusing


shadowislord:
Currently we have constructors taking a name, name + boolean, or name + two booleans. There's no need for this complexity. Its best to just have a single constructor specifying the asset manager, the name of the audio file, and whether its buffered or streamed. The documentation should clearly specify what are the differences between buffered or streamed sources and when to use either.

E.g.

``` java
audioNode = new AudioNode(assetManager, "Music/Song123.ogg", Type.Streaming)
```

or

``` java
audioNode = new AudioNode(assetManager, "Sound/Misc/Hit.ogg", Type.Buffered)
```

Also the streamCache option should always be true, since it is unknown whether the user wants the sound to be looping, its best not to guess and assume it should support looping always.

MeFisto94:
Should the AudioType be added as a field to the AudioNode so one can use
```
audioNode.getType() == AudioType.Streaming
```

Also, should the underlying AudioKey be changed aswell or should it remain in the "two-boolean" style?
And the third thing: Should the Type be an inner-class enum such as the old Status or rather an AudioType.java external file?

MeFisto94:
I am reopening this issue because of https://hub.jmonkeyengine.org/t/solved-ogg-memory-usage/38547/3 and https://hub.jmonkeyengine.org/t/solved-ogg-memory-usage/38547/5

Maybe we should rename the enum to `Uncompressed` and `Compressed`, since it's buffered either way and not really streamed anymore (or say it's only streamed the first time, then it's buffered).
The Problem is though that we can't easily change the names without having another deprecation state again.

Does anyone know a good way to re-word the javadoc in order to clarify this?

@pspeed42

PR #417, opened 2016-02-19
Simplified the AudioNode constructors (#342)

MeFisto94:
Yeah, I just tried to implement #342.
I was uncertain on how the "jme-style" would be for that enum, so I took an inner-class enum, however moving it out it just a matter of two key presses :)

I did not remove the constructors but only deprecated them to not break existing code.
I was also unsure about deprecating ```AudioNode(AssetManager assetManager, String name)```, simply because it's a constructor where one doesn't have to think about types at all.

When merged/alpha 2 drops, I will add the Wiki accordingly :)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/audio/AudioNode.java
+++ b/jme3-core/src/main/java/com/jme3/audio/AudioNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2012, 2016 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


Nehon:
Looks good to me
@shadowislord what do you think?

jme3-core/src/main/java/com/jme3/audio/AudioNode.java
```diff
@@ -113,6 +114,26 @@
     }
 
     /**
+     * <code>Type</code> indicates how to retrieve the audio data.
+     * It replaced the old "stream" and "streamCache" parameters.
+     * It defines whether the whole file is read and buffered or
+     * if it is read gradually from disk.
+     */
+    public enum Type {
```
shadowislord: The type is a property of `AudioData`, not `AudioNode`. 
See for example, `AudioData.DataType` enum which has the same purpose. Perhaps it should just be better documented.

MeFisto94:
I don't know if you get notified when I push a commit so:
I think I fixed it appropriately now :)

Issue #81, opened 2014-03-24
Exception after adding Skybox


ghost:
Reference on googlecode: https://code.google.com/p/jmonkeyengine/issues/detail?id=544

Nehon:
Fixed this commit https://github.com/MeFisto94/jmonkeyengine/commit/1c12b68bb0623b66b68531662ec81f51f12dc47e

PR #418, opened 2016-02-20
Validate SkyboxWizard's Input (#81)

MeFisto94:
In Issue #81, we simply passed the User Input to the SkyFactory which threw an Exception  in multiple cases:
a) When a User didn't specify all Textures
b) When they weren't a Square.

This is now fixed so you can't even proceed in the Wizard with these faults.
For this, only small changes to the Panel were needed.

This is how it looks like: (This is only the Netbeans example)
validation


Nehon:
good thanks

Issue #135, opened 2014-05-11
SDK tabs naming simplification


slawekmikula:
Simplify tab naming e.g. 
"AssetPackBrowser Window" to "AssetPack Browser" 
"FilterExplorer Window" to "Filter Explorer"
"SceneExplorer Window" to "Scene Explorer"

It can save a bit of space in multiple tabs in tab-group.

Nehon:
Fixed with commit https://github.com/MeFisto94/jmonkeyengine/commit/e051936f9be015b338992b2b10a51f15d8cf6d36

PR #420, opened 2016-02-21
Simplify Naming of SDK Windows (#135)

MeFisto94:
Yeah, really simple.
There was no comment on that Issue since May '14, but since I personally dock every window, it was annoying to scroll down, just to find the SceneExplorer.

I also tried to add more descriptive hovers (They appear when you mouseover the Window Title).
Feel free to suggest better hovers, I'll change the PR accordingly then


Nehon:
Thnaks you

PR #421, opened 2016-02-21
Fix some FilterExplorer NPEs (and add "Edit Filter")

MeFisto94:
Hey Guys,
I guess you have noticed that I am flooding you with PR's :P 

While working on the SDK I noticed that when a Filter is outside of the Asset Path, every call trying to access the FPP will throw an NPE, so that you'd have to restart the SDK.
So I simply fix every possible source of a NPE.

```getFilterPostProcessor()``` is a bit verbose, but I think it's better than when simply "nothing" happens. 
I also added the Edit Filter. Funnily enough it was already implemented, it was only missing the String (and hence you had an empty context menu entry before).

Have fun and enjoy it ;)


MeFisto94:
Embarassing that the build failed :O I blame it on Netbeans Compile-on-Save :P 
I fixed it now, it was the order of annotations.

Ultimately we should decide on a format (In-Code-Keys vs Bundle.properties)
Personally I'd go for the latter since it seperates code and text/localization.

PR #422, opened 2016-02-23
Two small ShaderNodes fixes

MeFisto94:
So whilst playing around with MatDefs I found this two issues:
a) When you only have one technique, you couldn't click into the ComboBox without a crash.
The reason behind this is that you cannot select the same item itself, so closing the box will lead to selected == null.
We fix this by simply taking the first item. If there are multiple available techniques we also print a warning regarding this.

b) When you have an incomplete MatDef, we had a few NPE's which crashed the sdk and hence made switching to another MatDef impossible.
I don't know if Exceptions on a broken (and hence user edited) MatDef are intentional but since the other code checks for null's, I assume it was simply missing :) 

BTW: Example MatDef for b
```
Technique Test {
}
```


Nehon:
Thank you.

PR #423, opened 2016-02-23
sdk filters : replaced all Jme...Filter by a single JmeFilter class

Dokthar:
I saw several classes doing exactly the same job, so why not just make one.


shadowislord:
Less code is always better.

normen:
Most of these were kept separate to be able to add special parameter editors like sliders and whatnot but if/when that becomes the case they can still be split.

PR #426, opened 2016-02-26
Batch node improvements

InShadow:
- Clarified in the JavaDoc that CullHint is not used for batched geometries.
- Added collideWith method that takes care of ignoring batched geometries.
- Removed some commented and not used code.
- Changed signature of isBatch method to be public final - to allow users to check the children if they are batched geometries or not.
- Corrected spelling mistake for gatherGeometries method.


shadowislord:
Looks fine to me, except I don't understand why `isBatch()` was made public. In what situation would the user need to use `isBatch()`?

InShadow:
I think that the users could have a need to iterate over the children of a batch node and find out which of them are batched geometries and which are not.
The use case would probably be, do something special with batched or non-batched nodes only.

Nehon:
Alright let's merge it

PR #430, opened 2016-02-28
Improved ShaderNodes

MeFisto94:
* Fixed a NPE when trying to select a Node. Now you can properly remove/edit nodes
 * "smartLog" now compares the hash across different instances (Two Material Previews led to duplicate log messages)
 * Using smartLog for CompilationErrors now to not spam the log with 60 FPS

The Lookup failed because there was no Instance of ```MatDefNavigatorPanel```.
Now everything works as intended (You can remove Nodes, change their properties, or at least select them without a crash.)

When you produce a shader compilation error, you have the same error printed three times (wireframe (geometry), solid (color) and a third). This doesn't happen anymore, since the hash is now static.

However I am uncertain because of such an example:
When you have two instances, one constantly printing a and one constantly printing b. Then smartLog wouldn't work. Since we have three Renderers with the same Material this is very unlikely.

### DIFF
--- a/sdk/jme3-materialeditor/src/com/jme3/gde/materialdefinition/EditableMatDefFile.java
+++ b/sdk/jme3-materialeditor/src/com/jme3/gde/materialdefinition/EditableMatDefFile.java
@@ -186,7 +186,7 @@ public class EditableMatDefFile {
             return "";
         } catch (Exception e) {
             Exceptions.printStackTrace(e);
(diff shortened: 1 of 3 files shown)


Nehon:
Well there could be more than 3 renderers in the future, if we use MRT.
Basically we'll have several color preview.
Anyway, we'll see when we're there.
Thanks for the fixes

PR #434, opened 2016-03-01
Fixed the Issue where removing a Control didn't set the dirty flag

MeFisto94:
See http://hub.jmonkeyengine.org/t/dirty-flag-not-set-when-deleting-model-controls/35216
btw: Fix Imports removed some unused imports and the "fireSave" Override is lent from JmeLight.java


shadowislord:
That's a lot of duplicate code. Maybe create a common subclass where this functionality can be implemented? Like AbstractSceneExplorerControlNode

MeFisto94:
I hope it's okay that I removed some more duplicate code (the whole destroy(), isDestroyable() and the getAction)

shadowislord:
Thanks, this is much better.

PR #435, opened 2016-03-01
Bone animation bind transform bug

AlrikG:
The constructor doesn't "Only copies the name and bind pose from the original.". It takes the object references from the original which leads to shared bind transforms.

For example several same models are loaded and each model should have different e.g. arm sizes. If all models shares the same bind object references, all models have the same arm size instead of different ones.

A simple clone() makes a real copy and fixes this problem. Now its prossible to have different sized and bone animated arms.

Hint:

enabling UserControl disable bone animations, so this is not the way I want. I want bone affected by animations and customized bone transforms (bind transforms).

Forum link: http://hub.jmonkeyengine.org/t/bug-bone-animations-share-their-bind-transforms/35217


shadowislord:
If you make those changes you have to fix the comment too:
https://github.com/AlrikG/jmonkeyengine/blob/BoneAnimationBindTransformBug/jme3-core/src/main/java/com/jme3/animation/Bone.java#L140-L149

AlrikG:
Yes you are right. I fixed it

AlrikG:
I have added a more precise comment because the constructor does more than the comment said

PR #436, opened 2016-03-02
Update BufferUtils.java

### LINKED ISSUE #395: NullPointerException when using BufferUtils.destroyDirectBuffer
I sometimes get this NPE:

    Caused by: java.lang.NullPointerException: null
	at com.jme3.util.BufferUtils.destroyDirectBuffer(BufferUtils.java:1296) ~[jME3-core.jar:na]

The actual code line 1296:

    Object cleaner = cleanerMethod.invoke(toBeDestroyed);

The ```BufferUtils.cleanerMethod``` static field is not yet initialized when executing the statement l.1296 - hence the NPE.

Indeed, the ```loadCleanerMethods()``` method seems to be incorrect. It goes like this:

        // If its already true, exit, if not, set it to true.
        if (BufferUtils.loadedMethods.getAndSet(true)) {
            return;
        }
        // This could potentially be called many times if used from multiple
        // threads
        synchronized (BufferUtils.loadedMethods) { /* initialization of static fields */ }

Unfortunately, the guard clause ```BufferUtils.loadedMethods.getAndSet(true)``` is not enough to prevent a concurrent thread to get to the ```return``` statement before the ```synchronized``` init block has been processed...

TehLeo:
Fixes #395 

Summary: fixes calling destroyDirectBuffer from multiple threads

### DIFF
--- a/jme3-core/src/main/java/com/jme3/util/BufferUtils.java
+++ b/jme3-core/src/main/java/com/jme3/util/BufferUtils.java
@@ -51,7 +51,6 @@ import java.nio.IntBuffer;
 import java.nio.LongBuffer;
 import java.nio.ShortBuffer;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


empirephoenix:
If we are at this, could we not remove the harmony code? That stuff is not supported since 5 years, it's like also it is on 1.5 level if I see correctly?

PR #445, opened 2016-03-09
Fixed collision group listeners not being notified

Fadorico:
In native bullet, group collision listeners we're never notified when a collision occurred for a group. This fixes the issue.

See http://hub.jmonkeyengine.org/t/physicscollisiongrouplistener-possibly-broken-in-native-bullet/35277/7

### DIFF
--- a/jme3-bullet-native/src/native/cpp/jmeClasses.cpp
+++ b/jme3-bullet-native/src/native/cpp/jmeClasses.cpp
@@ -40,6 +40,7 @@ jclass jmeClasses::PhysicsSpace;
 jmethodID jmeClasses::PhysicsSpace_preTick;
 jmethodID jmeClasses::PhysicsSpace_postTick;
 jmethodID jmeClasses::PhysicsSpace_addCollisionEvent;
+jmethodID jmeClasses::PhysicsSpace_notifyCollisionGroupListeners;
 
 jclass jmeClasses::PhysicsGhostObject;
 jmethodID jmeClasses::PhysicsGhostObject_addOverlappingObject;
@@ -137,6 +138,7 @@ void jmeClasses::initJavaClasses(JNIEnv* env) {
     PhysicsSpace_preTick = env->GetMethodID(PhysicsSpace, "preTick_native", "(F)V");
     PhysicsSpace_postTick = env->GetMethodID(PhysicsSpace, "postTick_native", "(F)V");
     PhysicsSpace_addCollisionEvent = env->GetMethodID(PhysicsSpace, "addCollisionEvent_native","(Lcom/jme3/bullet/collision/PhysicsCollisionObject;Lcom/jme3/bullet/collision/PhysicsCollisionObject;J)V");
+    PhysicsSpace_notifyCollisionGroupListeners = env->GetMethodID(PhysicsSpace, "notifyCollisionGroupListeners_native","(Lcom/jme3/bullet/collision/PhysicsCollisionObject;Lcom/jme3/bullet/collision/PhysicsCollisionObject;)Z");
     if (env->ExceptionCheck()) {
         env->Throw(env->ExceptionOccurred());
         return;
--- a/jme3-bullet-native/src/native/cpp/jmeClasses.h
+++ b/jme3-bullet-native/src/native/cpp/jmeClasses.h
@@ -46,6 +46,7 @@ public:
     static jmethodID PhysicsSpace_addCollisionEvent;
     static jclass PhysicsGhostObject;
     static jmethodID PhysicsGhostObject_addOverlappingObject;
+    static jmethodID PhysicsSpace_notifyCollisionGroupListeners;
 
     static jclass Vector3f;
     static jmethodID Vector3f_set;
--- a/jme3-bullet-native/src/native/cpp/jmePhysicsSpace.cpp
+++ b/jme3-bullet-native/src/native/cpp/jmePhysicsSpace.cpp
@@ -187,8 +187,28 @@ void jmePhysicsSpace::createPhysicsSpace(jfloat minX, jfloat minY, jfloat minZ, 
                 jmeUserPointer *up0 = (jmeUserPointer*) co0 -> getUserPointer();
                 jmeUserPointer *up1 = (jmeUserPointer*) co1 -> getUserPointer();
                 if (up0 != NULL && up1 != NULL) {
-                    collides = (up0->group & up1->groups) != 0;
-                    collides = collides && (up1->group & up0->groups);
+                    collides = (up0->group & up1->groups) != 0 || (up1->group & up0->groups) != 0;
+                    
+                    if(collides){
+                        jmePhysicsSpace *dynamicsWorld = (jmePhysicsSpace *)up0->space;
+                        JNIEnv* env = dynamicsWorld->getEnv();
+                        jobject javaPhysicsSpace = env->NewLocalRef(dynamicsWorld->getJavaPhysicsSpace());
+                        jobject javaCollisionObject0 = env->NewLocalRef(up0->javaCollisionObject);
+                        jobject javaCollisionObject1 = env->NewLocalRef(up1->javaCollisionObject);
+                        
+                        jboolean notifyResult = env->CallBooleanMethod(javaPhysicsSpace, jmeClasses::PhysicsSpace_notifyCollisionGroupListeners, javaCollisionObject0, javaCollisionObject1);
+                        
+                        env->DeleteLocalRef(javaPhysicsSpace);
+                        env->DeleteLocalRef(javaCollisionObject0);
+                        env->DeleteLocalRef(javaCollisionObject1);
+
+                        if (env->ExceptionCheck()) {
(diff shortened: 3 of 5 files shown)


shadowislord:
@normen 
Can you take a quick look at this? 
I thought collision groups were handled in the broadphase so there shouldn't be a need to check them again.

normen:
Yeah, it should be registered via setOverlapFilterCallback (see commented out code in PhysicsSpace.java), this needs to be done partly in the native com_jme3_bullet_PhysicsSpace.cpp, the callback itself can be routed back to java.

Fadorico:
So if I understand correctly the first if statement 

`if((node.getCollideWithGroups() & node1.getCollisionGroup()) > 0 || (node1.getCollideWithGroups() & node.getCollisionGroup()) > 0){`

should not be there because we know for sure at that point that those objects are supposed to collide?

normen:
Sorry, a few things get mixed up here. The check
if ((collisionObject.getCollideWithGroups() & collisionObject1.getCollisionGroup()) > 0
                            || (collisionObject1.getCollideWithGroups() & collisionObject.getCollisionGroup()) > 0)
needs to be added in the native jmePhysicsSpace.cpp, this avoids further checks of the collision right in the broadphase. Additionally the PhysicsCollisionGroupListeners need to be called in that method as well because they can change the result right in that method, also avoiding further checks if they so wish.

Fadorico:
The last commit takes into account what normen said and checks the group collisions in the broadphase (and modifies the result if needed). Here is the test case that was used for testing:

```
package test;

import com.jme3.bullet.BulletAppState;
import com.jme3.bullet.PhysicsSpace;
import com.jme3.bullet.collision.PhysicsCollisionGroupListener;
import com.jme3.bullet.collision.PhysicsCollisionObject;
import com.jme3.bullet.collision.shapes.MeshCollisionShape;
import com.jme3.bullet.collision.shapes.SphereCollisionShape;
import com.jme3.bullet.control.RigidBodyControl;
import com.jme3.renderer.RenderManager;
import com.jme3.scene.shape.Sphere;

public class CollisionGroupsTest extends SimpleApplication implements PhysicsCollisionGroupListener{

    private BulletAppState bulletAppState;

        CollisionGroupsTest app = new CollisionGroupsTest();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        flyCam.setMoveSpeed(50);
        
        bulletAppState = new BulletAppState();
        stateManager.attach(bulletAppState);
        bulletAppState.setDebugEnabled(true);
        bulletAppState.getPhysicsSpace().addCollisionGroupListener(this, PhysicsCollisionObject.COLLISION_GROUP_02);
        
        RigidBodyControl droppedSpherePhysics = new RigidBodyControl(new SphereCollisionShape(1), 1);
        Geometry droppedPhysicsSphere = new Geometry("Dropped sphere", new Sphere(16, 16, 1));
        droppedPhysicsSphere.addControl(droppedSpherePhysics);
        droppedSpherePhysics.setPhysicsLocation(new Vector3f(4, 8, 0));
        droppedSpherePhysics.setCollisionGroup(PhysicsCollisionObject.COLLISION_GROUP_01);
        droppedSpherePhysics.setCollideWithGroups(PhysicsCollisionObject.COLLISION_GROUP_02);
        getPhysicsSpace().add(droppedPhysicsSphere);
        
        // The top most sphere. This should collide with the dropped sphere but should be overriden by the collide method
        RigidBodyControl staticSpherePhysics1 = new RigidBodyControl(new MeshCollisionShape(new Sphere(16, 16, 1.2f)), 0);
        Geometry staticSphere1 = new Geometry("Static sphere 1", new Sphere(16, 16, 1.2f));
        staticSphere1.addControl(staticSpherePhysics1);
        staticSpherePhysics1.setPhysicsLocation(new Vector3f(2.5f, 0, 0f));
        staticSpherePhysics1.setCollisionGroup(PhysicsCollisionObject.COLLISION_GROUP_02);
        staticSpherePhysics1.setCollideWithGroups(PhysicsCollisionObject.COLLISION_GROUP_03);
        getPhysicsSpace().add(staticSphere1);
        
        // The top most sphere. This should not collide with the dropped sphere at all
        RigidBodyControl staticSpherePhysics2 = new RigidBodyControl(new MeshCollisionShape(new Sphere(16, 16, 1.2f)), 0);
        Geometry staticSphere2 = new Geometry("Static sphere 2", new Sphere(16, 16, 1.2f));
        staticSphere2.addControl(staticSpherePhysics2);
        staticSpherePhysics2.setPhysicsLocation(new Vector3f(2.5f, -4, 0f));
        staticSpherePhysics2.setCollisionGroup(PhysicsCollisionObject.COLLISION_GROUP_03);
        staticSpherePhysics2.setCollideWithGroups(PhysicsCollisionObject.COLLISION_GROUP_03);
        getPhysicsSpace().add(staticSphere2);
        
        // The box floor. The dropped sphere should collide and land on this
        RigidBodyControl node3Physics = new RigidBodyControl(new MeshCollisionShape(new Box(Vector3f.ZERO, 100f, 0.2f, 100f)), 0);
        Geometry node3 = new Geometry("Box", new Box(Vector3f.ZERO, 100f, 0.2f, 100f));
        node3.addControl(node3Physics);
        node3Physics.setPhysicsLocation(new Vector3f(0f, -6, 0f));
        getPhysicsSpace().add(node3);
    }

    private PhysicsSpace getPhysicsSpace() {
        return bulletAppState.getPhysicsSpace();
    }

    @Override
    }

    @Override
    public void simpleRender(RenderManager rm) {
    }

    @Override
    public boolean collide(PhysicsCollisionObject nodeA, PhysicsCollisionObject nodeB){
        Spatial collidedA = (Spatial)nodeA.getUserObject();
        Spatial collidedB = (Spatial)nodeB.getUserObject();
        System.out.println(collidedA.getName() + " collided with " + collidedB.getName());
        return false;
    }
}
```

shadowislord:
Looks good to me, except that the the logic is slightly different in this implementation compared to JBullet.
In JBullet we let the first listener decide the if the collision occurs but for this implementation we defer to the second listener if the first one says they should collide. We should decide what behavior we would like to have. 

Another option would be to always call both listeners and then combine their result like this:
`collide = listener1Result && listener2Result`.
The reason is that maybe you just want to use the listener as a listener and not to influence the collision result, so in that case you want both listeners to be called no matter what. Although, if the first listener opts out the collision then technically the objects are no longer colliding, so maybe it doesn't make sense to call the second listener?

@normen 
@Fadorico 
What do you think makes more sense?

Fadorico:
I think that having the first listener (if any) decide the fate of the collision isn't the right thing to do. Out of the two solutions above, I personally think that combining the results of both listeners makes the most sense. However, I'm not completely satisfied with this solution, so i'll write some use cases to show what I'm thinking.

For all cases, let listener1 be the first listener and listener2 be the second listener. Let's also assume listener1 is always the first listener to be notified of the collision.

**Use case 1: listener1 says the collision can't happen**
In this case, listener2 shouldn't be notified of the collision, since we already know at that point that listener1 doesn't want it to happen. We would then return a negative response saying the collision can't happen.

**Use case 2: listener1 says the collision can happen.**
In this case I think listener2 should be notified of the collision and ultimately decide its fate, since we know that listener1 is fine with the collision. This use case would then branch out into two possible scenarios.

**Use case 2.1: listener2 says the collision can happen as well**
If listener2 also agrees with the collision, then we can return a positive response and everything is rainbows and sunshine.

**Use case 2.2: listener2 says the collision can't happen**
Here's the case that is making me indecisive. If listener1 says the collision can happen but listener2 says otherwise, the collision shouldn't happen. So the right thing to do here would be to notify listener1 that the collision won't happen after all. However, with the current implementation, I don't see a clear way of doing this, as listener1 would have to undo what it did when it was notified.

So what I'm thinking is that their maybe should be two methods in PhysicsCollisionGroupListener; one that returns the result of "do I think the collision should happen" and another that notifies the listener of the final decision. Something like

```
public boolean shouldCollide(PhysicsCollisionObject node, PhysicsCollisionObject node1);
public void onCollide(PhysicsCollisionObject node, PhysicsCollisionObject node1);
```

But doing this changes the API, so I'm not sure what you all think about this idea.

shadowislord:
I think the approach you proposed where we just ask both listeners and then combine the result makes the most sense.

There's no need to overcomplicate this for two reasons:
1) Broadphase collision doesn't mean the objects are actually colliding, it just means bullet wants to know if they should be colliding or not. So using this listener to detect collision will not yield the expected result.
2) Users of the engine are expected to have some consistency with how they manage their collision groups and listeners... If the ball collision listener says balls should be colliding with the ground, but the ground collision listener says balls should not collide with the ground, well then that sounds like a user error to me and not something the engine should be responsible for handling.

Fadorico:
All right that sounds good then. I'll commit a quick change that will call both listeners and then combine the result.

shadowislord:
@Fadorico 
Do you think it makes sense to make the same change in JBullet as well to make sure it is consistent?

Fadorico:
I think it would make sense, yes.

shadowislord:
OK, looks good

PR #450, opened 2016-03-16
Added method to see if VertexBuffer is not used

riccardobl:
Since updateBufferData is not called for VertexBuffers that are not available in the shader, their updateNeeded state will stay set to true, while logically should be false, since those buffers won't ever be updated.

updateBufferData cannot be called on unused buffers without breaking the existing code (i.e. material switching).

This commit contains the most straightforward way i can think about for fixing this problem.

Refering to http://hub.jmonkeyengine.org/t/see-if-a-mesh-has-been-changed/35346/6


Nehon:
what does it exactly fix? What is the issue?

shadowislord:
Adding the `unused` field to VertexBuffer does not make any sense. The same `Mesh` could be used with a variety of shaders, some of which use that vertex buffer and some which do not.

riccardobl:
Ok, i wasn't aware of that. 
I solved with another approach.

PR #452, opened 2016-03-16
Changes for implementation of improved shadow rendering.

Perjin:
Texture arrays can now be set as render targets for color and depth buffers.
Also fixed an issue with the shader node generation, when using array variables the multiplicity was missing.
And made in the Light class frustumCheckNeeded and intersectsFrustum public, to be able to implement a custom light filter.

Forum: http://hub.jmonkeyengine.org/t/improved-shadow-renderer/33496/50

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/Light.java
+++ b/jme3-core/src/main/java/com/jme3/light/Light.java
@@ -117,8 +117,8 @@ public abstract class Light implements Savable, Cloneable {
      */
     protected String name;
     
-    boolean frustumCheckNeeded = true;
-    boolean intersectsFrustum  = false;
+    public boolean frustumCheckNeeded = true;
+    public boolean intersectsFrustum  = false;
 
     protected Light() {
     }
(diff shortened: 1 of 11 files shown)


shadowislord:
@Nehon is there any reason why this is targeting the PBR branch? There's nothing PBR specific here.

Nehon:
Well at first I thought it was the whole new shadows system and on Perjin post on the forum, he uses PBR. It actually the ground on which the shadows system is built, so it could be on master actually.
Though, maybe we should keep this for 3.2. In that case it can just be merged on this branch and wait for the 3.1 to be out when we merge this branch to master.

Perjin:
I can make a new pull request against the master branch if you want. (If I do this I can also split this in three pull requests one for the array render buffers, one for the shader nodes bug fix and a different one for the Light changes then you can pick what should get into 3.1.)

Nehon:
Well, if you can make the PR on master yes, but don't bother to split it. Sorry for the misdirection.

PR #454, opened 2016-03-20
Allow MotionEvents to be cloned (NPE)

MeFisto94:
Long story short:
Even though you are about to overhaul the cloning system you could still be effected by this:
Since `lookAt` and `rotation` had no initial values, they were serialized as null leading to a NPE in the cloneForSpatial:

```
at com.jme3.cinematic.events.MotionEvent.cloneForSpatial(MotionEvent.java:284)
```
Line 284 is `control.lookAt = lookAt.clone();` (it's still alpha-3)

The thing which first caught me is the behavior of BinaryOutputCapsule's defValue:
It's purpose is NOT to write `defVal` when the `value == null` but rather to simply not serialize, when `value` is already `defVal`

I don't know if it'd be a good addition to serialize defVal when value is null (since serializing null is mostly pointless, and if not, you'd have that as defVal), but that's a different story.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/cinematic/events/MotionEvent.java
+++ b/jme3-core/src/main/java/com/jme3/cinematic/events/MotionEvent.java
@@ -64,9 +64,9 @@ public class MotionEvent extends AbstractCinematicEvent implements Control, JmeC
     protected int currentWayPoint;
     protected float currentValue;
     protected Vector3f direction = new Vector3f();
-    protected Vector3f lookAt = Vector3f.ZERO;
+    protected Vector3f lookAt = null;
     protected Vector3f upVector = Vector3f.UNIT_Y;
-    protected Quaternion rotation = Quaternion.IDENTITY;
+    protected Quaternion rotation = null;
     protected Direction directionType = Direction.None;
     protected MotionPath path;
     private boolean isControl = true;
@@ -213,9 +213,9 @@ public class MotionEvent extends AbstractCinematicEvent implements Control, JmeC
     public void write(JmeExporter ex) throws IOException {
         super.write(ex);
         OutputCapsule oc = ex.getCapsule(this);
-        oc.write(lookAt, "lookAt", Vector3f.ZERO);
+        oc.write(lookAt, "lookAt", null);
         oc.write(upVector, "upVector", Vector3f.UNIT_Y);
-        oc.write(rotation, "rotation", Quaternion.IDENTITY);
+        oc.write(rotation, "rotation", null);
         oc.write(directionType, "directionType", Direction.None);
         oc.write(path, "path", null);
     }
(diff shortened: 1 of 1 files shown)


Nehon:
That's actually pretty dangerous to do that.
If rotation.set() or lookAt.set is called later in the code, we modify a JME constant.
Even the upVector shouldn't be initialized like that, actually, even if it's a bit particular be cause it's not supposed to be altered..
Here you should just change the defaultValue in the serialization to null, and test for null values in the cloneForSpatial method.

Also about the serialization process. if null is not a valid state for the object, it should be tested in the write method and not be serialized in that case. Though here it's definitely a bug, the default value shoudl be null as those attributes are optional.

MeFisto94:
So isn't it better to call Vector3f.clone() for the default values then?
Because when allowing those values to be `null` nearly every method is broken (atleast the setters for Vector3f and Quaternion)

Furthermore should we improve the read() to return clones as well or is the deserialized (read) object then cloned using cloneForSpatial?

Nehon:
ok, looked again into that and actually this is not good...
The setLookAt actually assign the given vector instead of setting it (using vector3f.set()).
This was intended so that when setting a lookAt vector (say for example, by giving it the worldTranslation of a spatial) the look at would be changed whenever the worldTranslation would be changed, because they essentially are the same instance of a vector. Resulting in the object on the path always looking at the target.

This is not really great and I wouldn't do it the same way today, because everywhere else in the engine, the logic would be to have an appState or a control update the position if needed, and not share the same instance...
But changing this today will break user code. So we have to stick with this logic.

So basically to fix this issue, the change should be to change the cloneForSpatial method replacing 
```
control.lookAt = lookAt.clone();
```
wilth 
```
control.lookAt = lookAt;
```
and 

```
control.rotation = rotation.clone();
```
wilth
```
control.rotation = rotation;
```

MeFisto94:
That however implies that you share the same instance across clones.
While this is useful for the outlined case with the same vector it's not for the rotation (e.g. when you clone the Motion Path and then change the curve to `Linear`). Then you'd have two different rotations.

We could use something like:
```java
control.rotation = (rotation == null ? null : rotation.clone());
```

However that might not protect us if someone is not calling setLookAt() on a freshly created instance, so maybe have Vector3f.ZERO.clone(); as the default value and also put that into the read() method?

Nehon:
Yes ok for the rotation.

However I still disagree for the lookAt.
it will work with a null vector. as clone.lookAt = lookAt won't fail if lookAt is null.
Also the read method should have null as default value.

As a side note. The current code in the read is wrong somehow. The default value is tested with "==" not "equals", so if you go Vector3f.ZERO.clone() and then set this in the default it will have no effect. The current value Vector3f.ZERO has no effect neither, so it should be changed.

Initializing the lookAt with Vector3f.ZERO.clone() will also work, however you won't be able to set a proper default value unless you keep a reference to this very same clone...
I'd rather nulll check tbh...

MeFisto94:
I just checked it: lookAt CAN be null, apart from the cloneForSpatial(), so we could just use null in the variable declaration.
So what should we actually exactly do?

Nehon:
As said before. leave the declaration as it is (null)
then assign the lookAt in the cloneForSpatial and cloneJme methods, don't clone it
Then in write and read use null as default value.

PR #457, opened 2016-03-25
Fix incorrect normal map shading in PBRLighting.

Perjin:
With the new Mikk Tangent Space generator this fix in the shader should produce correct shading results.
BiNormal/biTangent needs to be calculated in the fragment shader and the final normal has to be normalized.
http://i.imgur.com/gz9pRUP.png
https://hub.jmonkeyengine.org/t/normal-map-shading-problems/34947/91


Nehon:
Man... that's awesome...I completely forgot the binormal was computed in the vertex shader...

shadowislord:
Can we apply the same fix to `SPLighting.frag`?

Nehon:
yep I'll take care of it

PR #458, opened 2016-03-29
Fixes in lwjgl3

JavaSaBr:
When I switched my jME3-SpaceShift-Editor to lwjgl3, I got several problems.
1. The input of chars doesn't work for UI.
2. Mouse wheel doesn't works.
3. NPE on start of application.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
@@ -136,7 +136,7 @@ public class GlfwMouseInput implements MouseInput {
         glfwSetScrollCallback(context.getWindowHandle(), scrollCallback = new GLFWScrollCallback() {
             @Override
             public void invoke(final long window, final double xOffset, final double yOffset) {
-                onWheelScroll(window, xOffset, yOffset);
+                onWheelScroll(window, xOffset, yOffset * 120);
             }
         });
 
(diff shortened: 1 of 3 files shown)


jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
```diff
@@ -136,7 +136,7 @@ public void invoke(long window, double xpos, double ypos) {
         glfwSetScrollCallback(context.getWindowHandle(), scrollCallback = new GLFWScrollCallback() {
             @Override
             public void invoke(final long window, final double xOffset, final double yOffset) {
-                onWheelScroll(window, xOffset, yOffset);
+                onWheelScroll(window, xOffset, yOffset * 120);
```
DannyJo: Why 120? Feels a bit magic, could we put it as a constant with a proper name?

JavaSaBr: I compared this value with LWJGL2.

kvainer: I see it being used for Linux and Mac, but I cannot find it used on Windows:
https://github.com/LWJGL/lwjgl/blob/master/src/java/org/lwjgl/opengl/LinuxMouse.java#L48
https://github.com/LWJGL/lwjgl/blob/master/src/java/org/lwjgl/opengl/MacOSXNativeMouse.java#L53

shadowislord: I see it being used for Linux and Mac, but I cannot find it used on Windows:
https://github.com/LWJGL/lwjgl/blob/master/src/java/org/lwjgl/opengl/LinuxMouse.java#L48
https://github.com/LWJGL/lwjgl/blob/master/src/java/org/lwjgl/opengl/MacOSXNativeMouse.java#L53

JavaSaBr: I tested my Editor on Ubuntu 14.04 and Windows 7. This fix works fine.

shadowislord: Can you move the number into a constant field? Then I'll merge this.

JavaSaBr:
Updated

PR #459, opened 2016-03-29
Added handling of exception on SkeletonControl

JavaSaBr:
I had this problem with several model imported from Blender.

WARNING: Could not enable HW skinning due to error:
java.lang.NullPointerException
    at com.jme3.scene.Mesh.prepareForAnim(Mesh.java:439)
    at com.jme3.animation.SkeletonControl.switchToHardware(SkeletonControl.java:127)
    at com.jme3.animation.SkeletonControl.testHardwareSupported(SkeletonControl.java:160)
    at com.jme3.animation.SkeletonControl.controlRender(SkeletonControl.java:281)
    at com.jme3.scene.control.AbstractControl.render(AbstractControl.java:119)
    at com.jme3.scene.Spatial.runControlRender(Spatial.java:681)
    at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:667)
    at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:677)
    at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:677)
    at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:677)
    at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:677)
    at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:677)
    at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:677)
    at com.jme3.renderer.RenderManager.renderScene(RenderManager.java:656)
    at com.jme3.renderer.RenderManager.renderViewPort(RenderManager.java:1038)
    at com.jme3.renderer.RenderManager.render(RenderManager.java:1097)
    at com.jme3.app.SimpleApplication.update(SimpleApplication.java:260)
    at com.ss.editor.Editor.update(Editor.java:353)
    at com.jme3.system.lwjgl.LwjglWindow.runLoop(LwjglWindow.java:369)
    at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:452)
    at java.lang.Thread.run(Thread.java:745)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
+++ b/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
@@ -158,7 +158,12 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
             }
         }
 
-        switchToHardware();
+        try {
+            switchToHardware();
+        } catch (final Exception e) {
+            Logger.getLogger(SkeletonControl.class.getName()).log(Level.WARNING, "Could not enable HW skinning due to error:", e);
+            return false;
+        }
         
         try {
             rm.preloadScene(spatial);


shadowislord:
It should not be required to catch a null pointer exception in a normal Java program.
It is better to fix the underlying bug causing the null pointer instead.

gouessej:
It's not the real fix:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/scene/Mesh.java#L439

Your geometry has no indices, it should be supported.

JavaSaBr:
I agree, but I think what catch + warning is better that crash application.

shadowislord:
@JavaSaBr 
Is this still a problem? On the forum you mentioned your issues were resolved.

JavaSaBr:
This problem is actual in last revision. If I try load a model with bad animation, my application crashes.

shadowislord:
The only way this exception can happen is if the mesh has a HWBoneIndex buffer but no BoneIndex buffer. This would be a corrupt model.

empirephoenix:
Hm a check for that condition, and then a correct exception could be beneficial for this.
The current code makes "JME does not load my model" to "JME doesnt load model correctly, why?"
A proper exception would hopefully help more in this case. 
(And could still be caught by your usercode if a special treatment is necessary in your use case)

JavaSaBr:
Also, I develop my editor for jME and I don't want to have the crash after trying opening an invalid model in the editor.

empirephoenix:
Thats why I propose to check for the missing buffer early, so that it is not loaded partially at all if it is invalid.

JavaSaBr:
But I think my editor should not know details of model implementation in the jME, so I can't check it.

empirephoenix:
Either you don#t understand or do not want understand :P

The model should never bee laoded in the first palce at all. You should only get a AssetLoaderException with invalid model, HWBoneIndexBuffer exists but not BoneIndexBuffer.

That exception would be in the assetloader.loadModel not at the update loop (as a faulty model should never make it pass the loadCall and to the scenegraph).

JavaSaBr:
Ok, but I think we need to have the checking in the both places ;)

Nehon:
@JavaSaBr what is the loader? BlenderLoader?

JavaSaBr:
@Nehon j3oLoader

PR #460, opened 2016-03-29
Upgraded the loader of cursor for LWJGL3

JavaSaBr:
If you use lwjgl3, you don't need flip image of cursor. So I created the class CursorKey for configure flip of image.


JavaSaBr:
Maybe do you have some question about this change?

shadowislord:
The main issue I have with this is that it is now the user's problem to deal with differing coordinate systems between LWJGL2 and LWJGL3. I.e. when they switch to LWJGL3 they now have to change all their cursor loading code to pass this parameter.

IMO this flipping should be handled by the context and not by the user.

JavaSaBr:
OK, I will change thid

JavaSaBr:
I updated this pull request.

shadowislord:
Sorry but I don't think this is the right approach. We cannot expose `Application` on the `AssetManager` as that creates a circular dependency and also violates [loose coupling](https://en.wikipedia.org/wiki/Loose_coupling).

The approach I was thinking is to perform the flip in jme3-lwjgl3 for now, and then once LWJGL3 becomes the default backend we can remove the flipping entirely.

JavaSaBr:
Ok :)

PR #461, opened 2016-03-30
Texture arrays as render targets, bugfix for shader nodes and new getter and setter in Light class

Perjin:
These are necessary changes to implement improved shadow rendering.

Recreation of this pull request: [old pull request](https://github.com/jMonkeyEngine/jmonkeyengine/pull/452) for master branch instead of PBR branch.

Texture arrays can now be set as render targets for color and depth buffers.
Also fixed an issue with the shader node generation, when using array variables the multiplicity was missing.
Added getters and setters in the Light class for frustumCheckNeeded and intersectsFrustum, to be able to implement a custom light filter.

Forum: http://hub.jmonkeyengine.org/t/improved-shadow-renderer/33496/50

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/Light.java
+++ b/jme3-core/src/main/java/com/jme3/light/Light.java
@@ -184,6 +184,22 @@ public abstract class Light implements Savable, Cloneable {
         this.enabled = enabled;
     }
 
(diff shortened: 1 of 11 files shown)


shadowislord:
Looks good to me
@Nehon 
If you think the shader node changes are fine I think we should merge this.

PR #464, opened 2016-04-01
Get normals and tangent synced with mikk tangent space.

Perjin:
The normal maps calculated in mikk tangent space expect unnormalized (interpolated) tangent and world normals in the fragment shader, because of this they have to be normalized in the vertex shader.


Nehon:
Not sure it's a good idea tbh.
Normalizing in the vertex shader is always problematic as the interpolated value once in the frag shader will no longer be normalized, unless all normals of the triangle are pointing in the same direction. I know it's slower to normalize in the frag shader on a pixel basis but it's more accurate.

PR #468, opened 2016-04-02
'UseAlpha' parameter for the Lighting.j3md Material

Georgeto:
Readded the `UseAlpha` parameter for the Lighting.j3md Material, and gave it the functionality it always should have had, but never actually had.

The default value is `true`, to ensure backward compatibility.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/Lighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/Lighting.frag
@@ -103,7 +103,12 @@ void main(){
       vec4 diffuseColor = vec4(1.0);
     #endif
 
-    float alpha = DiffuseSum.a * diffuseColor.a;
+    #ifdef USE_ALPHA
+        float alpha = DiffuseSum.a * diffuseColor.a;
+    #else
+        float alpha = DiffuseSum.a;
+    #endif
+    
     #ifdef ALPHAMAP
        alpha = alpha * texture2D(m_AlphaMap, newTexCoord).r;
     #endif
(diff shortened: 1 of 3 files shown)


shadowislord:
Not sure why this is needed, since you can make alpha insignificant by disabling blending and not setting `AlphaDiscardThreshold`.

Georgeto:
My goal is to render my scene, using a transparent background (clear) color, as an 'transparent' image (it has an alpha channel). As the prouced image has an alpha channel, the alpha value of every pixel is relevant. So if i have an opaque object, which texture has an non opaque alpha channel, the resulting alpha value of the objet pixels will have an alpha value != 255, and the object will look weird in the rendered image.

entityviewerappall
`UseAlpha = true`

entityviewerapp1
`UseAlpha = false`

However i can understand, that if you are actually developing a game, this is not a common scenario. Or you could simply remove the alpha channel from your texture. So I should probably just use an modified version of the Lighting Material for my special use case, even though it will be annoying keeping it up-to-date.

PS: This feature is also requested by [FbxMaterial](https://github.com/jMonkeyEngine/jmonkeyengine/blob/2bdb3de2f5ff26d34219664e21398064aedd2cce/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/material/FbxMaterial.java#L254), so i am not the only one ;)

shadowislord:
I think a bigger problem with `FbxMaterial` is that it requires an `AlphaMap` parameter too.
Anyways, this could be useful for FBX, but then again, so many other things in FBX are broke I am not sure if this would help much..

empirephoenix:
So any news on this? 
I do not see it hurting anything, as it is not compiled into the final script if not actually used,
The transparent scenario is one I have as well for my inventory preview images :)

If noone has a good reason against this, I will merge this next time I review the pulls

Nehon:
Well, on my part, I'm not fond of it.
I removed this parameter because it was there for very specific use cases ans most of the time was not used, and was just adding clutter to the lighting material.

IMO this should be handled by doing a fork of the shader in the user's project with this special behavior.

empirephoenix:
opted for simpler better maintainable basis materials, sooner or later every project is kinda required to write a own shader anyway.

PR #471, opened 2016-04-05
Reversed raytest results when needed

NemesisMate:
In some cases, native-bullet returns the ray results on a reversed order so it leads to unexpected bugs.

More details can be found [here](https://hub.jmonkeyengine.org/t/native-bullet-raytest-randomly-giving-inverted-list/35511).

### DIFF
--- a/jme3-bullet/src/main/java/com/jme3/bullet/PhysicsSpace.java
+++ b/jme3-bullet/src/main/java/com/jme3/bullet/PhysicsSpace.java
@@ -49,6 +49,7 @@ import com.jme3.scene.Spatial;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


empirephoenix:
Is this a fix of the perceived Result, or was the actual bullet code looked at to determine if only those two orders can appear?   After this commit, we kind of guarantee that the result is always correctly ordered, so we should make sure, that is actually is. (As this could be a cause for quite subtle bugs for many users)

NemesisMate:
Is a fix of the perceived Result. I tried to look at the bullet code but I'm not skilled enough at the moment (and neither have time to get that skill) to ensure is working like that. However, I made a few tests (simple, can't ensure at 100%).

shadowislord:
I would recommend to just do a sort. We don't know what the intended order is in this case so better not make any assumptions.

NemesisMate:
Talking about assumptions, isn't there a way to know which is the intended order by asking bullet people? (I don't know how feasible is this)

shadowislord:
@NemesisMate 
Took a quick look at the source code and it doesn't seem like there is any implied order. Also consider that on the bullet side, the results are given via a callback and not a list of collision results.

empirephoenix:
Ok, rejected merging this, as no guarantee of the outcome is made
-> allow raw list as it comes from bullet
-> use Collections sort on the list, to ensure guaranteed order, no matter what bullet tries to do (we can make this default, as most users will probably want this anyway)

I would feel very happy, if you would resubmit with that smaller change in behaviour.

PR #472, opened 2016-04-06
Removed uneeded casting of assetkey on blender file setup

NemesisMate:
The casting to ModelKey what causing an assetkey exception if passing an AssetKey (ie: using loadAsset instead of loadModel) while it wasn't giving any extra method.


shadowislord:
It is intended that model loaders expect the key to be an instance of `ModelKey`. 
The `BlenderKey` simply provides additional parameters specific to `.blend` files.

PR #473, opened 2016-04-06
Added support to glBlendEquation

BorealFeast:
See RenderState#BlendEquation and RenderState#BlendEquationAlpha

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/RenderState.java
+++ b/jme3-core/src/main/java/com/jme3/material/RenderState.java
@@ -121,6 +121,92 @@ public class RenderState implements Cloneable, Savable {
         Always,}
 
     /**
+     * <code>BlendEquation</code> specifies the blending equation to combine
+     * pixels.
+     */
+    public enum BlendEquation {
+        /**
+         * Sets the blend equation so that the source and destination data are
+         * added. (Default) Clamps to [0,1] Useful for things like antialiasing
+         * and transparency.
+         */
+        Add,
+        /**
+         * Sets the blend equation so that the source and destination data are
+         * subtracted (Src - Dest). Clamps to [0,1] Falls back to Add if
+         * supportsSubtract is false.
+         */
+        Subtract,
+        /**
+         * Same as Subtract, but the order is reversed (Dst - Src). Clamps to
+         * [0,1] Falls back to Add if supportsSubtract is false.
+         */
+        ReverseSubtract,
+        /**
+         * Sets the blend equation so that each component of the result color is
+         * the minimum of the corresponding components of the source and
+         * destination colors. This and Max are useful for applications that
+         * analyze image data (image thresholding against a constant color, for
+         * example). Falls back to Add if supportsMinMax is false.
+         */
+        Min,
+        /**
+         * Sets the blend equation so that each component of the result color is
+         * the maximum of the corresponding components of the source and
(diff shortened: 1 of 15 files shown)


jme3-core/src/main/java/com/jme3/material/RenderState.java
```diff
@@ -1302,6 +1493,16 @@ public RenderState copyMergedTo(RenderState additionalState, RenderState state)
...
         }
+        if (additionalState.applyBlendEquation) {
+            state.blendEquation = additionalState.blendEquation;
+        } else {
+            state.blendEquation = blendEquation;
+        }
+        if (additionalState.applyBlendEquationAlpha) {
+            state.blendEquationAlpha = additionalState.blendEquationAlpha;
+        } else {
+            state.blendEquation = blendEquation;
```
shadowislord: Looks like copy and paste error. I guess this should be:
```
state.blendEquationAlpha = blendEquationAlpha;
```

jme3-core/src/main/java/com/jme3/renderer/opengl/GL2.java
```diff
@@ -71,8 +71,14 @@
...
     public static final int GL_VERTEX_PROGRAM_POINT_SIZE = 0x8642;
     public static final int GL_UNSIGNED_INT_8_8_8_8 = 0x8035;
+    public static final int GL_FUNC_ADD = 0x8006;
+    public static final int GL_FUNC_SUBTRACT = 0x800A;
+    public static final int GL_FUNC_REVERSE_SUBTRACT = 0x800B;
+    public static final int GL_MIN = 0x8007;
+    public static final int GL_MAX = 0x8008;
     
     public void glAlphaFunc(int func, float ref);
+    public void glBlendEquationSeparate(int colorMode, int alphaMode);
```
shadowislord: Why not implement this in `GL`? OpenGL ES 2.0 supports this function too. 
This would allow Android / iOS devices to take advantage of this feature.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -746,6 +746,18 @@ public void applyRenderState(RenderState state) {
...
+                if (gl2 != null && (state.getBlendEquation() != context.blendEquation || state.getBlendEquationAlpha() != context.blendEquationAlpha)) {
+                    int colorMode = convertBlendEquation(state.getBlendEquation());
+                    int alphaMode;
+                    if (state.getBlendEquationAlpha() == RenderState.BlendEquationAlpha.InheritColor) {
+                        alphaMode = colorMode;
+                    } else {
+                        alphaMode = convertBlendEquationAlpha(state.getBlendEquationAlpha());
+                    }
+                    gl2.glBlendEquationSeparate(colorMode, alphaMode);
+                    context.blendEquation = state.getBlendEquation();
```
shadowislord: Looks like
```
context.blendEquationAlpha = state.getBlendEquationAlpha();
```
is missing

shadowislord:
Also, one more thing is we require a test in jme3-examples for any features being added.

shadowislord:
One more thing, do you have any example use cases for this feature?

BorealFeast:
I've made a commit based on your comments.
One simple use case for this feature could be the render of negative type particles.

PR #480, opened 2016-04-21
Set 1 as default angular factor for bullet native

riccardobl:
Set the default angular factor for bullet native to the same value of jbullet.


riccardobl:
See https://hub.jmonkeyengine.org/t/solved-are-gimpact-collision-shapes-broken-in-bullet-native/35691/4?u=riccardo

PR #488, opened 2016-04-26
Reworked original FBX importer

Eirenliel:
Forum thread: https://hub.jmonkeyengine.org/t/fbx-importer/30309/39?u=eirenliel

Changes from original version:
* Load animations without any additional data (no SceneKeys or fba files) — animations are loaded from separate animation layers (currently not supported by Maya)
* Huge code refactoring. It is possible to work with now and add new features (looks like @shadowislord 's :))
* Added support for not baked transformations (like pre/post-rotations/translation/scale and pivots), all euler rotation orders are implemented too
* More accurate bone animation
* Up to 8 UV-layers
* ByPolygon material data is loaded correctly (without buffer overhead too)
* Textures are loaded by name from the same folder, if present and no better way found
* Added support of FaceCullMode.Off
* Support of axis signs
* Fixes, tons of them!

### DIFF
--- a/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/RotationOrder.java
+++ b/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/RotationOrder.java
@@ -1,0 +1,102 @@
+package com.jme3.scene.plugins.fbx;
+
+import com.jme3.math.FastMath;
+import com.jme3.math.Quaternion;
+import com.jme3.math.Vector3f;
+
+// TODO This class has some potential in it... Should investigate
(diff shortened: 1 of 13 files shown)


Eirenliel:
Oh my, forgot to set JRE to 1.7... Fixed.

shadowislord:
Looks good! I am going to merge this although might not have much time to spend on merging the two importers.

Eirenliel:
I already have few fixes + I want to move some features from your importer into this, so there will be another update. Furthermore I found some problems with some cases of importing bind poses and will investigate it as soon as I have time for this.

PR #497, opened 2016-05-19
Ordering + raw methods instead of reversing.

NemesisMate:
Continuing the discussion in the pull request: https://github.com/jMonkeyEngine/jmonkeyengine/pull/471, made the requested changes.

### DIFF
--- a/jme3-bullet/src/main/java/com/jme3/bullet/PhysicsSpace.java
+++ b/jme3-bullet/src/main/java/com/jme3/bullet/PhysicsSpace.java
@@ -788,7 +788,7 @@ public class PhysicsSpace {
      * PhysicsRayTestResults ordered by it hitFraction (lower to higher)
      */
     public List rayTest(Vector3f from, Vector3f to) {
-        LinkedList<PhysicsRayTestResult> results = new LinkedList<PhysicsRayTestResult>();
+        List<PhysicsRayTestResult> results = new ArrayList<PhysicsRayTestResult>();
         rayTest(from, to, results);
         
         return results;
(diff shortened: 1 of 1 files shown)


pspeed42:
Hmmm... I wonder why the code used LinkedList... it will be less optimal in many ways but especially for sorting.

NemesisMate:
Yes, I had the same wonder but without more knowledge I am reluctant to change it (that's why I asked [here](https://hub.jmonkeyengine.org/t/native-bullet-raytest-randomly-giving-inverted-list/35511/2) at first place)

pspeed42:
LinkedList is only more optimal in the cases where one is adding or
removing from the front or back of the list.  Else ArrayList is more
optimal in all cases... but especially for Collections.sort() which must
index into the list.  LinkedLists are horrible for indexing as it must
always iterate from the beginning.  Also, as for garbage, an ArrayList will
have much less garbage since LinkedList will create an internal linking
entry per item.

NemesisMate:
I know all that theory, what I don't know is why it uses a LinkedList in first place. If the only reason is because of it was was just adding/removing, then I suppose it must be changed now (hm... or changed only for the sorting case?). Maybe changing it depending on the case would be the better but at cost of obtaining different list types on the different methods (with/without sorting).
So, the different solutions would be:

1. Change LinkedList by ArrayList in both cases (the rayTest and the rayTestRaw).
2. Change to ArrayList just for the rayTest case (sorted).

If you are more eager to one of them just say it and I'll adapt it.

pspeed42:
Just use ArrayList in both cases.  Whatever the original reasons for
LinkedList they are not valid.  Either now or ever depends on historical
perspective but it's just as likely a misunderstanding about optimization
on the part of the original author.

empirephoenix:
Thanks for the work :)

PR #498, opened 2016-05-21
Add/remove all PhysicsControls even when a RigidBodyControl is present

TripleSnail:
The current implementation of addAll / removeAll doesn't remove rest of the PhysicsControls if the spatial has a RigidBodyControl.

This pull request removes the else-branch so that add / remove is always called for the spatial. Add / remove is no longer present in the `if (spatial.getControl(RigidBodyControl.class) != null)` -branch to prevent the RigidBodyControl from being added / removed twice, thus it only handles joints.

### DIFF
--- a/jme3-bullet/src/main/java/com/jme3/bullet/PhysicsSpace.java
+++ b/jme3-bullet/src/main/java/com/jme3/bullet/PhysicsSpace.java
@@ -487,9 +487,10 @@ public class PhysicsSpace {
      * @param spatial the rootnode containing the physics objects
      */
     public void addAll(Spatial spatial) {
+        add(spatial);
+
         if (spatial.getControl(RigidBodyControl.class) != null) {
             RigidBodyControl physicsNode = spatial.getControl(RigidBodyControl.class);
-            add(physicsNode);
             //add joints with physicsNode as BodyA
             List<PhysicsJoint> joints = physicsNode.getJoints();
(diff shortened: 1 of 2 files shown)


empirephoenix:
@normen can you maybee take a look at this at some point? I'm not entirely sure if this is/was a bug or expected behaviour.

empirephoenix:
I still do not know what to do with this, if you are still interested, could you try to explain in a bit more detail what this change does?

TripleSnail:
Sure

Suppose you have a Geometry with bunch of PhysicsCollisionObjects attached to it, and none of them is a RigidBodyControl. When you call:

`physicsSpace.addAll(yourGeom);`

Result: All the PhysicsCollisionObjects are added to physicsSpace.

Now what if yourGeom had at least one RigidBodyControl? After

`physicsSpace.addAll(yourGeom);`

Only the RigidBodyControl is added to physicsSpace. PhysicsCollisionObjects are NOT added. The same applies to removeAll.

My pull request changes that. Even if there's a RigidBodyControl, addAll would add all those PhysicsCollisionObjects (and removeAll would remove them).

Does this help?

empirephoenix:
Yes I think I understand now that the old code was doing. I still kinda wonder if there was a specific reason for this, or if this was just an oversight, but I cannot see any issue with merging this. So if noone does object within a few days, I would merge it.

empirephoenix:
Hm well, I would say noone objected

PR #505, opened 2016-06-05
fixed the NPE with controls

JavaSaBr:
If you can't load some control, you have a message about this, but in next iteration of render, you get crash with NPE.


empirephoenix:
Shouldn't this be fixed in the loader instead?

JavaSaBr:
I didn't find better place :)

empirephoenix:
Hm after revisiting this, I kinda agree, could you add a warning that shows if it skips a null?

JavaSaBr:
I have added it.

PR #510, opened 2016-06-12
Bugfix: Native Point2PointJoint.createJoint() incorrectly allocates btHingeConstraint

bertleft:
The rest of the native methods cast jointId to btPoint2PointConstraint so I've changed createJoint() to allocate a btPoint2PointConstraint.

Forum link: https://hub.jmonkeyengine.org/t/bugfix-for-point2pointjoints-native-createjoint/36147


empirephoenix:
Seems logical to me, do you found this by chance, or due to an actual error?  

I'm not 100% sure why it was that way originally.

Dokthar:
Looks better this way, better ask @normen .

> why it was that way originally

 copy & past

bertleft:
empirephoenix, I came across it tracking down a physics body explosion that occurred on my project when run under Linux but not OsX.

PR #511, opened 2016-06-13
Finally fixing this

empirephoenix:
Long overdue fix for: 

(in short calculating inertia on mass=0 objects can cause NAN depending on compiler parameters, these NAN will contaminate every other float they come in contact with, resulting in suddenly vanishing objects on jme side, as the nan position/rotation cannot be rendered and is outside frustum.

https://hub.jmonkeyengine.org/t/jmonkey-3-1-native-bullet-bug/33595/3
http://www.bulletphysics.org/Bullet/phpBB3/viewtopic.php?p=12158&f=9&t=

Signed-off-by: Kai Boernert <kai-boernert@visiongamestudios.de>


empirephoenix:
I would like a second opinion on this, maybe @normen is still deep enough in this to remember it?

empirephoenix:
Wait.

I thought
3.0 is 3.0
3.1 is 3.1
and master is 3.2?

pspeed42:
Yes, that's right.  This PR was submitted against master and so will go into some eventual 3.2.  I wasn't sure if you also wanted it to be in 3.1 or not.

PR #512, opened 2016-06-20
Removed warning if scale=1,1,1 in CylinderCollisionShape

riccardobl:
<strike>That method just prints a warning, so from my point of view there is no reason to leave the call there.
EDIT: I want to point out that this is an issue because when cylinder shapes are created frequently the logs are polluted with warnings about scaling when the developer isn't actually trying to scale the shape.</strike>
See below.


empirephoenix:
Agreed, both jbullet and native bullet just print a warning no need to trigger this manually.

However I would propose instead to check in the setScale if the new scale is not 1,1,1. That way ay generic handling of collision shapes would not generate warning if they are not actually trying to change the scale.

riccardobl:
Yeah, good idea.
I've updated the pr and made the change also for jbullet.

empirephoenix:
thanks

PR #514, opened 2016-06-24
userdata to null again once all data is removed.

NemesisMate:
Discussion about: https://hub.jmonkeyengine.org/t/userdata-map-not-going-back-to-null-value-once-any-data-is-set/36236

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -1524,19 +1524,22 @@ public abstract class Spatial implements Savable, Cloneable, Collidable, Cloneab
     }
 
     public void setUserData(String key, Object data) {
-        if (userData == null) {
-            userData = new HashMap<String, Savable>();
-        }
-
-        if(data == null){
-            userData.remove(key);
-            if(userData.isEmpty()) {
-                userData = null;
+        if (data == null) {
+            if (userData != null) {
+                userData.remove(key);
+                if(userData.isEmpty()) {
+                    userData = null;
+                }
             }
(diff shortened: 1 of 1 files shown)


empirephoenix:
lets wait a few days to see if someone wants to input something, else I will merge

gouessej:
Please improve your change. Currently, if userData is null before calling setUserData("any_key", null), the hashmap is created at the beginning whereas it is useless as you're going to drop it a few lines later. I'd rather write:
```
if (data == null) {
    if (userData != null) {
         userData.remove(key);
         if(userData.isEmpty()) {
             userData = null;
         }
    }
} else {
     if (userData == null) {
         userData = new HashMap<String, Savable>();
     }
     if (data instanceof Savable) {
         userData.put(key, (Savable) data);
     } else {
          userData.put(key, new UserData(UserData.getObjectType(data), data));
     }
}
```

Nehon:
I agree with you Julien. Though it's a bit out of the scope of this PR, because what you describe is the behaviour we have today, regardless of this PR.

Also, calling setUserData("any_key", null) when the userData is empty is not in the normal usage flow of  user data, usually, you want to delete things from a non empty list.

So basically, no big deal if it's merged like this. Though, If OP wants to jump on the occasion to refactor a bit and clean up the code, I'm fine with it too.

NemesisMate:
You are right, Julien, I think I was a little careless, thanks.

gouessej:
You're welcome. I didn't want your stuff to be rejected, I preferred suggesting a small change. Now it's ok for me of course.

empirephoenix:
Thanks for the work.

PR #519, opened 2016-06-30
NewtMouseInput no longer auto-centers mouse cursor

saloisio:
I've addressed the issue described here: [https://hub.jmonkeyengine.org/t/jogl-support-jogl2-that-is/20891/333](url)


gouessej:
Excellent job. Actually, GLWindow.confinePointer(boolean) shouldn't be called in this class.

PR #525, opened 2016-07-07
NewtMouseInput accepts null cursor

saloisio:
NewtMouseInput now handles null cursor by resetting mouse cursor to default image.


gouessej:
Good job as usual, discussion about this change here:
http://forum.jogamp.org/JOGL-2-support-for-Ardor3D-JMonkeyEngine-3-jzy3d-and-NiftyGUI-tp1706747p4036887.html

PR #526, opened 2016-07-09
Extracted BufferAllocation to be replaceable and allow java9 to run

empirephoenix:
https://hub.jmonkeyengine.org/t/java9-incompability/36317/10


empirephoenix:
thanks to alrik now also contains the first true native allocator :)

PR #528, opened 2016-07-13
Update to Bullet 2.83.7

riccardobl:


Ali-RS:
Thanks @riccardobl 
Can we still use `bulletAppState.setThreadingType(BulletAppState.ThreadingType.PARALLEL);`
see here : https://jmonkeyengine.github.io/wiki/jme3/advanced/bullet_multithreading.html#toolbar
it seams bullet uses experimental Bullet 3 OpenCL now. Is it possible to support it with JME new OpenCL wrapper ? 
thanks to @shamanDevel

riccardobl:
> Can we still use bulletAppState.setThreadingType(BulletAppState.ThreadingType.PARALLEL);

Yes, it has nothing to do with that. 
PARALLEL spawns a thread for each physics space in java and use them to call the update (while SEQUENTIAL does everything in the jme update loop), the code i removed was there to use a function of bullet<2.83 that allows to use multiple threads to resolve the collisions of a single physics space, but there is no way to enable this neither in the current release.

> it seams bullet uses experimental Bullet 3 OpenCL now. Is it possible to support it with JME new OpenCL wrapper ? 

It should be possible to use it without any wrapper, since we don't need to call anything from java.
But i don't know if this would be a good idea, since opencl drivers are not really well standardized yet...

Ali-RS:
Ah.. Yes, got that. Thanks for explanation.

shamanDevel:
I'm not sure if Bullet 3 OpenCL exposes the native handles to the OpenCL objects.
If this is the case, then yes, you can use the new OpenCL wrapper to access the internal Bullet objects.
However, you would have to create the wrapper objects explicitly using one of the three concrete implementations (JOCL, LWJGL2/3).

shadowislord:
Looks good, I guess this new version of bullet uses threading automatically? Or do we need to do something to enable it?

riccardobl:
It doesn't. Afaik there is no official way to enable the old threading mode, there are some forks that reintroduce it though.
For us it doesn't change anything, since the multi threading has never been enabled in jme.

riccardobl:
I added the -Ofast argument to the build script (info here: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html ), and i disabled  _FORTIFY_SOURCE that is on by default in latest Linux distros and enables some sort of protection for buffer overflow, but at the price of performances.
This resulted in a huge performance boost, try  it.

I hope there won't be any problem if I aggregate these changes to this Pr.

Dokthar:
no change for windows build ?

riccardobl:
True, i forgot it... added now.

shadowislord:
Can you remove the commented out code? IMO there's no point in keeping it if that functionality was removed from bullet.

riccardobl:
Done.

PR #533, opened 2016-07-20
BufferAllocator Round 2

empirephoenix:
Allows a flexible interface to use own Allocators.

This has two uses: 
1. the current one does not work at all under Java9 (and it brings a simpler one that does)
2. it allows to use better/faster native ones, like the ones included in lwjgl and jogl2.

The default behavior is trying to load the old reflection based one, and Fallback to the simpler if an exception occurs. This allows to use JME in java9, while not changing established behavior.


shadowislord:
Looks good, however it is using tabs instead of spaces.
Recommend to use 4 spaces instead to be consistent with the rest of the engine, then it would be A-OK

empirephoenix:
@shadowislord changed to use whitespaces

shadowislord:
@empirephoenix  `PrimitiveAllocator` and `BufferUtils` still use tabs

shadowislord:
LGTM
@pspeed42 do you have any concerns about this?

shadowislord:
Guess not..

gouessej:
sun.nio.ch.DirectBuffer is inaccessible in Java 1.9 but sun.misc.Cleaner is still accessible according to the JEP 260. Therefore, ReflectionAllocator would go on working if you stopped using the former but not the latter.

empirephoenix:
Good to know,
but thats the good thing with this change,
it's now super easy to write a specialized allocator for this case, and everyone can easily decide for himself what the requirements in the own game are.

I personally will probably switch to the lwjgl/jogl native allocator interfaces for performance reasons :)

PR #535, opened 2016-07-24
Added RenderState.BlendMode.Custom

papaspace:
BlendMode.Custom allows glBlendFuncSeparate() calls.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GL3.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GL3.java
@@ -41,7 +41,9 @@ import java.nio.IntBuffer;
 public interface GL3 extends GL2 {
 
     public static final int GL_DEPTH_STENCIL_ATTACHMENT = 0x821A;
+    public static final int GL_DST_ALPHA = 0x0304;
     public static final int GL_GEOMETRY_SHADER = 0x8DD9;
+    public static final int GL_ONE_MINUS_DST_ALPHA = 0x0305;
     public static final int GL_NUM_EXTENSIONS = 0x821D;
     public static final int GL_R8 = 0x8229;
     public static final int GL_R16F = 0x822D;
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GL3.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GL3.java
@@ -41,9 +41,7 @@ import java.nio.IntBuffer;
 public interface GL3 extends GL2 {
 
     public static final int GL_DEPTH_STENCIL_ATTACHMENT = 0x821A;
-    public static final int GL_DST_ALPHA = 0x0304;
     public static final int GL_GEOMETRY_SHADER = 0x8DD9;
-    public static final int GL_ONE_MINUS_DST_ALPHA = 0x0305;
     public static final int GL_NUM_EXTENSIONS = 0x821D;
     public static final int GL_R8 = 0x8229;
     public static final int GL_R16F = 0x822D;
--- a/jme3-android/src/main/java/com/jme3/renderer/android/AndroidGL.java
+++ b/jme3-android/src/main/java/com/jme3/renderer/android/AndroidGL.java
@@ -103,6 +103,10 @@ public class AndroidGL implements GL, GLExt, GLFbo {
     public void glBlendFunc(int sfactor, int dfactor) {
         GLES20.glBlendFunc(sfactor, dfactor);
     }
+    
+    public void glBlendFuncSeparate(int sfactorRGB, int dfactorRGB, int sfactorAlpha, int dfactorAlpha) {
+       GLES20.glBlendFuncSeparate(sfactorRGB, dfactorRGB, sfactorAlpha, dfactorAlpha);
+    }
 
     public void glBufferData(int target, FloatBuffer data, int usage) {
         GLES20.glBufferData(target, getLimitBytes(data), data, usage);
--- a/jme3-core/src/main/java/com/jme3/material/RenderState.java
+++ b/jme3-core/src/main/java/com/jme3/material/RenderState.java
@@ -1594,8 +1594,7 @@ public class RenderState implements Cloneable, Savable {
                 + "\noffsetFactor=" + offsetFactor
                 + "\noffsetUnits=" + offsetUnits
                 + "\nlineWidth=" + lineWidth
-                + (blendMode.equals(BlendMode.Custom)?
-                "\ncustomBlendFactors=("+sfactorRGB+", "+dfactorRGB+", "+sfactorAlpha+", "+dfactorAlpha+")":
-                "\n]");
+                + (blendMode.equals(BlendMode.Custom)? "\ncustomBlendFactors=("+sfactorRGB+", "+dfactorRGB+", "+sfactorAlpha+", "+dfactorAlpha+")":""
+                +"\n]");
     }
 }
--- a/jme3-core/src/main/java/com/jme3/material/RenderState.java
+++ b/jme3-core/src/main/java/com/jme3/material/RenderState.java
@@ -32,6 +32,7 @@
 package com.jme3.material;
 
 import com.jme3.export.*;
+import com.jme3.renderer.opengl.GL;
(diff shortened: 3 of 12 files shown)


Dokthar:
In the toString() you added a ternary operator but in one case there is no "\n]" at the end of the string.
Else it look good to me ... although i dont know gl stuff.

papaspace:
Thanks for checking. I just corrected this.

Dokthar:
you should check your parenthesis ... 
It should look like this : 
```java
(blendMode.equals(BlendMode.Custom)? "\ncustomBlendFactors=("+sfactorRGB+", "+dfactorRGB+", "+sfactorAlpha+", "+dfactorAlpha+")":"")
+"\n]";

```

papaspace:
Errmm, yes. Next time I'll check the toString() method for correctness -- Simple things fail first. Thanks for watching.

AlrikG:
Cool! Can you create a test case as an example to show how to use it?

empirephoenix:
So any opinions on this? Code quality wise it seems ok. What about the feature useability? 
Why is this required and cannot be done with the current possibilities (and hence allows the increase in code)

AlrikG:
I have the same problem as on site blow. Some particle effects doesn't look good on bright backgrounds and I want to change the way how they blended on my own custom rules.

http://blog.roblox.com/2014/10/transforming-the-look-of-roblox-with-new-particle-effects/comment-page-3/

shadowislord:
@empirephoenix 
One huge disadvantage is this ties the user API into OpenGL, since the custom blending mode now expects OpenGL arguments to be given. This means implementing the custom blending mode using Vulkan or DirectX might difficult. It's a tough call to make - and I am not sure if the benefits outweigh the costs..

empirephoenix:
Yes, I kinda agree, having a hard opengl depenency is pretty bad. So far we are sprisingly well Vulcan compatible (as glsl will run there) this would reduce this.

papaspace:
@AlrikG : Sorry for the late reply, I was on an extended vacation. I could create a test case o/r, but on one hand usage is quite simple and on the other hand some differences in the result only become relevant when rendering to texture and reusing that texture in your scene. Tell me if you still prefer having a test case.

@empirephoenix : "Why is this required and cannot be done with the current possibilities (and hence allows the increase in code)"
One situation that requires a custom blend mode has been discussed in [this forum thread](https://hub.jmonkeyengine.org/t/alpha-blending-question/36421), which has been successfully solved using a custom blend mode. Consider the following. While the predefined blend modes are sufficient in most cases, a fall back to pure OpenGL may be jME's biggest advantage over fixed-functionality engines.

@empirephoenix , @shadowislord : "since the custom blending mode now expects OpenGL arguments to be given."
This is not correct, as the custom blending mode expects integer arguments to be given. The GLRenderer takes these integers to define the blend mode during rendering, simply by invoking glBlendFuncSeparate(...). The respective "VulkanRenderer" would do the same, I suppose. From my point of view, this is no hard dependency on OpenGL.
However, in the future, a clean way of separating Vulkan from OpenGL would be to define the integers somewhere internally within the RenderState. So "upgrading" the custom blend mode to Vulkan [1] -once JME supports Vulkan- would only require a few steps but in the end OpenGL and Vulkan are very similar here. See: [Vulkan quick spec](https://www.khronos.org/registry/vulkan/specs/1.0/xhtml/vkspec.html#framebuffer-blendfactors), [Vulkan blend factors](https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCmdSetBlendConstants.html#framebuffer-blendfactors).


[1] I do not see a reason for mentioning DirectX here, as we all want a cross-platform engine and DirectX is limited to a single platform.

AlrikG:
Thank you but no I don't need a test case yet :). I still think its good idea to have such a feature

shadowislord:
@papaspace 
What I mean is that users are now expected to pass arguments into the method from the `GL` interface which is OpenGL specific. Ideally we would have those arguments exposed as jME enums, but this might be too much work given that not a lot of people have uses for these custom blending modes.

AlrikG:
I don't know why it is wrong to give the user a method to pass arguments from the GL interface. If a user wants to define a custom blend mode then he must use the GL interface. In my opinion it is better to have a method with OpenGL arguments than no possibility to define such custom blend modes.

pspeed42:
But we could translate them into JME enums... then there is no problem, really.  I mean, not 1000 separate blend modes but the JME-equivalent of the blend mode options.

I _do_ think it's worth it because I've already had to add two blend modes for my own purposes that I could have avoided by being able to call this new functionality.  I'm fortunate enough to be a core dev so I just added them.  Not every combination is worth that and not every developer will be willing to put that effort.

papaspace:
Thank you for your feedback. I would love to see this feature in JME, and obviously other people need it too. Since there seems to be an increasing number of core devs who recommend Enums, I will change this within the next couple of days.
(My experience with open software is almost zero and I am not sure about which general guideline to follow. The Enum-way seems to make sense if JME is targeting to be able to support renderers other than OpenGL at any time. That is an interesting fact I did not know yet. /monologue)

empirephoenix:
@papaspace Hi, any news here?

papaspace:
@empirephoenix Hi! Sorry, I didn't expect that high workload over the last couple of weeks. It's slowly getting better though. I promise that the changes will come.

AlrikG:
Cool! Maybe I have an improvement for your implementation: 
you could extend the enumeration by a private constructor with the corresponding  "public static final int" as parameter. This way you can avoid the swith case in convertBlendFunc().

Then you have something like:
"state.getCustomSfactorRGB().getBlendFunc()"

papaspace:
That's cool. I didn't even know that this was possible. Thanks for the tip!
An update is coming shortly.

AlrikG:
Great :) One more thing: Is it possible to replace "Zero(0)" by this "Zero(GL.GL_ZERO)" ?

papaspace:
I should have had more sleep, you're right

AlrikG:
Hm ok then I think we should wait for a comment from @shadowislord  or @pspeed42

jme3-core/src/main/java/com/jme3/renderer/opengl/GL3.java
```diff
@@ -41,7 +41,9 @@
 public interface GL3 extends GL2 {
 
     public static final int GL_DEPTH_STENCIL_ATTACHMENT = 0x821A;
+    public static final int GL_DST_ALPHA = 0x0304;
```
shadowislord: These two constants are available in both OpenGL 1.1 and OpenGL ES 2.0, so they should go into the `GL` interface

shadowislord:
I put one comment, once its fixed I think this is good to merge

papaspace:
@AlrikG I had to remove the integers in the enumeration. In fact, the Renderer (in this case GLRenderer) has to figure out the proper integer values on its own. Predefining a fantasy integer does not make sense and breaks the code.
@shadowislord I have previously found a version of GL.h where these two are commented out. Don't know where that came from, I apologize.

empirephoenix:
So I guess this is fine now? any objections to a merge?

empirephoenix:
Thanks for the work :)

papaspace:
You're welcome. Thank you for your valuable input! :)

AlrikG:
@papaspace  is there a test case? I only get plain colored quads from the particle emitter.

		mat.getAdditionalRenderState().setBlendMode(BlendMode.Custom);
		mat.getAdditionalRenderState().setCustomBlendFactors(BlendFunc.One, BlendFunc.Zero, BlendFunc.One, BlendFunc.Zero);
		mat.getAdditionalRenderState().setBlendEquation(BlendEquation.Add);

papaspace:
@AlrikG Unfortunately I am not familiar with DXT5. As it seems to work with PNG, the problem is probably in the DXT5 texture itself, or in the way the DXT5 texture is treated by LWJGL.

Are you sure that your DXT5 texture contains the proper values in the alpha channel? Does it work with the built-in BlendMode.Alpha?

AlrikG:
Yes it does work with BlendMode.Alpha and the DXT5 texture has a proper alpha channel. I also think its a DDS problem rather than a blend mode one.

PR #541, opened 2016-08-09
Added a cursor cache for LWJGL 2, resolves #537

### LINKED ISSUE #537: Mouse cursor disappearing after changing it multiple times
Mouse cursor disappearing after changing it multiple times. When you do it often enough and there is no way to get it back. We are using the same JmeCursor object instance. And setting it simply with com.jme3.input.InputManager inputManager.setMouseCursor().

We are changing it often depending on what the user interaction is. Seems that LibGDX has also stumbled upon this feature: https://github.com/libgdx/libgdx/pull/2841.

I also tested with LWJGL 3. And the same problem exists. The problem is present in both JME 3.0 and 3.1.

Forum discussion:
https://hub.jmonkeyengine.org/t/mouse-cursor-disappearing-after-changing-it-multiple-times/36458

tonihele:


tonihele:
If this is ok, I make a similar one to master. I guess I should have started there but...

PR #544, opened 2016-08-10
RotationController for Spatials

Brainsucker92:
I had several projects in the past where a Class for handling/controlling the rotation of Spatials was quite usefull. Unfortunately I didn't find any similar class in JME so far.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/control/RotationControl.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/RotationControl.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2009-2016 jMonkeyEngine
+ * All rights reserved.
+ *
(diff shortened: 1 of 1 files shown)


pspeed42:
Thanks for contributing but I don't think we'll include this in core.  For one, it's really small to implement inline when needed:
new AbstractControl() {
    public void controlUpdate( float tpf ) {
        spatial.rotate(some values * tpf);
    }
}
Code that's included has an incurred ongoing maintenance cost and the utility here does not offset that cost.

Second, for real games, rotation is generally controlled as part of the game object and controls are firmly on the "visualization" side.  JME often shows/does things in test and demos that are "quick and dirty" and might use something like this (but see the 5 lines of code that would work in that case) but it's kind of wrong to encourage that on core level when we can easily avoid it.

PR #552, opened 2016-09-06
Removed JmeCloneable interface from LodControl

Ali-RS:
It is already implemented in AbstractControl.


pspeed42:
Then you should remove Cloneable, too... but oh wait, you can't.  :)

It's better if we include the JmeCloneable interface everywhere that the Cloneable interface is.  So if you remove one then remove both... or don't remove either.

PR #555, opened 2016-09-11
fixed missing Cloneable in the Triangle.

JavaSaBr:
fixed missing Cloneable in the Triangle.


shadowislord:
If we are going to do this shouldn't we use the new `Cloner` system?

pspeed42:
Not sure it matters in this case since this is just simple cloning of a regular Java object.  I don't think the other jme math classes use the new cloner either.

PR #565, opened 2016-10-13
Add Town.zip scene

Ali-RS:
Added Town.zip to jme3-testdata


Ali-RS:
I uploaded .zip because [here](https://github.com/jMonkeyEngine/jmonkeyengine/blob/5970444b9c906fcdc64db38be356df869334f80b/jme3-examples/src/main/java/jme3test/asset/TestOnlineJar.java#L63) it needs to be located from zip. 
I can also upload extracted folder if you want.

Ali-RS:
@Nehon 
Any idea ?

Nehon:
mhhh..... well the test data is becoming bigger and bigger...
idk if we should add the town.zip into it... Maybe we could upload it along with one of the release and point to it... there is very small chance it will change some day anyway.. no real need to keep it in the source control... @pspeed42 @shadowislord what do you think?

and @Ali-RS definitely don't upload the extracted folder.

MeFisto94:
@Ali-RS Now you polluted this pr with your other changes.
Just as a hint: In the future you might want to create a new branch for each feature/pr you want to submit. That way it stays on the original level even when you change other things.

Also a good idea might be to have a branch `own_changes` or the like so you always have the original `master` to submit pr's without adding said changes.

But for now it's no real problem, we can fix that before merging, I just wanted to give you a bit info :)

Ali-RS:
Thanks, I will do as you said for my future PRs.

Regarding this PR, it seems core developers are not agree with putting town.zip file in testdata. Maybe we need to separate jme3-testdata from engine to a new repo in github. I think **[JmeResources](https://github.com/jMonkeyEngine/JmeResources)** is created for this purpose.

So i am going to close this for now.

PR #567, opened 2016-10-31
Fixes #566 - XML Serialization handles empty buffers properly now.

### LINKED ISSUE #566: DOMOutputCapsule is unable to serialize empty buffers
Hey Guys,
In DOMOutputCapsule Line 765 we're removing the last character (the space).
When we have an empty buffer, the last character isn't present.

There are multiple overloads of write where this occurs so I'll look through them and submit a pull request.

Edit: So a few things I noticed while editing the file:

- There already were some checks for empty buffers, but not for every buffer. They were implemented as `Math.max(0, buf.length() - 1);`. I changed that to `if (buf.length() > 0)` since that's more easy to understand. 
- I've added `@Override` to any Override.
- I've added the `//remove last space` comment where it was missing so it is consistent and obvious
- I've enclosed the if condition for the IOException in brackets so it's a bit more easily readable. (Since the long exception message looked like a block of multiple lines)

MeFisto94:
In DOMOutputCapsule Line 765 we're removing the last character (the space).
When we have an empty buffer, the last character isn't present.

- There already were some checks for empty buffers, but not for every buffer. They were implemented as `Math.max(0, buf.length() - 1);`. I changed that to `if (buf.length() > 0)` since that's more easy to understand. 
- I've added `@Override` to any Override.
- I've added the `//remove last space` comment where it was missing so it is consistent and obvious
- I've enclosed the if condition for the IOException in brackets so it's a bit more easily readable. (Since the long exception message looked like a block of multiple lines)

When this PR is approved/merged I will cherry pick this commit to v3.1

### DIFF
--- a/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMOutputCapsule.java
+++ b/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMOutputCapsule.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2016 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


empirephoenix:
Looks fine to, merging this as other already had a good Chance to veto by now

PR #571, opened 2016-11-19
Added cleanup for ChaseCamera

grizeldi:
Added function cleanupWithInput() which does the opposite than registerWithInput(), like said [here](https://hub.jmonkeyengine.org/t/properly-unregister-chasecamera-from-inputmanager/36813).


empirephoenix:
wont break anything existing, Looks fine to me, merging it

PR #572, opened 2016-11-24
Fix DefineList regression

shadowislord:
Fixes an issue where a material parameter with a 0 or 0.0 value would not be defined in the shader.


Nehon:
go ahead and merge it ;)

PR #576, opened 2016-11-24
Write/read the width and height of a quad

Fadorico:
Write the width and height of a quad on serialization. Otherwise getWidth() and getHeight() always return 0 when a quad was created from de-serialization.


empirephoenix:
Seems quite correct for me, will merge this if noone objects the next few days

PR #578, opened 2016-12-04
Fixes mouse cursor format for LWJGL3

FennelFetish:
LWJGL3 / GLFW expects images for cursors in a different format: RGBA instead of ARGB and vertically flipped. Putting the fix in GlfwMouseInput rather than JmeCursor to preserve compatibility with LWJGL 2.

[Forum Thread](https://hub.jmonkeyengine.org/t/mouse-cursor-flipped-and-wrong-colors/37484)


JavaSaBr:
Great!

empirephoenix:
So anyone objects this? else I think im gonna merge it.

Nehon:
go ahead

empirephoenix:
Thanks for the work

PR #579, opened 2016-12-07
recursively detects improperly modified spatials

AquariusPower:
It will look for all improperly modified spatials and show enough information to let the developer more precisely identify the problem from [here](https://hub.jmonkeyengine.org/t/a-more-precise-exception-report-about-modifying-the-scene-from-another-thread/34982).

Obs.: the second/newest commit is a proposal clarification method naming deprecation (so nothing will break) basically in favor of:
/* use only for debug purposes */
Spatial.getDebugName()

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Node.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Node.java
@@ -38,6 +38,7 @@ import com.jme3.export.JmeExporter;
 import com.jme3.export.JmeImporter;
 import com.jme3.export.Savable;
 import com.jme3.material.Material;
+import com.jme3.renderer.Camera;
 import com.jme3.util.SafeArrayList;
 import com.jme3.util.TempVars;
 import com.jme3.util.clone.Cloner;
@@ -47,7 +48,6 @@ import java.util.List;
 import java.util.Queue;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
 
 /**
  * <code>Node</code> defines an internal node of a scene graph. The internal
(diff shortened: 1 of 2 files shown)


pspeed42:
Seems like if we just exposed the modified flags then all of this could be done with one external utility class and a user parameter (for the, to me, unnecessary 'debug name')... and it wouldn't clutter up the code with a check that most experienced JME users avoid naturally.

I mean, I'll leave it to the others but I'm against this patch.

AquariusPower:
@pspeed42 both alternative suggestions are interesting (despite I still didnt find info about how it could be a 'user parameter'). Oh, these changes are solely intented towards new-comers.

pspeed42:
re: (despite I still didnt find info about how it could be a 'user parameter')

spatial.setUserData("myDebugName", "This is a debug string")

re "Oh, these changes are solely intented towards new-comers."

Yes, but new-comers will learn their lesson fast and then understand more why it's an issue... but these changes will dirty up the code and increase the maintenance burden (without actually teaching those users what's really wrong).  We would have to support these calls basically forever even if next version we came up with a better way to do updates where this wasn't an issue anymore.

This kind of debugging stuff should not be embedded right into the core classes.

AquariusPower:
You made me think about creating a JME debug class specialized in tracking problems we end users could create. It would never be a dependency to let things run. It could unsafelly access everything (even private), mainly whenever an uncaught exception happens, and would not even require the "modified flags" to be exposed.

But the Spatial.name is just for debug reasons as I learned, therefore it's method name is still confusing. Despite we should not use getName() as a coding base, if we do it, it will still work, but may break things by not reaching Spatials thru the expected way (ex. thru Lemur methods). I guess that will remain in the learning curve then... 

Good to know you are keeping it clear/simple to not clutter and create maintenance troubles.

I agree with you, I will just close it, thx!

Issue #96, opened 2014-03-24
MTLLoader and loadTexture


ghost:
Reference on googlecode: https://code.google.com/p/jmonkeyengine/issues/detail?id=597

PR #581, opened 2016-12-11
Fix some problems in issue #96 (SDK branch)

Ev1lbl0w:
Some problems launching examples of JmeTests (described in issue [#96](https://github.com/jMonkeyEngine/sdk/issues/96), SDK branch). Mostly dead links.


empirephoenix:
Thanks, I tested a few of those links and they seem to work. The older ones are quite obviously not correct anymore, so i skipped testing before after, as this can only improve the currently not working code.

PR #582, opened 2016-12-15
Fixes tab stops in BitmapText / BitmapFont (3.1)

FennelFetish:
The first explicitly set tab stop was always skipped. For all non-explicitly set tab stops it inserted a fixed spacing. This commit changes the behaviour so tabs are aligned to columns.
It also adds handling of tabs to BitmapFont.getLineWidth() which ignored tabs before.

[This was discussed in the forum.](https://hub.jmonkeyengine.org/t/tabs-in-bitmaptext/37495).
I made a pull request so this isn't forgotten.

Explanation of changes:
- BitmapFont: Added a constant DEFAULT_TAB_WIDTH. It is package private to not pollute namespace. Also added handling of tabs to getLineWidth().

- BitmapText: Only changed Javadoc. An important precondition is that the tab positions are sorted. I refrained from adding an Arrays.sort() call because the user shouldn't pay for what he doesn't need.

- LetterQuad: Removed faulty tab width calculation. It's in StringBlock now.

- StringBlock: Uses DEFAULT_TAB_WIDTH constant when instantiated. Removed the getters for tabWidth/tabPos. Added calculation of tab position at the bottom. The only way "tabPos" can leave this class now is through cloning. But since the array-reference is overwritten in setTabPosition(), it can safely be shared across cloned instances. The tabPos array is cloned in the setter to prevent modifications through the original reference.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/font/BitmapFont.java
+++ b/jme3-core/src/main/java/com/jme3/font/BitmapFont.java
@@ -86,6 +86,8 @@ public class BitmapFont implements Savable {
         Bottom
     }
 
+    static final float DEFAULT_TAB_WIDTH = 50.0f;
+
     private BitmapCharacterSet charSet;
     private Material[] pages;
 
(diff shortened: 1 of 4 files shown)


empirephoenix:
So, any objections to merging this? I left it laying around already quite a while due to other stuff. I will merge it if no further objections come within a week.

empirephoenix:
Thanks for your work

PR #585, opened 2016-12-28
Added LSF to showcased games

grizeldi:
Since the selection of games in JME's README.md is a bit outdatet, I thought adding some newer games wouldn't be a bad idea.
Opinions?


Nehon:
Cool thanks

PR #587, opened 2016-12-29
Added a method for changing a name of an animation.

JavaSaBr:
It needs to support name editing in my editor and I can't see any reasons why not. SDK uses a strange method for name editing:
```
                    Animation anim = control.getAnim(oldName);
                    Animation newAnim = new Animation(newName, anim.getLength());
                    newAnim.setTracks(anim.getTracks());
                    control.removeAnim(anim);
                    control.addAnim(newAnim);
```

### DIFF
--- a/jme3-core/src/main/java/com/jme3/animation/AnimControl.java
+++ b/jme3-core/src/main/java/com/jme3/animation/AnimControl.java
@@ -41,6 +41,7 @@ import com.jme3.renderer.ViewPort;
 import com.jme3.scene.Spatial;
 import com.jme3.scene.control.AbstractControl;
 import com.jme3.scene.control.Control;
+import com.jme3.system.Annotations.Internal;
 import com.jme3.util.TempVars;
 import com.jme3.util.clone.Cloner;
 import com.jme3.util.clone.JmeCloneable;
@@ -211,6 +212,7 @@ public final class AnimControl extends AbstractControl implements Cloneable, Jme
      * @param prevName the previous name of an animation.
      * @param newName  the new name of an animation.
      */
+    @Internal
     public void changeAnimName(String prevName, String newName) {
 
         if (!animationMap.containsKey(prevName)) {
--- a/jme3-core/src/main/java/com/jme3/animation/AnimControl.java
+++ b/jme3-core/src/main/java/com/jme3/animation/AnimControl.java
@@ -31,16 +31,20 @@
  */
 package com.jme3.animation;
 
-import com.jme3.export.*;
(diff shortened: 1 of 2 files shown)


Nehon:
This one is trickier than it looks.

First, we always had a modo to refrain to add code in the core engine that was not completely necessary, as in necessary in a game. And I'm 99,9% sure that you will never need to rename an application in a game.
So we always thought twice before adding code that would be only useful for the SDK, and refrain to do it whenever there was a work around. This one is convoluted I reckon, but does the job.
The only use case I know of changing the name of an anim in the engine is this one : in case of an editor/sdk.

As you realized it's not just a matter of adding an accessor in the Animation class, as the AnimControl uses the animation name as key in an internal map.
One way to do it, as you did, is to defer the name changing responsibility to the AnimControl. But what if another control also uses the name later? How would it be notified that the name changed? will it also have a name changing method? but then how will it notify the AnimControl?
Also, the setName you added in Animation is public, with no warning in the javadoc that it will completely screw up the animcontrol if you use it directly....
Best way would be to have the setName in Animation notify some registered controls that the name has changed an that they should clean things up on their side... but again... a very very convoluted thing to add to the core with 1 and only use case, that is not a game use case....

My point is, you'd better have some AnimUtils in your editor, that does the job. You can even go through reflection to access the name in the Animation class and the internal map in AnimControl, but this shouldn't be in the engine IMO.

JavaSaBr:
Ofcourse I can use Reflection API for this, but I believe that adding the method is not a bad option and I offer to add an annotation like a "InternalAPI" to the method.

JavaSaBr:
What do you think about the commit?
https://github.com/JavaSaBr/jmonkeyengine/commit/816c50761a606400ca8ef61f80876fa5242b1af7

Nehon:
I gave my opinion on the rename anim thing. It's not needed IMO.

for the internalApi thing, we already have something similar
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/system/Annotations.java
Though it's never used across the API at all...

I guess the plan was to add them and have the sdk check for them... 
If there is no support in the code editor it's plain useless IMO because users don't read doc, there is no way they go read annotations... 
Seems there are ways to have intellij raise errors or warnings for some annotations... but still... a far cry...
Maybe we could have gradle check for them during the build maybe...

I'll defer to the other core members @pspeed42 @shadowislord @normen so they can voice their opinion.

JavaSaBr:
I have added the annotation @Internal to new methods.

PR #592, opened 2017-01-03
New Physics based jme3 example game.

SkidRunner:
single class with physics marble game example.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java
+++ b/jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java
@@ -59,8 +59,6 @@ import com.jme3.scene.Spatial;
 import com.jme3.scene.shape.Box;
 import com.jme3.scene.shape.Sphere;
 import com.jme3.shadow.DirectionalLightShadowFilter;
-import com.jme3.system.AppSettings;
-import com.jme3.util.TempVars;
 import java.util.concurrent.Callable;
 
 /**
@@ -97,14 +95,7 @@ public class RollingTheMonkey extends SimpleApplication implements ActionListene
     private static final String INPUT_MAPPING_RESET     = "INPUT_MAPPING_RESET";
     
     public static void main(String[] args) {
-        AppSettings settings = new AppSettings(true);
-        settings.setTitle(TITLE);
-        settings.setResolution(1440, 810);
-        settings.setFullscreen(false);
-        
         RollingTheMonkey app = new RollingTheMonkey();
-        app.setShowSettings(false);
-        app.setSettings(settings);
         app.start();
     }
     
--- a/jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java
+++ b/jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java
@@ -328,9 +328,7 @@ public class RollingTheMonkey extends SimpleApplication implements ActionListene
             pickUp.rotate(pickUpSpeed, pickUpSpeed, pickUpSpeed);
         }
         
-        TempVars tempVars = TempVars.get();
-        
-        Vector3f centralForce = tempVars.vect1.set(Vector3f.ZERO);
+        Vector3f centralForce = new Vector3f();
         
         if(keyForward) centralForce.addLocal(cam.getDirection());
         if(keyBackward) centralForce.addLocal(cam.getDirection().negate());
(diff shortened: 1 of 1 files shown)


SkidRunner:
[RollingTheMonkey](https://github.com/SkidRunner/jmonkeyengine/blob/master/jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java)

empirephoenix:
Looks generally fine to me, I would however avoid the Tempvars useage and just allocate a new Vector3f each round. 
I assume that new jme users would seek this kind of example.
-> They might see TempVars remember some have true stuff about GC and jump to very wrong conclusions
-> Escape analysis in this example will probably lead to elimination of the allocation anyway.

However this is my personal opinion, others might think different.

Nehon:
I agree with Empire, we should not promote usage of tempvars in user code, it's best it stays an internal stuff.
Other than that, it's a nice example. Though it does not separate view from game logic (thought here is not much game logic :p), and maybe that's not the best message to users. But I guess we can't really expect that in a one class game :p

get rid of the tempVars usage and it's got to go for me.

SkidRunner:
if you noticed i put comments to encurage creation of controls. I'll clean it up. thank you

SkidRunner:
actually i could create a state and 2 controls as protected classes if you want most IDEs would show the seporate classes nested in the treeview but the game would still remain in one file. damn I haven't written a un nested multiclass file since Java 1.1...LOL

jayfella:
I would appreciate removing the presumed resolution and to display the settings page by default before merging.

SkidRunner:
> I would appreciate removing the presumed resolution and to display the settings page by default before merging.

presumed resolution removed.

@jayfella sorry that was just for my personal testing.

jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java
```diff
@@ -0,0 +1,417 @@
...
+    private static final float PLAYER_FORCE     = 80000 * PLAYER_ACCEL;  // F = M(4m diameter steel ball) * A
+    private static final Vector3f PLAYER_START  = new Vector3f(0.0f, PLAYER_RADIUS * 2, 0.0f);
+    
+    private static final String INPUT_MAPPING_FORWARD   = "INPUT_MAPPING_FORWARD";
+    private static final String INPUT_MAPPING_BACKWARD  = "INPUT_MAPPING_BACKWARD";
+    private static final String INPUT_MAPPING_LEFT      = "INPUT_MAPPING_LEFT";
+    private static final String INPUT_MAPPING_RIGHT     = "INPUT_MAPPING_RIGHT";
+    private static final String INPUT_MAPPING_RESET     = "INPUT_MAPPING_RESET";
+    
+    public static void main(String[] args) {
```
SkidRunner: removed AppSettings

jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java
```diff
@@ -0,0 +1,417 @@
...
+        
+    }
+    
+    @Override
+    public void simpleUpdate(float tpf) {
+        // Update and position the score
+        scoreText.setText("Score: " + score);
+        scoreText.setLocalTranslation((cam.getWidth() - scoreText.getLineWidth()) / 2.0f,
+                scoreText.getLineHeight(), 0.0f);
+
```
SkidRunner: removed TempVars

jme3-examples/src/main/java/jme3test/games/RollingTheMonkey.java
```diff
@@ -0,0 +1,417 @@
...
+        player.setPhysicsLocation(PLAYER_START.clone());
+        player.setAngularVelocity(Vector3f.ZERO.clone());
+        player.setLinearVelocity(Vector3f.ZERO.clone());
+        // Reset the score
+        score = 0;
+        // Reset the message
+        messageText.setLocalScale(0.0f);
+    }
+    
+}
```
SkidRunner: to address concerns of mvvm style coding I could create a protected class here
```
class PlayerBallControl extends RigidBodyControl implements PhysicsCollisionListener {
	public PlayerBallControl(float radius, float mass) {
		super(new SphereCollisionShape(radius), mass);
	}
	@Override
	public void collision(PhysicsCollisionEvent event) {
		if(!this.equals(event.getSource())) {
			return;
		}
		// TODO: handle collision event
	}
}
```
Of course also creating separate classes for player input and pickups.

PR #596, opened 2017-01-14
Changed isUnitVector method on Vector3f to use lengthSquared

NemesisMate:
Currently, Vector3f's isUnitVector() method is using length() instead of lengthSquared(), thus, a useless sqrt calculation.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Vector3f.java
+++ b/jme3-core/src/main/java/com/jme3/math/Vector3f.java
@@ -416,7 +416,7 @@ public final class Vector3f implements Savable, Cloneable, java.io.Serializable 
      * or false otherwise.
      */
     public boolean isUnitVector(){
-        float len = length();
+        float len = lengthSquared();
         return 0.99f < len && len < 1.01f;
     }
 


pspeed42:
It's funny how a good change will show me a piece of code that scares me for completely different reasons.  That if check is very scary.  +/- 1 centimeter within 1.0 means it's unit?!?  Yikes.  It's also completely different than the check in normalize() fortunately.  I guess there is no real reason for code to ever call isUnitVector() but still.

My math-brain is a little weak at the moment so I'm not sure if the lack of sqrt() makes this problem worse or better... that's the only reason I hesitate to accept it.  We may want to use the same check we use in normalize()... but that one actually might be _too_ strict.  Floating point comparison should always have _some_ epsilon I guess... but +/- cm is way too big in my opinion.

NemesisMate:
When talking about the +/- 1 cm I guess you are talking about the existing code (I didn't touch it). About the lack of the sqrt(), I don't have any clue (and that is why I just looked to other open source code and see what others think. And so, I saw: https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Vector3.java), where it is without the sqrt. Well, it is true that they have a much smaller number as error margin.
Why don't lower it a little bit?

Nehon:
to me the theory is that length is 1 more or less a thin epsilon due to float precision. so the squared value is also very close to being 1. So no need to have square root.
Though I also agree that epsilon here is way to big... i'd have picked like 10-5 at least or even lower...

pspeed42:
Yes, there is no reason to have sqrt() here.  We removed it incorrectly before.

...but the lack of a decent epsilon worries me because I'm not sure if it magnifies or minimizes the issue to switch it.  I'm also trying to think of a case where user code would be calling this often.  I mean we should fix it... but we should really fix it.

pspeed42:
Doing some quick checks, I guess it's ok.  0.99 * 0.99 = 0.9801... which is better for us.  (A change in behavior but a good one)  1.01 * 1.01 = 1.0201... which is also better for us.  (again a better change in behavior)
...still be nice to fix it properly, though.

pspeed42:
A fun read if you are interested in math: https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

stephengold:
I'm okay with the change as proposed. Other approaches include: (a) preserve the existing behavior by setting the limits to 0.9801 and 1.0201 OR (b) tighten the limits to something more reasonable, like 0.9999 and 1.0001 OR (c) deprecate the method and add a new one with epsilon as a parameter.

empirephoenix:
Wouldn't a correct approach base the epsilon on the absolute value range of the input? after all a fixed epsilon will not work for large values and be to large for tiny ones? Of course this would create a little cost in performance, but might lead to way more accurate calculations

pspeed42:
@empirephoenix did you read the included article?  And anyway, we know one of our test values is 1... so right there we at least know the range of magnitude in which we are dealing.  It should be possible to come up with a valid epsilon in that case but we'd probably have to base it on abs(val - 1f) instead of two separate comparisons.  Making the user pass an epsilon doesn't make much sense to me, either because of that.  We should change the method name if we allow an epsilon isCloseToUnit(float range)

And either way, if you are using isUnitVector() to see if you need to normalize then you (the royal you) should note that normalize is already doing its own more (too) strict check.  In fact, I think it's probably possible to call normalizeLocal() over and over and have it do work every time.

stephengold:
Anyone using this method is probably making a mistake. I'd eventually like to see it deprecated. Why not now?But if we're keeping it in 3.1, then we shouldn't change the behavior just for efficiency.

pspeed42:
I agree we should not fix this in 3.1... fortunately this patch is not for 3.1.  We can keep the method but we should _actually_ fix it in 3.2+.

stephengold:
The fix is to remove this method from Vector3f.

MeFisto94:
I have to disagree on that: This method is useful when you have a method which is looking for a direction only and you could throw an Exception when it isn't.

Actually this is a constructed case but if one would remove this method, it should be noted clearly that normalize() already checks if it's a normalized vector, because if not a developer would try to avoid normalizing the same value every frame. It is just waste.

pspeed42:
Note: I'm fine with keeping this method but your argument for keeping it is not a very strong one when isUnitVector() could easily be replaced with length() == 1 or lengthSquared() == 1 or whatever check the user wants to consider "good enough" for that check.

If we do keep it then we should fix it properly:  Subtract and compare against an appropriate epsilon.

empirephoenix:
So any ideas on this? As far as I can see, this one does probably not do any harm, but is also not really required, and might be thrown out with the whole method. However I remembered one use case for this: When I was tracking down a issue with native bullet, I spammed asserts all over the code, validating the the rotations are not fucked up. (there were nan's due to the bug). Using normalizeLocal instead would not have worked in that case.

empirephoenix:
killed due to being dead

PR #597, opened 2017-01-15
Added options of shaders with GLSL150

JavaSaBr:
Added options of shaders with GLSL150 to support OGL3 core profile on AMD cards.


JavaSaBr:
I am going to make several small PRs with changes like this...

Nehon:
Don't please, I'm working on a global solution in the engine.

PR #599, opened 2017-01-17
CapsuleCollisionShape: warn about scaling only if it's not identity

stephengold:
same diff riccardobl committed to 'master' on Jun 20, 2016
Without this change, a warning is generated every time you construct a CylinderCollisionShape.


Nehon:
thanks

PR #600, opened 2017-01-18
fix errors in how vector projection is calculated

stephengold:
same diff I committed to 'master' on Jan 14, 2017
Without this change, the magnitude of the project() result is incorrect unless 'other' is a unit vector.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Vector3f.java
+++ b/jme3-core/src/main/java/com/jme3/math/Vector3f.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2017 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


Nehon:
mhhh....mhhh.... what with this line ending stuff?

stephengold:
It's difficult for me to determine what's going on with the line endings. Not sure if it's my editor, my local repo, or GitHub that's changing them. But one of those three insists on adding carriage returns to Vector3f.java . Very embarrassing!

empirephoenix:
Check your git settings, if you default install on windows for example it is set to replace stuff and modify it. Also check your favourite IDE, at east in eclipse you can set which line ending to use.

stephengold:
"git config --get core.autocrlf" returns "true" for my repo. When I do "git config --list" on my repo, I don't see a setting for "core.autocrlf" so it's probably using the default, which for Windows7 is probably

stephengold:
It turns out the value I wanted was "false". I'm ready to close this request and start a new one.

PR #602, opened 2017-01-20
fix errors in how vector projection is calculated

stephengold:
same diff I committed to 'master' on Jan 14, 2017 only without altering the line endings in Vector3f.java
Without this change, the magnitude of the project() result is incorrect unless 'other' is a unit vector.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Vector3f.java
+++ b/jme3-core/src/main/java/com/jme3/math/Vector3f.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2017 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


Nehon:
mhhh..... what with the 7 commits? :))

stephengold:
It took me multiple tries to get the line endings right. Is this an issue?

Nehon:
well not really but once I merge this, 7 commits will be added to the branch.
That's no big deal but considering the extend of the change it's a bit too much.

 git is wonderful and allows you to basically rewrite history... while it can be dangerous, here is one of the cases where it's fine, because it's a pull request, and that you are pretty sure to be alone to work on it.
Here is a nice read.
https://www.atlassian.com/git/tutorials/rewriting-history

There are several options to consider : 
- Rebase interactive and squash/discard irrelevant your commits:
`git rebase -i <hash of the commit before the first in this branch>`
you'll be presented some kind of wonky unix like editor, but it does the job. I'll redirect you to the git rebase interactive documentation.

- reset soft your changes to the beginning of the branch and commit again, then push --force.
That's most probably what you intended to do with those reverts, or you should have used commit --amend which rewrites the previous commit. With git, you'll probably NEVER want to use the revert commit command, especially if it's to revert the commit before. It creates another commit as you can see.
`git reset --soft <hash of the commit before the first in this branch>`
This will reset all the commits while keeping the changes in the file (which in your case is actually the content of your last commit)
`git commit -m "fix errors in how vector projection is calculated"` 
Commit again
`git push --force`
This will rewrite history on the remote repo and update this PR.

Hope it helps

stephengold:
Happy now?

Nehon:
Wonderful thanks Steve

PR #605, opened 2017-01-23
Add cloneable support to shadow renderers and filters

JavaSaBr:
Add cloneable support to shadow renderers and filters.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/shadow/AbstractShadowRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/shadow/AbstractShadowRenderer.java
@@ -45,7 +45,6 @@ import com.jme3.math.Vector2f;
 import com.jme3.math.Vector3f;
 import com.jme3.post.SceneProcessor;
 import com.jme3.renderer.Camera;
-import com.jme3.renderer.Caps;
 import com.jme3.renderer.RenderManager;
 import com.jme3.renderer.Renderer;
 import com.jme3.renderer.ViewPort;
@@ -148,7 +147,7 @@ public abstract class AbstractShadowRenderer implements SceneProcessor, Savable 
 
     }
 
-    private void init(AssetManager assetManager, int nbShadowMaps, int shadowMapSize) {
+    protected void init(AssetManager assetManager, int nbShadowMaps, int shadowMapSize) {
         this.postshadowMat = new Material(assetManager, "Common/MatDefs/Shadow/PostShadow.j3md");
         shadowFB = new FrameBuffer[nbShadowMaps];
         shadowMaps = new Texture2D[nbShadowMaps];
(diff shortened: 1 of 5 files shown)


jme3-core/src/main/java/com/jme3/shadow/AbstractShadowRenderer.java
```diff
@@ -200,7 +204,7 @@ protected void initForcedRenderState() {
      *
      * @param postShadowMat
      */
-    protected final void setPostShadowMaterial(Material postShadowMat) {
```
Nehon: I'm not sure why I set this method to final in the first place, I don't think it will make a huge difference.
However, I'm really wondering why you need this method to no be final..?

jme3-core/src/main/java/com/jme3/shadow/DirectionalLightShadowRenderer.java
```diff
@@ -136,6 +136,12 @@ public void setLight(DirectionalLight light) {
     @Override
     protected void updateShadowCams(Camera viewCam) {
 
+        // it needs to support for editors
```
Nehon: Please remove this comment

jme3-core/src/main/java/com/jme3/shadow/PointLightShadowRenderer.java
```diff
@@ -100,7 +102,8 @@ protected void initFrustumCam() {
     protected void updateShadowCams(Camera viewCam) {
 
         if (light == null) {
-            throw new IllegalStateException("The light can't be null for a " + this.getClass().getName());
+            logger.warning("The light can't be null for a " + getClass().getName());
```
Nehon: I suppose not crashing and warning the user is alright

JavaSaBr:
I have removed the comment and have returned a final modification to the method, but I need to avoid an exception if I have no light in a shadow render :(

Nehon:
Yep I agree for the warning instead of the exception. it's better to not have shadows than crashing the whole app, and the warning should give a hint to the user.
I'm good with the PR, I'll marge it later tonight

JavaSaBr:
Thank you very much :)

Nehon:
I was thinking, now that we don't crash when the light is not set, maybe we should deprecate the current constructor and create a new one with the light as parameter....

JavaSaBr:
I have no idea why you want to add a constructor with a light as a parameter. I think we should have only a constructor without any parameters and we should have public methods to rebuild the render for any parameters at any time.

Nehon:
you have a point .... maybe that's doable. It's a bit out of scope of this PR though, so let's not do it here

Nehon:
Thank you ;)

PR #606, opened 2017-01-25
Added position to all lights

JavaSaBr:
I need to store a position of all kinds of light. I don't know how to implement it without core changes and without using light nodes. By the way, 3 of 5 light types using a position, so these changes help with reusing the code.
I provide directly editing of lights in my editor without a LightNode, because we should not have unnecessary nodes in your scene to edit a light. Maybe do you have else ideas about how to implement it?
http://imgur.com/a/rRZdE

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/Light.java
+++ b/jme3-core/src/main/java/com/jme3/light/Light.java
@@ -35,6 +35,7 @@ import com.jme3.bounding.BoundingBox;
 import com.jme3.bounding.BoundingSphere;
 import com.jme3.export.*;
 import com.jme3.math.ColorRGBA;
+import com.jme3.math.Vector3f;
 import com.jme3.renderer.Camera;
 import com.jme3.scene.Spatial;
 import com.jme3.util.TempVars;
@@ -103,7 +104,9 @@ public abstract class Light implements Savable, Cloneable {
     }
 
     protected ColorRGBA color = new ColorRGBA(ColorRGBA.White);
-    
+
+    protected Vector3f position = new Vector3f();
+
     /**
      * Used in LightList for caching the distance 
      * to the owner spatial. Should be reset after the sorting.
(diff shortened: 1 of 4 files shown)


pspeed42:
I guess you need to keep some MVC style "model" object for the lights so that your editor can manage them.  This is going to be true for a dozen different things so it's worthwhile solving the problem "How does the editor store things that don't get stored in the j3o?"... because 90% of the stuff an editor needs should be left out of the j3o.

pspeed42:
...because it makes zero sense for a DirectionalLight to have a position, for example.

JavaSaBr:
But MVC doesn't solve my issue because it doesn't give an answer to my question "how can I store this information" :( I have only a j3o file and I need to store positions of all lights, how can MVC help me with it? I have no idea :(

JavaSaBr:
you are right, a position for a direction light doesn't make sense, but we have duplicate code in 3 of 5 kinds of light.

pspeed42:
JME does not care about the position of your lights so it is not going to solve that problem for you.  JME specifically does NOT include features that are only useful for scene editors.

Store that data however you want... in some project json file, Java serialization, some other mechanism... but it doesn't really belong in the scene.  j3o is not going to be a very good project format for you for a hundred reasons.  Not the least of which is that each new version of JME might break your old projects.

JavaSaBr:
The option of using another file to store this information is terrible and I understand your point of view. It's just my attempt, and it will not break compatibility. If you will not approve these changes, I will understand it :)

empirephoenix:
I'm against this as well, sorry. If you insist on using j3o I would fake the data were necessary.
Eg just store a dummy spatial with all your extra data somewhere, and do a import dialog if a j3o is loaded that is missing this. (Aka found directional light, pleace place in scene for editing)

JavaSaBr:
The main problem is that lights don't have an user data like a spatial. What do you think about to add an user data to lights?

pspeed42:
Even JME's SDK stores additional data in separate files.  Your unwillingness to give your users an actual project file is going to be a constant struggle for you... and will likely break with new JME updates.

j3o is not really a transport format.  It's meant to be used to "compile" your scene for game distribution.  Even our own SDK relies on it a bit too much.

JavaSaBr:
@pspeed42  @empirephoenix so, what do you think about to add user data to the light class?

Nehon:
sorry I'm late to the party.
I do completely agree with Paul and empire.
Same for Userdata, and in a general way do yourself a favor and don't save in to the j3o something that doesn't belong here.

PR #608, opened 2017-01-27
Fixed saving a spatial of a motion event control.

JavaSaBr:
Fixed saving a spatial of a motion event control.


empirephoenix:
Makes sense for me

Nehon:
yep

PR #609, opened 2017-01-28
Fixed saving/cloning a state of enabled of a RigidBodyControl.

JavaSaBr:
Fixed saving/cloning a state of enabled of a RigidBodyControl.

### DIFF
--- a/jme3-bullet/src/common/java/com/jme3/bullet/control/RigidBodyControl.java
+++ b/jme3-bullet/src/common/java/com/jme3/bullet/control/RigidBodyControl.java
@@ -53,6 +53,7 @@ import com.jme3.scene.shape.Box;
 import com.jme3.scene.shape.Sphere;
 import com.jme3.util.clone.Cloner;
 import com.jme3.util.clone.JmeCloneable;
+
 import java.io.IOException;
 
 /**
@@ -143,6 +144,8 @@ public class RigidBodyControl extends PhysicsRigidBody implements PhysicsControl
         }
         control.setApplyPhysicsLocal(isApplyPhysicsLocal());
         control.spatial = this.spatial;
+        control.setEnabled(isEnabled());
+
         return control;
     }     
 
@@ -266,16 +269,20 @@ public class RigidBodyControl extends PhysicsRigidBody implements PhysicsControl
     public void render(RenderManager rm, ViewPort vp) {
     }
 
-    public void setPhysicsSpace(PhysicsSpace space) {
+    public void setPhysicsSpace(final PhysicsSpace space) {
+        final PhysicsSpace currentSpace = getPhysicsSpace();
         if (space == null) {
-            if (this.space != null) {
-                this.space.removeCollisionObject(this);
+            if (currentSpace != null) {
+                currentSpace.removeCollisionObject(this);
                 added = false;
             }
         } else {
-            if(this.space==space) return;
-            space.addCollisionObject(this);
-            added = true;
+            if (currentSpace == space) return;
+            // if this object isn't enabled, it will be added when it will be enabled.
+            if (isEnabled()) {
+                space.addCollisionObject(this);
+                added = true;
+            }
         }
         this.space = space;
     }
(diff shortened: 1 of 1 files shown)


jme3-bullet/src/common/java/com/jme3/bullet/control/RigidBodyControl.java
```diff
@@ -266,16 +269,20 @@ public void update(float tpf) {
     public void render(RenderManager rm, ViewPort vp) {
     }
 
-    public void setPhysicsSpace(PhysicsSpace space) {
+    public void setPhysicsSpace(final PhysicsSpace space) {
+        final PhysicsSpace currentSpace = getPhysicsSpace();
```
pspeed42: These changes seem to have nothing to do with the commit.

JavaSaBr: I think it's just more readable, I can revert this if it's important for you.

pspeed42: Either that or change your commit message to add "and a bunch of unrelated stuff because lulz."

pspeed42: Thinking on it some more.  I don't like and it needs to be removed.  setPhysicsSpace() is managing the state of this.space and shouldn't have to consult its own getter to know that.  To any reasonable Java developer, this.space is unambiguously "the current space"... and certainly the most current "current space" that it's ever possible for setPhysicsSpace() (the one managing that field) to deal with.  Calling getPhysicsSpace() is extra indirection that may obscure some random bit of horror that makes this method fail... as a developer, I'd have to go look.  And that's pointless since we are now (at best) indirectly asking about the one field that this method is supposed to be managing itself.

In general, your parches are more likely to be accepted if they only include the things you actually want to change and not a bunch of random other 'style' stuff.

tl;dr: some peoples "more readable" is another's unnecessary obfuscation.

pspeed42: I don't mean to sound ungrateful.  I appreciate the main fix.  It just wastes so much time to include other random stuff in what would have been a one line change otherwise.

JavaSaBr: I understand your point of view, I will revert these changes :)

JavaSaBr:
Do I need to revert some else?

pspeed42:
It looks good to me.  I'm no jme physics expert so hopefully someone else can look also.  It would be better if the commit were flattened but in this case I don't see that as a holdup.

Nehon:
It could be nice to squash your commits.
see this : https://www.atlassian.com/git/tutorials/rewriting-history

normen:
It looks good, in fact I didn't regard the enabled state in this case and I'm not 100% sure if its regarded everywhere else?

JavaSaBr:
I think when a user saves an object with disabled control, he expects that the control will be disabled when he will load the object again.

JavaSaBr:
https://github.com/jMonkeyEngine/jmonkeyengine/pull/612

normen:
Uh, sure, thats why I said it looks good. My implicit question was if you could check if the enabled state is considered in all other cases.

pspeed42:
Normen, could you double-check his flattened patch #612... or are you saying we should hold up applying that until the other checks are done?  My lack of familiarity with the code makes my uncomfortable to press the button.  (Alternately, if I had time to test it myself that would work too but I don't.)  It seems good to me but I don't know if there are additional implications.

JavaSaBr:
Normen, I didn't use bullet engine before, I detected this problem when I started implementing working with it in my editor. I tested my fix on cases of saving/loading my models and it works fine :)

normen:
Its the same patch. It looks fine. It can be applied. I am saying that somebody should look into other cases where enabled might not be regarded because I am now realizing that I probably didn't regard it in all cases.

PR #612, opened 2017-01-29
Fixed saving/cloning enabled state of a RigidBodyControl.

JavaSaBr:
Fixed saving/cloning enabled state of a RigidBodyControl.


empirephoenix:
Fine from my point,thanks, I might suggest to also take a look at other controls in jme, as this is quite likely to be inconsistent in many more places.

PR #615, opened 2017-02-06
Fix bullet crash when MeshCollisionShape is loaded from j3o [v3.1]

riccardobl:
https://hub.jmonkeyengine.org/t/core-dump-in-libbulletjme-so-when-reading-a-saved-rigidbodycontrol/37883


empirephoenix:
Thanks

PR #616, opened 2017-02-08
Fixed a exception in Material toString with blender materials.

delt0r:
Some blender materials have null Material Param keys. This shouldn't
throw an exception in a toString method. So rather than throwing a
OperaitonNotSupported exception a string is returned informative of the
null state of the key.


Nehon:
I agree about the comment, I don't like Exceptions in toString().

NemesisMate:
Agree too, I had to manage this exception all the time when printing the material for debug, is quite annoying (All the times I get it I think I have to change it but there it stills)

empirephoenix:
toString should never ever throw any exception, and since that method seems to be unused anywhere else than in the toString i merged it, thanks for the work

PR #618, opened 2017-02-13
Deprecated VRApplication. Now VRAppState should be used.

jseinturier:
Added a VREnvironment class that gather the VR related stuff. A VR environment has to be instanciated and initialized in order to know if the system can use VR display.

With this class, it is now possible to know if the system can be used with VR before any JMonkey use. It is so possible to decide if the VRAppState can be attached to the application or not.

An updated version of the sample application can be found [here](https://hub.jmonkeyengine.org/t/official-vr-module/37830/32?u=seinturier)


empirephoenix:
As it is a rather large change, does it change anything outside of the VR specific classes? If not it can be directly merged, as it will not influence anything else.

jseinturier:
No class outside the jme3-vr package are impacted by the changes made by this commit. This is the main goal of jme3-vr: to provide VR capability to JMonkey without modifying core classes or other module classes. I do not submit a pull request if any class that is exterior to jme3-vr is impacted.

empirephoenix:
Well in that case, thanks :)

PR #620, opened 2017-02-15
Removed the override that avoided a CompoundCollisionShape to be scaled

empirephoenix:
There is no explanation why this should be required, and NemesisMate says he tested it, and could not reproduce some of the older issues. It might have been solved with changes since then.


empirephoenix:
No objection after >one week, merged it

PR #622, opened 2017-02-18
added reusing position buffer on heightmap changing.

JavaSaBr:
I had a problem with direct memory cost when I very often make some changes of terrain heightmap(terrain editor). It took a lot of direct memory(8GB). So I have found a place on code where we can see, that we re-create a position buffer for each change. I have replaced it to reusing current position buffer.

### DIFF
--- a/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
+++ b/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
@@ -303,22 +303,29 @@ public class TerrainPatch extends Geometry {
 
     protected void setHeight(List<LocationHeight> locationHeights, boolean overrideHeight) {
 
+        final float[] heightArray = geomap.getHeightArray();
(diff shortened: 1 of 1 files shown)


pspeed42:
I assume we are 100% sure that the changes to the buffer are being done from the rendering thread?

JavaSaBr:
of course, I think this memory leak relates to LWJGL allocator, because we don't use deallocate method to old buffers.

empirephoenix:
Well not recreating the buffer makes sense in my opinion, if alone performance wise

JavaSaBr:
So, I have made fix of memory leaks in the PR:
https://github.com/jMonkeyEngine/jmonkeyengine/pull/623
but I think these changes make good optimizations... are you going to merge it?

empirephoenix:
From my point of view, not recreating buffers is a improvement, so if noone objects in the next few days, I will merge this

JavaSaBr:
Great :)

JavaSaBr:
@empirephoenix a few days passed ;)

empirephoenix:
Yes they did :)

PR #623, opened 2017-02-24
Implemented auto free LWJGL byte buffers.

JavaSaBr:
Implemented auto free LWJGL byte buffers.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/util/LWJGLBufferAllocator.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/util/LWJGLBufferAllocator.java
@@ -2,8 +2,12 @@ package com.jme3.util;
 
 import org.lwjgl.system.MemoryUtil;
 
+import java.lang.ref.PhantomReference;
+import java.lang.ref.ReferenceQueue;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * The implementation of the {@link BufferAllocator} which use {@link MemoryUtil} to manage memory.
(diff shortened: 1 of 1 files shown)


empirephoenix:
Thanks, I merged it without further testing, after all, it can't be any worse than the state before :)

JavaSaBr:
Thanks, I did some tests, it works fine.

AlrikG:
sounds good, but what was the purpose to replace this one?

- return MemoryUtil.memAlloc(size);
+final Long address = MemoryUtil.nmemAlloc(size);
+final ByteBuffer byteBuffer = MemoryUtil.memByteBuffer(address, size);

JavaSaBr:
What do you mean?

AlrikG:
I saw in the changes that you replaced "return MemoryUtil.memAlloc(size);" by the two new lines by I don't get it why did you do that. Is there any special reason for that? Don't get me wrong it is just out of interest :D

JavaSaBr:
I need to get address of allocated memory to make a deallocator.
http://imgur.com/a/wV43x

AlrikG:
ah ups my bad xD ... thanks for this PR :)

Ali-RS:
Thank you @JavaSaBr  :+1:

zzuegg:
IMHO it would be nice to have at least the possibility to manage the cleaning up myself. Like calling BufferAllocator.cleanup() myself rather then having a uncontrollable deamon.

JavaSaBr:
You have the method:  BufferUtils.destroyDirectBuffer(data);

empirephoenix:
It's no daemon, its the GC, it behaves similarly to the normal collections of the default BufferUtil

zzuegg:
Well, its not the GC, its clearly a deamon thread deallocating the native memory. It does behave similar to the default BufferUtil, and it is also fine that this is the default. But since we are tracking the objects it would be nice offer the possiblity of manual destruction without the need of tracking the buffer usages myself. Like combining the best of both worlds, mindless allocating/nulling and beeing able to trigger the cleanup at a time i know it will not jitter my frametimes.

zzuegg:
Maybe out of scope of this PR..

JavaSaBr:
If you will use BufferUtils.destroyDirectBuffer(), this "deamon" will skip these buffers.

empirephoenix:
You know how it goes :) WE are happy to hear from your PR :)

PR #625, opened 2017-02-27
fix mixed up javadoc description of euler angles

AlrikG:
I've noticed that the javadoc of the euler angle methods was incorrect and confusing, so I created this PR

- remove confusing description to fromAngles()
- fix incorrect description of toAngles()
- fix typo rall => roll

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Quaternion.java
+++ b/jme3-core/src/main/java/com/jme3/math/Quaternion.java
@@ -217,7 +217,7 @@ public final class Quaternion implements Savable, Cloneable, java.io.Serializabl
 
     /**
      * <code>fromAngles</code> builds a quaternion from the Euler rotation
-     * angles (y,r,p).
+     * angles (pitch,yaw,roll).
      *
      * @param angles
      *            the Euler angles of rotation (in radians).
@@ -233,8 +233,7 @@ public final class Quaternion implements Savable, Cloneable, java.io.Serializabl
 
     /**
      * <code>fromAngles</code> builds a Quaternion from the Euler rotation
-     * angles (x,y,z) aka (pitch, yaw, rall)). Note that we are applying in order: (y, z, x) aka (yaw, roll, pitch) but
-     * we've ordered them in x, y, and z for convenience.
+     * angles (x,y,z) aka (pitch,yaw,roll).
      * @see <a href="http://www.euclideanspace.com/maths/geometry/rotations/conversions/eulerToQuaternion/index.htm">http://www.euclideanspace.com/maths/geometry/rotations/conversions/eulerToQuaternion/index.htm</a>
      * 
      * @param xAngle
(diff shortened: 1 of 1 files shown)


empirephoenix:
This makes sense to me, and would explain some confusion I had years back with this methods :) Fine from my point of view.

AlrikG:
I just realized there is something more messed up with that euler methods.
I have tested with this simple example. If I put something into roll and read it back, then pitch and yaw are also modified.

```
float pitch = 0;
		float yaw = 0;
		float roll = FastMath.PI;

		System.out.println("==============INPUT==============");
		System.out.println("pitch " + pitch);
		System.out.println("yaw " + yaw);
		System.out.println("roll " + roll);

		Quaternion q = new Quaternion();
		q.fromAngles(pitch, yaw, roll); // (pitch, yaw, roll)
		float[] angles = q.toAngles(null); // (yaw, roll, pitch) => (pitch, yaw, roll)

		System.out.println("==============OUTPUT jME javadoc - wrong order==============");
		System.out.println("pitch " + angles[2]);
		System.out.println("yaw " + angles[0]);
		System.out.println("roll " + angles[1]);

		System.out.println("==============OUTPUT same order as input======");
		System.out.println("pitch " + angles[0]);
		System.out.println("yaw " + angles[1]);
		System.out.println("roll " + angles[2]);
```

empirephoenix:
Isn't that pi? 2pi is one rotation if im not wrong, so -pi == +pi or? The rest is artifacts from the conversions with floats

AlrikG:
yes but why is pitch and yaw rotated by -pi and roll is 0? Should not roll == pi and pitch==yaw==0 ?

pspeed42:
Euler angles are ambiguous.  A quaternion is not.  It is a compact representation of angles.  You are never guaranteed to get the same euler angles out that you put into it.  It's pure coincidence if you do.

pspeed42:
You've removed the description on the order that the angles are applied.  That's important because it's completely arbitrary and makes a big difference.  If you apply them in a different order then you get different results.  I mean, it was wrong but it's important. :)  Order is yaw then pitch then roll.  I think the person writing the docs got confused with the difference between them.  yaw = rotation around y.  pitch = rotation around x.  roll = rotation around z.

AlrikG:
OK then I will update the PR.

AlrikG:
is it fine now?

pspeed42:
The old text was wrong, I think.  We apply them in yaw, pitch, roll order.

AlrikG:
hm if I get the source code right, the order in which the angles are applied is (z,y,x) aka (roll,yaw,pitch). Do you agree with me?

pspeed42:
No, that's absolutely not the case.  Yaw is definitely first.  You look left/right before any other rotations are applied... otherwise you would induce side-effect roll with every other action.  (Imagine if you pitch first and then yaw, you will be inducing a roll as well.)

pspeed42:
Reread your post... it is possible that roll is first but it would be very strange.  Probably we should write a test that uses something like a fly camera but with the mouse wheel used to add roll.  Then we'd see for sure.

AlrikG:
ok then I build a test case if I have some time

pspeed42:
Note: if you have the time... it might easier to add mouse wheel to roll support to the Lemur Gem for camera movement since it already keeps separate yaw/pitch instead of applying quaternions every time.

empirephoenix:
dead

AlrikG:
Sorry but its not dead. I didn't had time for it because of planning my wedding, personal projects, and so on. But I definitely will work on this PR as soon as I have some time.

PR #626, opened 2017-02-28
Fixed some problems with getting an address of an LWJGL buffer.

JavaSaBr:
Fixed some problems with getting an address of an LWJGL buffer.


JavaSaBr:
@empirephoenix any questions?

empirephoenix:
Yes, :) What is the reason, that the memoryutils cannot be used? Do it cause errors when not having a bytebuffer but a flaotbuffer for example?

JavaSaBr:
The main change is to get an address using position '0' of a buffer because the method to get an address returns different addresses for different positions for the same buffer. Also, I don't want to use direct cast from Buffer to ByteBuffer, so I just added instanceof checks.

empirephoenix:
Oh i see, it did return the native pointer to the current position O.o yeah a wonder that did not crash instantly XD

PR #630, opened 2017-03-13
use the resizable parameter instead of a default 'true' value

rvandoosselaer:


empirephoenix:
Yes this looks right +1:  Thanks

PR #633, opened 2017-03-23
add PointSize to Unshaded matdef, test with SkeletonDebugger

stephengold:
In 3.0, SkeletonDebugger used Mesh.setPointSize() to highlight the joints. Now that setPointSize() is disabled, users of SkeletonDebugger need materials that set gl_PointSize. In JmeTests, SkeletonDebugger is always used with Unshaded.j3md, so that seems the logical place for it.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.vert
+++ b/jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.vert
@@ -16,7 +16,9 @@ varying vec2 texCoord1;
 varying vec2 texCoord2;
 
 varying vec4 vertColor;
-uniform float m_PointSize;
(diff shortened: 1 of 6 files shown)


jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.vert
```diff
@@ -16,6 +16,7 @@ varying vec2 texCoord1;
 varying vec2 texCoord2;
 
 varying vec4 vertColor;
+uniform float m_PointSize;
```
Nehon: This could also be between 
```
#ifdef HAS_POINTSIZE
   uniform float m_PointSize;
#endif
```

PR #637, opened 2017-04-06
fix SkeletonControl.getAttachmentsNode() for odd models (such as Jaime)

stephengold:
Formerly, getAttachmentsNode() assumed no local transform on any geometry with an animated mesh. However, Jaime violates this assumption. To make attachment nodes work more generally, each bone with attachments needs a target geometry. When the bone updates its transforms, it checks to see whether (1) the target geometry has the same parent as the attachments node and (2) whether its local transform is IDENTITY. If both these conditions are met (most models) the bone applies model transforms directly to the attachments node as before. If not, it adjusts the transform. My code assumes the parent of the attachment node is an ancestor of the target geometry. It also assumes all geometries with animated meshes have the same world transform. Formerly, getAttachmentsNode() failed if the spatial controlled by SkeletonControl was a Geometry. I've tried to allow for this possibility, though I didn't test.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
+++ b/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
@@ -70,15 +70,12 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
      * The skeleton of the model.
      */
     private Skeleton skeleton;
+
     /**
-     * List of targets which this controller effects.
+     * List of geometries affected by this control.
      */
-    private SafeArrayList<Mesh> targets = new SafeArrayList<Mesh>(Mesh.class);
-    /**
-     * Geometry with an animated mesh, for calculating attachments node
-     * transforms. A null means no geometry is subject to this control.
-     */
-    private Geometry targetGeometry = null;
+    private SafeArrayList<Geometry> targets = new SafeArrayList<Geometry>(Geometry.class);
+
     /**
      * Used to track when a mesh was updated. Meshes are only updated if they
      * are visible in at least one camera.
@@ -128,8 +125,9 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
         for (Material m : materials) {
             m.setInt("NumberOfBones", numBones);
         }
-        for (Mesh mesh : targets) {
-            if (mesh.isAnimated()) {
+        for (Geometry geometry : targets) {
+            Mesh mesh = geometry.getMesh();
+            if (mesh != null && mesh.isAnimated()) {
                 mesh.prepareForAnim(false);
             }
         }
@@ -141,8 +139,9 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
                 m.clearParam("NumberOfBones");
             }
         }
-        for (Mesh mesh : targets) {
-            if (mesh.isAnimated()) {
+        for (Geometry geometry : targets) {
+            Mesh mesh = geometry.getMesh();
+            if (mesh != null && mesh.isAnimated()) {
                 mesh.prepareForAnim(true);
             }
         }
@@ -220,9 +219,8 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
      */
     private void findTargets(Geometry geometry) {
         Mesh mesh = geometry.getMesh();
-        if (mesh.isAnimated()) {
-            targets.add(mesh);
-            targetGeometry = geometry;
+        if (mesh != null && mesh.isAnimated()) {
+            targets.add(geometry);
             materials.add(geometry.getMaterial());
         }
     }
@@ -248,10 +246,11 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
 
         offsetMatrices = skeleton.computeSkinningMatrices();
 
-        for (Mesh mesh : targets) {
-            // NOTE: This assumes that code higher up
-            // Already ensured those targets are animated
-            // otherwise a crash will happen in skin update
+        for (Geometry geometry : targets) {
+            Mesh mesh = geometry.getMesh();
+            // NOTE: This assumes code higher up has
+            // already ensured this mesh is animated.
+            // Otherwise a crash will happen in skin update.
             softwareSkinUpdate(mesh, offsetMatrices);
         }     
     }
@@ -325,8 +324,9 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
 
     //only do this for software updates
     void resetToBind() {
-        for (Mesh mesh : targets) {
-            if (mesh.isAnimated()) {
+        for (Geometry geometry : targets) {
+            Mesh mesh = geometry.getMesh();
+            if (mesh != null && mesh.isAnimated()) {
                 Buffer bwBuff = mesh.getBuffer(Type.BoneWeight).getData();
                 Buffer biBuff = mesh.getBuffer(Type.BoneIndex).getData();
                 if (!biBuff.hasArray() || !bwBuff.hasArray()) {
@@ -460,7 +460,8 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
         }
 
         updateTargetsAndMaterials(spatial);
-        Node n = b.getAttachmentsNode(targetGeometry);
+        int boneIndex = skeleton.getBoneIndex(b);
+        Node n = b.getAttachmentsNode(boneIndex, targets);
         /*
          * Select a node to parent the attachments node.
          */
@@ -485,12 +486,20 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
     }
 
     /**
-     * returns a copy of array of the targets meshes of this control
+     * Enumerate the target meshes of this control.
      *
-     * @return
+     * @return a new array
      */
-    public Mesh[] getTargets() {        
-        return targets.toArray(new Mesh[targets.size()]);
+    public Mesh[] getTargets() {
+        Mesh[] result = new Mesh[targets.size()];
+        int i = 0;
+        for (Geometry geometry : targets) {
+            Mesh mesh = geometry.getMesh();
+            result[i] = mesh;
+            i++;
+        }
+
+        return result;
     }
 
     /**
@@ -791,7 +800,6 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
      */
     private void updateTargetsAndMaterials(Spatial spatial) {
         targets.clear();
-        targetGeometry = null;
         materials.clear();
 
         if (spatial instanceof Node) {
--- a/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
+++ b/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2017 jMonkeyEngine
  * All rights reserved.
  *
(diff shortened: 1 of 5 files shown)


jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
```diff
@@ -210,15 +214,23 @@ public SkeletonControl(Skeleton skeleton) {
...
 
+    /**
+     * If specified the geometry has an animated mesh, add its mesh and material
+     * to the lists of animation targets.
+     */
+    private void findTargets(Geometry geometry) {
+        Mesh mesh = geometry.getMesh();
+        if (mesh.isAnimated()) {
+            targets.add(mesh);
+            targetGeometry = geometry;
```
Nehon: Mhh this won't work. Here you are using the last geometry in the subgraph for every bones.
You need to read the BoneIndex buffer to check if the bone affects this geometry. This should be done in the Bone class

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
```diff
@@ -443,9 +459,19 @@ public Node getAttachmentsNode(String boneName) {
                     + "in the skeleton.");
         }
 
-        Node n = b.getAttachmentsNode();
-        Node model = (Node) spatial;
-        model.attachChild(n);
+        updateTargetsAndMaterials(spatial);
+        Node n = b.getAttachmentsNode(targetGeometry);
```
Nehon: You should pass the entire list of geoms and let the Bone class find the one that it affects.

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
```diff
@@ -76,6 +75,11 @@
      */
     private SafeArrayList<Mesh> targets = new SafeArrayList<Mesh>(Mesh.class);
     /**
+     * Geometry with an animated mesh, for calculating attachments node
+     * transforms. A null means no geometry is subject to this control.
+     */
+    private Geometry targetGeometry = null;
```
Nehon: IMO not needed there

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
stephengold: I assume all the animated geometries in the model will have the same global transform, so it shouldn't matter which one I use. If my assumption is wrong, I will change the code as you suggest.

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
stephengold: true.

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
stephengold: Bote that a single bone can affect multiple geometries.

Nehon: IMO that's a wrong assumption.
What prevents a user from having several geometries with different transforms?
I mean this is the very issue we are trying to fix. We assumes geoms wouldn't have transforms, now we assume they have the same transforms. It looks like the same mistake to me.

Also you're right about the second statement... A bone can affects multiple geoms. Making the assumption in this particular case, and picking the first/last one, is less likely to cause trouble. Thought it doesn't justify the fact that it's wrong when a bone affects only one geom.

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
Nehon: I meant the targetGeomerty shouldn't be in SkeletonControl. It should only be in Bone and the target computed in Bone.
Even with your implementation it's just a matter of picking the last entry of the targets list, we don't need an additional variable.

stephengold: Pass the targets from updateTargetsAndMaterials() to getAttachmentsNode() as a return value?

Nehon: mhh wait... I did,'t realize but targets is actually a Mesh list not a Geometry list....
Maybe we could change this and just call geom.getMesh()

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
stephengold: I assume if a model has animated geometries with different transforms, then animations will affect them differently and they won't fit together when animated. I suppose one could create a model with two unconnected parts, each having a distinct set of bones, like a pair of insects. In that case giving the bone a complete list of animated geometries might be helpful. But in the multi-geometry models I'm familiar with (including Elephant, Ninja, and Sinbad) all the geometries have identical transforms.

stephengold: If I pass a list of geometries to the bone, and the bone affects more than one of them, will it be acceptable for the bone to arbitrarily use the first one it affects?

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
stephengold: If we generated a Geometry list, I believe we could dispense with both the 'targets' and 'materials' fields.

Nehon: True

jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
Nehon: Yes IMO it's acceptable.

stephengold:
I moved the target selection code from SkeletonControl to Bone. SkeletonControl's target list now holds geometries instead of meshes, and Mesh has a new public method to test for animation by a particular bone. I don't have a model with multiple transforms to test this new code, but it appears to work for Jaime and Oto, at least.

AlrikG:
Hi,
my AttachmentsNode doesn't work anymore. I'm also don't see any red box in the test case. Is this bug related to this PR?

unbenannt

Nehon:
Mhhh that's weird, it works for me. Could you tell us more on your system?

AlrikG:
specs:
Windows 7 64 Bit
AMD Radeon 290x (Driver Crimson 16.2.2)

Today I pulled the latest changes from git master and build jme with the command: "gradlew dist".
Then I updated my libs and tested it in my game. As you can see, the hand should hold a torch but. After that I checked the testcase via the freshly builded jMonkeyEngine3.jar and saw that the attachment was also missing.

qb 2017-09-19 10-03-36-373

stephengold:
(1) When you refer to "the testcase" do you mean TestAttachmentsNode?

(2) I'm curious about the structure of the scene graph in QB. Where is the SkeletonControl that controls the player's hand bone relative to the geometry that visualizes the torch?

AlrikG:
Hi @stephengold,

1. Yes I've tested with your new test case "TestAttachmentsNode". 
2. The hand model is a j3o file where the SkeletonControl and AnimationControl are attached to that model. The torch model has no animations and therefore no SkeletonControl and AnimationControl. Further more the torch is attached to a bone of the hand model.

stephengold:
I built a recent (c00407d) version of the 'master' branch of jMonkeyEngine. When I ran TestAttachmentsNode, the red box was visible and correctly positioned. I suspect the different behaviors we're seeing are related to our build procedures. You might try again with a fresh clone of jMonkeyEngine.

AlrikG:
Hi Stephen,
that's weird, I completely removed my clone and did a new checkout of the 'master' branch of jMonkeyEngine. Then I built it via cmd "gradlew dist". But I still don't see the red box :/ Do you guys have an idea what else the problem is?

Nehon:
Far fetched but I once had issues with amd cards not rendering some of the geometries. I had to rollback drivers to a previous version. For example the HelloJme example was displaying nothing. though if I added another box in the scene, this box would be displayed, but not the first one.
1. could you test this with your hardware ?
2. could you test the TestAttachementNode on an other hardware?

AlrikG:
I will not try a previous driver version because it works before and other people will have the same probem. But I tested it on different hardware:

(1) AMD Radeon r9 290x @ Crimson 16.12.2 ... not working
(2) AMD Radeon HD7850 @ Crimson 15.11 ... not working
(3) Intel Pentium G2020T ... working!

So it seams it doesn't work on AMD cards as you pointed out and the older driver version (15.11) also doesn't work :(

Nehon:
oh boy....AMD
What I meant by, "could you test this with your hardware" was to run the HelloJme test, not to rollback your drivers.

AlrikG:
oh sorry I misunderstood that.
It is the first AMD driver problem. Until now I only had nvidia driver problems :D

Nehon:
haha well we must not live in the same world because that's exactly the other way around for me...
But anyway...we need a solution for that.
Could you post the issue on the forum please? i'd like to ask people to test on their hardware and see how it goes

shadowislord:
Hi @AlrikG 

Can you add the following to `TestAttachmentsNode.main()`:

```java
settings.putBoolean("GraphicsDebug", true);
settings.putBoolean("GraphicsTrace", true);
app.setSettings(settings);
```

And post the log here. You only need to run it for a second or so (it will generate a lot of output).

Thanks.

AlrikG:
Hi @shadowislord 
this is the output [log.txt](http://alrik-online.de/alrik/log.txt)

stephengold:
@AlrikG, does the hand model set ignoreTransform=true ?

AlrikG:
Hi Stephen, no it dosn't

Nehon:
@stephengold it doesn't in TestAttachementNode anyway. It's another issue.

Jeddic:
I'm running into this problem as well since I updated to master for Spoxel, but only on my Retina Macbook Pro. It has a Geforce 650M graphics card so its not just an amd problem. I attached the output from setting the graphics debug on as well.

[output.log](https://github.com/jMonkeyEngine/jmonkeyengine/files/1385107/output.log)

AlrikG:
Are there any news? My torch is visible again when hardware skinning is disabled

shadowislord:
Hi @AlrikG , can you confirm that the problem was introduced in this commit? https://github.com/jMonkeyEngine/jmonkeyengine/commit/bf18ef304833a5428362e85fdd479b578e5f10df

AlrikG:
Hi @shadowislord, I can confirm the bug comes with the commit bf18ef3. The previous commit (381d69c) works.

Jeddic:
So I narrowed down the problem a little bit. It looks like there is a World Mat Override for "NumberOfBones" that gets applied to models added to the attachment node. If I set: 
    node.addMatParamOverride(new MatParamOverride(VarType.Int, "NumberOfBones", null)); on the attachment node then the geometry will appear. My guess is that since number of bones is set at the parent level it is forcing static geometry on an attachment node to calculate skinning which breaks on certain graphics cards. I'm going to test on my other dev machine to make sure this doesn't break something else. I'm not crazy for thinking that attachment nodes shouldn't be animated, right? @shadowislord @AlrikG

Nehon:
Haaaaaa
yeah it can be the issue.
Nice find

AlrikG:
@Jeddic I use the attachment nodes to attached animated weapons. Otherwise the attachment nodes seems useless to me :/

Nehon:
I think he meant attachement node should not be "skinned". And he's right. It's just meant to follow the bone it's attached to.

AlrikG:
ok I get it now  :)

Jeddic:
Yea, I wasn't very clear. My expectation was that at least by default anything in an attachment node shouldn't be skinned by default and should just follow the node. 

@Nehon I tried master and it seems to be working after your fix, thanks!

PR #641, opened 2017-04-13
Jme3 vr dev

jseinturier:
Removed more static imports

Merged with master revision

Created a VR input class that enable to handle VR tracker.


neph1:
Mostly refactoring from what I can see + breaking out a controller class? More OOP is positive in my opinion. I've had to spend some time familiarizing myself with the plugin to make my small fixes, so I should hopefully be able to contribute more to this in the future. I'll stay away from the architecture, but I can help fix issues.

jseinturier:
As there is no update of underlying OpenVR (still 1.0.6) and as the LWJGL3 OpenVR binding are not integrated within JMonkey, i plan now to make the code more Object oriented and more clear. My aim is to clean the code and the architecture before working on OSVR integration.

empirephoenix:
As I see it, this only changes Vr* classes and nothing else of the current core. Since you are maintaining this, I will merge it.

quazi-irfan:
@jseinturier would consider adding the feature to run VR application without having VR headset physically attached to the system?

empirephoenix:
I think there was a appstate, if you want to be able to write an application that con do both.

quazi-irfan:
Are you sure? I haven't tried harder. But @jseinturier said it in the forum that there is no 'HMD off' mode.

Link : https://hub.jmonkeyengine.org/t/can-not-run-vr-example-initializing-openvr-system-failed/38702/4?u=iamcreasy

PR #644, opened 2017-04-24
NullContext thread safety

rpax:
The fields `timeThen` and  `timeLate` are static. If multiple Applications are launched in different threads in headless mode, sync fails. This can be fixed removing the static modifier of the fields.


empirephoenix:
True, while usually multiple contexts are not that supported, there is no reason to block this, especially for the nullcontext. Thanks for the work.

PR #645, opened 2017-04-26
Added Xbox One S controller mappings

grizeldi:
Note that it has only been tested on linux.


grizeldi:
Just found a bug in the way all Xbox controllers are mapped, will fix, do not merge until then please.

grizeldi:
Also fixed right trigger being useless because it was mapped the same as the right stick.

PR #655, opened 2017-05-11
Updated broken link to point to wiki.

mitm001:
The original link was to a blog post that was turned into a wiki page. The blog post is no longer valid so pointed it to wiki instead.


empirephoenix:
Thanks for the work

PR #657, opened 2017-05-11
_FINE_ Logging level for Hardware skinning

rpax:
The _INFO_ level might to be too high for showing that the hardware skinning is enabled.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
+++ b/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
@@ -294,8 +294,9 @@ public class SkeletonControl extends AbstractControl implements Cloneable, JmeCl
 
                 if (hwSkinningSupported) {
                     hwSkinningEnabled = true;
-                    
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
```diff
@@ -294,8 +294,9 @@ protected void controlRender(RenderManager rm, ViewPort vp) {
 
                 if (hwSkinningSupported) {
                     hwSkinningEnabled = true;
-                    
-                    Logger.getLogger(SkeletonControl.class.getName()).log(Level.INFO, "Hardware skinning engaged for " + spatial);
```
iamcreasy: Probably moving this statement in the else clause would make the code more readable?

empirephoenix: Could the logger please be put into a static final field? Logger.getLogger(SkeletonControl.class.getName()) is relativly expensive

empirephoenix:
closed due to inactivity and required open changes. Anyone feel free to reopen after fixing is.

PR #663, opened 2017-05-24
Fixed IllegalArgumentException when removing a LodControl from a spatial

NemesisMate:
LodControl throws an exception when the spatial being added to isn't a geometry, however, when the controls is removed it calls setSpatial with a null value triggering the exception throw. Now checking for this null value case.


empirephoenix:
O.o thanks, apparently noone thought on removing the lodcontrol before :)

PR #669, opened 2017-06-06
Fix #646

### LINKED ISSUE #646: Blender Importer - DiffuseMap set as AlphaMap when alpha "influencer" is marked
To see the transparency of a texture in the blender viewport (and working with blender render), the [alpha influencer need to be marked](https://blender.stackexchange.com/a/11054). However, this causes the current blender importer implementation to add that texture as an AlphaMap too. This could be fixed if when both "influencers" are marked on a texture (color & alpha) it ignores the alpha (we can assume that that texture has an alpha channel).
With this solution, if the same texture is required to work as alpha and diffuse, it can be added twice and mark the influencers separately.

NemesisMate:
Ignoring the alpha influencer (mapping) if the color one was found first.

https://hub.jmonkeyengine.org/t/multiple-blender-importer-changes-fixes/38568

### DIFF
--- a/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/TextureHelper.java
+++ b/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/TextureHelper.java
@@ -614,8 +614,15 @@ public class TextureHelper extends AbstractBlenderHelper {
                     int texflag = ((Number) textureData.mtex.getFieldValue("texflag")).intValue();
                     boolean negateTexture = (texflag & 0x04) != 0;
 
+                    boolean colorSet = false;
                     for (int i = 0; i < mappings.length; ++i) {
                         if ((mappings[i] & mapto.intValue()) != 0) {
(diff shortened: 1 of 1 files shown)


empirephoenix:
I have sadly no idea how to validate this :(
Can you post a simple sample model that shows the differences?

NemesisMate:
Ok, I'll find time to post a sample model. However, you have more information in the [commit fix](https://github.com/jMonkeyEngine/jmonkeyengine/issues/646).

NemesisMate:
You can find a sample model [here](https://ufile.io/bmhuv). If you convert it to .j3o (using the importer) with the [alpha influencer](https://blender.stackexchange.com/questions/11053/why-is-the-alpha-channel-not-working/11054#11054) enabled you'll see how it also set the texture as alpha map (and so, adding transparency where it doesn't - the red part).
However, if you disable the influencer the conversion is done correctly (but in blender isn't seen as it must).

My fix solves this problem.

PR #670, opened 2017-06-06
Fixed #652

### LINKED ISSUE #652: Blender Importer - Generated textures with alpha have borders artifacts
When the importer get the textures from the blender file and set it to the model, it adds to these a gray "border". It doesn't add them, however, to the linkedData if **loadUnlinkedAssets** flag is set on the blender key.

It adds it too when "create j3m file" is selected on the sdk and the images are generated.

NemesisMate:
The issue was due to the texture blending with the object material. By default the material is gray (hence, a gray border). If the material color in blender is set to another color this is being the artifact's color. It can bee seen where the alpha in the texture isn't totally 0 or 1.
The current implementation weren't taking in account the material transparency when blending, so the fix is basically to use it just like blender does.

https://hub.jmonkeyengine.org/t/multiple-blender-importer-changes-fixes/38568

### DIFF
--- a/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/blending/TextureBlenderAWT.java
+++ b/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/blending/TextureBlenderAWT.java
@@ -174,6 +174,8 @@ public class TextureBlenderAWT extends AbstractTextureBlender {
         // float blendFactor = this.blendFactor * (1f - ((1f - pixelColor[3]) * materialColor[3]));
         // float oneMinusFactor = 1f - blendFactor;
 
+        float col;
+        
         switch (blendType) {
             case MTEX_BLEND:
                 result[0] = blendFactor * pixelColor[0] + oneMinusFactor * materialColor[0];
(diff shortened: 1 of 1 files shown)


empirephoenix:
Please close and open this, travis had some issues /home/travis/build/jMonkeyEngine/jmonkeyengine/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/blending/TextureBlenderAWT.java:240: error: cannot find symbol
                col = blendFactor * pixelColor[0];

Does this compile locally as it is in the git?

NemesisMate:
Ok, I missed the var declaration. Fixed now.

empirephoenix:
I'm still unsure about this (and the other blender related ones) on one side, I have no real way to confirm it is working, on the other hand, at least reading the code and the comment it makes sense to me. @Nehon what would your opinion be? just merge and see if complains happen?

NemesisMate:
To check the issue with this one you can use the example .blend I submited for this [other issue](https://github.com/jMonkeyEngine/jmonkeyengine/pull/669).

You can see that when you generate it with the blender importer, a gray border is added to the generated texture.

PR #672, opened 2017-06-18
Refactored Cylinder

### LINKED ISSUE #640: Cylinder.java: Wrong size of buffers for cylinders
```
Cylinder r = new Cylinder(3, 3, 1, 1, true);
IndexBuffer ib = r.getIndexBuffer();
System.out.println("ib " + ib.size() + ", v: " + r.getVertexCount());
```
Outputs "ib 90, v: 22". That means 30 triangles.
Expected value: 12 triangles for cylinder + tris for caps, (if not dealing with special case) 3 triangles per cap = 18 triangles and 17 vertices
(if dealing with special case) 14 triangles and 15 vertices
```
Cylinder r = new Cylinder(3, 3, 1, 1, false);
IndexBuffer ib = r.getIndexBuffer();
System.out.println("ib " + ib.size() + ", v: " + r.getVertexCount());
```
Outputs "ib 36, v: 12".  That is 12 Triangles.
Expected value: 12 Triangles and 9 vertices.

Saucistophe:
Refactored Cylinder generation to be more maintainable, readable, and fix #640.
I tested it against the former implementation, and got same results with all parameters.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
+++ b/jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
@@ -40,14 +40,11 @@ import com.jme3.math.FastMath;
 import com.jme3.math.Vector3f;
 import com.jme3.scene.Mesh;
 import com.jme3.scene.VertexBuffer.Type;
-import com.jme3.scene.mesh.IndexBuffer;
 import com.jme3.util.BufferUtils;
-import static com.jme3.util.BufferUtils.*;
 import java.io.IOException;
-import java.nio.FloatBuffer;
 
 /**
- * A simple cylinder, defined by it's height and radius.
+ * A simple cylinder, defined by its height and radius.
  * (Ported to jME3)
  *
  * @author Mark Powell
@@ -127,10 +124,10 @@ public class Cylinder extends Mesh {
      * mapped to texture coordinates (0.5, 1), bottom to (0.5, 0). Thus you need
      * a suited distorted texture.
      *
-     * @param axisSamples
-     *            Number of triangle samples along the axis.
-     * @param radialSamples
-     *            Number of triangle samples along the radial.
+     * @param axisSamples The number of vertices samples along the axis. It is equal to the number of segments + 1; so 
+	 * that, for instance, 4 samples mean the cylinder will be made of 3 segments.
+     * @param radialSamples The number of triangle samples along the radius. For instance, 4 means that the sides of the
+	 * cylinder are made of 4 rectangles, and the top and bottom are made of 4 triangles.
      * @param radius
      *            The radius of the cylinder.
      * @param height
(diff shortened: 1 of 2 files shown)


empirephoenix:
Cool that you are also at fixing this after finding it :)
One question, the return case for updateGeometrys, shouldn't that kinda be a IllegalStateException? Same for constructor, if I use values that can only be wrong, can a IllegalArgument be used? Or is there a case where just returning nothing is better?

Saucistophe:
@empirephoenix I'd have used an IllegalArgumentException if it were only for me, but I'm always afraid of breaking the existing behaviour for somebody somewhere. You think I should?

empirephoenix:
If I understood the issue correctly, the cylinder is generated incorrectly? It was not a bug in the Ray stuff?
So if a cylinder with invalid trinagles, and causes issues at other parts of the engine, I would favour exceptions

Saucistophe:
The issue was (probably) that too big a buffer was created for triangles because of incorrect calculations, and that lead to uninitialized, random data in the buffer which randomly generated invalid geometries that crashed rays.
I didn't test with erroneous parameters, since correct ones were enough to crash the ray thingy.

Let's go for exceptions, then!

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
```diff
@@ -127,10 +124,10 @@ public Cylinder(int axisSamples, int radialSamples,
      * mapped to texture coordinates (0.5, 1), bottom to (0.5, 0). Thus you need
      * a suited distorted texture.
      *
-     * @param axisSamples
-     *            Number of triangle samples along the axis.
-     * @param radialSamples
-     *            Number of triangle samples along the radial.
+     * @param axisSamples The number of vertices samples along the axis. It is equal to the number of segments + 1; so 
+	 * that, for instance, 4 samples mean the cylinder will be made of 3 segments.
```
Nehon: Please fix the extra space in javadoc... you must have issues with your formater if any.

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
```diff
@@ -201,194 +198,240 @@ public boolean isInverted() {
...
+	 * @param bottomRadius the radius of the bottom of the cylinder.
      * @param height the cylinder's height.
      * @param closed should the cylinder have top and bottom surfaces.
      * @param inverted is the cylinder is meant to be viewed from the inside.
      */
     public void updateGeometry(int axisSamples, int radialSamples,
-            float radius, float radius2, float height, boolean closed, boolean inverted) {
-        this.axisSamples = axisSamples;
+            float topRadius, float bottomRadius, float height, boolean closed, boolean inverted)
+	{
```
Nehon: brackets **always** at the end of the line.
May sounds like nitpicking, but it's a matter of consistency in the code base.

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
```diff
@@ -201,194 +198,240 @@ public boolean isInverted() {
...
+		// Ensure there's at least two axis samples and 3 radial samples, and positive dimensions.
+		if( axisSamples < 2
+			|| radialSamples < 3
+			|| topRadius <= 0
+			|| bottomRadius <= 0
+			|| height <= 0 )
+			throw new IllegalArgumentException("Cylinders must have at least 2 axis samples and 3 radial samples, and positive dimensions.");
```
Nehon: **Always** put brackets even when there is only one statement in a if, for, while, or whatever.

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
```diff
@@ -201,194 +198,240 @@ public boolean isInverted() {
...
+		// The normal in A and D is simply orthogonal to AD, which means we can get it once per sample.
+		//
+		Vector3f[] circleNormals = new Vector3f[radialSamples+1];
+		for (int circlePoint = 0; circlePoint < radialSamples+1; circlePoint++)
+		{
+            // The normal is the orthogonal to the side, which can be got without trigonometry.
+			// The edge direction is oriented so that it goes up by Height, and out by the radius difference; let's use
+			// those values in reverse order.
+			Vector3f normal = new Vector3f(height * circlePoints[circlePoint][0], height * circlePoints[circlePoint][1], bottomRadius - topRadius );
+			circleNormals[circlePoint] = normal.normalizeLocal();
```
Nehon: wow format went crazy here, please fix

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
```diff
@@ -201,194 +198,240 @@ public boolean isInverted() {
...
-        float[] sin = new float[radialSamples + 1];
-        float[] cos = new float[radialSamples + 1];
-
-        for (int radialCount = 0; radialCount < radialSamples; radialCount++) {
-            float angle = FastMath.TWO_PI * inverseRadial * radialCount;
-            cos[radialCount] = FastMath.cos(angle);
-            sin[radialCount] = FastMath.sin(angle);
+        // Vertices : One per radial sample plus one duplicate for texture closing around the sides.
+        int verticesCount = axisSamples * (radialSamples +1);
+		// Triangles: Two per side rectangle, which is the product of numbers of samples.
```
Nehon: why the indentation??

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
Saucistophe: Sorry about all that.
I usually use tabs and I did not check that it was the case, my bad.

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
Saucistophe: Tabs as well.

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
Saucistophe: Okay!

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
Saucistophe: Okay!

jme3-core/src/main/java/com/jme3/scene/shape/Cylinder.java
Nehon: ha ok ... yeah usually we format tabs as 4 spaces.

Saucistophe:
I added a test case; but what's strange is that I can't seem to make if fail with the old implementation of the cylinder.
If you check #640 , you'll find the code I've used to raise exceptions once out of ten; for some reason, the unit test is always OK. I think it's useful all the same, but you can dispose of it.

empirephoenix:
@Nehon does this work for you now?

Nehon:
It looks good, but I don't have time to test it, so one last question @Saucistophe : is it fully reto compliant with the old class?

Saucistophe:
@Nehon I've tested (visually) all the possible combinations on the following conditions:

- 2, 3, and 10 axis samples
- 3, 4 and 10 radial samples
- Inverted/not inverted
- Closed/not closed

I only tested with two random radiuses.
I do *not* have the same result for textures on closed cylinders: I spread equally the vertical coordinates of textures along the caps and sides.
Before, if you had 3 axis samples and 2 caps, the bottom was at 0, the bottom edge at 0.2, then 0.4, 0.6, 0.8 on sampling points, then 1 on top cap.
Now, it depends of the radii: if the bottom radius is 0.5, height 0.3 and top 0.2, you'll get 0, 0.5, 0.8 and 1.
I'm not sure this is clear - nor a good idea for that matter - but it does smooth the texture for thin and thick cylinders. Would you want a screen capture?

empirephoenix:
Yes, please add a capture. Changing the texture coordinates is not a really big issue in my opinion, as cylinders are usual not sued ingame anyway, but more for debugging or simplistic scenes

Saucistophe:
image
There you have it.

The new one is on the left. Notice how the sides are roughly one third of the surface, and one third of the texture. The cap is big, it uses a big part of the texture.
On the old one, the sides always use 3 quarters of the texture and the remaining quarter for the caps.

The old one always kept the same texture ratio but has poor repartition of details, the new one has no predefined ratio but provides better repartition of details.

empirephoenix:
I would say the left on is clearly better

Nehon:
Yeah I agree... though this means it's gonna break user code...
That said, I'm not sure a lot of users are using stock cylinder anyway... so let's merge this.

empirephoenix:
Alright :) Thanks for the work @Saucistophe

Saucistophe:
@Nehon My thoughts exactly. If someone complains, I can always add a flag or something.
@empirephoenix Thank you both for yours on the review :)

PR #673, opened 2017-06-22
added a possibility to handle shaders errors.

JavaSaBr:
Added a possibility to handle shaders errors.
I use this to handle incorrect shaders during shader editing in my Editor:
https://bitbucket.org/JavaSabr/jme3-spaceshift-editor/commits/d4128e6b06e0ae2ce16d845ada8fa7215819b6e7

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/CompileShaderException.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/CompileShaderException.java
@@ -32,56 +32,17 @@
 package com.jme3.renderer;
 
 /**
- * <code>RendererException</code> is raised when a renderer encounters
- * a fatal rendering error.
+ * <code>CompileShaderException</code> is raised when a renderer can't compile a shader.
  *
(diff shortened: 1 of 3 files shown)


empirephoenix:
Can you please reissue the pull request?  travis had some trouble yesterday, hopefully it will work today.

Nehon:
This is totally unneeded.
You just have to catch renderException and parser the exception message.
I'm against this change

JavaSaBr:
@Nehon Hi, so how can I fix a geometry which has an incorrect material in your case?

Nehon:
the geometry doesn't need fixing, the material does.
see this https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/util/MaterialDebugAppState.java

JavaSaBr:
@Nehon , wow, so I can use the:
`renderManager.preloadScene(dummyGeom);`
to validate a material, right? cool if it's true :+1:

PR #677, opened 2017-06-25
fixing documentation link

RemboL:
Noticed that this javadoc points to web documentation page that is not existing (probably from before migration), so I changed it to point to correct one.


empirephoenix:
Thanks for the work

Issue #357, opened 2015-09-29
Updating the frustum does not work as expected


slyh80:
There seems to be a problem when updating the frustum, camera and/or viewport.

Steps to reproduce:
1. Resize a viewport so it has another aspect ratio as before
2. Update camera aspect ratio
3. Notice that the resulting rendering is stretched

Please see this discussion for details, including code sample:
http://hub.jmonkeyengine.org/t/aspect-ratio-not-updated-when-resizing-viewport/33936/

slyh80:
I suspect class RenderManager (line 334) is part of the problem.
```
        for (ViewPort vp : viewPorts) {
            if (vp.getOutputFrameBuffer() == null) {
                Camera cam = vp.getCamera();
                cam.resize(w, h, true);
            }
            notifyReshape(vp, w, h);
        }
```

Note: I'm a layman regarding jME's internals.
But resizing all viewports to the window's(!) width and height does not make too much sense for me. Right?

stephengold:
As I noted on the Forum today:

https://hub.jmonkeyengine.org/t/solved-aspect-ratios-of-cameras-in-a-split-screen-render/38922

Camera.resize() with fixAspect=true doesn't take the camera's viewport dimensions into account when calculating what the frustum's aspect ratio is. Since the renderer frequently invokes resize(w, h, true) on user cameras, the resulting renders appear squashed unless

viewPortTop - viewPortBottom == viewPortRight - viewPortLeft

This turns out to be the root cause of issue #357. I ran the test code (submitted by slyh on Sep 28, 2015) with jME 3.1.0-stable. I reproduced the reported issue, then patched Camera.resize() and verified that the patch resolved the issue.

grizeldi:
Interesting. I managed to make this work not too long ago on 3.1 stable without patching the engine, can't remeber how though.

empirephoenix:
fixed on master

stephengold:
`TestAspectRatio` is failing on jMonkeyEngine 3.2.3-v3.2-6660 on Linux, with the blee cube rendering as a rectangle instead of a square, even though that build includes the fix to Camera.resize().

stephengold:
Still failing on Linux as of Git hash 4b62b70 in the v3.3 branch.

stephengold:
Some analysis here: https://hub.jmonkeyengine.org/t/betterlensflare/42779/4

Markil3:
Some notes:
[...]

PR #679, opened 2017-07-07
testcases and fix for issue #357

stephengold:
Here's the fix I came up with for issue #357. See my discussion in the forum and on that issue.


empirephoenix:
Can you please add forum link?

stephengold:
https://hub.jmonkeyengine.org/t/solved-aspect-ratios-of-cameras-in-a-split-screen-render/38922?u=sgold

empirephoenix:
Thanks for the work

PR #680, opened 2017-07-08
Added missed gl versions

JavaSaBr:
I think we need to have a possibility to set core profile of most GL version, not only 3.2. Some drivers work correctly only in the case when you use a core profile, but when we use a core profile, we have limits to use features from higher versions. For example Intel Graphics on Linux, which can't use features from 3+ Open GL without a core profile. At this moment, we can use only core profile for the version 3.2, it isn't enough for example if you want to use geometry/tesselation shaders.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/Caps.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/Caps.java
@@ -129,6 +129,26 @@ public enum Caps {
      */
     OpenGL40,
     /**
+     * Supports OpenGL 4.1
+     */
+    OpenGL41,
+    /**
+     * Supports OpenGL 4.2
+     */
+    OpenGL42,
+    /**
+     * Supports OpenGL 4.3
+     */
+    OpenGL43,
+    /**
+     * Supports OpenGL 4.4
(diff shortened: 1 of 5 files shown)


JavaSaBr:
Any questions?

empirephoenix:
Yes, this seems useful, on one of my machines pretty much everything works with opengl, except jme3 due to useing a GL30 function that is not supported in the selected profile.

JavaSaBr:
@empirephoenix  For example, I have an issue with Intel Graphics on Linux, when I use 2.0 GL render, I have a lot of issues with GL features of 3.0+ version if I use core profile 3.2 version, all features of 3.2 version work perfect, but I can't use features from 3.3+ version. I didn't see any workaround for this :(
I have no the problem with nvidia on linux for example, but I think we can have the same problem with AMD else

JavaSaBr:
I see that without core profile all drivers on linux work how they want to work :D

JavaSaBr:
@empirephoenix any questions else? :)

empirephoenix:
Merged, thanks for the work

Issue #661, opened 2017-05-16
SkeletonDebugger assumes all bones extend in the local +Y direction


stephengold:
When bone lengths are provided to SkeletonDebugger, it draws them extending in the local +Y direction, which is fine for models like Oto which don't violate that assumption. However, for models like Elephant and Ninja,  which have bones extending in directions other than +Y, SkeletonDebugger with bone lengths renders bizarre and confusing results. To render bones properly for all models, SkeletonDebugger would need bone directions as well as lengths. In other words: vectors.

![ninja-skele](https://cloud.githubusercontent.com/assets/2287609/26121403/f0db05ba-3a28-11e7-8869-ee86944bf86e.png)
![skeletondebugger](https://cloud.githubusercontent.com/assets/2287609/26121414/f667cad6-3a28-11e7-9bad-26257ec477a0.png)

stephengold:
I'm curious why there hasn't been a pull request.

shadowislord:
@stephengold the fix PR was merged, does it work for you?

stephengold:
The change doesn't resolve my issue. Now the bones are all drawn extending in the local (1, 1, 1) direction, which is even less useful than having them drawn in the +Y direction.

shadowislord:
OK, I reverted it

Nehon:
This will not be the case in the future ArmatureDebugger

PR #681, opened 2017-07-08
Resolving issue in ticket #661.  Creating skelton debugger that allow…

Bekreth:
…s bones to be extended in any direction instead of only the Y direction.


empirephoenix:
looks fine for me, any objections?

PR #684, opened 2017-07-16
added support multi-threading allocations for LWJGL Buffer Allocator.

JavaSaBr:
I see java crashes from time to time with jemalloc when I have high concurrency with memory allocations, so I have added thread-safe implementation of the allocator by default, but if your application doesn't have high concurrency, you can disable thread-safe allocator using a system property to improve performance with it.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglContext.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglContext.java
@@ -32,10 +32,10 @@
 
 package com.jme3.system.lwjgl;
 
+import static com.jme3.util.LWJGLBufferAllocator.PROPERTY_CONCURRENT_BUFFER_ALLOCATOR;
 import static org.lwjgl.opencl.CL10.CL_CONTEXT_PLATFORM;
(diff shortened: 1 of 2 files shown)


JavaSaBr:
@empirephoenix any questions? :)

Nehon:
We can't merge that. StampedLock is java8 only.
JME code base must be java7 complient.

JavaSaBr:
@Nehon it's in the lwjgl3 module, it has separated compilation and only for java8+

Nehon:
mhhh I didn't know... I'm not sure why... Is lwjgl3 not compliant with java7?

tonihele:
It requires java 8 or later: https://github.com/LWJGL/lwjgl3

They are also already using some features from java 9 if you check the latest release notes (3.1.2).

Nehon:
ok... well then I guess I have no objections to this PR

empirephoenix:
Merged, thanks for the work

gouessej:
@Nehon Does it mean that we are allowed to use Java 1.8 features in the rest of the source code too?

Nehon:
Only in jme3-lwjgl3. But I don't picture you, fiddling in that code 😉. 
I'd rather the rest of the engine remains Java 7 compliant.

PR #688, opened 2017-07-23
Fix of the one parameter of shadows render. 

JavaSaBr:
The parameter renderBackFacesShadows is true by default, but it doesn't have effects because the logic of making changes is inside in the setter of this parameter. So I have added a call of the setter during initialization.


Nehon:
mhh...  I wonder why I made it a Boolean with a big B...
Could you change it to a classic boolean?

JavaSaBr:
@Nehon done :)

Nehon:
Thanks

PR #689, opened 2017-07-23
Added alpha blending option to terrain lightings to allow alpha stencil textures.

NemesisMate:
More information [here](https://hub.jmonkeyengine.org/t/splatting-textures-with-alpha-using-terrainlighting/38459).


Nehon:
Thanks!!

PR #693, opened 2017-07-28
Update the water filter

JavaSaBr:
Just extended availability to edit/save/load the water filter.

### DIFF
--- a/jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
+++ b/jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
@@ -31,6 +31,7 @@
  */
 package com.jme3.water;
 
+import com.jme3.asset.AssetKey;
 import com.jme3.asset.AssetManager;
 import com.jme3.asset.TextureKey;
 import com.jme3.export.InputCapsule;
@@ -66,18 +67,19 @@ import java.io.IOException;
  */
 public class WaterFilter extends Filter implements JmeCloneable, Cloneable {
 
+    public static final String DEFAULT_NORMAL_MAP = "Common/MatDefs/Water/Textures/water_normalmap.dds";
+    public static final String DEFAULT_FOAM = "Common/MatDefs/Water/Textures/foam.jpg";
+    public static final String DEFAULT_CAUSTICS = "Common/MatDefs/Water/Textures/caustics.jpg";
+    public static final String DEFAULT_HEIGHT_MAP = "Common/MatDefs/Water/Textures/heightmap.jpg";
+
     private Pass reflectionPass;
     protected Spatial reflectionScene;
     protected Spatial rootScene;
     protected ViewPort reflectionView;
     private Texture2D normalTexture;
-    private TextureKey normalTextureKey;
     private Texture2D foamTexture;
-    private TextureKey foamTextureKey;
     private Texture2D causticsTexture;
-    private TextureKey causticsTextureKey;
     private Texture2D heightTexture;
-    private TextureKey heightTextureKey;
     private Camera reflectionCam;
     private Vector3f targetLocation = new Vector3f();
     private ReflectionProcessor reflectionProcessor;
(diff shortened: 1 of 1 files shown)


JavaSaBr:
@Nehon @pspeed42 what do you think about this? :)

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
```diff
@@ -382,6 +405,14 @@ public void read(JmeImporter im) throws IOException {
 
         useCaustics = ic.readBoolean("useCaustics", true);
 
+        causticsTexture = (Texture2D) ic.readSavable("causticsTexture", null);
+        heightTexture = (Texture2D) ic.readSavable("heightTexture", null);
+        normalTexture = (Texture2D) ic.readSavable("normalTexture", null);
+        foamTexture = (Texture2D) ic.readSavable("foamTexture", null);
+        needSaveReflectionScene = ic.readBoolean("needSaveReflectionScene", false);
+        reflectionScene = (Spatial) ic.readSavable("reflectionScene", null);
```
Nehon: Read the texture path, load the texture if it's relevant, else keep the texture null.

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
```diff
@@ -292,8 +303,20 @@ protected void cleanUpFilter(Renderer r) {
...
+
         OutputCapsule oc = ex.getCapsule(this);
 
+        final Spatial reflectionScene = getReflectionScene();
+        final boolean needSaveReflectionScene = isNeedSaveReflectionScene();
+
+        oc.write(causticsTexture, "causticsTexture", null);
+        oc.write(heightTexture, "heightTexture", null);
+        oc.write(normalTexture, "normalTexture", null);
+        oc.write(foamTexture, "foamTexture", null);
```
Nehon: Don't save the textures data. save the texture path and load them in the read method
Also you'll have to do something about this https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-effects/src/main/java/com/jme3/water/WaterFilter.java#L230
only load the default textures if the texture2D variables are null.

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
```diff
@@ -120,7 +122,9 @@
     private Vector3f center;
     private float radius;
     private AreaShape shapeType = AreaShape.Circular;
-    
+
+    private boolean needSaveReflectionScene;
```
Nehon: Mhh.... why do we need this?

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
```diff
@@ -292,8 +303,20 @@ protected void cleanUpFilter(Renderer r) {
...
 
+        final Spatial reflectionScene = getReflectionScene();
+        final boolean needSaveReflectionScene = isNeedSaveReflectionScene();
+
+        oc.write(causticsTexture, "causticsTexture", null);
+        oc.write(heightTexture, "heightTexture", null);
+        oc.write(normalTexture, "normalTexture", null);
+        oc.write(foamTexture, "foamTexture", null);
+        oc.write(needSaveReflectionScene, "needSaveReflectionScene", false);
+        oc.write(needSaveReflectionScene ? reflectionScene : null, "reflectionScene",
```
Nehon: This may break j3o/j3f backward compatibility.
Some people may have already saved post processes without the reflection scene. It would be nice to make sure those files still loads after this change. (to test it use the sdk and save a j3f file with a water filter, then try to load it in the engine)

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
JavaSaBr: the filter doesn't save the scene at all.

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
JavaSaBr: I'm sure that it doesn't break back compatibility, but I will recheck it again ;)

JavaSaBr:
@Nehon I have changed saving textures to use a texture keys, also I have re-checked back compatibility, it works fine ;)

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
Nehon: Yes I know, my point, is why do we need the flag? why not save it in any case.
Also... saving the scene here may be a problem for j3f in the SDK. Technically J3o works with object references, so an object data is effectively saved once even if it's referenced multiple times. But in case of j3f only the filter stack is saved, doing this you might save the whole scene into a j3f... Of course it's not a problem if you save all in one j3o...but the sdk doesn't do that.

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
```diff
@@ -70,11 +71,14 @@
     protected Spatial rootScene;
     protected ViewPort reflectionView;
     private Texture2D normalTexture;
+    private TextureKey normalTextureKey;
```
Nehon: You don't need to keep the texture key as a class field.

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
JavaSaBr: @Nehon so, I have added the flag to have a possibility to save a reflection scene if a user knows what he does¸ but in the most cases, we shouldn't do it. I need to have it in some case when the filter is a part of some scene.

JavaSaBr: @Nehon by default the filter doesn't save a scene.

jme3-effects/src/main/java/com/jme3/water/WaterFilter.java
```diff
@@ -754,6 +777,7 @@ public void setColorExtinction(Vector3f colorExtinction) {
      */
     public void setFoamTexture(Texture2D foamTexture) {
         this.foamTexture = foamTexture;
+        this.foamTextureKey = (TextureKey) foamTexture.getKey();
```
Nehon: Don't do this, just fetch the TextureKey in the write method when saving.

JavaSaBr:
@Nehon updated again :D

JavaSaBr:
@Nehon ok ;)

PR #696, opened 2017-08-01
fixed memory leak in terrains, optimized updating normals.

JavaSaBr:
Fixed the problem in LoD control:
https://hub.jmonkeyengine.org/t/terrain-editor-broken/38952/22
and one optimization in the method to update normals.


JavaSaBr:
@Nehon who is main reviewer changes in terrain system? :)

MeFisto94:
Well I don't think there is a dedicated maintainer for the terrain system.
However I think @pspeed42 might have a look since he implemented the new cloning system.

Btw it would've been better if you created two separate pull requests, one for the memory leak and one for the optimization, I guess, so it's more clear and could be merged separately.

JavaSaBr:
@MeFisto94 I think it's clear enough because these 2 fixes are in separated files and there aren't a lot of changes.

JavaSaBr:
@pspeed42 could you review these changes, please?

PR #702, opened 2017-08-08
Update com_jme3_bullet_objects_PhysicsRigidBody.cpp

Nebloksiam:
Reset deactivation timer.

From docu here: http://www.bulletphysics.org/mediawiki-1.5.8/index.php/Activation_States
..
ACTIVE_TAG
    Means active so that the object having the state could be moved in a step simulation. This is the "normal" state for an object to be in. Use btCollisionObject::activate() to activate an object, not btCollisionObject::setActivationState(ACTIVATE_TAG), or it may get disabled again right away, as the deactivation timer has not been reset. 
..


riccardobl:
I think it might be better to rebase it in a single commit if possible.

Nehon:
Github allows you to squash the commits when you merge a PR now.

PR #703, opened 2017-08-10
Added Inertia Manipulation Methods

empirephoenix:
because I want to get and set inertia


empirephoenix:
Hm whats up with the build servers again? Also do I see correctly that we finally can do windows specific builds trough appveyor? Nice :)

PR #704, opened 2017-08-14
Fixed sky shader

JavaSaBr:
Now you can't create a material file of any type for sky, because sky.j3md has the only TextureCubeMap type of used texture even if you don't want to use cube map texture.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/util/SkyFactory.java
+++ b/jme3-core/src/main/java/com/jme3/util/SkyFactory.java
@@ -44,8 +44,8 @@ import com.jme3.texture.Image;
 import com.jme3.texture.Image.Format;
 import com.jme3.texture.Texture;
 import com.jme3.texture.TextureCubeMap;
+
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
 
 /**
  * <code>SkyFactory</code> is used to create jME {@link Spatial}s that can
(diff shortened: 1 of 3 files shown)


JavaSaBr:
@Nehon could you review these changes, please? :)

Nehon:
I'm not fond of this change at all. It jut hack around the fact that we don't infer the texture type in the J3md.
- you don't need this if you use the sky factory, so you'll have to give me a very solid use case.
- If you ever find a solid use case, This should be done otherwise: Allow "Texture" as a type in J3md and infer the type at run time when the parameter is set.

JavaSaBr:
@Nehon Only one use case, I think we should have a possibility to present any material as j3m file, that all. If you think that it isn't enough for you, up to you, I can make my own j3md file for my cases. :)

Nehon:
Sky is peculiar and is supposed to work with the SkyFactory. I can't see why you'd want to save j3m. Especially that you can save a sky in a j3o file like any other spatial...

JavaSaBr:
@Nehon so ok, but my changes don't break anything and I think you can close this PR ;)

JavaSaBr:
@Nehon by the way, in your case a spatial stores data of a sky texture inside itself, I think it isn't the good option.

Nehon:
No, as long as a texture has a TextureKey, it's not saved along with the rest in the j3o.

JavaSaBr:
@Nehon ok

PR #705, opened 2017-08-19
Fixed: If Instancing is activated you only see shadows

b00nation:
When the pbr material was introduced someone forgot to keep the instancing in mind which uses :
            ViewProjectionMatrix
            ViewMatrix
I added these.


Nehon:
Thanks

PR #708, opened 2017-08-25
fixed two problems with asset manager:

JavaSaBr:
1. can't load classes from asset class loaders.
2. can't load resources from asset class loaders.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/export/SavableClassUtil.java
+++ b/jme3-core/src/main/java/com/jme3/export/SavableClassUtil.java
@@ -34,6 +34,7 @@ package com.jme3.export;
 import com.jme3.animation.Animation;
 import com.jme3.effect.shapes.*;
 import com.jme3.material.MatParamTexture;
+
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


JavaSaBr:
@empirephoenix @Nehon could you review this PR, please?

Nehon:
I don't get what this fixes, could you please explain the issue a bit more?

JavaSaBr:
@Nehon for example, I have a j3o model which includes materials with shaders and controls which don't exist in the main class loader, so when I try to load it, I have class not found exceptions and asset not found exceptions, because the process of loading assets from classpath doesn't look at additional class loaders of asset manager. The same issue with loading controls. 
Is it clear? :)

Nehon:
Ok I get it now... that actually looks like a bug...

JavaSaBr:
@Nehon thanks :)

Nehon:
Not merging it for now, waiting for others to review. If it's not merged in a couple of days, please ping me.

JavaSaBr:
@Nehon ok

empirephoenix:
} catch (InstantiationException e) {
                  } catch (IllegalAccessException e) {		                  } catch (IllegalAccessException e) {
                  }		                  }

What about this? as this would hide problems with initializers. I totally understand the need to do the continue for the not found case, but if a class could be loaded, why ignore errors on init?
Else than this I think this change makes sense.

JavaSaBr:
@empirephoenix I just wanted to make minimal changes, I didn't look at the code block :)

Nehon:
Seems that the code was already doing that... It's weird, that those exception are not re thrown...
IMO they were supposed to be thrown at first and has been silenced later for more proof loading....

JavaSaBr:
@Nehon I think changes of this code block should be in a separated PR ;)

Nehon:
Agree

empirephoenix:
Well fi that was the prior behaviour, I see no reason against this.

PR #709, opened 2017-08-27
Update ParticleEmitter.java

empirephoenix:
allow easy manipulation of particleemitter via subclasses, eg in my case I want a start color, end color and midcolor.


quazi-irfan:
>  start color, end color and midcolor.

@empirephoenix Must be a cool one to see :)

Issue #710, opened 2017-08-29
collideWith() sometimes reports collisions with phantom triangles


stephengold:
Mesh.collideWith() caches a BIHTree for any mesh containing 1 or more vertices, then uses the BIHTree to detect collisions. If the mesh does not consist of triangles (if the mode is Line or Points or anything but Triangles/TriangleStrip/TriangleFan) it generates a BIHTree anyway, treating each triad of vertices in the WrappedIndexBuffer as if it were a triangle. As a result, Mesh.collideWith() sometimes reports collisions with phantom triangles: triangles that don't exist in any mesh.

I believe I know how to correct this issue. I'll write a simple test case and generate a pull request.

stephengold:
Fixed in 'master'.

PR #711, opened 2017-08-29
test and fix for issue #710

stephengold:
I add a test-case for issue #710 and propose a simple fix.

### DIFF
--- a/jme3-core/src/test/java/com/jme3/scene/PhantomTrianglesTest.java
+++ b/jme3-core/src/test/java/com/jme3/scene/PhantomTrianglesTest.java
@@ -45,10 +45,11 @@ import org.junit.Test;
 
 /**
  * Verify that collideWith() doesn't reports collisions with phantom triangles.
+ * This was issue #710 at GitHub.
  *
  * @author Stephen Gold
  */
-public class Issue770Test {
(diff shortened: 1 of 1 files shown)


shadowislord:
I think it might make more sense to create a unit test for this instead of an example program. In this case there's nothing interesting going on visually and a unit test is automated.

Also it seems like a bunch of commits from master are appearing on this PR.. maybe you need to merge master into it?

jme3-core/src/test/java/com/jme3/scene/Issue770Test.java
```diff
@@ -0,0 +1,127 @@
...
+import com.jme3.math.Vector3f;
+import com.jme3.scene.shape.Quad;
+import org.junit.Test;
+
+/**
+ * Verify that collideWith() doesn't reports collisions with phantom triangles.
+ *
+ * @author Stephen Gold
+ */
+public class Issue770Test {
```
Nehon: I thought it was Issue 710?
Also maybe a more explicit name could help... we don't have direct access to github from code and issue 710 is not very explicit.

Nehon:
To remove the commits you didn't do you need to rebase your branch on master.
If you want I can guide you through this, ping me on slack when you are available :)

PR #715, opened 2017-09-03
Fixed NPE and ConcurrentModificationException

JavaSaBr:
Fixed NPE and ConcurrentModificationException during cloning TechniqeDef.


JavaSaBr:
@Nehon @empirephoenix could you review this, please?

empirephoenix:
as far I understand this, before they had shared lists between them? makes sense to create new ones then.

JavaSaBr:
@empirephoenix in the case with ShaderGenerationInfo, I had the same instances of lists in original ShaderGenerationInfo and in cloned version.

Nehon:
mhh that's weird I never had this issue....
Anyway yeah they should be cloned

PR #717, opened 2017-09-08
Use MPOs for skinning

shadowislord:
Fixes the annoying "Material instances cannot be shared when hardware skinning is used" error

### DIFF
--- a/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
+++ b/jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
@@ -32,8 +32,7 @@
 package com.jme3.animation;
 
 import com.jme3.export.*;
-import com.jme3.material.MatParam;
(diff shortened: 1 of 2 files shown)


jme3-core/src/main/java/com/jme3/animation/SkeletonControl.java
```diff
@@ -547,10 +507,9 @@ private void applySkinning(Mesh mesh, Matrix4f[] offsetMatrices) {
         fnb.rewind();
 
         // get boneIndexes and weights for mesh
-        indexReader.setBuffer(mesh.getBuffer(Type.BoneIndex).getData());
+        IndexBuffer ib = IndexBuffer.wrapIndexBuffer(mesh.getBuffer(Type.BoneIndex).getData());
```
Nehon: Won't that create a new instance of the wrapper on each frame ?

PR #724, opened 2017-09-10
Migrate material things to use jme clonable system

JavaSaBr:
I like Jme clone system and savable objects system and I like to use it with all jME objects, but I saw that MaterialDef's things don't implement it. So I have migrated these things to use JmeClonable system and implemented JmeClonable/Savable interfaces for all these things.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/Material.java
+++ b/jme3-core/src/main/java/com/jme3/material/Material.java
@@ -85,7 +85,7 @@ public class Material implements CloneableSmartAsset, JmeCloneable, Savable {
     private MaterialDef def;
     private ListMap<String, MatParam> paramValues = new ListMap<String, MatParam>();
     private Technique technique;
-    private HashMap<String, Technique> techniques = new HashMap<String, Technique>();
+    private HashMap<String, Technique> techniques = new HashMap<>();
     private RenderState additionalState = null;
     private RenderState mergedRenderState = new RenderState();
     private boolean transparent = false;
@@ -192,16 +192,19 @@ public class Material implements CloneableSmartAsset, JmeCloneable, Savable {
 
     @Override
     public Material clone() {
-        try {
-            return (Material) super.clone();
-        } catch (CloneNotSupportedException e) {
-            throw new RuntimeException(e);
-        }
+        final Cloner cloner = new Cloner();
+        final Material clone = jmeClone();
+        clone.cloneFields(cloner, this);
+        return clone;
     }
 
     @Override
     public Material jmeClone() {
-        return clone();
+        try {
+            return (Material) super.clone();
+        } catch (final CloneNotSupportedException e) {
+            throw new RuntimeException(e);
+        }
     }
 
     @Override
@@ -218,6 +221,7 @@ public class Material implements CloneableSmartAsset, JmeCloneable, Savable {
             paramValues.put(entry.getKey(), cloner.clone(entry.getValue()));
         }
 
+        techniques = new HashMap<>();
         technique = null;
         sortingId = -1;
     }
@@ -1068,7 +1072,7 @@ public class Material implements CloneableSmartAsset, JmeCloneable, Savable {
         }
 
         if (defName != null) {
-            def = im.getAssetManager().loadAsset(new AssetKey<>(defName));
+            def = (MaterialDef) im.getAssetManager().loadAsset(new AssetKey<>(defName));
         } else {
             def = (MaterialDef) ic.readSavable("material_def_embedded", null);
         }
--- a/jme3-core/src/main/java/com/jme3/material/Material.java
+++ b/jme3-core/src/main/java/com/jme3/material/Material.java
@@ -55,6 +55,8 @@ import com.jme3.texture.Texture;
 import com.jme3.texture.image.ColorSpace;
 import com.jme3.util.ListMap;
 import com.jme3.util.SafeArrayList;
+import com.jme3.util.clone.Cloner;
+import com.jme3.util.clone.JmeCloneable;
 
 import java.io.IOException;
 import java.util.*;
@@ -72,7 +74,7 @@ import java.util.logging.Logger;
  *
  * @author Kirill Vainer
  */
-public class Material implements CloneableSmartAsset, Cloneable, Savable {
+public class Material implements CloneableSmartAsset, JmeCloneable, Savable {
 
     // Version #2: Fixed issue with RenderState.apply*** flags not getting exported
     public static final int SAVABLE_VERSION = 2;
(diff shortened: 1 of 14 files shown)


JavaSaBr:
@Nehon could you look at this PR, please? :)

jme3-core/src/main/java/com/jme3/renderer/Caps.java
```diff
@@ -396,6 +397,15 @@
      */
     BinaryShader;
 
+    private static final Caps[] VALUES = values();
```
Nehon: Why this? What's the use case?
Why not just let the user do enum.values()[i] and handle the bounds check?

jme3-core/src/main/java/com/jme3/material/Material.java
```diff
@@ -990,7 +996,15 @@ public void render(Geometry geom, RenderManager rm) {
...
     public void write(JmeExporter ex) throws IOException {
         OutputCapsule oc = ex.getCapsule(this);
-        oc.write(def.getAssetName(), "material_def", null);
+
+        final String defAssetName = def.getAssetName();
+
+        oc.write(defAssetName, "material_def", null);
+
+        if (defAssetName == null) {
+            oc.write(def, "material_def_embedded", null);
```
Nehon: I am strongly against this.
The material definition is external to the scene graph. By doing this you allow matdefs to be saved along with a scene and be modified inside it, without having a reference j3md. If this had been the intention in the beginning material and material def wouldn't be separate.

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
```diff
@@ -195,4 +200,85 @@ public void addTechniqueDef(TechniqueDef technique) {
...
+
+        techniques = new HashMap<>();
+
+        for (final Map.Entry<String, List<TechniqueDef>> entry : oldTechniques.entrySet()) {
+            techniques.put(entry.getKey(), cloner.clone(entry.getValue()));
+        }
+    }
+
+    @Override
+    public void write(final JmeExporter ex) throws IOException {
```
Nehon: No. Matdef should be stored as j3md.

jme3-core/src/main/java/com/jme3/renderer/Caps.java
JavaSaBr: enum.values() recalculetes for each call.

jme3-core/src/main/java/com/jme3/material/Material.java
JavaSaBr: It needs for some cases when I generate a mat def on runtime, do you know other options of this?

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
JavaSaBr: you are right for most cases, but not for all cases :(

JavaSaBr:
@Nehon I need to have a possibility to save material def using binary exporter, and JmeClonable works more transparently for me in the case, so I sure that saving/reading the material def will be correct.

jme3-core/src/main/java/com/jme3/renderer/Caps.java
Nehon: Ok... but what's the use case?

jme3-core/src/main/java/com/jme3/material/Material.java
Nehon: This has nothing to do with serialization. Generating a MaterialDef at runtime is just a matter of creating a new MaterialDef.

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
Nehon: I am right in every case. It's been already discussed in the core team. Binary serialization should contain a scene. By doing this you open the special case where you could save only a materialDef in a j3o, which makes no sense and is very confusing for users. Save your materialDefs as j3md. There is even a serializer that does this in core. Look for J3mdExporter.

Nehon:
You think you need it, but you don't. I've made a Node editor, without changing a single line in the core engine. Everything is in place to make it work. The engine shouldn't be modified to fot the editors, it's the other way around.

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
JavaSaBr: @Nehon I don't talk about serialization j3md to j3o ;) you can use binary exporter to serialize jme objects to other places.

Nehon: Then this has nothing to do in the engine.

jme3-core/src/main/java/com/jme3/material/Material.java
JavaSaBr: @Nehon ok, how can I store a material with a generated material def? I can't store the generated material def as file... so the material can't use the material def without asset name.

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
Nehon: You often tend to change things in the engine that are only very particular uses cases that wouldn't be needed in a regular game. This increase a lot the clutter of code in the engine and is wrong in many ways. If you have special use case work around them in your own projects.

jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
```diff
@@ -715,6 +816,46 @@ public void read(JmeImporter im) throws IOException{
         usesNodes = ic.readBoolean("usesNodes", false);
         shaderNodes = ic.readSavableArrayList("shaderNodes", null);
         shaderGenerationInfo = (ShaderGenerationInfo) ic.readSavable("shaderGenerationInfo", null);
+
+        final int worldBindOrdinals[] = ic.readIntArray("worldBinds", EMPTY_INT_ARRAY);
+
+        worldBinds = new ArrayList<>(worldBindOrdinals.length);
+
+        for (final int ordinal : worldBindOrdinals) {
+            worldBinds.add(UniformBinding.valueOf(ordinal));
```
JavaSaBr: @Nehon it's an example of using Enum.valueOf(index);

jme3-core/src/main/java/com/jme3/renderer/Caps.java
JavaSaBr: @Nehon I have answered as a comment in the changes.

JavaSaBr:
@Nehon I agree with you about we don't need to make changes special for Editor, but these changes don't include specific things for my Editor, I just implemented two standard engine interfaces for some engine objects, that made it more flexible to some use cases.

jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
Nehon: Well if it's for serialization then it's not a good example. Doing it right would be to add the corresponding method in outputcapsule to save a list of Enum, and to save the values.
Saving the ordinal is very bad practice as if the order changes one day into the enum it would break the old models loading. We already have this in other places in the engine and it's very bad. Let's not do the same mistakes over and over again.
Anyway. This PR won't make it to core.

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
JavaSaBr: @Nehon I see only one work around, I need to copy-paste all material's classes :(

jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
JavaSaBr: @Nehon I can remake it to use transform from enum list to string list...

jme3-core/src/main/java/com/jme3/material/Material.java
Nehon: just set the assetName yes. that's the way it's intended to work. And save the material def as a j3md file.

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
Nehon: Well then you have to think more about how you could redesign your editor so that you don't need to do that... I really don't get what you are doing.

jme3-core/src/main/java/com/jme3/material/Material.java
JavaSaBr: @Nehon but I can't save j3md as file :(

jme3-core/src/main/java/com/jme3/material/MaterialDef.java
JavaSaBr: @Nehon I just want to have binary serialization for j3md, I didn't think that is a big problem for you, now I see that you have the strong position about it.

jme3-core/src/main/java/com/jme3/material/Material.java
pspeed42: re: "but I can't save j3md as file :("
Why can't you save it as a .j3md file as already discussed?

JavaSaBr: @pspeed42 because I build it on runtime... so I need to save it as "randomvalue".j3md in the asset, but it will be like trash file. I want to avoid it.

Nehon: Why would you save it before the user explicitly save it and provide a name?

JavaSaBr: @Nehon the main difference between your shader node editor and mine is that my editor edits the file like "shader node project file" which includes last preview material, last version of j3md and some additional things. Of course in the worst case, I can do it using some crap hack solution, but I think when some jme objects can be saved/cloned and other can't be, it's unexpected behavior for a user. I understand that you don't like to support saving/cloning of j3md things, but my point of view, all jme objects should implement JmeClonable and Savable interfaces.

Nehon: You don't need it to implement a proper editor. And if you have to hack around in your code to not clutter the core engine, so be it.

PR #727, opened 2017-09-13
Blender Importer: Load HDR and many other

MeFisto94:
When a blender file contained a hdr image, for example, the engine is able to load that as a texture, however the blender importer didn't do so.
This was because it didn't rely on the AssetManager to load images but instead called some Loaders directly.

This pull request fixes this by using the AM, which means any (even user-) registered AssetLoader can be used to load textures from blender files now.

I've made an API Change though: ImageLoader is now more of a TextureLoader. The AssetManager can only hand back `Texture` objects and since there is no point in converting:  `Texture->Image->Texture`, I added loadTexture() and deprecated loadImage.

Technically loadImage could be removed as well since it's never called, but I kept it in for now.


jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/ImageLoader.java
```diff
@@ -60,76 +79,58 @@
...
-                break;
-            default:
-                throw new IllegalStateException("Unknown image type: " + imageType);
+    public Texture loadTexture(AssetManager assetManager, BlenderInputStream inputStream, int startPosition, boolean flipY) {
+        inputStream.setPosition(startPosition);
+        TextureKey tKey;
+        Texture result = null;
+
+        Logger hdrLogger = Logger.getLogger(HDRLoader.class.getName());
+        hdrLogger.setLevel(Level.SEVERE); // When we bruteforce try HDR on a non hdr file, it prints unreadable chars
```
JavaSaBr: I think it should be static field...

Nehon:
@MeFisto94 Could you rebase this PR on master please?

MeFisto94:
I knew this wouldn't become easy :D I really tried to rebranch. Please tell me if we can just squash merge the changes or not. If not, I'll just create a new branch and cherry pick the commit there and manually solve the conflicts.

jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/ImageLoader.java
MeFisto94: Not sure if the assetManager official supports MultiThreading but making result static would definitely disallow us to load multiple blender textures at once

Nehon:
ha boy... what did you do?

MeFisto94:
I have no clue. I tried to rebase to master, but the UI wanted me to solve conflicts three times (rebasing always stopped and I had to re-continue it) :D

jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/ImageLoader.java
Nehon: I think he's talking about the logger that can definitely be static even in a multi threading context

Nehon:
yeah well, with git, when you are in the situation to fix conflicts on a file that you never changed in your PR that means that you did something wrong and you should do gir rebase --abort.
Now it's a bit too late though.
So yeah I think you can create a new branch an cherry-pick it's gonna be easier

PR #730, opened 2017-09-15
Fix #456

### LINKED ISSUE #456: Distribution package does not include jbullet dependencies
When generating the distribution package, the `jme3-examples:dist` task does not copy `jbullet.jar` and `stack-alloc.jar` into the `dist/lib` folder. It seems this happens because Gradle does not consider local JAR dependencies as dependencies of the project.

shadowislord:
When using a `flatDir` repository Gradle is able to copy the dependencies correctly


Nehon:
good to merge?

PR #733, opened 2017-09-16
Fixes #731: BatchNode now support batching geometries without normals

### LINKED ISSUE #731: BatchNode.java: NullPointer with Mesh without normals
NullPointer thrown if a mesh without normals is added to BatchNode
```
protected void updateSubBatch(Geometry bg) {
        Batch batch = batchesByGeom.get(bg);
        if (batch != null) {
            Mesh mesh = batch.geometry.getMesh();
            Mesh origMesh = bg.getMesh();
            VertexBuffer pvb = mesh.getBuffer(VertexBuffer.Type.Position);
            FloatBuffer posBuf = (FloatBuffer) pvb.getData();
            VertexBuffer nvb = mesh.getBuffer(VertexBuffer.Type.Normal);
            FloatBuffer normBuf = (FloatBuffer) nvb.getData();  // <- NPE
```

NemesisMate:


Nehon:
ha... I was working on a fix....You beat me to it.
However i'd like to do it a more generic way and not have 2 versions of the doTransform method, one with tangents and one without.
Also in your change you iterate twice over the buffers instead of once. This code is executed on each frame so it has to be as fast as possible.

NemesisMate:
Oh!, of course, I just saw the issue on github and decided to post the solution I'm using for a LemurBatchNode (gui node's panels hasn't got any normals) but yeah, I'm agree it can be done better. :P

Nehon:
yep saw the issue this morning too.
Thanks a lot to have jumped on this though ;)
I'm gonna try to simplify the code.

Nehon:
So here is the fix I pushed
https://github.com/jMonkeyEngine/jmonkeyengine/commit/ef5952bcad5696ed39b296875158b1d1667112ae
This way we have a lot less code for each case, and still only one iteration over the buffers.

Nehon:
hanks for the contribution nonetheless

PR #735, opened 2017-09-17
Some fixes in materail def things.

JavaSaBr:
1. Small fixes with indents
2. The important fix with working of varying variables.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/shader/Glsl100ShaderGenerator.java
+++ b/jme3-core/src/main/java/com/jme3/shader/Glsl100ShaderGenerator.java
@@ -234,27 +234,48 @@ public class Glsl100ShaderGenerator extends ShaderGenerator {
         comment(source, shaderNode, "Begin");
         startCondition(shaderNode.getCondition(), source);
 
-        List<String> declaredInputs = new ArrayList<String>();
+        final List<String> declaredInputs = new ArrayList<>();
+
         for (VariableMapping mapping : shaderNode.getInputMapping()) {
+
+            final ShaderNodeVariable rightVariable = mapping.getRightVariable();
+            final ShaderNodeVariable leftVariable = mapping.getLeftVariable();
 
             //Variables fed with a sampler matparam or world param are replaced by the matparam itself
             //It avoids issue with samplers that have to be uniforms.
-            if (isWorldOrMaterialParam(mapping.getRightVariable()) && mapping.getRightVariable().getType().startsWith("sampler")) {
-                nodeSource = replace(nodeSource, mapping.getLeftVariable(), mapping.getRightVariable().getPrefix() + mapping.getRightVariable().getName());
+            if (isWorldOrMaterialParam(rightVariable) && rightVariable.getType().startsWith("sampler")) {
+                nodeSource = replace(nodeSource, leftVariable, rightVariable.getPrefix() + rightVariable.getName());
             } else {
-                if (mapping.getLeftVariable().getType().startsWith("sampler")) {
+
+                if (leftVariable.getType().startsWith("sampler")) {
                     throw new IllegalArgumentException("a Sampler must be a uniform");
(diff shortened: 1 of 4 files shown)


JavaSaBr:
@Nehon could you review this, please? :)

Nehon:
Ok... I don't have time lately, so I'm gonna give you 1 try.
The indent part is useless, The indent is there just to have a debug readable version of the shader, we don't really care if that's spaces or tabs.
Though i'm curious about the "important varying fix". What does this fix? There is no problem with varying and shader nodes so you'd better have found a huge bug.
As a rule of thumb, please explain your changes systematically when making a pull request.

JavaSaBr:
@Nehon about the fix of varying variables... I was wrong, sorry, I can't reproduce this problem without my fix in my editor, only in SDK editor. So I have reverted this change.

But I have pushed new changes about declaring variables:
When I add a shader node to technique definition without my changes I have the result:
https://imgur.com/jxu8D2U
you can see that the generator doesn't declare input variables of the shader node if we don't have input mappings for this, so we have an invalid shader in the result:
The same case with my changes:
https://imgur.com/lgbR3Lw
so now this shader is valid.

About indent... if you are against changes with formatting, of course, I can revert it, there are differences:
https://imgur.com/aPbVmP9
https://imgur.com/vHWubKH

Nehon:
Give me a j3md, and a jme3 test case outside of your editor. If the j3md is valid and if it fails to generate the shader properly, I'll consider your PR.

JavaSaBr:
@Nehon 
```

MaterialDef tempMatDef {
    MaterialParameters {
        Vector4 Color
    }

    Technique {
        WorldParameters {
            WorldViewProjectionMatrix
            Time
            Tpf
        }

        VertexShaderNodes {
            ShaderNode CommonVert {
                Definition: CommonVert: Common/MatDefs/ShaderNodes/Common/CommonVert.j3sn
                InputMappings {
                    worldViewProjectionMatrix = WorldParam.WorldViewProjectionMatrix
                    modelPosition = Global.position.xyz
                }
                OutputMappings {
                    Global.position = projPosition
                }
            }
        }

        FragmentShaderNodes {
            ShaderNode ConstructColor {
                Definition: ConstructColor: Common/MatDefs/ShaderNodes/Basic/ConstructColor.j3sn
                InputMappings {
                    red = WorldParam.Time
                }
                OutputMappings {
                }
            }
            ShaderNode ColorMult {
                Definition: ColorMult: Common/MatDefs/ShaderNodes/Basic/ColorMult.j3sn
                InputMappings {
                    color1 = MatParam.Color
                    color2 = ConstructColor.outColor
                }
                OutputMappings {
                    Global.color = outColor
                }
            }
        }

    }
}

```

JavaSaBr:
@Nehon You can try to load it as material in jME ;)

Nehon:
the problem is here: 
```
ShaderNode ConstructColor {
                Definition: ConstructColor: Common/MatDefs/ShaderNodes/Basic/ConstructColor.j3sn
                InputMappings {
                    red = WorldParam.Time
                    //MISSING INPUTS FOR green, blue and alpha
                }
                OutputMappings {
                }
            }
```
green, blue and alpha MUST be linked with input.
If they aren't you have to raise an error for the user to fix it.
Declaring the variable as your fix does will end in unexpected result and basically a silent fail of the shader...which is the worst thing the user can have.

JavaSaBr:
I understand this problem, but I don't agree that it's worth than exceptions... Also I'm going to add a possibility to define default values for unmapped input parameters, it will require the changes...

Nehon:
An error is worth an exception. 
There are enough silent fails when you debug shader to add this one.
Adding default values in the j3sn would be a nice addition. But it has to be done right.

JavaSaBr:
I'm going to create PRs step by step, I want to avoid big PRs.

PR #737, opened 2017-09-19
fixed NPE.

JavaSaBr:
Fixed NPE during saving MatParamOverride. I found that MatParamOverride can have a null value in some cases. For example, if you look at the constructor of SkeletonControl, you can see that this control creates two MPO with null values by default.
```

        this.numberOfBonesParam = new MatParamOverride(VarType.Int, "NumberOfBones", null);
        this.boneMatricesParam = new MatParamOverride(VarType.Matrix4Array, "BoneMatrices", null);
```

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/MatParam.java
+++ b/jme3-core/src/main/java/com/jme3/material/MatParam.java
@@ -37,6 +37,7 @@ import com.jme3.math.*;
 import com.jme3.shader.VarType;
 import com.jme3.texture.Texture;
 import com.jme3.texture.Texture.WrapMode;
+
 import java.io.IOException;
 
 /**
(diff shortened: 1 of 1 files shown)


JavaSaBr:
@Nehon what do you think about cases when an MPO has a null value?

jme3-core/src/main/java/com/jme3/material/MatParam.java
```diff
@@ -301,20 +302,17 @@ public void write(JmeExporter ex) throws IOException {
         OutputCapsule oc = ex.getCapsule(this);
         oc.write(type, "varType", null);
         oc.write(name, "name", null);
-        if (value instanceof Savable) {
-            Savable s = (Savable) value;
-            oc.write(s, "value_savable", null);
+        if (value == null) {
+        } else if (value instanceof Savable) {
```
Nehon: Please don't use empty if statements.
I'd rather have an early return like 
```
if(value==null){
    return;
}
if(value instanceof Savable){
...
```
Or at least put the if/else chain in a if (value != null) statement.

JavaSaBr: @Nehon it was a suggestion from IntelliJ IDEA. =/

Nehon: So? IDE's are dumb, you are in control of your code.

JavaSaBr:
@Nehon I have updated the condition.

MeFisto94:
Yeah the SDK is also affected by this. 
Just saving any previously converted model, even without displaying it, leads to something like:

```
java.lang.NullPointerException
	at com.jme3.material.MatParam.write(MatParam.java:316)
	at com.jme3.material.MatParamOverride.write(MatParamOverride.java:140)
	at com.jme3.export.binary.BinaryExporter.processBinarySavable(BinaryExporter.java:401)
	at com.jme3.export.binary.BinaryOutputCapsule.write(BinaryOutputCapsule.java:720)
	at com.jme3.export.binary.BinaryOutputCapsule.writeSavableArrayList(BinaryOutputCapsule.java:757)
	at com.jme3.export.binary.BinaryOutputCapsule.writeSavableArrayList(BinaryOutputCapsule.java:345)
	at com.jme3.scene.Spatial.write(Spatial.java:1605)
	at com.jme3.scene.Node.write(Node.java:734)
	at com.jme3.export.binary.BinaryExporter.processBinarySavable(BinaryExporter.java:401)
	at com.jme3.export.binary.BinaryOutputCapsule.write(BinaryOutputCapsule.java:720)
	at com.jme3.export.binary.BinaryOutputCapsule.writeSavableArrayList(BinaryOutputCapsule.java:757)
	at com.jme3.export.binary.BinaryOutputCapsule.writeSavableArrayList(BinaryOutputCapsule.java:345)
	at com.jme3.scene.Node.write(Node.java:735)
	at com.jme3.export.binary.BinaryExporter.processBinarySavable(BinaryExporter.java:401)
	at com.jme3.export.binary.BinaryExporter.save(BinaryExporter.java:210)
	at com.jme3.gde.core.assets.AssetDataObject.saveAsset(AssetDataObject.java:234)
	at com.jme3.gde.core.assets.SpatialAssetDataObject.saveAsset(SpatialAssetDataObject.java:111)
	at com.jme3.gde.core.assets.actions.ConvertModel$1.run(ConvertModel.java:67)
```

PR #738, opened 2017-09-19
Fixed file lock leaks in GLSLLoader.

JavaSaBr:
We didn't close readers of shader streams in the GLSLLoader. It provided some file lock leaks and then I couldn't change these files on windows without restarting jME application.


JavaSaBr:
@Nehon I don't use Windows OS, but this problem came from some users of my editor, in the case with game application it isn't the important problem, but I think we need to fix it :)

PR #739, opened 2017-09-19
fixed asset linked node to reuse shared data between loaded models.

JavaSaBr:
I looked at the problem more deeply which was described by @NemesisMate here:
https://hub.jmonkeyengine.org/t/linkedassets-arent-sharing-meshes-when-imported/38494?u=nemesismate
So I think we need to use asset manager to load children of the asset linked node, but at this moment, BinaryImporter doesn't process correctly recursive loading models in this case, because the importer has a state which was inconsistent in my tests. I offer one of some solutions of this problem, of course, core team can have more correctly vision of this.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryLoader.java
@@ -1,10 +1,7 @@
 package com.jme3.export.binary;
 
 import com.jme3.asset.AssetInfo;
-import com.jme3.asset.AssetManager;
-import com.jme3.export.InputCapsule;
-import com.jme3.export.JmeImporter;
-import com.jme3.export.Savable;
+import com.jme3.asset.AssetLoader;
 
 import java.io.IOException;
 import java.util.ArrayDeque;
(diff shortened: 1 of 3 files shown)


JavaSaBr:
@Nehon @shadowislord @pspeed42 what do you think about this?

JavaSaBr:
@Nehon @shadowislord @pspeed42 I will try to explain here the main problem and my fix:
When we look at the case, when we have a scene with a lot of asset link nodes which are related to the same model, we expect to see that all link nodes will load models with shared data between them. But at this moment the link nodes load different models with different geometry data. So this PR fixes this behavior.
My fix includes two main changes:
1. We used binary importer directly inside reading asset link node before, now we use asset manager to load linked models.
2. BinaryImporter doesn't support recursive loading models, so I implemented a stack of importers to avoid this problem.

Nehon:
Tbh, I'm not familiar with the asset link system. I never used it even in the sdk. @shadowislord could you review this please.?

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryLoader.java
```diff
@@ -0,0 +1,78 @@
...
+ * The default loader to load binaries files.
+ *
+ * @author JavaSaBr
+ */
+public class BinaryLoader implements JmeImporter {
+
+    /**
+     * The thread local importers.
+     */
+    private final ThreadLocal<Deque<BinaryImporter>> threadLocalImporters;
```
shadowislord: The `ThreadLocal` is not needed because `AssetLoader`s are already loaded per-thread

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryLoader.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.IOException;
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * The default loader to load binaries files.
+ *
+ * @author JavaSaBr
+ */
+public class BinaryLoader implements JmeImporter {
```
shadowislord: Why does this need to extend `JmeImporter? Shouldn't it implement `AssetLoader` instead?

JavaSaBr:
ok, I will update this PR.

JavaSaBr:
@shadowislord I have updated this PR.

PR #747, opened 2017-09-27
add an empty constructor to Grid mesh

rvandoosselaer:
Add an empty constructor with logical values for serialization purposes. (e.g. the BinaryImporter has issues instantiating the Grid mesh.)
[binaryImporter_stacktrace.txt](https://github.com/jMonkeyEngine/jmonkeyengine/files/1337060/binaryImporter_stacktrace.txt)


jme3-core/src/main/java/com/jme3/scene/debug/Grid.java
```diff
@@ -45,6 +45,10 @@
  */
 public class Grid extends Mesh {
 
+    public Grid() {
+        this(5, 5, 0.5f);
```
Nehon: If it's for serialization why not an empty constructor? the mesh data will be populated in the read() method so basically here you do the work twice and the first time with arbitrary data.

pspeed42: Also, I can't remember for sure but it's possible that I modified SpiderMonkey's serialization to work with private/protected constructors.  In which case we could make this private and leave it empty.

pspeed42: ...and even if not, I think it's only a one or two line change to fix SM to work with private/protected constructors.  We already do a similar thing for private fields anyway.

rvandoosselaer: I think it can be used more then only for serialisation; now you can instantiate the mesh and it will have some default behaviour (2 by 2 world units in size). Just like the com.jme3.scene.debug.Wirebox for example. But I do get both your points. I'll have a peek at SM and see for a private constructor, or just an empty one.

Nehon: Any news on this one?

rvandoosselaer:
hi @Nehon , sorry this completely slipped my mind. I updated the PR, it's now an empty constructor for serialisation.

Nehon:
thanks

Issue #697, opened 2017-08-05
copy pasted comment needs to be updated in FlyByCamera.java


ericchase:
[FlyByCamera.java](https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java#L133)

![image](https://user-images.githubusercontent.com/30473157/28991592-9252dab6-794f-11e7-9a70-b133c1e3890a.png)

stephengold:
Fixed in master by PR #748 at commit 082fea9 .

PR #748, opened 2017-09-28
FlyByCamera: comments, annotations, & imports; address GitHub issue #697

stephengold:
In addition to correcting the cut-and-paste error referenced in issue #697, many other shortcomings of the inline documentation were corrected. Two @Overrides annotations were added. A star import was expanded into multiple statements.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
+++ b/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2017 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
```diff
@@ -32,62 +32,88 @@
 package com.jme3.input;
 
 import com.jme3.collision.MotionAllowedListener;
-import com.jme3.input.controls.*;
+import com.jme3.input.controls.ActionListener;
+import com.jme3.input.controls.AnalogListener;
+import com.jme3.input.controls.KeyTrigger;
+import com.jme3.input.controls.MouseAxisTrigger;
+import com.jme3.input.controls.MouseButtonTrigger;
```
Nehon: Not an issue, but so you know, star imports do not consume more memory or performance. It's really juste a matter of taste. Also it depends on the IDE settings, mine will use star import as long as there are at least 3 imports from the same package... Maybe at some point we should settle on this and make a clear statement in the contribution guilde line, and reject PR that don't follow the rule. I personally don't mind, but every single PR has a change in import style :p

stephengold: I understand and would support a more detailed set of style guidelines ... along with better enforcement of the guidelines we have.

PR #752, opened 2017-10-07
Add Oculus Rift support

ZNixian:
This PR adds Oculus Rift / LibOVR support, via the LWJGL bindings.

See [the forum thread](https://hub.jmonkeyengine.org/t/libovr-oculus-rift-support/39427) for more information.

### DIFF
--- a/jme3-vr/build.gradle
+++ b/jme3-vr/build.gradle
@@ -2,7 +2,7 @@ if (!hasProperty('mainClass')) {
     ext.mainClass = ''
 }
 
-def lwjglVersion = '3.0.0'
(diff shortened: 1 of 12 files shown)


ZNixian:
@neph1 Should I remove that altogether, then?

neph1:
No. I think the whole logic for deciding whether to use the mirror view or not is really convoluted. I'd like to find out first why it checks if swap buffers is set. Like I said, reversing the setting worked for me. But that's for a different issue and PR to do. I just wanted to mention that I get it too.

neph1:
Tested and approved. Nice job!

PR #759, opened 2017-10-26
Updated some things from shader node system.

JavaSaBr:
1. Added default values for shader node variables.
2. Fixed using texture uniforms in vertex shader nodes.
3. Added two collections to shader node definition for alternative implementations of shader generators.
4. Other small changes.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
+++ b/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
@@ -655,6 +655,15 @@ public class TechniqueDef implements Savable, Cloneable {
         return worldBinds;
     }
 
+    /**
+     * Sets the list of world bindings.
+     *
+     * @param worldBinds the list of world bindings.
+     */
+    public void setWorldBinds(final ArrayList<UniformBinding> worldBinds) {
+        this.worldBinds = worldBinds;
+    }
+
     public void write(JmeExporter ex) throws IOException{
         OutputCapsule oc = ex.getCapsule(this);
         oc.write(name, "name", null);
--- a/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
+++ b/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
@@ -188,6 +188,7 @@ public class TechniqueDef implements Savable, Cloneable {
         defineTypes = new ArrayList<VarType>();
         paramToDefineId = new HashMap<String, Integer>();
         definesToShaderMap = new HashMap<DefineList, Shader>();
+        worldBinds = new ArrayList<>();
     }
     
     /**
@@ -513,10 +514,12 @@ public class TechniqueDef implements Savable, Cloneable {
             }
         }
 
-        if (getWorldBindings() != null) {
-           for (UniformBinding binding : getWorldBindings()) {
-               shader.addUniformBinding(binding);
-           }
+        final List<UniformBinding> worldBindings = getWorldBindings();
+
+        if (!worldBindings.isEmpty()) {
+            for (UniformBinding binding : worldBindings) {
+                shader.addUniformBinding(binding);
+            }
         }
         
         return shader;
@@ -625,14 +628,10 @@ public class TechniqueDef implements Savable, Cloneable {
      * to the list of world parameters, false otherwise.
      */
     public boolean addWorldParam(String name) {
-        if (worldBinds == null){
-            worldBinds = new ArrayList<UniformBinding>();
-        }
-
         try {
-            worldBinds.add( UniformBinding.valueOf(name) );
+            worldBinds.add(UniformBinding.valueOf(name));
             return true;
-        } catch (IllegalArgumentException ex){
+        } catch (IllegalArgumentException ex) {
             return false;
         }
     }
@@ -653,15 +652,6 @@ public class TechniqueDef implements Savable, Cloneable {
      */
     public List<UniformBinding> getWorldBindings() {
         return worldBinds;
-    }
-
-    /**
-     * Sets the list of world bindings.
-     *
-     * @param worldBinds the list of world bindings.
-     */
-    public void setWorldBinds(final ArrayList<UniformBinding> worldBinds) {
-        this.worldBinds = worldBinds;
     }
 
     public void write(JmeExporter ex) throws IOException{
@@ -830,10 +820,8 @@ public class TechniqueDef implements Savable, Cloneable {
             e.printStackTrace();
         }
 
-        if (worldBinds != null) {
-            clone.worldBinds = new ArrayList<>(worldBinds.size());
-            clone.worldBinds.addAll(worldBinds);
-        }
+        clone.worldBinds = new ArrayList<>(worldBinds.size());
+        clone.worldBinds.addAll(worldBinds);
 
         return clone;
     }
--- a/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
+++ b/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
@@ -514,12 +514,8 @@ public class TechniqueDef implements Savable, Cloneable {
             }
         }
 
-        final List<UniformBinding> worldBindings = getWorldBindings();
-
-        if (!worldBindings.isEmpty()) {
-            for (UniformBinding binding : worldBindings) {
-                shader.addUniformBinding(binding);
-            }
+        for (final UniformBinding binding : getWorldBindings()) {
+            shader.addUniformBinding(binding);
         }
         
         return shader;
--- a/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
@@ -659,7 +659,11 @@ public class J3MLoader implements AssetLoader {
 
         if(isUseNodes){
             //used for caching later, the shader here is not a file.
-            
+
+            if (technique.getWorldBindings() == null) {
+                technique.setWorldBinds(new ArrayList<UniformBinding>());
+            }
+
             // KIRILL 9/19/2015
             // Not sure if this is needed anymore, since shader caching
             // is now done by TechniqueDef.
--- a/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
@@ -31,7 +31,6 @@
  */
 package com.jme3.material.plugins;
 
-import com.jme3.material.logic.*;
 import com.jme3.asset.*;
 import com.jme3.material.*;
 import com.jme3.material.RenderState.BlendEquation;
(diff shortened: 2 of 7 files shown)


JavaSaBr:
@Nehon this PR with minimal changes to support my alternative implementation of shader generator(points #3 and #4).

jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
```diff
@@ -200,8 +202,26 @@ protected void readShaderNodeDefinition(List<Statement> statements, ShaderNodeDe
...
+                    }
+                } else if (line.startsWith("Imports")) {
+                    for (final Statement sub : statement.getContents()) {
+                        try {
+                            final String[] values = sub.getLine().split("[ \\{]");
+                            shaderNodeDefinition.getImports().add(values[0]);
+                        } catch (final RuntimeException e) {
+                            throw new MatParseException(e.getMessage(), sub, e);
+                        }
+                    }
```
Nehon: Why do we need this?
Defines and Imports can be defined in the shader node code... no need to do this in the shader node definition.

jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
```diff
@@ -660,19 +690,13 @@ public VariableMapping readInputMapping(Statement statement1) throws IOException
                 throw new MatParseException("Could not find a Material Parameter named " + right.getName(), statement1);
             }
             if (shaderNode.getDefinition().getType() == Shader.ShaderType.Vertex) {
-                if (updateRightFromUniforms(param, mapping, vertexDeclaredUniforms, statement1)) {                  
+                if (updateRightFromUniforms(param, mapping, vertexDeclaredUniforms, statement1)) {
+                    updateMaterialTextureType(statement1, mapping, left, param);
```
Nehon: mhhh I'm not sure vertex texture fetch is available de facto with opengl2.0 or opengles 1.0.
So we may need to check this at generation time. and warn the user and provide a way to fallback to another solution if relevant...

jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
```diff
@@ -659,7 +659,11 @@ private void readTechnique(Statement techStat) throws IOException{
 
         if(isUseNodes){
             //used for caching later, the shader here is not a file.
-            
+
+            if (technique.getWorldBindings() == null) {
+                technique.setWorldBinds(new ArrayList<UniformBinding>());
```
Nehon: Not sure why we need this? isn't the collection initialized when there is a world bind?
Seems the setter in techniqueDef has only been implemented for this use... are you using it somewhere else?

JavaSaBr: @Nehon I have NPE with access to this collection when I don't have world bindings on technique definition(for example when I want to save shader node material during editing it, when I removed all world bindings(temporary)). For example, you can look at line 58 in the class J3mdTechniqueDefWriter, we don't check null here. I think we don't work with a lot of instances of technique definitions, this optimization isn't necessary to have null here by default.

jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
JavaSaBr: What happened when you will have two shader nodes with the same import statements? But I agree that it isn't good to have "hardcoded(imports/defines)" additional information here, can I change it to using more generic solution?

jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
Nehon: well in that case why not just instantiate the list when it's declared instead of adding a setter?

jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
JavaSaBr: @Nehon but in any case, you can't do it with any openGL version, maybe do you have a better idea how to allow it for users?

jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
JavaSaBr: @Nehon I agree, I will update it ;)

jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
Nehon: I need to check this part, I know there is a logic in the global shader system with this, but I can't test for now.

jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
Nehon: Well... this one is though... because we can only know when we load the definition...
I guess keep it like this. And let's leave the check to the user willing to use this feature...

JavaSaBr:
@Nehon I have updated creating world bindings collection in TechniqueDef, and I have changed the solution with ShaderNodeDefinition to have a possibility to add additional information to there(for example for some specific shader generator).

jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
```diff
@@ -513,10 +514,12 @@ private Shader loadShader(AssetManager assetManager, EnumSet<Caps> rendererCaps,
...
         }
 
-        if (getWorldBindings() != null) {
-           for (UniformBinding binding : getWorldBindings()) {
-               shader.addUniformBinding(binding);
-           }
+        final List<UniformBinding> worldBindings = getWorldBindings();
+
+        if (!worldBindings.isEmpty()) {
+            for (UniformBinding binding : worldBindings) {
```
Nehon: the empty check is not necessary here. If it's empty it won't get into the for loop anyway.
just go "for (UniformBinding binding : getWorldBindings()) {"

JavaSaBr:
@Nehon for example, I store the list of imports(if it needs) to a definition to handle it correctly from my shader generator.

jme3-core/src/main/java/com/jme3/material/TechniqueDef.java
JavaSaBr: I will update it

JavaSaBr:
@Nehon I have updated iterating of world binding collections.

Nehon:
No import should not be handled like that.
I checked and there is no de duplication indeed in the system. When you import twice the same lib the shader fails to compile.
But imports are handled on a lower level in JME so i'd rather have them de duplicated at this moment. This way it could also work for classic shaders (in the case you have imports in a shader libs for example)
I'm gonna work on this.

JavaSaBr:
@Nehon so I don't want to replace any your logic of this, I use this solution only in my shader generator, and now ShaderNodeDefinition doesn't include the "imports statement", only abstract "additional values" which you can use up to you.

By the way, if we talk about my solution of this, I can do the analysis of imported shaders, can do deduplication, can add missed world bindings automatically during shader generating and something else...

Nehon:
ok I looked into the code and the imports are injected when we load the shaderNode source.
So basically we'd need to parse the file for #imports store them in a set and replace them when relevant in the next source where it's present...
So several regexps, to pass the code in... 
How fast is you AST generation? Maybe runing it once and for all would be faster in the end.
Note that if we use the AST we'll use it for everything to not waste the time we take to parse the file.
So we may consider changing this part for example -> https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/shader/Glsl100ShaderGenerator.java#L504-L528

JavaSaBr:
@Nehon When I finish working on AST generator(I need to refactor/to optimize some things else), I will make a separated PR with performance tests ;) but what about my change in ShaderNodeDefinition? Do I need to revert it in the result?

Nehon:
Default value are nice. So we can keep this.
I'm more skeptical about the additionalValue... feels wrong to me.

JavaSaBr:
@Nehon I don't know other option how to extend the definition class to have some additional information here :(

Nehon:
My point is, if additional information is imports, it has nothing to do in the definition.
Also additional information is so vague... if we really need to add something at some point I'd rather have something explicit.
But imports can be handled differently really...
Either we leverage your AST generation to rework the whole generation process, and extract the information from the node source code either I do it with regexp... but as said.. if in the end multiple regexp are slower than one AST generation... let's go for AST...

JavaSaBr:
My AST parser is very fast, but in the case with imports, when you load a shader file with imports, you will get a source code with already included imports

Nehon:
Arf... You're right...
Then we need the loader to work differently

JavaSaBr:
I can add a flag to key class to have different logic for this, what do you think?

Nehon:
Yep, that's what I thought. Something like à loadFilesSeparately flag in the asset key. It would return a map with file name as key and the source as value. This way the generator can store the imports in a map and add them on top of the file when the shader is generated.
I will handle this if you don't mind.  I'll push something soon

JavaSaBr:
Ok, I will wait for your commit with it, but I have other cases with additional values... I want to have a possibility to add some help information to a definition for some reasons.

Nehon:
Mhh. Like what ? There is already the description section. What else do you need ?

JavaSaBr:
For example, I want to describe required material parameters for a shade node(for defines from imported shaders)

Nehon:
How would it look like in the j3sn? Give me an example.

Le sam. 28 oct. 2017 à 14:46, Alexandr Brui <notifications@github.com> a
écrit :

JavaSaBr:
I need to place list of strings somewhere in j3sn, for example, in some library we use define NUM_BONES, I need to show for a user, that some shade node requires NumBones material parameter... I can't show it only looking at the source code, because the same shader node can have several versions of shader sources

Nehon:
ok... what's wrong with the description section?
like 
description {
      Required material parameters : Int NumBones, Int Whatever...
      ....
}

Nehon:
I meant the Documentation section not description sorry

JavaSaBr:
I want to show it in UI as specific information, in your case, I need to parse documentation to extract only this part =\ at this moment, I don't know about other cases with additional values.

Nehon:
I don't want to add any data that is useless to the engine.
Note that the documentation section already has a specific syntax like @input @output etc...
I'm ok with @matParam Int NumBones for example
It's easy to parse, and it's what is done for (the sdk was parsing for these @input and @output to display documentation for example).
Note that the documentation is not loaded by the engine, so you can put pretty much anything you want in it.

JavaSaBr:
Ok, then I wait for your change to load shaders without including imports.

Nehon:
Done, https://github.com/jMonkeyEngine/jmonkeyengine/commit/171007693bdc621318da8498c72470077234b9f3

JavaSaBr:
@Nehon I have tested your solution, it's ok for me, I have re-created a new PR:
https://github.com/jMonkeyEngine/jmonkeyengine/pull/760
with the summary commit with reverted shader node definition class.

PR #769, opened 2017-11-13
AST based shader generator

JavaSaBr:
Added new implementation of ShaderGenerator which is based on AST of GLSL code. 
Features:
1. Generating result shader without using regex.
2. Auto-resolving names of additional methods in shader nodes to avoid name collisions.
3. Auto-resolving names of local variables of shader nodes to avoid name collisions.
4. Auto-resolving world bindings from imported shaders inside shader nodes.
5. Auto-resolving defines inside shader nodes. For example:
We have an using of definition in the shader node code like this:
```
#ifdef SD_IS_SET_inColor                    
        vertexColor *= inColor;
    #endif
```
the generator transforms it into this in the result shader:
```
 #ifdef <Namespace>_SD_IS_SET_inColor                    
        <Namespace>_vertexColor *= inColor;
    #endif
```
and if the input parameter is set, the generator adds the definition to the top of the result shader:
`#define <Namespace>_SD_IS_SET_inColor 1`


JavaSaBr:
added auto-resolving definitions like:

#ifdef HAS_OUTPUT_vertexColor
  attribute vec4 inColor;
#endif

#if defined(HAS_INPUT_normalMap) || defined(HAS_INPUT_parallaxMap)
    attribute vec4 inTangent;
#endif

tlf30:
@JavaSaBr this has been sitting here several years now, what is the status in this?

JavaSaBr:
Hi @tlf30, my feature was rejected.

PR #771, opened 2017-11-21
Fixed the problem with LightControl.

JavaSaBr:
Fixed the problem with the reference to light, when we had incorrect reference to the light after loading/cloning spatial of this control.

Example:
I have a node with a point light and a light control with the reference to the light. When I load the node late, I have a cloned version of the node with a cloned version of the light and a cloned version of the light control, but the cloned control related to original light.


JavaSaBr:
@pspeed42 could you review this, please? :)

Issue #772, opened 2017-11-23
scale of a physics shape is applied 2x by BulletDebugAppState


stephengold:
When physics objects with scaled shapes are visualized by a BulletDebugAppState, the visualizations generally don't match the physics objects. This issue came up in the Forum back in 2014: 

https://hub.jmonkeyengine.org/t/scaling-a-capsule-shape/28361

The issue is present both in 3.1-stable and in 3.2-6419.

Here is an updated test case using box shapes:
```
package mygame;

import com.jme3.app.SimpleApplication;
import com.jme3.app.state.ScreenshotAppState;
import com.jme3.bullet.BulletAppState;
import com.jme3.bullet.collision.shapes.BoxCollisionShape;
import com.jme3.bullet.control.RigidBodyControl;
import com.jme3.material.Material;
import com.jme3.math.ColorRGBA;
import com.jme3.math.Vector3f;
import com.jme3.scene.Geometry;
import com.jme3.scene.shape.Box;

public class Main extends SimpleApplication {

    // scaling factors for the geometry and the physics shape
    float geometryScale = 2f;
    float physicsScale = 2f;

    public static void main(String[] args) {
        Main app = new Main();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        ScreenshotAppState screenShotState = new ScreenshotAppState();
        stateManager.attach(screenShotState);

        flyCam.setMoveSpeed(10f);
        flyCam.setZoomSpeed(20f);

        Material gray = new Material(assetManager,
                "Common/MatDefs/Misc/Unshaded.j3md");
        gray.setColor("Color", ColorRGBA.Gray);

        float radius = 0.4f;
        Box mesh = new Box(radius, radius, radius);

        Geometry geometry = new Geometry("box", mesh);
        rootNode.attachChild(geometry);
        geometry.setLocalScale(geometryScale);
        geometry.setMaterial(gray);

        BulletAppState bulletAppState = new BulletAppState();
        getStateManager().attach(bulletAppState);
        bulletAppState.setDebugEnabled(true);
        bulletAppState.setSpeed(0f);

        BoxCollisionShape shape
                = new BoxCollisionShape(new Vector3f(radius, radius, radius));
        shape.setScale(new Vector3f(physicsScale, physicsScale, physicsScale));

        RigidBodyControl rbc = new RigidBodyControl(shape, 0f);
        bulletAppState.getPhysicsSpace().add(rbc);
[...]

PR #773, opened 2017-11-23
fix for issue #772: populate bullet debug mesh using unscaled shape

stephengold:


empirephoenix:
While maybee not optimal solution, certainly better than leaving this another 10 years open

stephengold:
Thank you for your attention to this matter.

PR #774, opened 2017-11-24
Fixed detecting varyings during loading shader nodes.

JavaSaBr:
Fixed detecting varyings during loading shader nodes in the case, when we have several output parameters with the same name but with different namespaces.


JavaSaBr:
@Nehon I can't set reviewers for my PRs, so could you look at this fix, please? :)

Nehon:
What issue does it fix exactly, do you have an example where it fails?

JavaSaBr:
@Nehon what do you need to reproduce it:
It's without my fix:
https://imgur.com/jkNgFHI
It's my fix:
https://imgur.com/9wiynFr
You can see that I have one missed varying without my fix.

Nehon:
ok, I understand.

PR #775, opened 2017-11-24
Cinematics saving/loading

neph1:
I believe this PR makes saving/loading of Cinematics possible. At least it works with my simple test case. I have a feeling it won't work in all cases, though. See description in c83d087 for details
This is not so large as it seems. For most classes it's just the addition of constructors. For some reason git wants to replace the whole class. I'm not sure all the changes in commit 5954822 are required when it comes to constructors.
CameraEvent is a new class derived from the anonymous inner class in Cinematic that was causing issues with serialization.
Cinematic and AnimationEvent are the two classes that have changes in them.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/animation/AnimChannel.java
+++ b/jme3-core/src/main/java/com/jme3/animation/AnimChannel.java
@@ -68,7 +68,11 @@ public final class AnimChannel {
     private float blendAmount = 1f;
     private float blendRate   = 0;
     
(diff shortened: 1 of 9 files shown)


empirephoenix:
I'm not reviewing this as I have no idea, but this will help whoever does the review:
https://github.com/jMonkeyEngine/jmonkeyengine/pull/775/files?w=1
?w=1 enables whitespace ignoring in githubs editor

Nehon:
Thanks

PR #777, opened 2017-11-27
Fixed finding definitions during material loading.

JavaSaBr:
Fixed finding definitions during material loading.
The problem is:
I have different shader definitions with the same name(Split/Add/Mult for different types) in my shader node material, but when I try to load it I see that loader looks at only the name of these definitions, so I have incorrect loaded material in the result.


Nehon:
oh ok.. I see the problem.
I'll merge this once the build is green.

JavaSaBr:
@Nehon yeah, my artist creates some shader node materials in Blender and I try to reproduce it in jME using our shader node system, it's good option to find some missed things in our system :)

PR #782, opened 2017-12-04
Added a method to force loading documentation.

JavaSaBr:
At this moment, I don't have a possibility to load documentation of shader node definitions in a material definition. 
Also, a shader node definition doesn't contain enough information to load the same definition with documentation separately. So, I have added a method with a thread-local flag to force loading documentation.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
+++ b/jme3-core/src/plugins/java/com/jme3/material/plugins/ShaderNodeLoaderDelegate.java
@@ -38,15 +38,9 @@ import com.jme3.material.MatParam;
 import com.jme3.material.MaterialDef;
 import com.jme3.material.ShaderGenerationInfo;
 import com.jme3.material.TechniqueDef;
-import com.jme3.shader.Shader;
-import com.jme3.shader.ShaderNode;
-import com.jme3.shader.ShaderNodeDefinition;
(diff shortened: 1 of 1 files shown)


pspeed42:
The engine doesn't need documentation... so the engine shouldn't have documentation features built into it.

JavaSaBr:
@pspeed42 the documentation is already in the engine, but in some cases, I can't read it.

Nehon:
Use the flag in the ShaderNodeDefinitionKey.

JavaSaBr:
@Nehon yeah, I know about this, but how can I do it in the case of loading material definition?

Nehon:
You can't. and if you need it, you have to find a way on your own. Reload the shadernodedefinition separately when you want to display the documentation for example.
This is clearly not worth a change in the engine.

JavaSaBr:
@Nehon I agree, but I don't see any option how to load it separately, could you provide me advice, please?

Nehon:
Well you want to display the doc on a rollover from what I've seen on the forum, right?
I don't know how your editor is done, but I guess you can know over which node you are, and from the node get the path to the shader node definition. then with the path, create a key, clear it from the cache, then set the flag to true then load the definition again. this time you'll have the doc.

JavaSaBr:
@Nehon wow, sorry, I have missed the field "path" :)
https://imgur.com/MeUOqiH
thanks.

PR #785, opened 2017-12-06
Animated cursor support for LWJGL 3

### LINKED ISSUE #784: Support animated cursors with LWJGL 3
Support animated cursors with LWJGL 3. This was dropped in LWJGL 3, and according to LWJGL developer (http://forum.lwjgl.org/index.php?topic=5572.0) this is not going to come to LWJGL 3. And instead we need to synthesize this ourselves.

Seems not a too big of a task to do. We are heavily dependent on this and I can try to make this happen the way it was suggested to be done by Spasi.

tonihele:
Resolves #784 

I also changed the LinkedLists to ArrayDeques as they should be "faster", anyway somewhat more recommended. Minor formattings :( But only minor.

This is also not discussed in the forums.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
@@ -29,7 +29,6 @@
  * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-
 package com.jme3.input.lwjgl;
 
 import com.jme3.cursors.plugins.JmeCursor;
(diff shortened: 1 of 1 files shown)


tonihele:
Also in case this is all fine and dandy, is there a possibility for this late addition to be added to the 3.2?

Nehon:
@empirephoenix @JavaSaBr could you review this PR please?

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
```diff
@@ -69,20 +68,24 @@
 
     private static final int WHEEL_SCALE = 120;
 
-    private LwjglWindow context;
+    private final LwjglWindow context;
     private RawInputListener listener;
     private boolean cursorVisible = true;
+    private Long[] currentCursor;
```
JavaSaBr: Why do you use Long[] instead of long[]?

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
```diff
@@ -69,20 +68,24 @@
...
     private int mouseY;
     private int mouseWheel;
     private boolean initialized;
     private GLFWCursorPosCallback cursorPosCallback;
     private GLFWScrollCallback scrollCallback;
     private GLFWMouseButtonCallback mouseButtonCallback;
-    private Queue<MouseMotionEvent> mouseMotionEvents = new LinkedList<MouseMotionEvent>();
-    private Queue<MouseButtonEvent> mouseButtonEvents = new LinkedList<MouseButtonEvent>();
+    private final Queue<MouseMotionEvent> mouseMotionEvents = new ArrayDeque<MouseMotionEvent>();
+    private final Queue<MouseButtonEvent> mouseButtonEvents = new ArrayDeque<MouseButtonEvent>();
```
JavaSaBr: It's ok, but I think we can use the construction ArrayDeque<>() instead of ArrayDeque<MouseButtonEvent>() ;)

JavaSaBr:
@Nehon I have tested these changes, it works fine.

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
tonihele: Because I use it in the Map as well :(

JavaSaBr: you can use long[] in all places.

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
tonihele: Sure! Diamonds for the win :)

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
tonihele: Hmm, yesh, it is an object. Should go in there. Instinctively added value type because Map...

empirephoenix:
Looks fine to me (apart from the already discussed stuff), looks like the cursor api really improved, since I saw it last time in lwjgl2

Nehon:
@tonihele could you update the PR with the requested changes? I'll merge it right and it should make it in 3.2 alpha3

tonihele:
Thanks! I'll get back to this in about 10 hours or so.

Nehon:
how many minutes? :D

tonihele:
Right on schedule :)

Nehon:
I almost waited....  ;) 
Right once the build is green I'll merge the PR.
Thanks ;)

PR #786, opened 2017-12-09
Implementing value mappings for shader nodes.

JavaSaBr:
It's a good option in some cases when we want to set values to some input variables of a shader node directly without using material parameters. It looks like:
```
ShaderNode Add {
                Definition: Add: Common/MatDefs/ShaderNodes/Math/FloatVertexNodes.j3sn
                ValueMappings {
                    first = 1.5
                    second = 2.5
                }
            }
```


Nehon:
Doesn't look like what we talked about.
We don't need another ValueMappings section, this can directly be done in the InputMappings section.
Also it was just about values, it was more about glsl expression ie
```
InputMappings{
         color1 = % mix(vec4(1.0), vec4(0.5), 0.5) % 
         color2 = % vec4(1.0) %
         float1 = % 1.0 %
         color3 = Global.outColor
}
```
It doesn't look like you did this.

JavaSaBr:
@Nehon so, you think we need to provide a possibility to input values and expressions, right? if yes, I will re-implement it :)

Nehon:
yes that's what we talked abut in the private message.
Also please use the inputMappings section and not an additional section.
The build fails on a strange error... 
:jme3-core:compileJava/home/travis/build/jMonkeyEngine/jmonkeyengine/jme3-core/src/main/java/com/jme3/shader/ShaderNode.java:288: error: for-each not applicable to expression type
        for (final ValueMapping valueMapping : valueMapping) {
                                               ^
  required: array or java.lang.Iterable
  found:    ValueMapping

I think its confused because you name the iteration variable the same as the collection.`
https://github.com/jMonkeyEngine/jmonkeyengine/pull/786/files#diff-5d722bff374c84f550a4c22b059e134aR288

JavaSaBr:
@Nehon ok, I will do it :)

JavaSaBr:
@Nehon I have updated PR with the implementation of this:
```
ShaderNode Add {
                Definition: Add: Common/MatDefs/ShaderNodes/Math/FloatVertexNodes.j3sn
                InputMappings {
                    first = %%1.5 * 5.0%%
                    second = %%pow(2.0, 2.0)%%
                }
            }
```

JavaSaBr:
@Nehon Hi, what do you think about this approach? I don't expect of merging it to release 3.2, but I want to know it's ok or no, if it's ok, I will start to implement integration with UI of this approach.

Nehon:
sorry ut slipped my mind, gonna have a look asap

JavaSaBr:
thank you very much :)

JavaSaBr:
@Nehon Hi, you have released 3.2, so could you look at the PR, please? :)

PR #787, opened 2017-12-12
Blender ImageLoader: Support Loading of HDR Files and many other

MeFisto94:
See jMonkeyEngine/jmonkeyengine#727

When a blender file contained a hdr image, for example, the engine is able to load that as a texture, however the blender importer didn't do so.
This was because it didn't rely on the AssetManager to load images but instead called some Loaders directly.

This pull request fixes this by using the AM, which means any (even user-) registered AssetLoader can be used to load textures from blender files now.

I've made an API Change though: ImageLoader is now more of a TextureLoader. The AssetManager can only hand back Texture objects and since there is no point in converting: Texture->Image->Texture, I added loadTexture() and deprecated loadImage.

Technically loadImage could be removed as well since it's never called, but I kept it in for now.

### DIFF
--- a/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/ImageLoader.java
+++ b/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/ImageLoader.java
@@ -31,12 +31,14 @@
  */
 package com.jme3.scene.plugins.blender.textures;
 
+import com.jme3.asset.AssetManager;
+import com.jme3.asset.TextureKey;
 import com.jme3.scene.plugins.blender.file.BlenderInputStream;
 import com.jme3.texture.Image;
+import com.jme3.texture.Texture;
 import com.jme3.texture.plugins.AWTLoader;
-import com.jme3.texture.plugins.DDSLoader;
-import com.jme3.texture.plugins.TGALoader;
-import java.io.InputStream;
+import com.jme3.texture.plugins.HDRLoader;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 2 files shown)


empirephoenix:
Great, this would also allow me to get rid of my currently ridiculous blenderloader hack.
However there seem to be some issue with the build server, are maybee some clases / a commit missing?

empirephoenix:
One part requiring a bit thought is the handling of extension case, eg mac and linux/android differenciate between .png and .Png

MeFisto94:
I am not sure what happened, since 727 passed without a problem actually. I had many issues with git though until I hard reset my fork's master and used a copy of the ImageLoader class actually.

The Problem is I don't like the brute force approach but for embedded images there seems no easy way to know the file type.

The extensions are no problem here, unless jme handles them differently. The idea here is that the asset manager gets "dummy" assetkeys, and then a assetloader says "yes, I can load "png".equalsIgnoreCase(fileName)", so it depends on how we handle that internally.

Most files will be loaded through the AWTLoader though, so not sure what they expect.

As a side note: Linux is Case-Sensitive for Filenames, but is it really for extensions?

empirephoenix:
Yes: https://i.imgur.com/vpn6tPI.png
Yes, so we need a full all possibilities brute force, since the blender loader is not meant to be used for production, but for initial importing into the engine, performance concerns are luckily not relevant that much.

MeFisto94:
It's slightly offtopic but: All those files are associated with your pdf viewer nonetheless. And linux has the "file" command, which seems to check the file headers and/or mimetype.

But back to the topic: The question is how jme handles which `AssetLoader` to take?

empirephoenix:
Linux does not care for file endings at all, it looks at the magic bytes in the file at the header.
However in jme the system really only takes a look at the extensions, so some kind of ugly workaround is inevitable here I fear. Brute forcing the extension might be our best way.

MeFisto94:
Sure, but I was referring to how does jme handle the extension->Loader problem, and I've found:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/445f7ed010199d30c484fe75bacef4b87f2eb38e/jme3-core/src/main/java/com/jme3/asset/ImplHandler.java#L270
(which is called by DesktopAssetManager.registerLoader):
https://github.com/jMonkeyEngine/jmonkeyengine/blob/445f7ed010199d30c484fe75bacef4b87f2eb38e/jme3-core/src/main/java/com/jme3/asset/DesktopAssetManager.java#L131

So the extensions are no problem here, as they are case-sensitive in jme anyway

empirephoenix:
True, for the loader determination, there should be no problems, only for actually locating the file.

MeFisto94:
This shouldn't be a problem either, since as you can see, we already have an InputStream which we are simply passing around (This is for embedded image files, hence we cannot guess the extension and have to bruteforce)

empirephoenix:
Ah I get it, I thought this was also an issue for loading external textures, if it is only for the embedded ones, case for the extension is indeed not important.

PR #789, opened 2017-12-23
avoid ClassCastException in Uniform.setValue()

stephengold:
In 3.1, a delayed ClassCastException would occur com.jme3.shader.Uniform#setValue() whenever:
  1. a ColorRGBA material parameter was overridden by a Quaternion/Vector4f 
  2. a Quaternion material parameter was overridden by a ColorRGBA/Vector4f OR
  3. a Vector4f material parameter was overridden by a ColorRGBA/Quaternion .

To the shader, these all represent Vector4 values, but to Java they are incompatible objects.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/shader/Uniform.java
+++ b/jme3-core/src/main/java/com/jme3/shader/Uniform.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2017 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


stephengold:
With the pre-existing code, the class-cast exception typically occurs during RenderManager.render(), too late for the application code to catch it. But perhaps that's preferable to silently correcting the type mismatch.

Nehon:
Or that yes... in the end in the shader it will be a vec4...

Nehon:
hey, looked into this a bit more...
maybe it would be better to have a pivot Vector4f (from the tempVars)variable to avoid a lot of code.
something like this 
```
TempVars vars = TempVars.get();
                Vector4f vec4 = vars.vect4f1;
                //handle the null case
                if (this.value == null) {
                    try {
                        this.value = value.getClass().newInstance();
                    } catch (InstantiationException | IllegalAccessException e) {
                        throw new IllegalArgumentException("Cannot instanciate param of class " + value.getClass().getCanonicalName());
                    }
                }
                //feed the pivot vec 4 with the correct value
                if (value instanceof ColorRGBA) {
                    ColorRGBA c = (ColorRGBA) value;
                    vec4.set(c.r, c.g, c.b, c.a);
                } else if (value instanceof Vector4f) {
                    vec4.set((Vector4f) value);
                } else {
                    Quaternion q = (Quaternion) value;
                    vec4.set(q.getX(), q.getY(), q.getZ(), q.getW());
                }

                //feed this.value with the collected values.
                if (this.value instanceof ColorRGBA) {
                    ((ColorRGBA) this.value).set(vec4.x, vec4.y, vec4.z, vec4.w);
                } else if (value instanceof Vector4f) {
                    ((Vector4f) this.value).set(vec4);
                } else {
                    ((Quaternion) this.value).set(vec4.x, vec4.y, vec4.z, vec4.w);
                }
                vars.release();
```

stephengold:
Sure, I'll do it that way.

Nehon:
don in https://github.com/jMonkeyEngine/jmonkeyengine/pull/791

stephengold:
Thanks for followup. I'm sorry I didn't implement your suggestion promptly. I've been busy with family visiting this week.

Nehon:
No problem. I figured you were busy. And it's totally  natural. I had some time and I wanted to integrate it in beta 2,so I did it.

PR #790, opened 2017-12-26
Upgraded lwjgl3 module and refactored some core interfaces.

JavaSaBr:
These changes are for jME 3.3.
1. Upgraded LWJGL to 3.1.5
2. Refactored LWJGL3 module
3. Refactored some core interfaces(just added Javadoc in the most cases)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/audio/openal/AL.java
+++ b/jme3-core/src/main/java/com/jme3/audio/openal/AL.java
@@ -12,37 +12,36 @@ public interface AL {
     /**
      * Boolean False.
      */
-    static final int AL_FALSE = 0;
+    int AL_FALSE = 0;
 
     /**
      * Boolean True.
      */
-    static final int AL_TRUE = 1;
+    int AL_TRUE = 1;
 
-    /* "no distance model" or "no buffer" */
-    static final int AL_NONE = 0;
+    /* "no distance model" or "no buffer" */ int AL_NONE = 0;
 
     /**
      * Indicate Source has relative coordinates.
      */
-    static final int AL_SOURCE_RELATIVE = 0x202;
+    int AL_SOURCE_RELATIVE = 0x202;
 
     /**
      * Directional source, inner cone angle, in degrees. Range: [0-360] Default:
      * 360
      */
-    static final int AL_CONE_INNER_ANGLE = 0x1001;
+    int AL_CONE_INNER_ANGLE = 0x1001;
 
     /**
      * Directional source, outer cone angle, in degrees. Range: [0-360] Default:
      * 360
      */
-    static final int AL_CONE_OUTER_ANGLE = 0x1002;
+    int AL_CONE_OUTER_ANGLE = 0x1002;
 
     /**
      * Specify the pitch to be applied at source. Range: [0.5-2.0] Default: 1.0
      */
-    static final int AL_PITCH = 0x1003;
+    int AL_PITCH = 0x1003;
 
     /**
      * Specify the current location in three dimensional space. OpenAL, like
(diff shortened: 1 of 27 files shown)


jme3-core/src/main/java/com/jme3/audio/openal/AL.java
```diff
@@ -12,37 +12,36 @@
     /**
      * Boolean False.
      */
-    static final int AL_FALSE = 0;
```
Nehon: uh what?
What's the logic behind removing static final for constants?

pspeed42:
Probably because he let his IDE uglify it.  On interfaces, I guess all fields are static final by default but I think it's still a bad practice to remove them.  It confuses users and makes constants harder to find using normal search tools.  I suggest rejecting the patch until that's fixed.

Nehon:
Oh yeah on interfaces it's directly static final...
Well anyway I agree with you we should keep the final static as it makes it clear those are constants.

JavaSaBr:
@Nehon so, I will recover "public static final" in the core interfaces, anything else? :)

Nehon:
Appart from this it's ok

JavaSaBr:
@Nehon done.

jme3-core/src/main/java/com/jme3/audio/openal/AL.java
```diff
@@ -19,8 +19,7 @@
      */
     static final int AL_TRUE = 1;
 
-    /* "no distance model" or "no buffer" */
-    static final int AL_NONE = 0;
+    /* "no distance model" or "no buffer" */ int AL_NONE = 0;
```
Nehon: missed this one.

jme3-core/src/main/java/com/jme3/audio/openal/AL.java
```diff
@@ -92,8 +91,7 @@
      * Range:  [0.0 - 1.0]
      *
      * Logarthmic
-     */
-    static final int AL_MIN_GAIN = 0x100D;
+     */ int AL_MIN_GAIN = 0x100D;
```
Nehon: and this one

jme3-core/src/main/java/com/jme3/audio/openal/AL.java
```diff
@@ -136,8 +134,7 @@
      * Source is Static if a Buffer has been attached using AL_BUFFER
      * Source is Streaming if one or more Buffers have been attached using alSourceQueueBuffers
      * Source is undetermined when it has the NULL buffer attached
-     */
-    static final int AL_SOURCE_TYPE = 0x1027;
+     */ int AL_SOURCE_TYPE = 0x1027;
```
Nehon: and this one

jme3-core/src/main/java/com/jme3/input/JoyInput.java
```diff
@@ -39,12 +39,12 @@
     /**
      * The X axis for POV (point of view hat).
      */
-    public static final int AXIS_POV_X = 254;
+    int AXIS_POV_X = 254;
```
Nehon: and this one

jme3-core/src/main/java/com/jme3/input/JoyInput.java
```diff
@@ -39,12 +39,12 @@
...
     
     /**
      * The Y axis for POV (point of view hat).
      */
-    public static final int AXIS_POV_Y = 255;
+    int AXIS_POV_Y = 255;
```
Nehon: and this one

JavaSaBr:
@Nehon have pushed changes.

Nehon:
@pspeed42 are you ok with the "public" not being in the Interfaces?

JavaSaBr:
@Nehon the last comment was 24 days ago, I think nobody against it.

pspeed42:
I'm not ok with public not being on the interfaces.   Please put them back and stop using "instant fix" on your editor.

JavaSaBr:
@pspeed42 ok, I will revert public words.

pspeed42:
Things like this:
static final int AXIS_POV_X = 254;

Should be:
public static final int AXIS_POV_X = 254;

I prefer it for the methods as well but I'm less strict about it.  I prefer method signatures to match when possible with the implementations.

JavaSaBr:
@pspeed42 done.

Nehon:
To me there should be 2 commits only:
- upgrade lwjgl3 module
- refactor core opengl interfaces.

That's the moment to use the git rebase -i command, or learn about it.

I can squash it to one commit with github, but IMO there should be 2 commits.

PR #793, opened 2018-01-06
Updated version.

mitm001:


stephengold:
While a stable version 3.2 of the Engine is available, it's not (by itself) a "complete game development suite". That phrase refers to the SDK, not the Engine.

I think this change should wait until there's a stable version 3.2 of the SDK.

mitm001:
Actually, the SDK mention shouldn't be there. Its mentioned in the next paragraph.

mitm001:
Go to https://github.com/jMonkeyEngine/sdk/releases to download the jMonkeyEngine SDK. Read the wiki for a complete install guide. Power up with some SDK Plugins and AssetPacks and you are off to the races. At this point you're gonna want to join the forum so our tribe can grow stronger.

Should read,

Go to https://github.com/jMonkeyEngine/sdk/releases to download the jMonkeyEngine SDK. Read the wiki for a complete install guide. At this point you're gonna want to join the forum so our tribe can grow stronger.

Since the plugins and assetpacks are nonexistent and likely to never be fixed.

I can fix that also if you want.

stephengold:
There are plenty of NetBeans plugins available. Good point about asset packs, however.

Nehon:
So, do you guys agree on this, is it ok to merge?

stephengold:
I think this change should wait until there's a stable version 3.2 of the SDK.

grizeldi:
It's out now.

Nehon:
true

Issue #792, opened 2018-01-06
prior to 1st mouse movement, InputManager returns wrong position 


stephengold:
After my JME application starts but before I move my mouse, inputManager.getCursorPosition() always returns (0,0), even if the mouse cursor is nowhere near that corner of the screen.  Meanwhile, LWJGL's Mouse.getX() and Mouse.getY() returns a more reasonable position.

As soon as my mouse moves, inputManager.getCursorPosition() starts returning reasonable positions that agree with LWJGL.

I see the same behavior with both 3.1.0-stable and 3.2.0-stable, so if this is a bug, it's not a new one.

Here is my test app:
```
package mygame;

import com.jme3.app.SimpleApplication;
import com.jme3.font.BitmapText;
import com.jme3.math.Vector2f;
import org.lwjgl.input.Mouse;

public class Main extends SimpleApplication {

    BitmapText bitmapText;

    public static void main(String[] args) {
        Main app = new Main();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        flyCam.setEnabled(false);
        inputManager.setCursorVisible(true);

        guiFont = assetManager.loadFont("Interface/Fonts/Default.fnt");
        bitmapText = new BitmapText(guiFont, false);
        bitmapText.setLocalTranslation(0, cam.getHeight(), 0);
        bitmapText.setSize(guiFont.getCharSet().getRenderedSize());
        guiNode.attachChild(bitmapText);
    }

    @Override
    public void simpleUpdate(float tpf) {
        float x = Mouse.getX();
        float y = Mouse.getY();

        Vector2f pos = inputManager.getCursorPosition();
        String message = String.format(
                "inputManager says %s     lwjgl says (%.1f, %.1f)",
                pos, x, y);
        bitmapText.setText(message);
    }
}
```

stephengold:
Now we need fixes for LWJGL3 and JOGL, assuming they're still supported.

Nehon:
IS the issue also for those back ends?

stephengold:
Yes!

Nehon:
Something similar to https://github.com/jMonkeyEngine/jmonkeyengine/commit/26ae3ec806c3bad233ae2c9192fc6051aafb5aeb
should be done then

stephengold:
I looked at LWJGL3, but I don't understand it well enough to code a fix for it.

tonihele:
My PR would solve this for LWJGL 3.

PR #795, opened 2018-01-10
initialize InputManager.cursorPos to fix issue #792 (if using LWJGL)

stephengold:
It's not pretty code, but it improves the behavior of InputManager when LWJGL is in use and shouldn't break anything (that isn't already broken) when it isn't.

For the longer term, a cleaner solution might be to modify MouseInput.setInputListener() so that it returns the current mouse position.

Let me know what you think!

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/InputManager.java
+++ b/jme3-core/src/main/java/com/jme3/input/InputManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2018 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


stephengold:
I'm pretty sure that would work, and it would be a cleaner solution. It would break any custom MouseInput subclasses, of course.

empirephoenix:
I would not worry about that, after all anyone who would subclass this classes, should be able to add a empty init method as well, or copy paste your implementation

stephengold:
I'm fine either way, just waiting to hear from @nehon to decide whether to rework this PR.

Nehon:
IMO the best solution is to send a crafted MouseMotionEvent when a listener is set in the LwjglMouseInput https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/LwjglMouseInput.java#L162 or if it's not initialized at this poitn send the event in the initialized method https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/LwjglMouseInput.java#L84

Just do something like this https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/LwjglMouseInput.java#L125-L127

This way, no change in the MouseInput api, and no change in the InputManager. just using th esystem in place

jme3-core/src/main/java/com/jme3/input/InputManager.java
```diff
@@ -150,6 +153,23 @@ public InputManager(MouseInput mouse, KeyInput keys, JoyInput joystick, TouchInp
         }
 
         firstTime = keys.getInputTimeNanos();
+
+        try {
+            Class mouseClass = Class.forName("org.lwjgl.input.Mouse");
+            Field initializedField = mouseClass.getDeclaredField("initialized");
```
JavaSaBr: I think this class shouldn't know about "org.lwjgl.input.Mouse" or other implementations.

Issue #538, opened 2016-07-28
Typo in assert


AppChecker:
Hi.

In file [jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java, line 230](https://github.com/jMonkeyEngine/jmonkeyengine/blob/46891b32c69e8696947a12b4069d8c7383f87d3f/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java#L230)

bp.clientObject is checked twice:

```
assert (
bp.clientObject instanceof com.bulletphysics.collision.dispatch.CollisionObject && 
bp.clientObject instanceof com.bulletphysics.collision.dispatch.CollisionObject
);
```

Probably, it should be:

```
assert (
bp.clientObject instanceof com.bulletphysics.collision.dispatch.CollisionObject && 
bp1.clientObject instanceof com.bulletphysics.collision.dispatch.CollisionObject
);
```

This possible defect found by [statis analyzer AppChecker](https://cnpo.ru/en/solutions/appchecker.php)

empirephoenix:
Quite likely

demoth:
so i'm taking it

demoth:
well, in master branch this code is commented out

DomenicDev:
I will create a pull request for this issue.

Nehon:
Thanks ;)
Seems I can't assign you for some reason so I'll just assign myself so that other people know that it's taken care of

PR #803, opened 2018-01-13
Fixed typo in PhysicSpace described in issue #538

DomenicDev:
This pull request fixes the typo in the assert statement in the class PhysicSpace.
The typo is contained in both packages: jme3-bullet and jme3-jbullet but note that in first the whole method is commented out anyway.


Nehon:
thanks

Issue #647, opened 2017-05-01
Ogre importer ignores receive_shadows


stephengold:
It looks like com.jme3.scene.plugins.ogre.MaterialLoader#readMaterialStatement() parses the receive_shadows statement in the .material file and then forgets about it. As a result, models imported from Ogre always have ShadowMode=Inherit (which may have the effect of Off) even when the original model had "receive_shadows on".

I discovered this bug  in JME 3.1.0-stable using a model exported from MakeHuman.

thoced:
I will try, I will try to solve this problem. but I do not promise anything
Thoced

DomenicDev:
Seems like this issue can be closed since the commit has now been successfully merged into master.

PR #804, opened 2018-01-14
Ogre importer ignores receive_shadows  issue #647

thoced:
The receive_shadow flag is normally interpreted.
Thoced

### DIFF
--- a/jme3-plugins/src/ogre/java/com/jme3/scene/plugins/ogre/MaterialLoader.java
+++ b/jme3-plugins/src/ogre/java/com/jme3/scene/plugins/ogre/MaterialLoader.java
@@ -69,6 +69,7 @@ public class MaterialLoader implements AssetLoader {
     private boolean twoSide = false;
     private boolean noLight = false;
     private boolean separateTexCoord = false;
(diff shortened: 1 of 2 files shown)


Nehon:
@stephengold is it ok for you?

stephengold:
The code looks plausible. Has the change been tested?

thoced:
yes I tested, the shadowmode is well applied to materials

stephengold:
This morning it occurred to me that RenderQueue.ShadowMode.Receive might not be the best setting for a model, since it doesn't include the capability to cast shadows.

thoced:
Indeed, the problem is that the OgreScript does not export a flag to indicate that we want to project shadows.

We could force with the ShadowMode.CastAndReceive parameter but that would not be correct either if the user does not want the shadow to be projected.

Maybe we should look in the Blender ogre export script and add this possibility

stephengold:
An excellent idea.

stephengold:
And thank you, @thoced for your contribution to the project!

Nehon:
IMO it's lot of work for not that much. The user can easily fix this issue in the engine once the model is loaded.

Issue #783, opened 2017-12-05
Wrong class instanciation in DOMInputCapsule


dorian-moffroid-prolexia:
The `readSavableFromCurrentElem` method of `DOMInputCapsule` seems to have an issue at the line 976:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/445f7ed010199d30c484fe75bacef4b87f2eb38e/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMInputCapsule.java#L976-L982

In this way, when a `defVal` is not `null`, the method will not use the `class` attribute in the XML file. This can work in several case but when the `class` attribute in the XML file is not the same as the default value class, the instanciated class is wrong.

Here is a modified version of `TestParticleExportingCloning` sample that shows the issue:
[TestParticleExportingCloning.txt](https://github.com/jMonkeyEngine/jmonkeyengine/files/1531236/TestParticleExportingCloning.txt)

This may solve the issue:
```java
String className = currentElem.getNodeName();
if (currentElem.hasAttribute("class")) {
    className = currentElem.getAttribute("class");
} else if (defVal != null) {
    className = defVal.getClass().getName();
}
tmp = SavableClassUtil.fromName(className, null);
```

Issue #764, opened 2017-11-05
EmitterSphereShape: Infinity loop


TehLeo:
Infinity loop occurs when using EmitterSphereShape with user specified center.
The issue is in the method EmitterSphereShape:getRandomPoint(Vector3f):
```
 @Override
    public void getRandomPoint(Vector3f store) {
        do {
            store.x = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
            store.y = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
            store.z = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
        } while (store.distance(center) > radius);
    }
```

PR #805, opened 2018-01-14
[Taken]Fix issue #783 & #764

jmecn:
Fix issue #783 & #764

### DIFF
--- a/jme3-core/src/main/java/com/jme3/effect/shapes/EmitterSphereShape.java
+++ b/jme3-core/src/main/java/com/jme3/effect/shapes/EmitterSphereShape.java
@@ -96,10 +96,12 @@ public class EmitterSphereShape implements EmitterShape {
     @Override
     public void getRandomPoint(Vector3f store) {
         do {
-            store.x = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
-            store.y = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
-            store.z = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
-        } while (store.distance(center) > radius);
+            store.x = (FastMath.nextRandomFloat() * 2f - 1f);
+            store.y = (FastMath.nextRandomFloat() * 2f - 1f);
+            store.z = (FastMath.nextRandomFloat() * 2f - 1f);
+        } while (store.lengthSquared() > 1);
+        store.multLocal(radius);
+        store.addLocal(center);
     }
 
     @Override
--- a/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMInputCapsule.java
+++ b/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMInputCapsule.java
@@ -974,13 +974,12 @@ public class DOMInputCapsule implements InputCapsule {
             ret = referencedSavables.get(reference);
         } else {
             String className = currentElem.getNodeName();
-            if (defVal != null) {
-                className = defVal.getClass().getName();
-            } else if (currentElem.hasAttribute("class")) {
+            if (currentElem.hasAttribute("class")) {
                 className = currentElem.getAttribute("class");
+            } else if (defVal != null) {
+                className = defVal.getClass().getName();
             }
             tmp = SavableClassUtil.fromName(className, null);
-            
             
             String versionsStr = currentElem.getAttribute("savable_versions");
             if (versionsStr != null && !versionsStr.equals("")){


jme3-core/src/main/java/com/jme3/effect/shapes/EmitterSphereShape.java
```diff
@@ -95,11 +95,10 @@ public void cloneFields( Cloner cloner, Object original ) {
...
     public void getRandomPoint(Vector3f store) {
-        do {
-            store.x = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
-            store.y = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
-            store.z = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
-        } while (store.distance(center) > radius);
+        store.x = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
+        store.y = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
+        store.z = (FastMath.nextRandomFloat() * 2f - 1f) * radius;
+        store.addLocal(center);
```
shamanDevel: Now it creates a random point within a box of the desired side length (=radius) around the center.
The Do-while loop is still necessary to skip the samples outside the sphere

TehLeo:
Yes, just as shaman pointed out, it got to pick an uniform point on a sphere. The current method, picks a random point in a box, repeatedly until the point is in the sphere. It forgets to add the center.

Here are two methods:

```
 do {
    store.x = (FastMath.nextRandomFloat() * 2f - 1f);
    store.y = (FastMath.nextRandomFloat() * 2f - 1f);
    store.z = (FastMath.nextRandomFloat() * 2f - 1f);
} while (store.dot(store) > 1);
store.multLocal(radius);
store.addLocal(center);
```

Or a method without a while loop. (ref: https://math.stackexchange.com/questions/87230/picking-random-points-in-the-volume-of-sphere-with-uniform-probability  )
```
float l = FastMath.pow(FastMath.nextRandomFloat(), 1f/3f);
float u = FastMath.nextRandomFloat()*2f-1f;
float o = FastMath.nextRandomFloat()*FastMath.TWO_PI;
		
store.z = radius*l*u;
u = FastMath.sqrt(1f-u*u);
		
store.x = radius*l*u*FastMath.cos(o);
store.y = radius*l*u*FastMath.sin(o);
```

jme3-core/src/main/java/com/jme3/effect/shapes/EmitterSphereShape.java
Nehon: mhhh... wouldn't it be simpler to do;
```
     store.x = FastMath.nextRandomFloat();
     store.y = FastMath.nextRandomFloat();
     store.z = FastMath.nextRandomFloat();
     store.normalizeLocal(); // <- random direction
     store.multLocal(radius); // <- mult by the radius to be inside the sphere radius
     store.addLocal(center); // <- relocal aroudn the center.
```
No loop needed and the normalize ensure that the point is inside the sphere.

TehLeo: It would be simpler, but would not be uniformly distributed.

Nehon: Uh yeah first the x2-1 should stay to allow negative values. But then just take à random length between 0 and radius, and you break the uniformity.

shamanDevel: I might be wrong, but I think there is needs to more probability mass on the points with bigger radius than on the inner points. So if you normalize and then scale with a value between 0 and radius, that distribution must not be uniform, but something like u'=sqrt(1-u^2), as TehLeo suggested.

jmecn:
I see. Next time I will make separated PR.
Just modified the code, please check it.

TehLeo:
I've checked your changes. It creates random points within a sphere volume.
However, the points are not uniformly distributed. 

See the two methods I posted in the comment above.

jmecn:
ok, let me see..

TehLeo:
Almost there. Except the line:

`store.multLocal(FastMath.nextRandomFloat() * radius);`

I have to say you are really trying to change the code I provided, which is fine, but try to understand the problem as well.

The random point was already selected uniformly within unit sphere volume within the while loop.
What is left to do is multiply it by radius and add the center.

Thus change that line back to.
`store.multLocal(radius);`

jmecn:
Oh my.. I finally understand that line now.

Nehon:
@TehLeo good for you?

TehLeo:
Yes, all is fine with #764 .

Nehon:
@jmecn Could you please rebase -i your PR so that it's only 2 commits one for each issue?
Do you know how to use rebase -i?

jmecn:
@Nehon never use rebase before. But I think I can have a try.

Nehon:
Here is a nice an thorough explanation https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

jmecn:
@Nehon sir I have conbined the commits, please check my code.

Nehon:
Perfect, nice work.
I'll merge this PR once the build pass

jmecn:
Why this pr build failed? any reason?

empirephoenix:
Seems like an error with the buildserver (all buildlogs are null/empty) , I suggest to just commit a extra space or similar somewhere, so that it is triggered again.

Nehon:
No don't commit again I'll relaunch the build

Issue #682, opened 2017-07-10
AwtKeyInput Missing Keycodes for "PrintScreen"


MeFisto94:
Hey Guys,
As the SDK received an issue request that taking a screenshot while the SceneComposer is open leads to `unsupported key:154` being shown.

After some research, I found out that https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-desktop/src/main/java/com/jme3/input/awt/AwtKeyInput.java#L375 or rather https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-desktop/src/main/java/com/jme3/input/awt/AwtKeyInput.java#L605 is responsible for that.

So if jme has a Key related to PrintScreen or if there are plans on allowing PrintScreen as a key tell me, so I can work on it.

empirephoenix:
Is this key the same on all platforms? If so, I see no reason why we should not support it.

grizeldi:
The original report comes from linux.

empirephoenix:
I can check linux, can you check what windows does?

grizeldi:
Sadly I can't. Ubuntu here.

empirephoenix:
XD, so does ubuntu does this then?

grizeldi:
Yup, it was me reporting it to MeFisto ;P

gouessej:
You can use KeyEvent.VK_PRINTSCREEN but there is no such key under Mac OS X as far as I know.

DomenicDev:
I would like to take this issue.

I have got a question though: What Code does the Key "PrintScreen" belongs to?

MeFisto94:
A bit background here:
In order for jme to handle the PrintScreen Key, it has to be added to https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/input/KeyInput.java, which however requires a numeric value.

Furthermore what about KEY_LAST? It would need to be increased and hence breaking compatibility for anyone relying on this being 0xE0 instead of referring to KEY_LAST (which can be ignored maybe, because it would be misuse).

DomenicDev:
So it's okay to just increase LAST_KEY to 0xE1 and give PRINT_SCREEN the value of 0xE0 ?

pspeed42:
[...]

PR #806, opened 2018-01-14
Adds missing key code for Print Screen (#682)

DomenicDev:
This pull request adds the missing key code for the Print Screen key. 
However, LWJGL will interpret this key as SYSRQ unless the buttons are separated.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/KeyInput.java
+++ b/jme3-core/src/main/java/com/jme3/input/KeyInput.java
@@ -446,12 +446,15 @@ public interface KeyInput extends Input {
      */
     public static final int KEY_UNLABELED = 0x97;
     /**
-     * Print Screen key.
-     * NOTE: Only use this if your Print Screen button
-     * is separated else LWJGL will interpret this key
-     * as <code>KEY_SYSRQ</code>
+     * PrtScr key.
+     * Note: for use on keyboards with a PrtScr key that is
+     * separate from the SysRq key. Most keyboards combine
+     * SysRq and PrtScr so if the intent is to actually
+     * capture the user's desire to capture the screen
+     * then SysRq is the most likely scan code.
+     * Use PrtScr to catch the rest (laptops, mini-keyboards, etc.)
      */
-    public static final int KEY_PRTSCREEN = 0x9A;
+    public static final int KEY_PRTSCR = 0x9A;
     /**
      * Enter key (num pad).
      */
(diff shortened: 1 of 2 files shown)


Nehon:
I'm a bit confused reading the issue's comments. @pspeed42 is this PR ok?

MeFisto94:
I'm not that satisfied with the javadoc explanation but I couldn't come up with a good example either, so where are our "writer / documentation monkeys"? :)

DomenicDev:
Sorry if the documentation isn't that good. If you could tell me what to write instead I would apply it of course. I am just willing to help.

pspeed42:
re: "/**
+     * Print Screen key.
+     * NOTE: Only use this if your Print Screen button
+     * is separated else LWJGL will interpret this key
+     * as <code>KEY_SYSRQ</code>
+     */
+    public static final int KEY_PRTSCREEN = 0x9A;"

It's just not true.  It has nothing to do with LWJGL.

If you have a SysReq key on your keyboard then it probably also has a Ptr Scr ABOVE IT.  ie: back in the old days you'd have to hit a special key to make the sys req key a print screen key.  Modern operating systems don't make you hit shift or control anymore... but it's the sys req key though.  Keys are keys and not their interpretation.  For example the '[' is not a curly brace key.  It's exactly the same thing.

Some KEYBOARDS will have a separate PRT SCR button.  Since this is not the Sys Req key, it will have a different scan code.  Different keys, different scan codes.  Scan code = key.  Scan code != interpretation of key.  '[' has a scan code... '{' does not unless some keyboard has a separate key for '{'.

Here is what I suggest.  Change the constant to "PRTSCR" (arbitrarily since most of the keyboards in my house with a dedicated print screen button have Prt Scr.  The laptops have PrntScrn and one keyboard mixes the two approaches.)

Then change the comment to:
-----
PrtScr key.
Note: for use on keyboards with a PrtScr key that is separate from the SysRq key.  Most keyboards combine SysRq and PrtScr so if the intent is to actually capture the user's desire to capture the screen then SysRq is the most likely scan code.  Use PrtScr to catch the rest (laptops, mini-keyboards, etc.).
----
Note: the different abbreviated forms above are on purpose and should be kept as is.  If we keep using the words "print screen" then it too strongly associates with "the thing I want to do" and has little or nothing to do with "key scan codes that will be sent".

DomenicDev:
Alright! Thank you very much for that exclusive explanation! Really, thank you!
I will create a new commit soon and change the JavaDoc corresponding to your description.

PR #818, opened 2018-01-31
Fixed loading linked child nodes.

JavaSaBr:
If we can't locate a linked child, we should print a warning about this, but don't prevent load the parent model. We did it in the case when a loaded child is null, but if the asset manager can't load the child, it doesn't return null, it throws an exception.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/AssetLinkNode.java
+++ b/jme3-core/src/main/java/com/jme3/scene/AssetLinkNode.java
@@ -31,7 +31,9 @@
  */
 package com.jme3.scene;
 
+import com.jme3.asset.AssetLoadException;
 import com.jme3.asset.AssetManager;
+import com.jme3.asset.AssetNotFoundException;
 import com.jme3.asset.ModelKey;
 import com.jme3.export.InputCapsule;
 import com.jme3.export.JmeExporter;
@@ -172,13 +174,15 @@ public class AssetLinkNode extends Node {
         for (final Iterator<ModelKey> iterator = assetLoaderKeys.iterator(); iterator.hasNext(); ) {
 
             final ModelKey modelKey = iterator.next();
-            final Spatial child = assetManager.loadAsset(modelKey);
+            try {
 
-            if (child != null) {
+                final Spatial child = assetManager.loadAsset(modelKey);
                 child.parent = this;
+
                 children.add(child);
                 assetChildren.put(modelKey, child);
-            } else {
+
+            } catch (final AssetNotFoundException | AssetLoadException ex) {
                 Logger.getLogger(this.getClass().getName()).log(Level.WARNING,
                         "Cannot locate {0} for asset link node {1}", new Object[]{modelKey, key});
             }


JavaSaBr:
@Nehon any thoughts? :)

Nehon:
@MeFisto94 this may impact the SDK, could you review please?

JavaSaBr:
@Nehon @MeFisto94 I don't think so that it can impact the SDK.

MeFisto94:
Sorry for the late reaction, somehow I skipped the GitHub E-Mails and did not see I was tagged.
I'll have a look and will tell you :)

MeFisto94:
Okay, so this will not be a problem for the SDK in general.
The old code having this null check actually indicates that this was meant that way.

IMO I am not sure if silently ignoring failures is a good thing, though.
We might get posts on the hub like "fix your shit broken engine" when it's just about an AssetNotFoundException.
This is also where it affects the SDK: Without an Exception, there will be no popup telling the user that something went wrong, only a small red text which they'll ignore much like they also ignore the stack trace on exceptions ^^

So SDK-wise, I'm fine with it, Engine-wise  an asset key property to toggle this behavior might be nice, but that's up to you guys :)

JavaSaBr:
@MeFisto94 @Nehon  if we have warnings for missed linked objects, we can load a scene to do with it something, if we don't have it, we should restore missed reference to return availability to open the scene, I think it's not ok.

Nehon:
Maybe we'd need an error reporting in the asset manager... So that "non fatal" error could be properly reported.
here with just a warning, an editor cannot know that something went wrong and the user will never be notified.

JavaSaBr:
@Nehon so, I will change the severity level of this warning to error.

empirephoenix:
Hm having the assetmanager returing a object with your loaded result + additional informations and errors could greatly help for all kind of tools. Eg. informations on missing bones for animations, textures not found, missing tangentialstuff and other not necessarly critical errors.

JavaSaBr:
@empirephoenix I have no ideas how it can be in the current approach of asset manager...

empirephoenix:
It cant I fear. It would require every method of assetmanger to return a wrapper object.

What might be possible:
add a setAdditionalInformationCallback into assetmanager
every loader could now emit filename, information evt into this callback.
Tools editors or interested people could not set that callback before loading something to give detailed information.
A default value if not sued would be a logger.info handler, that way each application has at least limited additional knowledge and each loader would be able to get rid of direct logger calls. 
@Nehon what would be your opinion on something like this? (out of scope for this Pr of course, more a seperate thing)

pspeed42:
Didn't the SDK implement its own asset manager wrapper for similar (and other) reasons?  Wouldn't an IDE carefully manage its own asset manager so that it could clear the cache surgically, etc..

...what is the "in game" use-case for runtime fix-up of broken assets?

empirephoenix:
noticing upon load, that a users install is corrupt for example

pspeed42:
An exception does that just fine.  We're talking about behavior other than throwing an exception for bad data, ie: "bad data" is no longer "exceptional" behavior.... ie: bad data is to be expected during normal runtime.

That seems like an editor-specific feature to me and is easily accommodated in editor-specific code.

empirephoenix:
Hm I see tow problems with this.
in an editor I want an model with a missing texture to load, eg the loader should not throw a critical exception
In my deployed game, this should be an exception.
Without some indirection, it is not possible to fulfill both.

Without the first, importing assets into jme becomes even worse, without the second, I would have a game running with missing textures.
Currently every loader seems to do this differently.

pspeed42:
Repeating myself: an editor would implement its own asset manager and catch the exceptions to do something nicer... or even do deeper rewrite but wrapping and catching is the easiest.  The game would use the regular asset manager.

In general, the engine will not implement editor-specific features but especially when they are pretty trivial to work around.

empirephoenix:
Ok lets put a few examples what I mean:

DDSLoader@206
Why is a missing flag in the texture only worth a logger statement? If we do that right, it should be an exception right? What way would be there for an editor to warn about this properly? except from parsing the whole logger output?

Sceneloader@107
There are some additional warnings emited to the console, how is decided that they are non critical? What way would be there for an editor to warn about this properly? except from parsing the whole logger output?

GltfLoader@103
Loading a GLTF version not supported by jme is only worth a warning? not an exception?
also:
`logger.log(Level.WARNING, "Couldn't find any matching material definition for material " + materialIndex); `
I would say not being able to load a material is not an ignore and continue case.

Now I can hear you saying, but for production use only the j3o ect file are relevant (and I kna agree)
but:
```
[...}
j3m loader: logger.log(Level.WARNING, "Unknown texture option \"{0}\" encountered for \"{1}\" in material \"{2}\"", new Object[]{value, key, material.getKey().getName()});
[...]
logger.log(Level.WARNING, "Cannot locate {0} for material {1}", new Object[]{textureKey, key});
[...]
```

Or even the binaryimporter
`logger.logp(Level.SEVERE, this.getClass().toString(), "readObject(int id)", "NULL class object: " + alias);`



I would be totally fine, if we would say either it loads 100% or it is an exception, especially for the case of production importers, eg j3o, j3m and other jme native formats.

(And any tool can then handle those however it seems fitting)
But currently there are dozens of cases, where we simply do no throw an exception if (depending on view) critical errors occur. 
So even if we say no editor only features, we suppress errors that might be critical for game use.

pspeed42:
Without having time to read in detail, I would probably think that all of those cases should be an exception.

JavaSaBr:
so, any editor should implement its own asset manager and all loaders, right?

JavaSaBr:
ok, I think I have nothing to add.

pspeed42:
Here... maybe I can play the game, too.

Why do you keep insisting I told you to rewrite the whole engine?

pspeed42:
For reference: https://en.wikipedia.org/wiki/Straw_man
...and few things will "trigger" me as fast as putting words in my mouth to make me look stupid.

empirephoenix:
To be honest, I did understand that as that I have to write my own assetmanager for my tools as well, as the current one is not usable for non production game cases. I just don't have a issue with doing this as much.

Nehon:
uh wow... this escalated pretty fast.
I agree there is a lack of consistency in the AssetManager on how to handle errors. For example it will load dummy textures when a texture is missing, but only when loading from a j3o.
However (I'm not 100% sure of this) it will throw and exception when you try to load a missing image from a j3m.
I think the point is to make sure the j3o is recoverable. 
But his makes only sense at dev time already, because once your game is shipped you can't fix this.
So this PR actually raise a bigger question about how the asset manager should work.

Also re:  "An editor would implement its own asset manager and catch the exceptions to do something nicer."
If every error is an exception you d'ont even need a custom asset manager, you can catch the errors.
However if you want to recover from "non fatal" issues (like the one this PR is trying to fix), you have to go deeper than that. Many exceptions are thrown by the Loaders directly, aborting the loading process, So if you want to recover, you need a custom Loader too. And this can become very tedious.

Note that the custom asset manager in the SDK doesn't do that, it just handles netbeans specific case and add some convenience methods to manage the cache (though I think that all cache clearing methods have been backported to the engine's asset manager)

So IMO, our choices are:
- Make the asset manager only "Game runtime" oriented and throw exceptions whenever an issue is encountered (no recovery at all)
- Make it super forgiving with error management and recovery processes, to ease up the development at the risk of introducing overhead and clutter, for something that is not needed at game runtime.

Note that for now we are in a weird in between state.

pspeed42:
There could be a larger conversation about how rigid the asset manager currently is, also.  How far do we want to open things up?

Should it act strictly as probably originally intended, ie: like the old games that bundled all of their assets into a compiled resource file and you always knew everything was there?

Or should it be more forgiving and allow the developer to configure fallback for things like missing textures?  (For example, it bugs the hell out of me that I can't have it fallback to a global texture pool if it doesn't find the model-specific one... or that I can't even probe to see if an asset exists.)  I think it could probably be more configurable... but whatever the case, it should decide what it wants to be and then do that in the best and most consistent way.

If someone can figure out a way to let me more easily have two FNT loaders at once, that would be great, too. ;)

pspeed42:
And note: you can already do some of that by overriding and wrapping the existing loaders for things like textures... but note that would be a lot easier if the asset manager was more "introspectable" as far as what is already registered.

It's possible we should/could extend the existing asset event listener (https://javadoc.jmonkeyengine.org/com/jme3/asset/AssetEventListener.html) support to allow overriding default behavior... or just require that it be the thing that actually throws the exception (though that doesn't easily allow for path redirection).

PR #820, opened 2018-01-31
Added Key varnames to KeyNames

Pesegato:


pspeed42:
I don't think this is generally useful and can be equivalently done through reflection for those very rare cases that it is necessary.  As a rule, JME tends to not add things to the engine that are only useful for "editors"... and in this case the information is already available anyway.  KeyInput.class.getFields().

PR #824, opened 2018-02-12
Refactoring upgrading lwjgl3 v.3

JavaSaBr:
A compacted version of this PR:
https://github.com/jMonkeyEngine/jmonkeyengine/pull/790

### DIFF
--- a/jme3-core/src/main/java/com/jme3/audio/openal/AL.java
+++ b/jme3-core/src/main/java/com/jme3/audio/openal/AL.java
@@ -12,37 +12,37 @@ public interface AL {
     /**
      * Boolean False.
      */
-    static final int AL_FALSE = 0;
+    public static final int AL_FALSE = 0;
 
     /**
      * Boolean True.
      */
-    static final int AL_TRUE = 1;
+    public static final int AL_TRUE = 1;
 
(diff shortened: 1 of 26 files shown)


JavaSaBr:
@Nehon attempt #3 :)

JavaSaBr:
@Nehon anything else?

AlrikG:
Hi JavaSaBr,
Sorry if you've already answered this question somewhere, but why did you remove the public access modifier from the GL class's statics and methods?

JavaSaBr:
Hi, because the public is default access for interfaces.

AlrikG:
Ok thanks :) I overlooked that. But this also means that you can remove the static and final modifiers from constants. Am I right? => https://docs.oracle.com/javase/tutorial/java/IandI/interfaceDef.html

pspeed42:
Did you read back where we had him put it back because we prefer it the verbose way in this case?
a) it's easier to cut-paste
b) it's easier to grep/search through code
c) it avoids conversations like these where you have to point to the spec to prove that you can do it.

AlrikG:
No sorry I only looked into the changes instead of the PR #790

PR #826, opened 2018-02-17
deprecate Control.cloneForSpatial()

stephengold:


Nehon:
I agree with this. 
Though maybe we should search for usage in the engine, and remove the method where its implemented (not that most remaining implementation are most probably not used, the jmeClone/cloneField implementation is used)

stephengold:
Thanks. I'll generate a separate PR to get rid of all uses in the Engine.

PR #827, opened 2018-02-20
make TerrainTestCollision more user-friendly

stephengold:
+ Change the initial camera position so that the terrain is visible.
 + Display a hint about the T key.
 + Fix a bug where the T key had to be pressed twice before it would work.


stephengold:
See also commits d636049a6ffdc1b225cf4d9a4abb7ec7c17bd51c and ceea055cc40d476810ef32b20c16de5a7c4042d5 , which addressed the same issues in other terrain tests.

PR #828, opened 2018-02-21
Jme3 vr

jseinturier:
I propose within this PR an AWT component rendering within app state

This implementation differs from the previous one as it's relies on AppState (compatible with all JME application) and as it use only produced framebuffer (no link with native objects).

With this implementation, it is possible to use AWT component ad rendering target and so to integrate JMonkey rendering within AWT / Swing application. This capability was no more available with LWJGL3 as the version 3.1.x does not provide anymore AWT link.

The old AWT integration cannot be removed at this time as many users of JMonkey use AWT canvas, however, those that want to use LWJGL3 version of Jmonkey (for example to use VR capabilities) have no more access to AWT capabilities. This is the reason why my commit add redundancy with some classes...


empirephoenix:
Is see this correctly, that only a new integration way is added, but nothing is removed? (So can be merged without destroying anything?)
Btw is it possible to render swing ui components above the jme window with this integration? (the other has some issues there)

jseinturier:
The idea of this commit is to have AWT component rendering capability respire the underlying renderer. More over, the use of an app state enable to use the component rendering without creating a specific application.

My problem is that the existing AWT related classes seems to only work with lwjgl3 and as i dis not want to break somethings i've committed classes with same lames bu within différents packages.

jseinturier:
Sorry for the mistakes within my previous comment. I wanted to says "despite" instead of "respire" and the actual AWT classes works only with lwjgl2 and not lwjgl3 or other renderer.

Nehon:
@shadowislord could you look into this PR please?

jseinturier:
Hello Guys,

Sorry but nobody has accepted or rejected my pull request. Is it possible to accept or reject it ? Thanks a lot.

empirephoenix:
Since I see no removals and only additions, so I guess this is ok from my point of view.

stephengold:
Many of the people with integration rights are "on leave" from the JMonkeyEngine project. I'm relatively inexperienced, and I don't know enough about VR to decide whether this PR should be integrated or not. Same thing with PRs #932 and #937.

jseinturier:
I'm sorry with this mistake but in fact this PR is not related with VR. This PR deal with the fact that when using LWJGL 3, the AWTCanvas and the related classes that enable to render Jmonkey within an AWT/SWING component are no more available.

This PR add new classes that enable to display JMonkey rendering within any AWT component and so that enable to intégrateur JMonkey rendering within any AWT/SWING application. The interest is that or does notre relies on specific LWJGL version.

The link with VR is that i'm also contributing to the VR module and a used the sale branch to make the pull request. Sorry

jseinturier:
I've reactivated the forum topic that i have opened months ago [( AWT / Swing rendering with LWGJL 3.1.X)](https://hub.jmonkeyengine.org/t/awt-swing-rendering-with-lwgjl-3-1-x/40109/2) I hope that some guys will enable you to authorize (or reject) this PR.

jseinturier:
Hello,

Does anyone can have a look to this PR ? I would like to know if it is acceptable ?

empirephoenix:
Well since there were no prior objections, and apparently noone else feels like knowing enough about the integration, I just merged it.

PR #832, opened 2018-02-28
Add constructor to Plane

pdid17:
I propose an additional constructor to com.jme3.math.Plane that allows construction of a plane from another common mathematical definition of a plane. The plane is defined by a normal vector and a single point on the plane.


Nehon:
mhh why did you close it?

Nehon:
ok I get why now ;)

Issue #797, opened 2018-01-11
Spatial.clone breaks rigidbodies


riccardobl:
It seems that the user object is never set for cloned rigidbodies. 

Test code:

```java
import com.jme3.app.SimpleApplication;
import com.jme3.bullet.BulletAppState;
import com.jme3.bullet.control.RigidBodyControl;
import com.jme3.material.Material;
import com.jme3.scene.Geometry;
import com.jme3.scene.shape.Box;

public class Test  extends SimpleApplication{
    public static void main(String[] args) {
        Test t=new Test();
        t.setShowSettings(false);
        t.start();
    }   

    Geometry cloned;
    Geometry original;
    float DELAY=2.f;

    @Override
    public void simpleInitApp() {
        BulletAppState bas=new BulletAppState();
        stateManager.attach(bas);        
        Box b=new Box(10,10,10);
        Material mat=new Material(assetManager,"Common/MatDefs/Misc/Unshaded.j3md");
        

        original=new Geometry("Test",b);
        original.setMaterial(mat);
        original.addControl(new RigidBodyControl(0));
        rootNode.attachChild(original);
        bas.getPhysicsSpace().addAll(original);
  
        cloned=original.clone();
        bas.getPhysicsSpace().addAll(cloned);
        rootNode.attachChild(cloned);
    
    
    }

    @Override
    public void simpleUpdate(float tpf){
        DELAY-=tpf;
        if(DELAY<=0){
            DELAY=2.f;
            if(cloned.getControl(RigidBodyControl.class).getUserObject()==null){
                System.out.println("/!\\ Cloned RigidBody has null user object");
            }else{
                System.out.println("Everything is fine");
            }
        }
    }
}
```

Full gradle project: http://bit.ly/2APrAGN

``` 
$ gradle execute
[...]
/!\ Cloned RigidBody has null user object
/!\ Cloned RigidBody has null user object
/!\ Cloned RigidBody has null user object
/!\ Cloned RigidBody has null user object
```

pspeed42:
I don't know bullet well but the getUserObject() name implies that this would be a user object you'd supply by calling setUserObject()... which I don't see anywhere in the code.  Can you confirm that your original.getUserObject() is not null?

It's possible that it's not cloned properly but your example wouldn't show this, I think.

pspeed42:
Oh, I see... it should be the spatial.

[...]

PR #837, opened 2018-03-05
setUserObject called when cloning controls - bug fixed (issue #797)

thoced:
Hi, 
I think I corrected this problem, can you see if this is correct ?
 When testing with RigidBodyControl, the null value was placed in the setUserObject. The spatial (in the case of RigidBodyControl), was not copied.

Thanks

### DIFF
--- a/jme3-bullet/src/common/java/com/jme3/bullet/control/CharacterControl.java
+++ b/jme3-bullet/src/common/java/com/jme3/bullet/control/CharacterControl.java
@@ -110,6 +110,7 @@ public class CharacterControl extends PhysicsCharacter implements PhysicsControl
         control.setUpAxis(getUpAxis());
         control.setApplyPhysicsLocal(isApplyPhysicsLocal());
         control.spatial = this.spatial;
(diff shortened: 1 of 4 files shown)


Nehon:
@pspeed42 what do you think of this fix? 
related to https://github.com/jMonkeyEngine/jmonkeyengine/issues/797

pspeed42:
Is it weird that the object isn't cloned?  I feel like the usage implies that the default behavior should be to use JME cloner to clone the object in the cloneFields() method.

From the usage, it looks like the fix really should have been in cloneFields() to use the cloner to clone the user object (which would have reused the spatial if it was the spatial).  Of course, since the super.clone() method is not called you probably still need to copy the reference as is done in this patch but I think this is only a partial solution.

pspeed42:
Why are we calling setUserObject()?  I think I lost the bubble on that one.

...I think it's an odd side-effect that setUserObject() will be called for clones.

stephengold:
I don't believe this is the right fix for the issue. I'd like to close this PR without integrating it.

PR #838, opened 2018-03-05
added class with material's constants.

oualidouazrou:
[Thread](https://hub.jmonkeyengine.org/t/suggestion-to-the-predefined-materials-and-shaders)
@Nehon


Nehon:
Thanks

PR #839, opened 2018-03-08
Added Texture Formats

TehLeo:
Added texture formats R16F, R32F, RG16F, RG32F.
Added documentation for Format.Image elements.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/texture/Image.java
+++ b/jme3-core/src/main/java/com/jme3/texture/Image.java
@@ -258,6 +258,7 @@ public class Image extends NativeObject implements Savable /*, Cloneable*/ {
          * half-precision floating point red, green, and blue.
          * 
          * Requires {@link Caps#FloatTexture}.
+		 * May be supported for renderbuffers, but the OpenGL specification does not require it.
          */
         RGB16F(48,true),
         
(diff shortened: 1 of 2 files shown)


jme3-core/src/main/java/com/jme3/texture/Image.java
```diff
@@ -253,77 +253,238 @@
          * Requires {@link Caps#SharedExponentTexture}.
          */
         RGB9E5(32,true),
+
+        @Deprecated
+        Reserved11(0),
```
shadowislord: Adding this here changes the ordinals for the formats, which would break serialization (unfortunately, it uses the enum ordinal...)
New formats always have to be added at the bottom, and any existing formats can only be renamed, not removed

shadowislord:
What's the value of these formats? Where would they be used? Why are existing formats inadequate?

TehLeo:
R16F, R32F, RG16F, RG32F are standard formats required by opengl spec.
They are useful for various things, such as lookup tables, etc. I am currently using them in my wip volumetric cloud rendering.

I thought changing the order of the enum might cause issues, but did it to sort values in more meaningful way. I'll redo the request to include the formats at the end.

TehLeo:
With the new request: The order of the enum elements is preserved. I have added the 4 formats at the end and added comments for the integer formats.

Nehon:
@shadowislord is that ok for you ?
Can I merge this PR?

shadowislord:
@Nehon 
OK. It seems those formats are already there though

Nehon:
Thanks

stephengold:
This PR is being discussed for possible inclusion in JME v3.2.3:
  https://hub.jmonkeyengine.org/t/next-release-s-of-the-engine/40856/126

PR #853, opened 2018-04-18
Fix a typo which makes pbr shader fail to compile when glossiness is enabled

grizeldi:
There was a singe s in glossiness which makes the whole shader not compile if you have a glossiness map set.


Nehon:
thanks

Issue #859, opened 2018-05-08
Add an "aimAtTarget" Method to Ray


Bekreth:
The ray class should contain a method to take in a Vector3f as a target to aim at and calculate the vector to use for the `setDirection` method.

MeFisto94:
Hey, thanks for your Contribution.
Actually what you want would be easily achieveable with standard methods already:
```java
Quaternion q = new Quaternion();
q = q.lookAt(target); // I am not sure whether this is a static method or not, if it returns something or not

// Transform the "Forward" Vector to look into the direction:
Vector3f v = q.mult(Vector3f.UNIT_Z); // Now you have your direction
Ray r = new Ray(Vector3f.ZERO, v);
```

It's a bit more verbose, but technically it's already there

PR #860, opened 2018-05-08
Add `aimAtTarget` method to Ray (issue #859)

Bekreth:
Addressed issue 859.  Added a method to aim a ray at a given target location.  Additionally, updated some of the Javadocs to make it clear that the `setDirection` method takes in a relative vector for direction.


pspeed42:
I'm against adding this method and I'm kind of against the javadoc change.  For one thing, the method is only replacing one line of code already.  For another, it's poorly implemented since it creates double-garbage.

As to the javadoc, a "direction" is a very specific thing: "a unit vector pointing in some direction"... ALL of JME's documentation will be WAAAY easier to read if you just know that.  It seems redundant after a while to mention it everywhere.

Bekreth:
I guess you just have to decline it then. I raised this PR because I felt the documentation was misleading.

PR #862, opened 2018-05-18
Update GLImageFormats to fix framebuffer exceptions in android

joliver82:
As stated at forum topic https://hub.jmonkeyengine.org/t/illegalstateexception-framebuffer-has-erronous-attachment-if-using-filterpostprocessor-or-shadows/40463 there's an error in the depth format defined for opengl ES 2.0 which should be GL.GL_UNSIGNED_SHORT instead of GL.GL_UNSIGNED_BYTE solving "Framebuffer has erronous attachment" IllegalStateException


Nehon:
Great thanks!

PR #864, opened 2018-05-30
bullet native android recompilation to fix at least vehicle physics

joliver82:
Recompiled bullet native libraries for android, changed Android.mk and Application.mk to properly compile with android ndk r16b and to avoid bullet opencl to be compiled because there is no opencl in android. Android ndk16b is used instead of ndk17 not to lose hardware compatibility because armeabi, mips and mips64 were removed in ndk17

Forum reference: https://hub.jmonkeyengine.org/t/native-physics-not-properly-working-on-android/40504


joliver82:
Removed to upload a new compilation of android natives to have @stephengold jme-bullet improvements/fixes

PR #865, opened 2018-06-03
Corrected the angleBetween method

fba-rio:
The formula used in the angleBetween method was wrong.


pspeed42:
No, that's definitely not right.  You will be getting the angle between the tips of the vectors which is definitely not the angle between the vectors.  Think of the hands on a clock.  You want the angle that would move one hand to the other hand's position... not the angle between the tips of the hands.

pspeed42:
And truthfully, 9999 times out of 10,000, if you are using this method then you are already doing something wrong somewhere else.

fba-rio:
Oops, you're right. Sorry for my mistake.

PR #872, opened 2018-07-16
Feature 871

### LINKED ISSUE #871: Update LWJGL to 3.1.6
Update LWJGL to 3.1.6.

Also there is a bug in the context creation. The if the context is re-created at any point, it will crash. This is due to adding the error listener to GLWF that has already been freed but the pointer still exists in GLWF. This is reproducible always when running the examples with the example chooser and choosing i.e. BrickWall. First time ok, second time crash.

tonihele:
Resolves #871

### DIFF
--- a/jme3-lwjgl3/build.gradle
+++ b/jme3-lwjgl3/build.gradle
@@ -2,7 +2,7 @@ if (!hasProperty('mainClass')) {
     ext.mainClass = ''
 }
 
-def lwjglVersion = '3.1.5'
+def lwjglVersion = '3.1.6'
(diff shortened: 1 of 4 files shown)


empirephoenix:
Wild guess, what about line changes? some git clients are per installer setup to change them, you probably want a leave as is setup?

tonihele:
I'm using the jMonkeyEngine SDK straight with GIT. I didn't have this problem last time I contributed.

tonihele:
I'll redo this... no white spaces

PR #874, opened 2018-07-17
Initializes the mouse position correctly

### LINKED ISSUE #792: prior to 1st mouse movement, InputManager returns wrong position

tonihele:
Resolves #792 for LWJGL 3

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
@@ -148,14 +148,6 @@ public class GlfwMouseInput implements MouseInput {
         int x = (int) Math.round(xpos);
         int y = currentHeight - (int) Math.round(ypos);
 
-        if (mouseX == 0) {
-            mouseX = x;
-        }
-
-        if (mouseY == 0) {
-            mouseY = y;
-        }
-
         xDelta = x - mouseX;
         yDelta = y - mouseY;
         mouseX = x;


tonihele:
I updated a bit, a minor fix that the delta comes correctly if you move to the corner. Since the mouse position is now always initialized, we don't need that assumption anymore that 0, 0 means unset.

tonihele:
One thing though... I noticed that usually the mouse cursor is placed in the center of the newly created window (a.k.a. your app), but not always. I'm not sure what governs this. But when it is not placed in the window, we will get a mouse position that is negative or bigger than our window. And the mouse listener of course only picks up on the window it self, so it might be that the delta of the first movement is huge. This is also always true if you travel away from your window and come back from another angle, so I'm not sure if this is a feature or a bug then. The initial thing can be handled, but not the deliberate later mouse escapes which at least is a feature. I hope I could explain this clearly :)

For full screen of course everything is working better now.

PR #875, opened 2018-07-22
Fix AudioNode issues when using velocityFromTranslation and small refactoring

riccardobl:
This PR fixes some issues in the AudioNode that caused wrong velocity calculation when using velocityFromTranslation.

In short, when cloning the node, previousWorldTranslation wasn't reset to its initial value but cloned as reference, and the first velocity update was a subtraction with a NaN.

I've also modified the cloneFields method to reset the velocity to 0 if velocityFromTranslation is set, to prevent cloned nodes from preserving the velocity of the original in case they are never moved.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/audio/AudioNode.java
+++ b/jme3-core/src/main/java/com/jme3/audio/AudioNode.java
@@ -116,13 +116,13 @@ public class AudioNode extends Node implements AudioSource {
          */
         Stopped,
     }
-   
+
     /**
      * Creates a new <code>AudioNode</code> without any audio data set.
      */
-    public AudioNode() {        
+    public AudioNode() {
     }
-    
+
     /**
      * Creates a new <code>AudioNode</code> with the given data and key.
      *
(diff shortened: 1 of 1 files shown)


riccardobl:
I moved the flag up to line 725, but on second thought this was going to skip the position update for nodes that do not use velocity from translation... 

> It's seems that the original issue is just a clonning issue.

See rows 732 and 738 in the original code, if velocityFromTranslation is enabled, the first update will always turn velocity into NaN.


I've made some changes, it should be fine now.

pspeed42:
Why does this have a getParent() == null check?  That doesn't make sense to me... and seems like a big change in behavior (I almost never attach my AudioNodes to anything, for example).  Definitely too risky for 3.2.2... and the check should maybe be removed.  (Or at least explained.)

riccardobl:
Well, it can't really be a change in behavior, since it never worked before.
But i think the null check can be removed, tbh i don't remember if there is a reason for it to be there in the first place.

pspeed42:
Position worked before and now it doesn't, right.  Velocity may have been broken before but position definitely wasn't... at least not when I've used it in the past.

PR #881, opened 2018-08-23
Fixed call to onJoyButtonEvent()

fba-rio:
The method onJoyButtonEvent() was being called on every update(). Now it is called only when the state of a button changes (from not pressed to pressed and vice-versa).


stephengold:
Redone as PR #980.

Issue #877, opened 2018-08-10
Cannot handle more than 1 joint - Physical Hinges and Joints


DanielMartensson:
Hi!

I was following this tutorial https://wiki.jmonkeyengine.org/jme3/advanced/hinges_and_joints.html and I'm doing a cradle with 6 joints and 6 balls.

The problem is that if I have more than 1 joint, the other joints is going to fall down. 

Here is the code. Try to run it and add some hinge joints by removing the comments down there.

Right now the code
`
bulletAppState.getPhysicsSpace().add(Wire1Joint1);`

Is activaded and you should see a pendelum with a square.

Try to activate

`bulletAppState.getPhysicsSpace().add(Wire1Joint2);`

```
package mygame;


import com.jme3.app.SimpleApplication;
import com.jme3.asset.AssetManager;
import com.jme3.bullet.BulletAppState;
import static com.jme3.bullet.PhysicsSpace.getPhysicsSpace;
import com.jme3.bullet.collision.shapes.BoxCollisionShape;
import com.jme3.bullet.collision.shapes.CollisionShape;
import com.jme3.bullet.control.RigidBodyControl;
import com.jme3.bullet.joints.HingeJoint;
import com.jme3.math.Vector3f;
import com.jme3.renderer.RenderManager;
import com.jme3.scene.Node;
import com.jme3.scene.Spatial;


/**
 * This is the Main Class of your Game. You should only do initialization here.
 * Move your Logic into AppStates or Controls
 * @author normenhansen
 */
public class Main extends SimpleApplication {

    private BulletAppState bulletAppState;
    
    public static void main(String[] args) {
        Main app = new Main();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        flyCam.setMoveSpeed(20);
        
        bulletAppState = new BulletAppState();
        stateManager.attach(bulletAppState);
        bulletAppState.setDebugEnabled(true);
        
        // Load Cradle
        //Spatial Frame = assetManager.loadModel("Scenes/CradleScene.j3o");
        // Create node for that Cradle
        Node FrameNode = new Node();
        // Attach the cradle at the node
        //FrameNode.attachChild(Frame);
        rootNode.attachChild(FrameNode);
        
        // Create the static nodes for left wire
        //Node LeftWire1StaticNode = PhysicsTestHelper.createPhysicsTestNode(assetManager, new BoxCollisionShape(new Vector3f( .1f, .1f, .1f)),0);
[...]

PR #882, opened 2018-08-24
add a test for JME issue #877

stephengold:
I haven't found a fix for the issue. Here's a readable test case which works with 3.1.0-stable but not with master.


stephengold:
Note: issue #877 was resolved by pull request #962.

PR #885, opened 2018-08-31
avoid JDK10 so Travis CI can succeed on OSX

stephengold:
Currently the default OS X Build Environment at Travis has JDK10. However, it's tricky to make our build work with both JDK7 and JDK10 because 'javah' was replaced by 'javac -h'. Specifying "osx_image: xcode9.3" bypasses this issue for now.


stephengold:
Note to self: this PR is already integrated into the `v3.2` branch.

PR #890, opened 2018-09-04
rm auto-generated C++ headers from repo and rm them during "build clean"

stephengold:
There are 3 groups of C++ source files in the  jme3-bullet-native subproject:
  1. Bullet library headers and implementations, extracted from a downloaded ZIP archive
  2. JNI headers, auto-generated by running the javah utility on Java source in the jme3-bullet subproject
  3. JME headers and implementations, to turn JNI calls into Bullet library calls and vice versa

Currently group 1 files are not ignored by Git because the .gitignore file was not updated when we switched from Bullet 2.83.7 to Bullet 2.86.1 . Also, "build clean" fails to remove these files.

Currently group 2 files are checked into the Git repository. Instead, they should be ignored by Git and regenerated as needed. Also, "build clean" fails to remove these files.

This pull request addresses the issues:
 + group 1 files are ignored by Git and removed by "build clean"
 + group 2 files are ignored by Git, regenerated as needed, and removed by "build clean"


stephengold:
In order to prevent AppVeyor from failing every pull request, an IF command has been added to the YML that aborts the on_success section if secure variables are not set. (For security reasons, these variables are never set in pull-request builds.)

Issue #889, opened 2018-09-03
disabled physics control gets added to a physics space


stephengold:
One of the invariants of physics controls is that if the control is disabled, the physics object cannot be "in" (added to) any physics space.

Care was taken to preserve this invariant for RigidBodyControl, CharacterControl, and VehicleControl, but not for AbstractPhysicsControl nor GhostControl.

PR #891, opened 2018-09-06
test case and fix for issue #889

stephengold:
A cleaner setPhysicsSpace() method fixes the issue.

Applied the change to AbstractPhysicsControl, CharacterControl, GhostControl, RigidBodyControl, and VehicleControl (though only AbstractPhysicsControl and GhostControl were affected by the issue) in order to increase clarity and reduce the amount of unique code.

At some point in the past, the issue probably affected all physics controls. Whoever fixed RigidBodyControl somehow missed fixing AbstractPhysicsControl and GhostControl.

If all physics controls were based on AbstractPhysicsControl, there'd be a lot less cut-and-paste, and issues like this one would be easier to catch and fix globally. But such a change would break code that assumes certain physics controls are based on PhysicsCollisionObject.


empirephoenix:
Looks fine for me, definitely better readable :)

PR #926, opened 2018-10-02
Load animated cursor NPE

Nebloksiam:
Potential NPE when loading animated cursors (.ani)


Nebloksiam:
I have some animated cursors (.ani) in my project. None of them could be loaded. The animated cursor from jme-testdate does work, I don't know why.

I've attached a test case

[CursorTest.zip](https://github.com/jMonkeyEngine/jmonkeyengine/files/2439542/CursorTest.zip)

PR #930, opened 2018-10-03
Particle tile number/UV calculation change

JESTERRRRRR:
Particles with a non random animation set currently only work for image sets with rows and columns equal, though you can set them separately and make them unequal. 3x4, 8x2 and so on give incorrect values for imgY. 

Forum link: https://hub.jmonkeyengine.org/t/particle-emitter-possible-bug/40971/2

Small change means you can have any number of rows and columns.


stephengold:
Thanks for this!

PR #937, opened 2018-10-05
Updating OpenVR bindings

jseinturier:
Updated OpenVR JNA binding from 1.0.9 to 1.0.16
Added OpenVR/LWJGL binding using KWJGL 3.2.0


neph1:
Maybe this is the best compromise for the time being. I'd like for lwjgl to be the default choice and that code sharing is maximized (which it seems to be from the code). The reason is since the rest of the engine is lwjgl based, it makes sense. I agree that jme is sometimes slow to adapt to new versions, but with gradle it's easy for a user to choose any version they like. It also means we can defer a lot of support questions to lwjgl, it's well documented, etc etc. I assume that over time, someone will migrate oculus bindings as well.

jseinturier:
Maybe this PR can be accepted ?

neph1:
I don't have time to work more on this for the time being. But will pick up on it in due time. Thanks.

stephengold:
Thanks, both of you!

PR #950, opened 2018-11-13
OrientedBoxProbeArea copy constructor fixed

Toboi:
The copy constructor of the OrientedBoxProbeArea wouldn't take the given transform into account. Now it does.


stephengold:
Good catch! If there are no objections, I think I will integrate this one.

Issue #915, opened 2018-09-27
CI build process doesn't update libbulletjme.so


stephengold:
Up until about a year ago, the Linux shared objects (SOs) were automatically committed to the Git repository by Travis after any change that affected them.  This is no longer happening.

For instance, the commit dc4189afc3bb625f8be8044b53667ebedb4fd46e triggered automatic commits of `libbulletjme.dylib` and `bulletjme.dll` (both 32-bit and 64-bit) but failed to update `libbulletjme.so` (either version). `libbulletjme.so` hasn't been updated since e12261eaa5673ef6aa698b078a697d5fc10e1a7d back in August 2017.

Travis job 2286.1 completely successfully. It executed `private/upload_native.sh`, and the outcome was `nothing to commit, working directory clean`.

I'm unsure where to go from here.

stephengold:
Here's a clue from job 2286.1:
```
> Task :jme3-bullet-native:downloadBullet
> Task :jme3-bullet-native:unzipBullet
> Task :jme3-bullet-native:unzipBulletIfNeeded
> Task :jme3-bullet-native:compileBulletjmeLinux32SharedLibraryBulletjmeCpp FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Failed to capture fingerprint of input files for task ':jme3-bullet-native:compileBulletjmeLinux32SharedLibraryBulletjmeCpp' property 'headerDependencies' during up-to-date check.
> Could not list contents of '/usr/include/driver.h'. Couldn't follow symbolic link.
* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Get more help at https://help.gradle.org
BUILD FAILED in 52s
```

For reasons I don't understand, the job continued through the failure.

stephengold:
By googling the error message, I learned that Ubuntu sometimes relocates header files between releases.
I notice our `.travis.xml` changed the build environment (from default to "precise") on 19 August 2017. Circumstantial evidence, but perhaps sufficient to justify experimentally reverting that change.

stephengold:
With that change, the `compileBulletjmeLinux32SharedLibraryBulletjmeCpp` task completes silently during the `oraclejdk8` build.  However, now upload_native.sh reports `No native changes.` and the openjdk7 build fails with a `javax.net.ssl.SSLException`.

Btw, according to the Travis blog, support for the "precise" environment ended back in April.

[...]

PR #951, opened 2018-11-14
Android native physics

joliver82:
Recompiled bullet native libraries for android to match bullet 2.86.1 and latest updates from @stephengold. Changed Android.mk, Application.mk and build.gradle to properly compile with android ndk r16b and to avoid bullet opencl to be compiled because there is no opencl in android. Android ndk16b is used instead of newer ndk not to lose (almost useless) hardware compatibility because armeabi, mips and mips64 were removed in ndk17

Forum reference: https://hub.jmonkeyengine.org/t/native-physics-not-properly-working-on-android/40504

This PR overwrite cancelled PR at https://github.com/jMonkeyEngine/jmonkeyengine/pull/864


stephengold:
Thank you for doing this. Did you run any tests?

joliver82:
I rebuilt the game I'm developing right now which uses vehicle control and some joints and works as expected and even performs a little better than my previous native compilation. Compilation included in jme3 now breaks at least vehicle control.

I've not tested anything more specific

Also... I've noticed issue https://github.com/jMonkeyEngine/jmonkeyengine/issues/915 would it be possible to include also automatic compilation in CI? With the mk's and gradle modifications I've made,  it's as easy as running "grade updatePreCompiledBulletLibs" for jme3-bullet-native-android subproject

stephengold:
Thanks for your answer regarding testing.

I'll leave integration for another day in case anyone else wishes to comment.

Regarding continuous integration, I'm unsure how difficult that would be to set up. I had the impression that building jme3-bullet-native-android required installation of some tools. (Building on my desktop fails with an unenlightening `execCommand == null!` diagnostic.) Those tool installs would probably have to automated.

joliver82:
The "exexCommand == null!" error may just mean you don't have android ndk installed in the path defined in gradle.properties previously /opt/android-ndk-r10c and now /opt/android-ndk-r16b

stephengold:
That sounds plausible. So, to set up continuous integration, installation the android ndk in that path would need to be automated.

Issue #957, opened 2018-11-21
Triangle setters should null out the cached center and normal


stephengold:
`com.jme3.math.Triangle` generates a center vector and a normal vector and caches them. However, altering the triangle's vertices (using `set(int,Vector3f)`, `set(int, float, float, float)`, `set1(Vector3f)`, `set2(Vector3f)`, `set3(Vector3f)` or `set(Vector3f, Vector3f, Vector3f)`) does not update or invalidate the cached vectors. As a consequence, `getCenter()` and `getNormal()` may return incorrect values.

stephengold:
Fixed in `master` branch.

pspeed42:
Existing apps may be relying on this bug accidentally.  Recommend for 3.3 and not 3.2.2.

PR #958, opened 2018-11-21
test and fix for JME issue #957

stephengold:
+ Fixed the issue by nulling out both cached vectors.
+ Added an automated test for the issue.
+ Updated the copyright dates.
+ Improved the JavaDoc.
+ Added Override annotations.


stephengold:
Integrated without review.

PR #965, opened 2018-12-03
Added AppStateManager.getState(class, failOnMiss) 

Ali-RS:
that can optionally throw an exception if the state does not exist.

Inspired from @pspeed42  SiO2 GameSystemManager

https://github.com/Simsilica/SiO2/blob/df02d9dd08943744e6633e476dec38eb6eb44dd7/src/main/java/com/simsilica/sim/GameSystemManager.java#L264


pspeed42:
Yeah, I'm fine with this.  It's useful to have apps fail fast in most cases... I only wish I'd thought of all of this when there was still time to make it the default behavior. :)

quazi-irfan:
@stephengold Since this is a not a breaking change - shouldn't it be on 3.2 release?

pspeed42:
It's a new feature.  3.2.2 should be a bug fix release only.

Issue #964, opened 2018-12-02
Add a way to set CF_NO_CONTACT_RESPONSE for rigid bodies


riccardobl:
Long story short, i have a physic projectile that needs to explode when hitting everything except one rb.
Of course i could do that by messing around with collision groups, but it turned out that bullet provides a more convenient way for my use case.

Enabling CF_NO_CONTACT_RESPONSE will make the rb behave pretty much like a ghost control that is subjected to forces, the collision events will still be triggered, but it will pass through everything.

For my personal use i've just made a patch that exposes the setCollisionFlags method https://github.com/riccardobl/jmonkeyengine/commit/f1840b2ba2dfdbfc29094423366d7de66491bb94

Seeing over the years others issues that would have been easily fixed by this, I think the engine would benefit from having a setContactResponse(boolean) method in PhysicsCollisionObject.

Thoughts?

stephengold:
Sounds useful.

Adding this would create yet another incompatibility between `jme3-bullet` and `jme3-jbullet`, but I don't think that's a major concern at this point.

Why in `PhysicsCollisionObject` instead of `PhysicsRigidBody`?

If you create a pull request, remember that the flag will have to be handled whenever a physics object is loaded, saved, or cloned.

riccardobl:
> Why in PhysicsCollisionObject instead of PhysicsRigidBody?

It is a property of btCollisionObject in bullet.

stephengold:
Are there circumstances where one might wish to setContactResponse() on a ghost or character object?

riccardobl:
It is already set by jme for the ghost controls, i think it should work for characters and soft  bodies.

stephengold:
I'll implement this for `jme3-bullet`.

stephengold:
I'm satisfied that the new feature works in `jme3-bullet`. In `jme3-jbullet`, it throws an appropriate exception on any attempt to set it to false.

stephengold:
I'm reopening this. Now that we've rediscovered the source code for `jbullet.jar`, it should be possible to implement this feature for JBullet as well.

stephengold:
Implemented (for `PhysicsRigidBody`) at 234bd476f3af6ada000eb1ede76c7f48995df304 .

PR #966, opened 2018-12-05
implement setContactResponse(boolean) for PhysicsRigidBody (issue #964)

stephengold:
Implementation for PhysicsCharacter will be a separate PR.

### DIFF
--- a/jme3-bullet-native/src/native/cpp/com_jme3_bullet_collision_PhysicsCollisionObject.cpp
+++ b/jme3-bullet-native/src/native/cpp/com_jme3_bullet_collision_PhysicsCollisionObject.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2018 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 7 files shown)


riccardobl:
I know i was the one who suggested the name for this method, but on second thought i think it would be better to invert the logic and call it setIgnoreContactResponse or setNoContactResponse.

Also i think hard coded flags should be avoided, by either doing [this](https://github.com/riccardobl/jmonkeyengine/blob/f1840b2ba2dfdbfc29094423366d7de66491bb94/jme3-bullet/src/main/java/com/jme3/bullet/collision/PhysicsCollisionFlags.java)  or by making set(No|Ignore)ContactResponse native.

stephengold:
Riccardo: Thanks for reviewing this PR. I appreciate your help.

While I see value in being consistent with the flag name, I'm reluctant to alter the method names because of the extra mental effort to interpret "set ignore" or "set not" as compared with "set".

I agree that hardcoded flag values are undesirable. That was lazy of me. I'll address it before integrating this PR.

I've also decided that `isContactResponse()` should be a `PhysicsCollisionObject` method rather than just a `PhysicsRigidBody` one, so I need to update this PR anyway.  (The setter will remain specific to `PhysicsRigidBody`.)

stephengold:
In the process of testing load/save/clone, I came across some unrelated bugs in `jme3-bullet`, but I'm satisfied that load/save/clone work for contact response. Once I've documented those issues, I'll look at `PhysicsCharacter`.

PR #968, opened 2018-12-09
Added a (hacky) fix for a Java Swing/AWT + GLFW interaction issue…

Ali-RS:
… on Linux.

Read more on https://hub.jmonkeyengine.org/t/experimenting-lwjgl3/37275

### DIFF
--- a/jme3-desktop/src/main/java/com/jme3/app/SettingsDialog.java
+++ b/jme3-desktop/src/main/java/com/jme3/app/SettingsDialog.java
@@ -442,6 +442,10 @@ public final class SettingsDialog extends JFrame {
                 if (verifyAndSaveCurrentSelection()) {
                     setUserSelection(APPROVE_SELECTION);
                     dispose();
+                    
+                    // System.gc() should be called to prevent "X Error of failed request: RenderBadPicture (invalid Picture parameter)"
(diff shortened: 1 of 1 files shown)


Ali-RS:
@empirephoenix  any thought ?

empirephoenix:
I had the BadRenderer issue a lot of times in smaller Tests, but never in my full Application, this would explain this. While this is extremly ugly, if it makes stuff work 99% of the time it is still a great improvement. So from my point of view this might be ok. Might another approch be, to drain the XWindow error queue on JME-LWJGL3 init?

empirephoenix:
Thanks for investigating this

Ali-RS:
@empirephoenix  thanks for merging it. 

Regarding 

> Might another approch be, to drain the XWindow error queue on JME-LWJGL3 init?

I took a look at it but was unable to understand what is going on there. Sorry about that :( 

I just clone this PR from what @Renanse  did in this PR : https://github.com/Renanse/Ardor3D/commit/3984929e9763f2e80bcbac04f875912e00a26086

stephengold:
Is this fix suitable for JME v3.2.2 ?

Ali-RS:
Don't see any strong reason why it should not be :)

PR #971, opened 2018-12-09
implement setContactResponse(boolean) for PhysicsCharacter (issue #964)

stephengold:
This deals with the 2nd half of the issue. Once this is integrated and tested, we should close the issue.


stephengold:
I thought there was a reason to keep a copy of the state, but on closer inspection, there isn't one. I'll remove it. I'll also remove the corresponding state in `PhysicsRigidBody`.

Good catch!

Issue #970, opened 2018-12-09
RigidBodyControl doesn't read/write velocities


stephengold:
The following test fails in master `jme3-bullet` because `PhysicsRigidBody.write()` doesn't serialize the linear and angular velocities and `PhysicsRigidBody.read()` doesn't set them. I haven't tested `jme3-jbullet` to see if it has this issue, but a quick code inspection leads me to believe it does.

This one looks like an easy 8-line fix.

Here's my test case:
```
package jme3test.bullet;

import com.jme3.app.SimpleApplication;
import com.jme3.asset.AssetNotFoundException;
import com.jme3.asset.ModelKey;
import com.jme3.asset.plugins.FileLocator;
import com.jme3.bullet.collision.shapes.CollisionShape;
import com.jme3.bullet.collision.shapes.SphereCollisionShape;
import com.jme3.bullet.control.RigidBodyControl;
import com.jme3.bullet.objects.PhysicsRigidBody;
import com.jme3.export.JmeExporter;
import com.jme3.export.binary.BinaryExporter;
import com.jme3.math.Matrix3f;
import com.jme3.math.Vector3f;
import com.jme3.scene.Node;
import com.jme3.scene.Spatial;
import com.jme3.scene.control.Control;
import java.io.File;
import java.io.IOException;

/**
 * Test saving/loading a RigidBodyControl.
 *
 * @author Stephen Gold sgold@sonic.net
 */
public class TestIssueZZZ extends SimpleApplication {

    private int fileIndex = 0;

    public static void main(String[] args) {
        TestIssueZZZ app = new TestIssueZZZ();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        assetManager.registerLocator(".", FileLocator.class);

        CollisionShape shape = new SphereCollisionShape(1f);
        RigidBodyControl rbc = new RigidBodyControl(shape, 1f);
        setParameters(rbc);
        verifyParameters(rbc);
        RigidBodyControl rbcCopy = (RigidBodyControl) saveThenLoad(rbc);
        verifyParameters(rbcCopy);

        stop();
    }

    /**
     * Clone a body that implements Control by saving and then loading it.
     *
     * @param sgc the body/control to copy (not null, unaffected)
     * @return a new body/control
     */
    private PhysicsRigidBody saveThenLoad(PhysicsRigidBody body) {
        Control sgc = (Control) body;
        Node savedNode = new Node();
        /*
[...]

PR #973, opened 2018-12-13
test and fix for issue #970

stephengold:
Confirmed the bug in both `jme3-bullet` and `jme3-jbullet` using the test code. Added the test code to `jme3-examples`. Added the missing read/write code to both `jme3-bullet` and `jme3-jbullet`. Re-ran the test to confirm the bug is fixed in both `jme3-bullet` and `jme3-jbullet`.


empirephoenix:
Makes sense and looks fine for me :) Lets give it a little protest time before merging :)

PR #979, opened 2018-12-27
Fixed how AppSetting mergeFrom checks for key mapping. 

mitm001:


mitm001:
See https://hub.jmonkeyengine.org/t/how-to-disable-audio-for-simpleapplication/39305/20

Issue #978, opened 2018-12-18
add getSpeed()/setSpeed() methods to SimpleApplication


stephengold:
`LegacyAppliction` provides accessors for most of its protected fields, but not for `speed`. So each app that uses `speed` must provide its own.

pspeed42:
Apps should not use speed.  It's a broken concept because it equates animation time and visualization time with game time.  It automatically presumes and FORCES spatials to be game objects.  The whole concept will be removed eventually... so I'm against expanding the API since we should already add deprecated tags to these methods.

stephengold:
I thought `speed` was a multiplier used to convert real time to game time. Equating those metrics would imply `speed=1`, which is effectively what we'd have if the concept were removed. So your 2nd sentence makes no sense to me.

I also don't see `speed` forcing spatials to be game objects; to me, that looks like an orthogonal issue. Care to elaborate?

stephengold:
I don't agree that `speed=0` is the only interesting value. Aside from `speed=1` there's also the range from about 0.1 to 0.6, which provides a slow-motion effect (extra drama for victory/defeat situations). And in my demos I use `speed=1e-12` to produce a pause effect while still updating controls and appstates.

pspeed42:
To LegacyApplication speed=0 is the same as paused=true.  Speed = anything else does nothing in LEGACY APPLICATION.

Speed should be moved to SimpleApplication where it can live in the same someday-deprecated state the SimpleApplication does... where it pretends that game time is the same as animation time is the same as screen update speed.  None of which is true.  Else you are saying if a GUI button is emitting sparkly particles as part of the general UI then those should ALSO slow down or hang when the game itself is running at half speed.

There is a more complicated question of "which time" that JME completely ignores and will have to be built into future concepts of "pause"... or just left up to the game.

stephengold:
Okay, so `speed` belongs in `SimpleApplication` not `LegacyApplication`. In that case, `setSpeed()` and `getSpeed()` also belong in `SimpleApplication`.

Right?

pspeed42:
Yes, I guess SimpleApplication is fine since it actually uses it... and then we can deprecate it with all of the rest of SimpleApplication when we get there.

stephengold:
Enhancement added to `master` branch at hash c6158bba68c92f70182009968f56b06bb32a178b (pull request #986)

PR #986, opened 2018-12-29
Added getter and setter for LegacyApplication's speed

8Keep:
#978


pspeed42:
I don't think this is a good change as it makes something public through the API that should not be public... and in apps that want it to be public it is trivial for them to add.  (Plus, I'm kind of against the 'speed' idea anyway, my own last name not withstanding.)

pspeed42:
...not only that, speed = 0 seems to be the only interesting value in LegacyApplication and SimpleApplication does its own speed processing anyway.  I think the idea of speed should be completely removed from LegacyApplication.

8Keep:
I'm available to do whatever should be done. I guess we wait for Stephens response.

stephengold:
Let's discuss at issue #978.

stephengold:
8Keep, I took at look at your pull request, and it's perfect. Pspeed and I just need to agree on whether issue #978 was a good idea or not.

8Keep:
No problem! If you come across any more menial fixes with no time to make a pull request, I'd love to take them to get more practice.

stephengold:
Would you be willing to rework your PR to meet the new requirements (SimpleApplication instead of LegacyApplication)?

8Keep:
Yes

8Keep:
Let me know if this is acceptable.

stephengold:
Hm. I don't think we should move the `speed` variable from `LegacyApplication` just yet. Some users may be depending on it. Would you mind putting it back in?

stephengold:
There's still one change in `LegacyApplication` that needs to be reverted.

8Keep:
Lots of commits for very little. Should I create a new PR with just the end changes?

stephengold:
No need for a new PR. GitHub has a "Squash and merge" option which I will now employ...

Issue #972, opened 2018-12-11
TestModelExportingCloning fails with new Animation system 


Ali-RS:
Running following test fails 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/cde3c39c4f0c131c8d536ef369991de1967ef3c1/jme3-examples/src/main/java/jme3test/model/anim/TestModelExportingCloning.java#L42

with this exception

```
java.lang.IllegalArgumentException: Object is not cloneable, type:class com.jme3.anim.tween.action.ClipAction
	at com.jme3.util.clone.Cloner.clone(Cloner.java:262)
	at com.jme3.util.clone.Cloner.clone(Cloner.java:160)
	at com.jme3.anim.AnimComposer.cloneFields(AnimComposer.java:191)
	at com.jme3.util.clone.Cloner.clone(Cloner.java:255)
	at com.jme3.util.clone.Cloner.clone(Cloner.java:160)
	at com.jme3.util.clone.ListCloneFunction.cloneFields(ListCloneFunction.java:66)
	at com.jme3.util.clone.ListCloneFunction.cloneFields(ListCloneFunction.java:43)
	at com.jme3.util.clone.Cloner.clone(Cloner.java:228)
	at com.jme3.util.clone.Cloner.clone(Cloner.java:160)
	at com.jme3.scene.Spatial.cloneFields(Spatial.java:1446)
	at com.jme3.scene.Node.cloneFields(Node.java:725)
	at com.jme3.util.clone.Cloner.clone(Cloner.java:255)
	at com.jme3.util.clone.Cloner.clone(Cloner.java:160)
	at com.jme3.scene.Spatial.clone(Spatial.java:1359)
	at com.jme3.scene.Node.clone(Node.java:684)
	at com.jme3.scene.Node.clone(Node.java:61)
	at com.jme3.scene.Spatial.clone(Spatial.java:1389)
	at jme3test.model.anim.TestModelExportingCloning.simpleInitApp(TestModelExportingCloning.java:66)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:220)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:475)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:574)
	at com.jme3.system.lwjgl.LwjglWindow.create(LwjglWindow.java:423)
[...]

PR #1002, opened 2019-01-08
fix for issue #972 (TestModelExportingCloning fails)

stephengold:
In addition to making `Action`, `BlendableAction`, and `ClipAction` into `JmeCloneable` classes, I also made `AbstractTween` into a `JmeCloneable` class.


Ali-RS:
Thanks for the fix @stephengold  :)

PR #1015, opened 2019-02-04
Added isSimilar() to Quaternion, Vector2f, Vector3f and Vector4f

Ali-RS:
...that allow for comparison of values within some epsilon.

Inspired from @pspeed42 's [SimMath](https://github.com/Simsilica/SimMath) library.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Quaternion.java
+++ b/jme3-core/src/main/java/com/jme3/math/Quaternion.java
@@ -1279,6 +1279,29 @@ public final class Quaternion implements Savable, Cloneable, java.io.Serializabl
         }
         return true;
     }
+    
+    /**
+     * Returns true if this quaternion is similar to the specified quaternion
+     * within some value of epsilon.
+     */
+    public boolean isSimilar(Quaternion other, float epsilon) {
+        if (other == null) {
+            return false;
+        }
+        if (Math.abs(other.x - x) > epsilon) {
+            return false;
+        }
+        if (Math.abs(other.y - y) > epsilon) {
+            return false;
+        }
+        if (Math.abs(other.z - z) > epsilon) {
+            return false;
+        }
+        if (Math.abs(other.w - w) > epsilon) {
+            return false;
+        }
+        return true;
+    }
 
     /**
      * 
--- a/jme3-core/src/main/java/com/jme3/math/Vector2f.java
+++ b/jme3-core/src/main/java/com/jme3/math/Vector2f.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2018 jMonkeyEngine
+ * Copyright (c) 2009-2019 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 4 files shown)


stephengold:
Is Float.NAN similar to every other floating-point number?

MeFisto94:
I have a `similar` code in every of my own project as well, however it's called `MathTools.approximateEquals(Vector3f, Vector3f)` and what it does is call `FastMath.approximateEquals(a.x, a.y) && ....`, so I don't care about an Epsilon or corner cases as outlined by @stephengold, they are then a problem for engine developers, so that might be an alternative implementation which leverages existing code.

Ali-RS:
That's how @pspeed42 is doing here : https://github.com/Simsilica/SimMath/commit/385d71d3413529cdd5ce9fa6ca8f6a391fea72e1 
I am creating an equivalent to that in JME.

pspeed42:
I mean, it's possible that I did something screwy as I had a very specific use-case in mind.  (Also, for double a variable-epsilon'ed comparison is even more useful than for float.)  So I hadn't really thought about NaN but followed some patterns I've followed before.

At a quick read, I think the logic still works, though.  NaN and anything else (including NaN) kicks out false no matter the epsilon... which I think is ok in the case of a comparison that is looking for a specific precision.

It's possible I've missed something, though.  In my use-case, I was trying to see if something moved "more than a little" or rotated "more than a little" in a case where both halves of the comparison were derived from essentially the same path of logic (one frame to the next).  Most of my "little" comes from float-double-float kinds of conversions.  And my use-case doesn't have to worry about quaternion flipping, etc.

stephengold:
The way the methods are written, if each test is `false` then the method returns `true`, meaning "similar". So I think the logic is broken for `NaN`.

Ali-RS:
@stephengold will adding this check inside isSimilar() fix it : 

```
 if (!isValidVector(other)) {
            return false;
 }
```

pspeed42:
re: "The way the methods are written, if each test is `false` then the method returns `true`, meaning "similar". So I think the logic is broken for `NaN`."

Since the tests are for delta > epsilon, can you explain in more detail how the method is broken?  (Note: NaN > epsilon is always true.)

Ali-RS:
> Since the tests are for delta > epsilon, can you explain in more detail how the method is broken? (Note: NaN > epsilon is always true.)

@pspeed42 

https://www.baeldung.com/java-not-a-number

```


System.out.println("NaN == 1 = " + (NAN == 1));
System.out.println("NaN > 1 = " + (NAN > 1));
System.out.println("NaN < 1 = " + (NAN < 1));
System.out.println("NaN != 1 = " + (NAN != 1));
System.out.println("NaN == NaN = " + (NAN == NAN));
System.out.println("NaN > NaN = " + (NAN > NAN));
System.out.println("NaN < NaN = " + (NAN < NAN));
System.out.println("NaN != NaN = " + (NAN != NAN));
--
NaN == 1 = false
NaN > 1 = false
NaN < 1 = false
NaN != 1 = true
NaN == NaN = false
NaN > NaN = false
NaN < NaN = false
NaN != NaN = true
```
But that's not the case with Float.compare(f1, f2) : http://java2s.com/Tutorials/Java/Data_Types/How_to_compare_two_Java_float_values.htm

> int compareTo(Float anotherFloat) compares two Float objects numerically. Float.NaN is considered to be equal to itself and greater than all other float values including Float.POSITIVE_INFINITY

pspeed42:
D'oh.  My quick testing was in groovy which apparently gives different results in this case.

Ali-RS:
The other way to fix it is to use

```
if (Float.compare(Math.abs(other.x - x) , epsilon) == 1) {
            return false;
}
```

which one do you prefer ?

pspeed42:
That's probably better... but I think > 0 is more appropriate for a compare() result.

Ali-RS:
Okay, I am going to update the pr.

Ali-RS:
Updated the PR. 

btw, what does this have to do with my pr https://github.com/jMonkeyEngine/jmonkeyengine/issues/1  !!

stephengold:
I believe it does.

Ali-RS:
@stephengold could this be merged ?

Ali-RS:
Thanks :)

PR #1016, opened 2019-02-07
Added removeAction(), removeLayer() and removeCurrentAction() to AnimComposer

Ali-RS:


empirephoenix:
Can you explain what this does? I have no idea what it is supposed to solve/fix/improve :)

Ali-RS:
It's an improvement for AnimComposer which is aimed for new animation system for JME 3.3. 
I believe it's not completed yet, and @Nehon is the one who works/was working on this system. 

```
    public void removeLayer(String name) {
        layers.remove(name);
    }
```
Having an option to makeLayer() you should also provide an option to remove it. Shouldn't it ? 
The same goes for removeAction(). 

   ```
  if (actionName == null) {
            // Clear current action from specified layer
            l.currentAction = null;
            return null;
        }
```
this one let you stop an specific layer from interpolation it's current action with out removing layer. Which is meant for : 
 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/eeac6d10eb3a5a128dd7988775b35cd5a756681c/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java#L162 

As @Nehon is not around so I guess I should ask @stephengold  and @pspeed42 to take a look at it.

Ali-RS:
Added an appropriate javadoc for setCurrentAction() method.

Ali-RS:
Moved removing of current action to a separate method.

Ali-RS:
Can this be merged ?

stephengold:
How about JavaDoc for the added methods `removeAction()`, `removeCurrentAction()`, and `removeLayer()`?

Ali-RS:
@stephengold ,  JavaDoc added.

Ali-RS:
@stephengold  thanks so much for merging the PR :)

Issue #1019, opened 2019-02-09
Missing features for AnimComposer and Action tween


Ali-RS:
AnimComposer and Action tween can be improved to support these functionalities : 
looping (on/off), fastForward(time), getTime(), getRemaining(), cancel(), getPercentRemaining(), ... 
As they are available in Lemur AnimationState, TweenAnimation 

https://github.com/jMonkeyEngine-Contributions/Lemur/blob/master/src/main/java/com/simsilica/lemur/anim/TweenAnimation.java

Ali-RS:
Looping is added in https://github.com/jMonkeyEngine/jmonkeyengine/pull/1845

PR #1038, opened 2019-03-12
Added feature request #1019

Ali-RS:
Note anim looping=false works only for forward animation (speed > 0). For backward animation (speed < 0) the behaviour is like before (keep looping).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -67,10 +67,10 @@ public class AnimComposer extends AbstractControl {
     public Action setCurrentAction(String name) {
         return setCurrentAction(name, DEFAULT_LAYER);
     }
-    
+
     /**
      * Run an action on specified layer.
-     * 
+     *
      * @param actionName The name of the action to run.
      * @param layerName The layer on which action should run.
      * @return The action corresponding to the given name.
@@ -80,13 +80,13 @@ public class AnimComposer extends AbstractControl {
         if (l == null) {
             throw new IllegalArgumentException("Unknown layer " + layerName);
         }
-        
+
         Action currentAction = action(actionName);
         l.time = 0;
         l.currentAction = currentAction;
         return currentAction;
     }
-    
+
     /**
      * Remove current action on specified layer.
      *
@@ -97,16 +97,17 @@ public class AnimComposer extends AbstractControl {
         if (l == null) {
             throw new IllegalArgumentException("Unknown layer " + layerName);
         }
-        
+
         l.time = 0;
         l.currentAction = null;
     }
 
     /**
-     * 
+     *
      * @param name The name of the action to return.
-     * @return The action registered with specified name. It will make a new action if there isn't any.
-     * @see #makeAction(name) 
+     * @return The action registered with specified name. It will make a new
+     * action if there isn't any.
+     * @see #makeAction(name)
      */
     public Action action(String name) {
         Action action = actions.get(name);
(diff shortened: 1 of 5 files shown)


stephengold:
I'm unclear. Are you asking for help other than integration of your PR into `master` branch?

Ali-RS:
@stephengold  if you think this is an issue to not be able to turn of looping if animation is backward then yes I need help to fix it before we merge it to master.
(In my case I do not need this feature, Actually I do not need back ward animation at all in my game)

Ali-RS:
I guess it is because time never become equals to currentAction.getLength() in 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/9357231911213bc93970ac6f0d74602608b1c1b8/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java#L318
and for that reason this line never returns false
https://github.com/jMonkeyEngine/jmonkeyengine/blob/9357231911213bc93970ac6f0d74602608b1c1b8/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java#L240

Ali-RS:
I think supporting this feature will need to do some dirty code so @stephengold I think it is better to add a note in setLooping() method javadoc and mention that setting loop has no effect on backward animations.

stephengold:
The current behavior seems unexpected and unintended: a bug. I think it would be better to make it behave as expected than to simply document the bug. Particularly since `AnimComposer` is unreleased, so there's no compatibility issue with previous releases.

Ali-RS:
@stephengold  above changes should solve the issue.  
I will appreciate if you take a look at it and probably run the test (I already provided in TestAnimMigration).

stephengold:
I don't yet feel qualified to review this one.

Ali-RS:
@pspeed42  may I also ask about your idea. 
Can I merge it (after fixing the conflict)  ?

pspeed42:
Without spending some time expanding my view of the surrounding code, I'm a little concerned. It could just be based on ignorance.

Does not the animation code already loop?  It seems to.

Ali-RS:
> Does not the animation code already loop?
Yes, it already loops but there is no proper way to disable it.

pspeed42:
So you want an animation that plays once and then stops?  And this is not already supported?

I was 99% sure the wave animation in spacebugs runs and then stops on its own.

Ali-RS:
> So you want an animation that plays once and then stops? 

Yes
> And this is not already supported?
Yes

> I was 99% sure the wave animation in spacebugs runs and then stops on its own.
Hmm, unless you set the speed to 0 or replace the animation in somewhere it will loop waving, I am sure.

pspeed42:
I can't check what I did right now but I will look tonight.

Ali-RS:
Okay, thanks so much.

Ali-RS:
Okay, I refactored my code in a way that do not need these features anymore :)
I am going to close this PR tomorrow.

pspeed42:
Aw.  Sorry... leave it open.  I'm still going to look but I ran out of time last night.

Ali-RS:
Okay no problem. Thanks for your attention so.

barrybecker4:
So how does one detect the end of animation using AnimComposer? we could use AnimEventListener before, but that is now deprecated and does not work with AnimComposer.

It would be great if the deprecation annotation in the code could give migration instructions.
Also when closing an issue like this, a brief explanation of why this feature is no longer needed would be nice.

Ali-RS:
The preferred way is to use a `CallMethod` action at the end of the sequence.

see this example

https://github.com/jMonkeyEngine/jmonkeyengine/blob/252d16647d6b096a7b3329a361f760e06ae46116/jme3-examples/src/main/java/jme3test/bullet/TestBoneRagdoll.java#L180-L184

and this post
https://hub.jmonkeyengine.org/t/solved-when-animcomposer-reaches-the-end-of-an-action/41652/2

PR #1039, opened 2019-03-12
Add tests for com.jme3.math.FastMath

James-Adam:
Hi, 
I've analysed your code base and noticed that `com.jme3.math.FastMath` is not fully tested.

I've written some tests for the functions in this class with the help of [Diffblue](https://www.diffblue.com/) [Cover](https://www.diffblue.com/products). Hopefully, these tests should help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other particular classes that you consider important.

### DIFF
--- a/jme3-core/src/test/java/com/jme3/math/FastMathTest.java
+++ b/jme3-core/src/test/java/com/jme3/math/FastMathTest.java
@@ -33,8 +33,13 @@ package com.jme3.math;
 
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
 import org.junit.Ignore;
+
+import java.lang.Math;
+
+import org.junit.Rule;
+import org.junit.rules.ExpectedException;
 
 /**
  * Verifies that algorithms in {@link FastMath} are working correctly.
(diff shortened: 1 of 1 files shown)


stephengold:
This looks fine.  Thank you.  I'll leave this PR open a few days in case someone else wants to review it.

I'm a little surprised your efforts didn't uncover any bugs in `FastMath`.

If you're interested in writing more tests, I suggest you look next at the `com.jme3.math.Vector3f` class, the `com.jme3.renderer.Camera` class, or the `com.jme3.scene` package. But if you're anything like me, you'll find room for improvement almost anywhere you look.

Ali-RS:
Also include this package if you are interested `com.jme3.anim`  in jme3-core module :
https://github.com/jMonkeyEngine/jmonkeyengine/tree/master/jme3-core/src/main/java/com/jme3/anim

James-Adam:
I'm happy to contribute!

I'm very interested at taking a look at some of the classes/packages suggested, and will be looking to have tests for these written over the next few days. 
I am specifically taking a look at `com.jme3.math.Vector3f` currently; this class is around 50% covered as it stands currently, but with no dedicated tests. @stephengold Would you be interested in seeing a new test class written to cover as much of this class as possible regardless of which methods are covered serendipitously by other test classes, or would you prefer tests which cover only the uncovered lines?

stephengold:
Uncovered code paths should take priority. But I wouldn't object to being systematic and thorough unless/until testing has a noticeable impact on the build process.

stephengold:
By the way, the fields of a `Vector3f` are public, so it's never necessary to use `Vector3f.getX()` or `Vector3f.getY()` or `Vector3f.getZ()`.

PR #1047, opened 2019-03-19
Add RGTC compressed formats

riccardobl:
From https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_compression_rgtc.txt

```
    These formats are designed to reduce the storage requirements
    and memory bandwidth required for red and red-green textures by
    a factor of 2-to-1 over conventional uncompressed luminance and
    luminance-alpha textures with 8-bit components (GL_LUMINANCE8 and
    GL_LUMINANCE8_ALPHA8).

    The compressed signed red-green format is reasonably suited for
    storing compressed normal maps.

```


stephengold:
Looks fine to me, but I'm ignorant when it comes to OpenGL. @pspeed42 what do you think?

stephengold:
One question: how much testing has this PR had?

pspeed42:
My OpenGL-fu is not great these days but these changes look fine to me.

riccardobl:
I took it out of my game, where i use only RGTC2 for normalmaps, i'll add an example and the glsl code needed to reconstruct Z.

Issue #433, opened 2016-03-01
LWJGL 3 context cannot be created on Mac OS X


shadowislord:
Mac OS X requires window management code to run on the main thread. Unfortunately, jME3 creates a separate thread to initialize the window. 
This will be a breaking change which is only included for applications opting to use LWJGL 3.

zzuegg:
As an addon info, on osx, the main thread has to be thread 0.
It requires the vm argument -XstartOnFirstThread

shadowislord:
Fixed in https://github.com/jMonkeyEngine/jmonkeyengine/commit/e541a5a147e0dc8acefee694fce4d7f8a616a5be

PR #1048, opened 2019-03-19
Fix waitFor not being implemented in lwjgl3

tlf30:
So this seems like a big bug to me. I noticed that for some reason for the lwjgl 3 display context, the wait for was being ignored. There was a comment about a OSX fix, but with no information to go with it. 

It seems to me that it is important for the waitFor to work, as it broke one of my applications that I was swapping from lwjgl2 to lwjgl3. 

Does anyone know about this, or can provide insight into this? 
This PR adds support for the waitFor in the same way that it was implemented in lwjgl2.
Also, if someone can confirm that this still works for them, that would be great. 

PS: I hope I opened this PR correctly. If not I'm sorry in advance, every project has their preferred way of doing things. 

Thank you,
Trevor Flynn

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -425,7 +425,9 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
                 waitFor(true);
             }
         } else {
+            //Fix for OS X, OpenGL must be run on main thread.
             mainThread = Thread.currentThread();
+            run();
         }
     }
 
@@ -637,6 +639,7 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
         needClose.set(true);
 
         if (mainThread == Thread.currentThread()) {
+            //If on main thread, no reason to wait.
             return;
         }
 
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -29,7 +29,6 @@
  * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-
 package com.jme3.system.lwjgl;
 
 import com.jme3.input.JoyInput;
@@ -118,8 +117,6 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
     private GLFWErrorCallback errorCallback;
     private GLFWWindowSizeCallback windowSizeCallback;
     private GLFWWindowFocusCallback windowFocusCallback;
-
-    private Thread mainThread;
 
     private double frameSleepTime;
     private long window = NULL;
@@ -303,7 +300,9 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
     protected void setWindowIcon(final AppSettings settings) {
 
         final Object[] icons = settings.getIcons();
-        if (icons == null) return;
+        if (icons == null) {
+            return;
+        }
 
         final GLFWImage[] images = imagesToGLFWImages(icons);
 
@@ -418,9 +417,10 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
             return;
         }
 
-        // NOTE: this is required for Mac OS X!
-        mainThread = Thread.currentThread();
-        run();
+        new Thread(this, THREAD_NAME).start();
+        if (waitFor) {
+            waitFor(true);
+        }
     }
 
     /**
@@ -629,11 +629,6 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
     @Override
     public void destroy(boolean waitFor) {
         needClose.set(true);
-
-        if (mainThread == Thread.currentThread()) {
-            // Ignore waitFor.
-            return;
-        }
 
         if (waitFor) {
             waitFor(false);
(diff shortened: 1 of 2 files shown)


tlf30:
Not sure why the compare is not working correctly, only a couple lines changed...

pspeed42:
Probably you checked the file in with a different line ending.  You may want to fix it as it's harder to see what actually changed.

tlf30:
Ah, I did not think about that. I'll fix it right now

MeFisto94:
FYI: https://github.blog/2018-05-01-ignore-white-space-in-code-review/
As pointed out below:
https://github.com/jMonkeyEngine/jmonkeyengine/pull/1048/files?w=1 also works

Note though that this still means we don't want "unclean" commits, as git tools might not ignore whitespaces and it's not clear which line ending/white space is the correct one.

Actually we should have linters/auto formatters as bots which comment on the PRs :)

pspeed42:
Or I sort of remember that it's possible to set line ending style on the repo or something... but that may make a whole bunch of source change because I don't believe it's consistent right now.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -1,646 +1,641 @@
...
-
-    @Override
-    public void create(boolean waitFor) {
-        if (created.get()) {
-            LOGGER.warning("create() called when display is already created!");
-            return;
-        }
-
-        // NOTE: this is required for Mac OS X!
-        mainThread = Thread.currentThread();
```
MeFisto94: I know what they are doing here and your method might break OSX support. Do some research on the hub please:
For OpenGL to work on Mac OS X it has to be started on the MainThread (which is the Thread the Application is started with, hence Thread.currentThread(). Your change destroys this. So it changes semantics.

WaitFor() might be useless because if you're already on the MainThread, you don't have to wait for it.
But it's an inconsistent behavior as lwjgl2 probably returns from the Application#start() call.
So a proper solution would be to document this or provide AppSettings to enable "forking", unless on OS X.
Or always use the CurrentThread, which to me is expectable if you hand control over to a framework.

tlf30:
OK I have line endings fixed.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
tlf30: Would it be better to check if we are on OS X, then ignore the wiatFor. Also, does this mean that `jme-lwjgl` does not work on OS X?

tlf30:
I would also like to note, that with LWJGL3, the headless context does return from start, which makes inconsistent behavior within the library itself.

tlf30:
OK, I just made a new commit. If waitFor is false, we run it on the main thread, otherwise, create a new thread. How about this solution?

tlf30:
I do not have an apple computer, perhaps someone who does can confirm that this fix works?

tlf30:
Just for documentation purpose, it was issue #433 that has the os x workaround info in it. 

I have tested this proposed fix on my apps in windows, and I think it is ready but still needs tested in OS X. Any thoughts? 

Thanks,
Trevor

stephengold:
Perhaps someone at the Forum would be willing to help us.

Jeddic:
What exactly needs to be tested here? Is there some sort of test case to run?

tlf30:
Just needs to be tested in a working application. Behavior should not change on OS X and that needs to be confirmed.

MeFisto94:
Sorry for not reacting on this so long, I've just re-encountered this without the PR:
Currently, start() blocks until the application is closed, where it should:
- Return instantly in the case of start(false)
- Return as soon as the context is initialized in start(true).

I think your PR addresses the first point, but not the second one? [Which would at least be one improvement]

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
MeFisto94: Yes, this is probably better, but I don't like it cause we'd have an OS dependant behavior then. I need to triple check that with lwjgl3, but as pointed out in #1193, VR has the same problem.

lwjgl2 however is capable of working on Mac OS X, which is related to the way it creates the window, I guess. Which is why we need to double check things and re-check the above assumption. I guess I need to try that on mac.

MeFisto94:
Okay, so after Review in #1248, I think this PR cannot work like this, it needs a more complicated approach, which I didn't manage to do.

The thing is: the GLFW docs say you must not access these functions (most of initInThread) from outside the main thread. They say it's because _some_ platforms require that. I don't know if that will lead to exceptions or work on all those other platforms (i.e. not working on mac os soley), but it's discouraged by the docs.

I think we have to accept that behavior, because everything else gives additional problems, starting with context restarts and a proper app shutdown. 

For now I'd recommend discussing on #1248 about our options.

tlf30:
@MeFisto94 thank you for looking into this. I am not sure why the documentation is so unclear on LWJGL's side on this, but I have been using a different thread for quite some time now in my applications be using this approached: 
```java
                Thread t = new Thread() {
                    @Override
                    public void run() {
                        jme.start();
                    }
                };
                t.start();
```

This was taken right from a working application. I have only tested it in windows though. Using JME3 LWJGL3. 

Perhaps this only works in Windows...?

MeFisto94:
> why the documentation is so unclear on LWJGL's side on this

Because it's a limitation of glfw and their Docs state this clearly.

> Perhaps this only works in Windows...?

And Linux, but it seems to just crash on Mac OS. So while it's forbidden you can get away with that on most platforms.
But as you said in your comment: having start() blocking isn't that bad.

stephengold:
I'm confused about the status of this PR. Does it need code changes? Testing? Or just integration?

Jeddic:
From what I've gathered, there isn't a good way to maintain the previous functionality with lwjgl3 on all platforms. Creating a new thread will crash on OSX. So this isn't really a simple fix, because we can't match the lwjgl2 functionality.

MeFisto94:
What Jeddic says, we're violating the docs/contract when creating a new thread. I tried a different way but failed.

It's a problem: Currently one cannot use start(false) and thus the thread calling into app.start() is locked forever, which is bad for Editors/SDKs. With this PR, that would work on Windows and Linux but crash the JVM on Mac OS X.

I don't know if some (if Mac OS waitFor == true) would work, but we might need to find a solution to the glfw problems anyway

tlf30:
That is what this PR does. It keeps Mac is on main thread, and forks for everything else.

PR #1049, opened 2019-03-19
Fix and Improve TerrainPicking

MeFisto94:
by not adding "ghost" collisions to the CollisionResults (which weren't removed).
Improve TerrainPicking by allowing to do more ray tracing (previously, it stopped on the first hit, now it will stop on the first hit within range).

So: I got the report that terrain picking yields collision results even when no collision has happened. This was due to the way the code was structured: The user had to check the result of collideWith to know if a collision happened or not. This is wrong however, because the passed collisionResults struct could already contain results from previous operations (e.g. when picking the scenegraph recursively) and thus the limit check was wrong (it just picked the "closest" collision, which in that scenario makes no sense) and especially the desired result cannot be identified and removed.

During working on that patch I changed semantics: Previously the first hit was returned, even when it's out of range. The code didn't continue to raycast. Now it returns the first hit WITHIN the range or otherwise continues picking. I don't know if that is valid, because I can only imagine situations in which the next raycast is even more out of range. What do you guys think of that situation?

I'm obviously asking you guys to test picking if it has any downsides (the semantics change should at most only waste CPU cycles, I hope).

Depending on the decision about the semantics, I might add another commit which encapsulates the distance check into it's own method to reduce redundancy.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/terrain/TerrainTestCollision.java
+++ b/jme3-examples/src/main/java/jme3test/terrain/TerrainTestCollision.java
@@ -64,6 +64,9 @@ import com.jme3.terrain.heightmap.ImageBasedHeightMap;
 import com.jme3.texture.Texture;
 import com.jme3.texture.Texture.WrapMode;
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Creates a terrain object and a collision node to go with it. Then
  * drops several balls from the sky that collide with the terrain
@@ -83,7 +86,7 @@ public class TerrainTestCollision extends SimpleApplication {
     protected BitmapText hintText;
     PointLight pl;
     Geometry lightMdl;
-    Geometry collisionMarker;
+    List<Geometry> collisionMarkers;
     private BulletAppState bulletAppState;
     Geometry collisionSphere;
     Geometry collisionBox;
@@ -103,6 +106,7 @@ public class TerrainTestCollision extends SimpleApplication {
 
     @Override
     public void simpleInitApp() {
+        collisionMarkers = new ArrayList<>();
         bulletAppState = new BulletAppState();
         bulletAppState.setThreadingType(BulletAppState.ThreadingType.PARALLEL);
         stateManager.attach(bulletAppState);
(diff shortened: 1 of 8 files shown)


pspeed42:
re: "Now it returns the first hit WITHIN the range or otherwise continues picking. I don't know if that is valid, because I can only imagine situations in which the next raycast is even more out of range."

...if the ray is not vertical, and the limit is long, it may collide multiple parts of the terrain, eh?  I'm not sure how that matches (or not) with what you've said above.  I got confused.

MeFisto94:
>...if the ray is not vertical, and the limit is long, it may collide multiple parts of the terrain, eh? I'm not sure how that matches (or not) with what you've said above. I got confused.
That's correct, I was under the impression that the multiple parts become further away the more we progress, but that assumption is wrong, to me it looks like it's just checking iteratively for every possible x and y coordinate/patch/what ever, so yeah, it may collide with multiple parts of the terrain.

Which brings up another question: For now it only always ever returned one collision and not multiple. It would be better to have multiple but if we're going to do that, unit tests might be appropriate, to be sure it works.

stephengold:
A change like this should include a testcase.

MeFisto94:
Good point. The issue happened when the terrain is at a level < y=0 and picking is performed from 0, 0, 0 with a limit < y. Then cr.size() has to be 0 and closestCollision = null.

Maybe we can combine that case with a ray in -1, -1, 0 direction, so it can hit multiple mountains, but that implies generating some terrain somehow (as asset?)

stephengold:
Existing terrain tests in `jme3-examples` use the following heightmaps from jme3-testdata:
 + `Textures/Terrain/splat/mountains128.png`
 + `Textures/Terrain/splat/mountains512.png`

Other heightmaps available from jme3-testdata:
 + `Textures/BumpMapTest/Simple_height.png`
 + `Textures/Terrain/splat/fortress512.png`
 + `Textures/Terrain/splat/mountains1024.png`
 + `Textures/Terrain/splat/pools.png`

jayfella:
What is the status of this?

stephengold:
Needs review:
  https://hub.jmonkeyengine.org/t/the-contribution-mega-thread/41589/21

jchappell24:
This is becoming a battle that you have to deal with. Not sure what to do
with it and the amount of time for you to get back to me with a good enough
reason to get my money back but Time is of the essence and you don't need
me to do anything about it, so please send me the link to the video and I
will be in touch.

stephengold:
It seems this PR has all the review it's going to get.

PR #1050, opened 2019-03-20
Add tests for com.jme3.scene.mesh.VirtualIndexBuffer

James-Adam:
This PR contains a suite of tests for `com.jme3.scene.mesh.VirtualIndexBuffer`, once again utilising [Diffblue](https://www.diffblue.com/) [Cover](https://www.diffblue.com/datasheet) to assist with the testing workflow. 
The rest of the requested classes and packages are on my radar; i'll be looking into them next. 
Specifically, `com.jme3.renderer.Camera` is the next class i'll be writing tests for. I hope these tests help!

### DIFF
--- a/jme3-core/src/test/java/com/jme3/scene/mesh/VirtualIndexBufferTest.java
+++ b/jme3-core/src/test/java/com/jme3/scene/mesh/VirtualIndexBufferTest.java
@@ -1,0 +1,203 @@
+package com.jme3.scene.mesh;
+
+import com.jme3.scene.Mesh.Mode;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import java.nio.IntBuffer;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertArrayEquals;
+
+public class VirtualIndexBufferTest {
+
+    @Rule public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void testHybrid() {
+        thrown.expect(UnsupportedOperationException.class);
+        new VirtualIndexBuffer(8, Mode.Hybrid);
+    }
+
+    @Test
+    public void testRemaining() {
+        final VirtualIndexBuffer bufferPoints = new VirtualIndexBuffer(8, Mode.Points);
+        assertEquals(8, bufferPoints.remaining());
+        bufferPoints.get();
+        assertEquals(7, bufferPoints.remaining());
+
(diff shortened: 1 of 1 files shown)


riccardobl:
How much is your bot "assisting"? Because there are some parts that don't make logical sense to me, while i can see why a machine could see them as "correct".

It also seems to me that your tests are not testing for validity, but rather locking the current behavior, i can see how this might be useful in some cases (assuming that the current behavior is correct...) but im not entirely sure if it is in this specific case.

James-Adam:
The tool tells you what the current behaviour of the code is in the form of unit tests. It doesn't guess what the intended behaviour _should_ be. Therefore, I carefully reviewed the tests to check whether there is any potentially buggy behaviour in the code under test that comes apparent by the tests. 
I have found many functional bugs that way, but it requires that the reviewer knows the intention of the code. Since you are more expert in your code than I am I would be grateful if you could say which parts don't make logical sense to you.
What should definitely not be done is committing these tests without reviewing them, thus setting in stone buggy behaviour.

The benefit of the tool is that it automatically writes tests that exercise parts of the code that are not tested yet. It particularly comes up with tests for cases I would usually miss when writing tests manually. When I encounter a peculiar behaviour of the code under test, I can easily understand what is going on by debugging through it. That way, it helps me a lot understanding my code, improving my test suite and weeding out bugs that I was not aware of before.

riccardobl:
> Since you are more expert in your code than I am I would be grateful if you could say which parts don't make logical sense to you.

It is not my code, i wan't even aware that we had this class, but here is what i've found

- in testGet the bot is testing for indices that are above numIndices

- testRemaining_Hybrid is not testing remaining but the constructor...

- I'm not sure why testSize is testing only two mesh modes

- the tests for put and getBuffer are not really needed, since they don't validate the functionality of the class.

-  in general, in my opinion, the tests should try to compare a constant index buffer with the values generated by the virtual index buffer for every mesh mode, instead of testing for random indices


testPosition looks fine.

James-Adam:
Thanks for taking the time to point out these issues to me! Some of these issues were in part a lesser understanding of the intended behaviour of the class than I would have liked. 
I've taken some time to implement your feedback into these tests (and at the same time learned a lot more about the use of a Virtual Byte Buffer!) Please let me know whether this is now closer to the style of testing you expect from a class such as this.

riccardobl:
Thanks, it looks fine now.
I'll leave this open for a while in case anyone else wants to comment.

Issue #901, opened 2018-09-12
Native Bullet initializes collision margin to 0 for all shapes


stephengold:
The Bullet 2.83 Physics SDK Manual says:

> It is best not to modify the default collision margin, and if you do use a positive
> value: zero margin might introduce problems.

and

> ## Don’t set the collision margin to zero
> Collision detection system needs some margin for performance and stability. If the gap is noticeable,
> please compensate the graphics representation.

And yet, in the jme-bullet project, collision shapes initialize their margins to zero by default.

We should change the default to a positive value, such as 0.04, which is the Bullet default.

stephengold:
Seeing as this issue is impacting TestCcd, I'll submit a PR to fix this.

stephengold:
Fixed in `master` branch at 1ad324aa57f134078abaa8d2764bd940b6d2a337

Issue #1043, opened 2019-03-18
TestCcd broken for jme3-bullet


stephengold:
In current `master` branch, `jme3test.bullet.TestCcd` behaves differently with `jme3-bullet` than it does with `jme3-jbullet`:

The green projectiles (generated by clicking the left mouse button) have `setCcdMotionThreshold(0.1f)` and are launched at 40 wu/sec velocity. When they're fired toward the obstacle (which is 0.1 wu thick) in `jme3-jbullet`, they consistently bounce off the obstacle.

In `jme3-bullet`, however, the behavior seems to depend on where the projectile collides with the obstacle: in some locations the projectile bounces back, in some it's deflected but passes through, and in some the projectile seems to pass through without deflection.

riccardobl:
It seems that passing a value grather that 0 to setCcdSweptSphereRadius makes the test behave as expected (the default value is 0), im not sure about what having it set to 0 mean for bullet..

stephengold:
I tested some old releases. The test has been broken since at least JME 3.1.0-stable.

stephengold:
The test works with Minie v0.2.10 but not with Minie v0.2.9. This leads me to suspect this issue is related to collision margins, which default to 0 in Minie v0.2.9 (and in all current JME releases). Minie v0.2.10 changed the default collision margin to 0.04 .

stephengold:
See issue #901.

stephengold:
Increasing the collision margin of the obstacle to 0.0001 is sufficient to resolve this issue. I'm undecided whether to modify `TestCcd` or address the larger issue #901 in `CollisionShape`.

stephengold:
So far we have 3 possible fixes:
 (1) increase the collision margin of the obstacle in `TestCcd`
 (2) implement a default collision margin >0 in` jme3-bullet`
 (3) increase the swept-sphere radius in `TestCcd`

@riccardobl: Which fix do you think is best for the project?

riccardobl:
I'd say that we should definitely fix it by increasing the default margin, since as you reported in issue #901 the official documentation advise against using 0 margin.

MeFisto94:
Definitely against 1, because in my game, ccd doesn't work either. thus the real solution probably is 2.
Regarding 3: I could swear I was having troubles with a set swept-sphere radius. Probably only under steep angles and high speed, but still.

Edit yep:
```java
[...]

PR #1053, opened 2019-03-22
fix issue #901 (collision margins initialized to 0)

stephengold:
This adds a new public static method to set the default collision margin.

I'm unsure whether `jme3-jbullet` has any issue with 0 collision margin. To maintain consistency between libraries I modified both `jme3-bullet` and `jme3-jbullet`.

I ran all the physics tests in `jme3-examples` with both physics libraries and didn't notice any changes, except that `TestCcd` now works with `jme3-bullet`. Refer to issue #1043 for details.


stephengold:
I just realized the default collision margin needs a getter too. One moment...

PR #1061, opened 2019-04-05
Add tests for com.jme3.renderer.Camera

James-Adam:
This PR contains a suite of tests for `com.jme3.renderer.Camera`, once again utilising [Diffblue](https://www.diffblue.com/) [Cover](https://www.diffblue.com/datasheet) to assist with the testing workflow. I've also added additional` .equals()` methods for Camera.java and Plane.java to assist testing of the this class.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Plane.java
+++ b/jme3-core/src/main/java/com/jme3/math/Plane.java
@@ -291,4 +291,13 @@ public class Plane implements Savable, Cloneable, java.io.Serializable {
             throw new AssertionError();
         }
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Plane plane = (Plane) o;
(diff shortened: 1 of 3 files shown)


riccardobl:
Sorry, this test is using autogenerated garbage data that make no sense (negative radius spheres, zero vector as directions, viewports with 0 area...etc...).

jme3-core/src/main/java/com/jme3/math/Plane.java
```diff
@@ -291,4 +291,13 @@ public Plane clone() {
             throw new AssertionError();
         }
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
Ali-RS: Also please use curly Braces for if statement. 
```
if (this == o) {
     return true;
}
```

It is recommended to use the default code formatter of your IDE if you are in Netbeans or Eclipse.

Issue #1065, opened 2019-04-09
ClassCastException in Uniform class


rvandoosselaer:
Fix in: #1066 

There is an error in the class cast checking in the Uniform class, line 376.

```java
...
//feed this.value with the collected values.
if (this.value instanceof ColorRGBA) {
    ((ColorRGBA) this.value).set(vec4.x, vec4.y, vec4.z, vec4.w);
} else if (value instanceof Vector4f) {
    ((Vector4f) this.value).set(vec4);
} else {
    ((Quaternion) this.value).set(vec4.x, vec4.y, vec4.z, vec4.w);
}
...
```

the check is looking for `value` instead of `this.value`.

Stacktrace:
```
java.lang.ClassCastException: com.jme3.math.Vector4f cannot be cast to com.jme3.math.Quaternion
        at com.jme3.shader.Uniform.setValue(Uniform.java:379) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.material.Material.updateShaderMaterialParameters(Material.java:847) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.material.Material.render(Material.java:1018) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.renderer.RenderManager.renderGeometry(RenderManager.java:614) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.renderer.queue.RenderQueue.renderGeometryList(RenderQueue.java:266) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.renderer.queue.RenderQueue.renderQueue(RenderQueue.java:305) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.renderer.RenderManager.renderViewPortQueues(RenderManager.java:877) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.renderer.RenderManager.flushQueue(RenderManager.java:779) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
        at com.jme3.renderer.RenderManager.renderViewPort(RenderManager.java:1108) ~[jme3-core-3.3.0-alpha1.jar:3.3-alpha1]
[...]

PR #1066, opened 2019-04-09
fixed wrong reference check #1065

rvandoosselaer:
See #1065


tlf30:
Is there an issue describing the problem that is being fixed here?

rvandoosselaer:
> Is there an issue describing the problem that is being fixed here?
Yes, didn't link the issues properly. I described the problem in #1065

pspeed42:
Nice catch.  Those kinds of typos can slip right by.

PR #1067, opened 2019-04-10
Added an easy way to add sub steps to the profiler

Jeddic:
This is the implementation I use in spoxel to dig into performance within app states. The default profiler only gives timings for the App State Manager, but this code automatically profiles each of the app states added to the App State Manager. You can set further breakpoints within the update methods of an app state by calling:

`profiler.appSubStep("EffectState", "Run Client Effects");`

In this case "EffectState" is an app state and "Run Client Effects" is one of the profiling points i've set.

### DIFF
--- a/jme3-android/src/main/java/com/jme3/app/DefaultAndroidProfiler.java
+++ b/jme3-android/src/main/java/com/jme3/app/DefaultAndroidProfiler.java
@@ -135,6 +135,11 @@ public class DefaultAndroidProfiler implements AppProfiler {
         }
     }
 
+    @Override
+    public void appSubStep(String... additionalInfo) {
+        
+    }
+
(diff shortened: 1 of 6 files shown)


pspeed42:
Neat.  The path manipulation weirded me out at first but then I noticed the whole class is smelly in that regard.  So good job navigating that!

For the record: my notes on the smelliness that have nothing to do with this patch:
1) A stateful "path" field that seems to only be used as a temporary string builder... probably should have been called tempPath to make that clearer.
2) many cases where locally scoped variables "path" collide with that name causing confusion.
3) the getPath() is called all over the place and routinely creates new StringBuilders all the time... making me question the value of the temp path string builder.

Jeddic:
Yea, the whole path implementation was a little weird but it works for how simple the profiler is. It worked for me with spoxel and really helped narrow down some of the performance problems I had hidden in my code. I added a missing file change to the commit.

Issue #363, opened 2015-10-07
BitmapFont positions text incorrectly when VAlign is set to "Center" or "Bottom"


slyh80:
If the user sets the vertical align of a `BitmapFont` to either `VAlign.Center` or `VAlign.Bottom` it positions the text incorrectly. I'll only discuss `VAlign.Bottom` here.

For `VAglin.Bottom` the text is positioned _below_ the defined text box and renders downwards from there. Instead it should position the text at the bottom border of text box and render upwards.

Screenshot:
https://www.dropbox.com/s/2idcx1oxg4jrwl9/valign-bottom-incorrect.png?dl=0

The right text should be inside the blue box; the last line "touching" the bottom border.

The bug is probably caused by in incorrect calculation of `totalHeight` in method `Letters.validateSize()`.

Code to reproduce:

```
BitmapText label = font.createLabel("This is a test text that has a bottom valign.");
Rectangle textbox = new Rectangle(380, 240, 110, 100);
label.setBox(textbox);
label.setVerticalAlignment(VAlign.Bottom);

Geometry backgroundBox = new Geometry("", new Quad(textbox.width, -textbox.height, true));
backgroundBox.setMaterial(mat);
backgroundBox.setLocalTranslation(textbox.x, textbox.y, 0);
getGuiNode().attachChild(backgroundBox);

getGuiNode().attachChild(label);
```

stephengold:
I was able to reproduce this issue in current `master` branch.

stephengold:
Thank you for pursuing this issue!

stephengold:
Re-target to v3.3.0, due to risk of breaking existing applications.

PR #1068, opened 2019-04-10
Calculate the total line height of a text

rvandoosselaer:
This PR fixes: #363 

The `totalHeight` was calculated using this line:

```java
totalHeight = Math.max(totalHeight, -l.getY1());
```

This always resulted in `totalHeight = 0`. I changed the calculation to using the lineHeight of the font and multiplying it by the line count.

```java
totalHeight = font.getLineHeight(block) * block.getLineCount();
```

Since the new height calculation needs the proper line count, I inversed the order of calling the methods in the `update()` method. First the line count is set, then the alignment is calculated.

I also added a test case with some different alignment settings: `TestBitmapFontAlignment`.

Screenshot-20190411-102416

### DIFF
--- a/jme3-core/src/main/java/com/jme3/font/Letters.java
+++ b/jme3-core/src/main/java/com/jme3/font/Letters.java
@@ -35,6 +35,7 @@ import com.jme3.font.BitmapFont.Align;
 import com.jme3.font.BitmapFont.VAlign;
 import com.jme3.font.ColorTags.Range;
 import com.jme3.math.ColorRGBA;
+
 import java.util.LinkedList;
 
 /**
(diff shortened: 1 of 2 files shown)


tlf30:
Thank you for the fix! 
The code looks good to me but I did not test it, thoughts anyone?

pspeed42:
It looks like you might be only the second person in the world that has made a change to BitmapText classes without breaking something else.  I mean, time will tell, I guess. :)  Looks good to me, though (and I'm the only other person whose done it... and I did it by failing 5-10 times before that.)

stephengold:
Re-target to v3.3.0 due to the risk of breaking existing applications.

PR #1059, opened 2019-04-02
Jdk 11 Build Support

tlf30:
OK, this PR adds build support for JDK 11. This has been tested on both JDK 8 and JDK 11. 

Can everyone please look over this and TEST it on both JDK 8 and 11 to confirm that it works for them, and on Linux, OS X, and Android if you have them (I have only tested on Windows 10).

Take note, this does NOT add modules to JME3, it simply makes it able to compile using JDK 11. 

I would love some feed back, let me know if I forgot anything.

Current Platforms Tested:
- [x] Windows 10
- [x] Linux
- [ ] OS X
- [x] Android

Thank you,
Trevor Flynn

PR #1072, opened 2019-04-18
partial revert PR #1059 (JDK 11 support) which duplicated .class files

stephengold:
This PR proposes reverting the part of PR #1059 which caused class files to be stored in the source-code directories. This will break JDK10/11 builds, where `javah` no longer exists. However it will allow development on JDK7/8/9 to proceed while a cleaner solution for JDK10/11 builds is developed.

### DIFF
--- a/jme3-android-native/decode.gradle
+++ b/jme3-android-native/decode.gradle
@@ -44,19 +44,13 @@ task copyTremorFiles(type: Copy) {
     into outputDir
 }
 
-// Generate headers via javac -h
+// Generate headers via javah
 task generateJavahHeaders(type: Exec) {
-    def files0 = fileTree("src/main/java/").filter { it.isFile() && it.getName().endsWith(".java") }.files
-    def files1 = fileTree("src/common/java/").filter { it.isFile() && it.getName().endsWith(".java") }.files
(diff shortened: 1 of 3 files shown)


riccardobl:
What about the changes made to jme3-android-native/openalsoft.gradle and jme3-android-native/decode.gradle, i think those should be reverted aswell.

stephengold:
ok

tlf30:
@stephengold did you attempt to put the `-d` flag on the `javac -h` command? I take it from this PR that it does not work?

tlf30:
@stephengold I have confirmed that adding the `-d` flag does work. Instead of reverting the changes, I think it would be better to just fix the issue now. 
`args '-d', classBuildDir`
Where `classBuildDir`: `String classBuildDir = "${buildDir}" + File.separator + 'classes'`

stephengold:
@tlf30 I created this pull request as a temporary stopgap. If you have a better solution, please submit it as a pull request ASAP, and I'll withdraw this PR.

stephengold:
@tlf30 Will you submit your solution as a PR?

tlf30:
Sorry, just got out of bed. I'll do it now

tlf30:
@stephengold Done

stephengold:
I withdraw this PR in favor of #1073.

PR #1073, opened 2019-04-19
Fixed classes location while building natives (JDK11 Support)

tlf30:
This PR fixes the issue of class files being placed in the source tree during native binary builds. 
See #1072 for more info.


stephengold:
Thank you.

PR #1077, opened 2019-04-25
Added ambient light support in PBR shader

Ali-RS:
to control the brightness and color of the light probe. Very useful for simulating day/night light and any use case where you don't rebake the probes every time the ambient color needs to change.


Ali-RS:
This uses a cleaner solution over my previous PR: #1075

grizeldi:
This will also be very useful for lowering probe intensity so that the point lights stand out more.

On the other hand, what happened with the indent formatting in the shader file?

Ali-RS:
> On the other hand, what happened with the indent formatting in the shader file?

Some of them had no indent, some them had 5 and most of them had 2 space, so I updated them to use 2 spaces.

PR #1078, opened 2019-04-25
Basic support for Transform Feedback

TehLeo:
Adds basic support for Transform Feedback (TF).
As part of discussion on discord
- new class BufferObject was added which manages BufferObjects
- VertexBuffer now extends BufferObject
- jme3.shader.BufferObject was renamed to UniformBufferObject and now extends BufferObject

To support basic transform feedback.
- setTransformFeedbackOutput(TransformFeedbackOutput output) method was added to the Renderer interface

The use of this method is similar to how rendering of framebuffers is handled in jme. This method is called to enable transform feedback, after which rendered geometry's TF ouput will be captured to the specified output buffers. Calling it will null, disables transform feedback.

In order to use TF, a shader has to specify the output variables which will be outputted to TF. This is accomplished by specifying the output variables in Technique block of .j3md file. Syntax is:
```
TransformFeedback [Off | Interleaved | Separate | InShader] varying1 varying2 varying3 ...
```
For example a vertex/geometry shader contains the ouput:
```
out mat4 TransformFeedback;
```
A technique can be written as:
```
Technique {
    VertexShader GLSL130: jme3test/renderer/TestTR.vert
    GeometryShader GLSL150: jme3test/renderer/TestTR.geom
    RenderState {
        RasterizerDiscard On
    }
    TransformFeedback Interleaved TransformFeedback
}
```
- added Render State RasterizerDiscard - when enabled Fragment Processing stage is skipped, 
- added QueryObject, retrieve the number of written primitives for TF operation
- added TF testcase jme3test.renderer.TestTransformFeedback

Awaiting your feedback.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/DetailedProfiler.java
+++ b/jme3-core/src/main/java/com/jme3/app/DetailedProfiler.java
@@ -57,6 +57,9 @@ public class DetailedProfiler implements AppProfiler {
             for (StatLine statLine : data.values()) {
                 for (Iterator<Integer> i = statLine.taskIds.iterator(); i.hasNext(); ) {
                     int id = i.next();
+                    //FixMe
+                    //throws with app.settings.putBoolean("GraphicsDebug", true);
+                    //reason: query has not started/ended or is already running
                     if (renderer.isTaskResultAvailable(id)) {
                         long val = renderer.getProfilingTime(id);
                         statLine.setValueGpu(val);
--- a/jme3-core/src/main/java/com/jme3/material/RenderState.java
+++ b/jme3-core/src/main/java/com/jme3/material/RenderState.java
@@ -34,6 +34,8 @@ package com.jme3.material;
 import com.jme3.export.*;
 import com.jme3.scene.Mesh;
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * <code>RenderState</code> specifies material rendering properties that cannot
@@ -432,7 +434,7 @@ public class RenderState implements Cloneable, Savable {
          */
         Invert
     }
-
+    
     static {
         NULL.cullMode = FaceCullMode.Off;
         NULL.depthTest = false;
@@ -446,6 +448,7 @@ public class RenderState implements Cloneable, Savable {
         ADDITIONAL.applyColorWrite = false;
         ADDITIONAL.applyBlendMode = false;
         ADDITIONAL.applyPolyOffset = false;
+        ADDITIONAL.applyRasterizerDiscard = false;
     }
     boolean wireframe = false;
     boolean applyWireFrame = true;
@@ -485,6 +488,8 @@ public class RenderState implements Cloneable, Savable {
     BlendFunc dfactorRGB = BlendFunc.One;
     BlendFunc sfactorAlpha = BlendFunc.One;
     BlendFunc dfactorAlpha = BlendFunc.One;
+    boolean rasterizerDiscard = false;
+    boolean applyRasterizerDiscard = true;
             
     public void write(JmeExporter ex) throws IOException {
         OutputCapsule oc = ex.getCapsule(this);
@@ -515,6 +520,7 @@ public class RenderState implements Cloneable, Savable {
         oc.write(dfactorRGB, "dfactorRGB", dfactorRGB);
         oc.write(sfactorAlpha, "sfactorAlpha", sfactorAlpha);
         oc.write(dfactorAlpha, "dfactorAlpha", dfactorAlpha);
+        oc.write(rasterizerDiscard, "rasterizerDiscard", rasterizerDiscard);
 
         // Only "additional render state" has them set to false by default
         oc.write(applyWireFrame, "applyWireFrame", true);
@@ -526,7 +532,7 @@ public class RenderState implements Cloneable, Savable {
         oc.write(applyPolyOffset, "applyPolyOffset", true);
         oc.write(applyDepthFunc, "applyDepthFunc", true);
         oc.write(applyLineWidth, "applyLineWidth", true);
-
+        oc.write(applyRasterizerDiscard, "applyRasterizerDiscard", true);
     }
 
     public void read(JmeImporter im) throws IOException {
@@ -557,8 +563,8 @@ public class RenderState implements Cloneable, Savable {
         dfactorAlpha = ic.readEnum("dfactorRGB", BlendFunc.class, BlendFunc.One);
         sfactorRGB = ic.readEnum("sfactorAlpha", BlendFunc.class, BlendFunc.One);
         dfactorAlpha = ic.readEnum("dfactorAlpha", BlendFunc.class, BlendFunc.One);
-
-
+        rasterizerDiscard = ic.readBoolean("rasterizerDiscard", false);
+        
         applyWireFrame = ic.readBoolean("applyWireFrame", true);
         applyCullMode = ic.readBoolean("applyCullMode", true);
         applyDepthWrite = ic.readBoolean("applyDepthWrite", true);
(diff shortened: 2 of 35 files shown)


jme3-core/src/main/java/com/jme3/app/DetailedProfiler.java
```diff
@@ -57,6 +57,9 @@ public void appStep(AppStep step) {
             for (StatLine statLine : data.values()) {
                 for (Iterator<Integer> i = statLine.taskIds.iterator(); i.hasNext(); ) {
                     int id = i.next();
+                    //FixMe
+                    //throws with app.settings.putBoolean("GraphicsDebug", true);
+                    //reason: query has not started/ended or is already running
```
riccardobl: Can you elaborate? Is this caused by the PR?

jme3-core/src/main/java/com/jme3/material/RenderState.java
```diff
@@ -34,6 +34,8 @@
 import com.jme3.export.*;
 import com.jme3.scene.Mesh;
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Objects;
```
riccardobl: Unused imports

jme3-core/src/main/java/com/jme3/renderer/Caps.java
```diff
@@ -552,7 +590,37 @@ public static boolean supports(Collection<Caps> caps, Shader shader){
...
+     * can be supported by the renderer.
+     * 
+     * @param caps The collection of renderer capabilities {@link Renderer#getCaps() }.
+     * @param query The query object to check
+     * @return True if it is supported, false otherwise.
+     */
+    public static boolean supports(Collection<Caps> caps, QueryObject query) {
+        switch(query.getType()) {
+            case SamplesPassed:
+                //GL1.5 is available or (GL_ARB_occlusion_query)
```
riccardobl: So.. are we supposed to test if 1.5 is available?

jme3-core/src/main/java/com/jme3/renderer/Renderer.java
```diff
@@ -33,9 +33,11 @@
 
 import com.jme3.material.RenderState;
 import com.jme3.math.ColorRGBA;
+import com.jme3.scene.BufferObject;
 import com.jme3.scene.Mesh;
+import com.jme3.scene.TransformFeedbackOutput;
 import com.jme3.scene.VertexBuffer;
-import com.jme3.shader.BufferObject;
+import com.jme3.shader.UniformBufferObject;
```
riccardobl: Unused import

jme3-core/src/main/java/com/jme3/renderer/RenderContext.java
```diff
@@ -275,6 +296,11 @@
     
     public ColorRGBA clearColor = new ColorRGBA(0,0,0,0);
     
+    /**
+     * Started queries.
+     */
+    public int[] startedQueries = new int[6/*QueryObject.Type.values().length*/];
```
riccardobl: What is the point of this? Just to remember which queries where started? Do we really need it?

jme3-core/src/main/java/com/jme3/renderer/QueryObject.java
```diff
@@ -0,0 +1,263 @@
...
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.jme3.renderer;
+
+import com.jme3.util.NativeObject;
+import java.util.EnumMap;
```
riccardobl: Unused import?

jme3-core/src/main/java/com/jme3/renderer/QueryObject.java
```diff
@@ -0,0 +1,263 @@
...
+         * or by checking if it is available.
+         * The query can start by calling the begin method.
+         */
+        CanBegin,
+        /**
+         * A started query can only end.
+         */
+        CanEnd
+    }
+
```
riccardobl: Is this really needed?

jme3-core/src/main/java/com/jme3/app/DetailedProfiler.java
TehLeo: No its not caused by PR, the line below renderer.isTaskResultAvailable(id)
issues 
`gl.glGetQueryObjectiv(taskId, GL.GL_QUERY_RESULT_AVAILABLE)`
This method is not allowed to be called unless 
`gl.glBeginQuery(GL.GL_TIME_ELAPSED, taskId);` and 
`gl.glEndQuery(GL.GL_TIME_ELAPSED);`
are called. In other words it asks if result is ready without first starting and ending a query thus causing a glerror.

To resolve: This file will have to be rewritten with the use of the newly introduced QueryObject not only to fix this issue but also to not cause issues with user defined Queries.

riccardobl: Ok. Then if you can plase open an issue for this.

jme3-core/src/main/java/com/jme3/renderer/Caps.java
riccardobl: What about GL ES?

jme3-core/src/main/java/com/jme3/renderer/RenderContext.java
TehLeo: Yes, unless a query is started issuing a glEndQuery or checking for result yield a glerror. This is here to prevent that. Furthermore, take a look at the following 
```
void glBeginQuery(GLenum target, GLuint id);
void glEndQuery( GLenum target);
```
Do you see something off about the functions above?
The off thing about them is that glEndQuery does not take query id as a parameter!

Thus in other words lets say I make two timer queries:
```
queryT1 = new QueryObject(app.getRenderer(), QueryObject.Type.TimeElapsed);
queryT2 = new QueryObject(app.getRenderer(), QueryObject.Type.TimeElapsed);
```
If you run the following code:
```
queryT1.begin();
queryT2.begin();
//some work
queryT2.end();
queryT1.end();
```
You will correctly get a RenderException (at line `queryT2.begin();`)
```
com.jme3.renderer.RendererException: Query of type TimeElapsed already started.
```
Without the render exception you would get gl error. In other words you are not allowed to nest the same type of query.

jme3-core/src/main/java/com/jme3/renderer/QueryObject.java
TehLeo: Refer to the answer above, in short, to prevent gl errors.

jme3-core/src/main/java/com/jme3/renderer/RenderContext.java
riccardobl: Sure, but this is never exposed outside of the core. So maybe we don't need those checks ...

jme3-core/src/main/java/com/jme3/material/RenderState.java
TehLeo: thanks for point out, i'll fix these with next commit

jme3-core/src/main/java/com/jme3/renderer/RenderContext.java
TehLeo: Are you saying that users will not be able to use QueryObject?

riccardobl: As usual there is nothing stopping you from using internals in jme, but imo this is not an end user feature so maybe we can avoid having all those extra checks.

TehLeo: Why not have the checks? Those checks are not expensive. You are not starting 10 thousands  of queries per frame. They only bring robustness. So I do not see a reason for them not to be there.
Also even if it was used just in internals, there's always a risk that you can nest some queries and this would be a nightmare for figuring out such error.
Eg. Imagine you use TimerQuery while you also have DetailedProfiler attached. And there you go... nested queries.

Issue #99, opened 2014-03-24
blendMode=”multiply” in Nifty renders incorrectly


ghost:
Reference on googlecode: https://code.google.com/p/jmonkeyengine/issues/detail?id=609

stephengold:
5 years later, I'm still working around this issue.

stephengold:
Forum discussion: https://hub.jmonkeyengine.org/t/nifty-labels-inside-a-scrollpanel-control/27402

stephengold:
With the help of the Internet Archive, I re-created the workaround from 2014:
   https://hub.jmonkeyengine.org/t/text-within-a-nifty-scroll-panel-acquires-a-black-surround/41768

stephengold:
I discovered a possible fix for this issue: replace 
```
    } else if (blendMode == BlendMode.MULIPLY) {
        return RenderState.BlendMode.Modulate;
```
with
```
    } else if (blendMode == BlendMode.MULIPLY) {
        return RenderState.BlendMode.Alpha;
```
in both `JmeBatchRenderBackend.java` and `RenderDeviceJme.java`.

I imagine such a change might have undesirable side effects, but so far I haven't observed any.

tonihele:
Ok, I tested this. I don't know what to make of these results. I took our custom styling off from the ScrollPanel. I typed the JME BlendMode on the result images. I could imagine that the Nifty's multiply translates to JME's Modulate. We should have some sort of basic test case.
For our part. I didn't see any ill-effects either way. But we don't seem to use the blend mode multiply at all by default.
https://github.com/nifty-gui/nifty-gui/wiki/Effects#blendMode
![image](https://user-images.githubusercontent.com/8344766/57016496-23b17b80-6c23-11e9-8e43-f0c792e0ee20.png)
![image](https://user-images.githubusercontent.com/8344766/57016504-2ad88980-6c23-11e9-810a-503e005e5377.png)

stephengold:
Thanks for trying this. So your custom scroll panel doesn't use blendMode="multiply"? I'm wondering if it has any application outside of scroll panels...

tonihele:
Well, the in the pictures it uses. I dropped our custom styling, so I assume it used. And changing the value it translated to in JME has the following outcomes. But I don't know what to make of that.

I guess the attribute can be used in effects etc.?

stephengold:
I've decided the only available fix is too risky for JME 3.2.4

stephengold:
[...]

PR #1082, opened 2019-04-30
fix for issue #99

stephengold:
I tried various ways of rendering blendMode=”multiply” in NiftyGUI, and this is the one that worked best. This PR will need considerable testing to ensure it doesn't break anything.


stephengold:
I've decided this is too risky for 3.2.4

Issue #1074, opened 2019-04-20
Renderer interface lacks getters for many properties


stephengold:
The `com.jme3.renderer.Renderer` interface defines `setAlphaToCoverage()`, `setBackgroundColor()`,  `setClipRect()`, `setDefaultAnisotropicFilter()`, `setDepthRange()`, `setLinearizeSrgbImages()`, `setMainFrameBufferSrgb()`, and `setViewPort()`. However, it does not provide getters for any of these properties.

Runtime access to these properties might aid in debugging render issues.

stephengold:
Further investigation indicates that many of these methods are used dynamically by the `RenderManager`:  `setBackgroundColor()`, `setClipRect()`, `setDepthRange()`, and `setViewPort()`. The corresponding properties are probably of little interest to JME users, so perhaps skip them.

Properties of interest to users include "alphaToCoverage" and "defaultAnisotropicFilter". Not sure yet about "linearizeSrgbImages" and "mainFrameBufferSrgb".

quazi-irfan:
I'll take a look at it tonight.

stephengold:
It looks like "linearizeSrgbImages" and "mainFrameBufferSrgb" are set during initialization of the `JmeContext` and seldom changed thereafter. I infer that their settings might be of interest to users. I'll leave this issue open in case someone wants to implement them in the style of PR #1093.

PR #1093, opened 2019-05-21
Attempt to fixing 1074

karanehra:
This is my first attempt at resolving #1074.

Added getters for "alphaToCoverage" and "defaultAnisotropicFilter" in the Renderer interface.
Added respective method stubs to GLRenderer and NullRenderer implementing Renderer.

I'm still trying to implement the exact method in the GLRenderer. Any help appreciated.


stephengold:
Needs tests to show that the new methods work as intended.

riccardobl:
The indentations is very wrong, besides that this seems ok.

stephengold:
Some of the changes are unnecessary. Please revert the changes to `jme3-core/build.gradle` and `TestAspectRatio.java`.

The files in `jme3-core/.idea` appear to be IDE-specific and should not be part of this pull request. Please remove them. To avoid committing such files in the future you may want to add `**/.idea` to your `.gitignore` file.

As Riccardo pointed out, the added code in `GLRenderer.java` and `NullRenderer.java` isn't correctly indented. Please reformat this code, either to match the surrounding code or else to conform to the [Google style](https://google.github.io/styleguide/javaguide.html#s4-formatting).

karanehra:
Added the requisite changes.

stephengold:
A few more requests...

1. `TestAlphaToCoverage.java` should also verify the behavior of `getAlphaToCoverage()` *before* `Caps.Multisample` is added: verify that it returns `false`.
 
2. Since `TestAlphaToCoverage.java` doesn't display anything relevant, I suggest adding a `stop()` at the end of `simpleInitApp()` to terminate the test.

3. The purpose and expected result of `TestAlphaToCoverage.java` should be documented. Please add a Javadoc comment similar to the one in `TestAspectRatio.java`.

karanehra:
Done sir.

riccardobl:
TBH this test seems unnecessary, but if we really want to keep it, i think it should go in tests (not in examples). Actually is it possible to make it an unit test?

karanehra:
I actually tried that using the help of @stephengold . But we were having a lot of problems going that way.

louhy:
Care to share which problems? Maybe there isn't a better option, but maybe someone can think of a workaround - we're here to help! :grin:

karanehra:
Reverted all unnecessary changes. @riccardobl i mentioned you in the relevant thread. We would appreciate your help.

karanehra:
@louhy what's your handle at hub.jmonkeyengine? I'll mention you too.

louhy:
@karanehra Handle same as here - but I'll see it.

riccardobl:
@karanehra which thread exactly?

karanehra:
@riccardobl the relevant thread at hub.jmonkeyengine

riccardobl:
I can't see any relevant thread and i didn't get any mention :man_shrugging:

stephengold:
The thread is a private message thread between me and @karanehra:
   https://hub.jmonkeyengine.org/t/eager-to-contribute/41866/61

If it's not visible to @riccardobl and @louhy at this point, then I don't know how to make it visible.

@karanehra tried to make an automated test using `junit`: modified the `build.gradle` to add the necessary `testCompile` and `testRuntime` dependencies, explicitly loading the `lwjgl` native library, and instantiating a `GLRenderer`. However, we discovered that the renderer would need a `Display` in order to be properly initialized. At that point, I suggested writing a `SimpleApplication` instead.

riccardobl:
I can't access it, i think you mentioned someone else lel.
I see, it might be possible to configure the CI to use software render, then you would have the display. But i think this is eventually something for the future.

louhy:
Ahh, sounds like what they call "headless" mode. Seen it done in JavaFX land but not sure how easy (or possible) to do it for GL. Would be cool but probably a mini-project on its own.

Issue #1097, opened 2019-05-26
OpenGL version naming is inconsistent


riccardobl:
The version naming in AppSettings is a bit inconsistent, in detail
```
LWJGL_OPENGL3 should be renamed in LWJGL_OPENGL32
LWJGL_OPENGL30 should be added for gl 3.0
LWJGL_OPENGL4 should be renamed in LWJGL_OPENGL40
```
The changes can be applied without breaking backward compatibility by deprecating LWJGL_OPENGL3 and LWJGL_OPENGL4 and fixing the renderers to deal with the new naming, see the commented lines in https://github.com/jMonkeyEngine/jmonkeyengine/pull/1096/commits/e31fc2c96fa8c8e58761eb1c969534a169d3a3ac

louhy:
I can clean that up for you.

louhy:
Hmm, noticed a weird name problem on commit, linked to an old account for some reason so it's saying the wrong author. Will try to fix...

louhy:
Should be good now. 

I'm used to just committing myself or reviewing others changes, so first try with the PR thing on GitHub - let me know if I goofed. :)

riccardobl:
Hey thanks. The changes need to be ported also on the lwjgl3 renderer.
When you are ready, plase submit a pr following these instructions https://help.github.com/en/articles/creating-a-pull-request#creating-the-pull-request

louhy:
Gotcha - already submitted the PR but feel free to reject it etc. 
I think I see what you mean, will take a look at that one tomorrow if you don't mind. 
And no problem - it's time I learn how this OSS collaboration stuff works in practice so I can help out. :)

riccardobl:
Oh i see, you submitted the pr on my fork, it should submitted on this repo instead. :+1:

louhy:
Can do - yeah I was wondering about that since you hadn't merged that yet, I should probably wait until yours is merged, I guess?

riccardobl:
Done, you can reapply the changes to the current master.

louhy:
Should be ready. I did notice there are some internal JME references to the deprecated constants. Should also replace LWJGL_OPENGL3 with LWJGL_OPENGL32 and LWJGL_OPENGL4 with LWJGL_OPENGL40 where they're referenced in subprojects (vr, desktop, etc.) at some point.

riccardobl:
Fixed in https://github.com/jMonkeyEngine/jmonkeyengine/pull/1099

PR #1099, opened 2019-05-27
#1097-Fixed inconsistent OpenGL version constants

louhy:


riccardobl:
:thinking: the lwjgl3 patch seems incomplete. The new constants should be added here  too
https://github.com/jMonkeyEngine/jmonkeyengine/blob/f45598b16e164e8988698920d0ae72950286bc3d/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java#L78-L111

louhy:
Yeah. Actually now I'm thinking there's more needed too. Ex. this in LwjglContextVR.java should probably accept also LWJGL_OPENGL30. Will take some more time to go over this carefully. 

```
if (settings.getRenderer().equals(AppSettings.LWJGL_OPENGL2)
                || settings.getRenderer().equals(AppSettings.LWJGL_OPENGL3)) {
```

louhy:
**IMO, now I think I need to do more here.** Those constants are public API so I'm afraid that if someone switches to the non-deprecated constants, any internal code that checks them may only be looking for the deprecated ones. Will look deeper at it this afternoon or tonight.

louhy:
Wasn't so extensive after all, just vr and desktop. Hope that helps.

riccardobl:
:thinking:  I have an idea 
```java
   @Deprecated
    public static final String LWJGL_OPENGL3 = "LWJGL-OpenGL3";

    public static final String LWJGL_OPENGL32 = LWJGL_OPENGL3;
```
If you print it, it will say LWJGL-OpenGL3 instead of *32 ... but i think we can live with that

louhy:
That should work too, but I believe I've corrected the other references - do you have any concerns with how it is now? It (slightly) bothers me that 3 and 3.2 are conflated like this.

riccardobl:
Now that you brought up the issue, i see that this might be a problem if a library wants to check the renderer and relies on the deprecated constants.

We can either 
1. live with the bad naming where 30 is 3.0 and 3 is actually 3.2...
2. make the new constants refer to the old ones even if the string content is still wrong
3. fix the naming and accept that this change might break some usecases

:thinking:

riccardobl:
I think the VR renderer can use this patch https://github.com/jMonkeyEngine/jmonkeyengine/commit/a68d8b50e31d9aceaee4a1dfba28e9c2a70b2cfc since it seems just a modified version of lwjgl2.
This would both fix the issue and enable the other versions like with the lwjgl2 renderer

louhy:
I'm not following #2, but have you seen this commit? https://github.com/jMonkeyEngine/jmonkeyengine/pull/1099/commits/06c2ab902cd184b00be7f2e9b40065479c1e55b8 

Maybe I went down a 4th route? I think it's #3 but hopefully without breakage, NB isn't finding checks of these constants anywhere else. VR was updated to check for both the deprecated 3 and the new 3.2. I'm not familiar with the VR subproject's requirements, so I didn't want to assume it supported anything other than the versions explicitly specified. 

The only thing I can imagine going wrong is if someone hard coded the constant strings somewhere in an external project, but that's just... wrong.

riccardobl:
> The only thing I can imagine going wrong is if someone hard coded the constant strings somewhere in an external project, but that's just... wrong.

That's what im talking about, it breaks libraries that check the renderer with getRenderer().equals(LWJGL_OPENGL3) if the main project is using LWJGL_OPENGL32.

louhy:
You're saying VR checked for LWJGL_OPENGL3, so if the main project uses LWJGL_OPENGL32, that's a version change from what the VR library expected? 

If I understood that correctly, my reasoning and understanding was that although the library was asking for LWJGL_OPENGL3, JME was always assuming 3.2 capabilities anyway, so VR should support both LWJGL_OPENGL3 (which means 3.2) and LWJGL_OPENGL32 (which means 3.2).

Sorry if I'm being dense and you want to move on. I'm sure you're more familiar with how the renderers work and maybe I'm missing under-the-hood context. 

If we just need to make this change I can do that. Was that the only one? (Should LWJGL_OPENGL40  = LWJGL_OPENGL4? LWJGL_OPENGL30 = LWJGL_OPENGL3?)
```java
public static final String LWJGL_OPENGL32 = LWJGL_OPENGL3;
```

riccardobl:
Let's say you want to use the new naming in your app, for lwjgl you will set the renderer like this

```
settings.setRenderer(LWJGL_OPENGL32);
```

let's say you have an external library that does something like this 

```
if(settings.getRenderer().equals(LWJGL_OPENGL3)){
  //enable something
}
```
(even though i don't think this is something commonly done in jme...)

you would have to update that aswell.

If we were to change the pr to do this 
```
LWJGL_OPENGL32 = LWJGL_OPENGL3
LWJGL_OPENGL40 = LWJGL_OPENGL4
```
It would be 100% backward compatible and we won't need the double case for lwjgl2 or both LWJGL_OPENGL32,LWJGL_OPENGL3 and LWJGL_OPENGL40,LWJGL_OPENGL4 in RENDER_CONFIGS for lwjgl3. So this might be a better solution? :thinking:

louhy:
Ah, you're talking about external stuff with multiple references, could be a pain. (In JME itself I didn't find much of that.) Yeah it's a bit cleaner in some ways. 

Still going to have
LWJGL_OPENGL30 = "LWJGL-OpenGL30";
plus
LWJGL_OPENGL3 = "LWJGL-OpenGL3";

Which seems a little wild: 3 means 3.2 and 30 means 3.0 
You sure?

louhy:
Ok maybe there isn't a perfect solution, see if that looks reasonable to you.

riccardobl:
Yeah, i think this is good enough. As you say, not perfect, but at least now the constant names make sense.
Let's leave this open for a while and see if others have opinions on the matter

PR #1100, opened 2019-05-28
Add support for gltf morph names.

tlf30:
OK, so I am putting a RFC (Request For Comments) here to get some feedback. 

glTF-Blender-IO just added support for shapekey names (names on mesh morphs) to the import/export in blender. (Info found here: https://github.com/KhronosGroup/glTF-Blender-IO/issues/362) 

Seeing as mesh morphs are becoming a more popular feature, I think it would be nice if we had a way to facilitate them. I have made some changes and added glTF support for mesh morph names. This is a rough work in progress, but I wanted some feedback on the core jme changes here to see what everyone thinks.

PS: Sorry in advance about the Geometry formatting changes, those will be reverted before the final revision of the PR.


oxplay2:
i might be wrong, but i dont see later usage of "targetNames " in:

JsonArray targetNames = meshObject.getAsJsonArray("targetNames");

stephengold:
What's holding up the revert of the format changes?

tlf30:
@stephengold I am home from work right now, but right now I am getting caught up on chores from being gone. 
I will get this PR finished up before I leave again. I leave on June 19th for another two weeks.

pspeed42:
@stephengold Note that this PR is in no way complete anyway.  The final version probably won't even have any changes to Geometry.java at all.  This was a "request for comment" and so my understanding was that it was only a partial start.  Probably would have been better done as a fork until ready to be a PR but whatever.

tlf30:
Yes, there are several things that will change here. I just wanted to know where to make the changes and if they were acceptable to make. The logic behind the named shape keys will be moved to the Mesh class, and I need to update this branch with the finished changes to the gltf plugin, and the formatting changes reverted.

tlf30:
@stephengold @pspeed42 
I think this is ready for review. 

Thanks,
Trevor Flynn

stephengold:
I looked this over, and I don't see any major issues, just minor ones with English grammar and formatting. @pspeed42, do you plan to review this PR?

pspeed42:
I would like to review it... as soon as I figure out what it's for. :)  I've never used morph names before so I have a learning curve.

jme3-core/src/main/java/com/jme3/scene/Mesh.java
```diff
@@ -1529,6 +1529,25 @@ public void addMorphTarget(MorphTarget target) {
     public MorphTarget[] getMorphTargets() {
         return morphTargets.getArray();
     }
+    
+        /**
+     * Get the name of all morphs in order.
+     * Morphs without names will be null
+     * @return an array
+     */
+    public String[] getMorphTargetNames() {
```
Ali-RS: Not sure why `getMorphTargetNames()` is needed? 
We can already access morph names using `getMorphTargets()`.

jme3-core/src/main/java/com/jme3/scene/Geometry.java
```diff
@@ -601,6 +601,39 @@ public void setMorphState(float[] state) {
...
+    /**
+     * Set the state of the morph with the given name.
+     * 
+     * If the name of the morph is not found, no state will be set.
+     * 
+     * @param morphTarget The name of the morph to set the state of
+     * @param state The state to set the morph to
+     */
+    public void setMorphState(String morphTarget, float state) {
+        if (mesh == null || mesh.getMorphTargets().length == 0) {
```
Ali-RS: You can use `!mesh.hasMorphTargets()` instead of `mesh.getMorphTargets().length == 0`

Ali-RS:
What about replacing this block 

```
        int index;
        boolean found = false;
        for (index = 0; index < nbMorphTargets.length; index++) {
            if (nbMorphTargets[index].getName().equals(morphTarget)) {
                found = true;
                break;
            }
        }
        if (found) {
            return morphState[index];
        }
```

with this? 
```

 int index = mesh.getMorphIndex(morphName); // need to add `getMorphIndex` into Mesh class
 if(index < 0) {
 return -1;
 } else {
 return morphState[index];
 }
```

tlf30:
@Ali-RS I do not think it would be a good idea to add a `getMorphIndex` method, as it is an implementation dependent method. In the future (possible random scenario) if we used a hashmap to store morphs then the method would no longer be relevant and need to be removed. (Just an example, not saying it will happen)

`hasMeshTargets` looks good. I'll throw that change in and fix the javadoc bugs.

tlf30:
@Ali-RS so after thinking it over I did go with your recommendation. I think in the future as we get more support for shape key names, it would be good to move away from storing morph shapes as an array and move to an hashmap, but for now, as the implementation stands, and probably will stand for some time, it is the best approach to just keep is as an array and make it as easy as possible to use in that form.  

I think all the final changes are done, thoughts?

jme3-core/src/main/java/com/jme3/scene/Geometry.java
```diff
@@ -610,25 +610,8 @@ public void setMorphState(float[] state) {
...
-        for (index = 0; index < nbMorphTargets.length; index++) {
-            String name = nbMorphTargets[index].getName();
-            if (name != null && name.equals(morphTarget)) {
-                found = true;
-                break;
-            }
-        }
-        if (found) {
+        int index = mesh.getMorphIndex(morphTarget);
+        if(index > 0) {
```
Ali-RS: I gues it should be `if(index >= 0)` 
also an space is needed at `if(`

jme3-core/src/main/java/com/jme3/scene/Geometry.java
```diff
@@ -669,24 +652,12 @@ public void setDirtyMorph(boolean dirtyMorph) {
...
-        boolean found = false;
-        for (index = 0; index < nbMorphTargets.length; index++) {
-            if (nbMorphTargets[index].getName().equals(morphTarget)) {
-                found = true;
-                break;
-            }
-        }
-        if (found) {
+        int index = mesh.getMorphIndex(morphTarget);
+        if(index < 0) {
```
Ali-RS: An space is needed at `if(`

jme3-core/src/main/java/com/jme3/scene/Mesh.java
```diff
@@ -1552,6 +1552,22 @@ public void addMorphTarget(MorphTarget target) {
...
+     * Get the index of the morph that has the given name.
+     * @param morphName The name of the morph to search for
+     * @return The index of the morph, or -1 if not found. 
+     */
+    public int getMorphIndex(String morphName) {
+        int index = -1;
+        MorphTarget[] nbMorphTargets = getMorphTargets();
+        for (int i = 0; i < nbMorphTargets.length; i++) {
+            if (nbMorphTargets[i].getName().equals(morphName)) {
+                index = i;
```
Ali-RS: Probably will be nice to add a `break;` after line 1566

tlf30:
@Ali-RS thanks for the catch on those. (I was working on it late last night 😪 )

Ali-RS:
Sorry to be picky, but do not you think method `Geometry.getMorphNames()` is extra?  It just does a direct call to `Mesh.getMorphTargetNames()`.

pspeed42:
Yeah, seems unnecessary to me.  I'm not sure I understand why Geometry needs to know about morph names since it's a mesh thing.

tlf30:
@Ali-RS @pspeed42
While I agree with you, I was trying to be consistent with the way morphs are implemented. The Geometry class has a similar function to get the morph values:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/1d5fe496e040cc377c97755b133cdec566aa1db5/jme3-core/src/main/java/com/jme3/scene/Geometry.java#L626

pspeed42:
Yes, the morph states are the states of the morphs... which is geometry-specific.  That's why the array returned is empty... it's for putting values into.  At least that's my understanding.

tlf30:
OK, `getMorphNames` is gone

Ali-RS:
Changes look good to me, I guess it's ready to be merged, let's leave it open for a few more days in case someone wants to review.

pspeed42:
I'm kind of clueless.  Can someone describe how this would be used in practice?  If it's already been explained then just pointing me to it would be fine.  Thanks.

tlf30:
@pspeed42 
When you have mesh morphs, (or otherwise known as shape keys) they get exported in an arbitrary order from whatever modelling software created them. Applications such as Blender allow the user to assign names to these morphs. When using the morph in JME3, previously it required going though each morph and figuring out which index correlated to which morph, then hard coding it into the application. Now we can use the name of the morph to determine which morph we are accessing. There is no more trial and error, or hard coding indexes of morphs.

An example would also be if someone added a new morph to a model, they would have to repeat the process of figuring out which morph is which in JME, but now they can just access them all by their names. 

I hope that clears it up for you

Ali-RS:
To add on what tlf30 said, morph state/weight can be modified in two ways. 
1- With MorphTrack animation
2- By manually setting the weight of morph (a common use case facial expression and lip-syncing)

I think the aim of this PR is to facilitate manual modification of morph weights.

pspeed42:
Thanks, guys.

Issue #1051, opened 2019-03-21
Null check in pr #875 changed the behavior for unattached audio nodes


riccardobl:
As reported by pspeed, pr https://github.com/jMonkeyEngine/jmonkeyengine/pull/875 added a nullcheck that changed the behavior of positional AudioNodes that aren't attached to any parent.

While this is not the most common way to use audio node, i think it would probably be better to restore the previous behavior, since it doesn't affect the topic of the pr.

PR #1109, opened 2019-06-03
Fix for 1051

karanehra:
After reading through issue #1051 and PR #875 i removed the nullcheck from AudioNode.java.
@riccardobl is there anything else that needs to be done to resolve this?


riccardobl:
I think that's all

Issue #1106, opened 2019-06-02
Blender import warning


rvandoosselaer:
When importing models from blender into JME, I constantly get this warning:

> The texture Diffuse Color has linear color space, but the material parameter DiffuseMap specifies no color space requirement, this may lead to unexpected behavior.
Check if the image was not set to another material parameter with a linear color space, or that you did not set the ColorSpace to Linear using texture.getImage.setColorSpace().

See multiple forum posts about this issue:
- https://hub.jmonkeyengine.org/t/sdk-import-error/38227

This is the check that logs the warning:
```java
if (paramDef.getColorSpace() == null && value.getName() != null && value.getImage().getColorSpace() == ColorSpace.Linear) {
            logger.log(Level.WARNING,
                    "The texture {0} has linear color space, but the material "
                            + "parameter {2} specifies no color space requirement, this may "
                            + "lead to unexpected behavior.\nCheck if the image "
                            + "was not set to another material parameter with a linear "
                            + "color space, or that you did not set the ColorSpace to "
                            + "Linear using texture.getImage.setColorSpace().",
                    new Object[]{value.getName(), value.getImage().getColorSpace().name(), name});
```
The warning is only logged when a linear colorspace is set on the image and the MatParam doesn't have a colorspace.

I looked a bit around the jme3-blender code and found I this:
in the file TextureBlenderAWT:146

```java
Image result = depth > 1 ? new Image(Format.RGBA8, width, height, depth, dataArray, ColorSpace.Linear) : new Image(Format.RGBA8, width, height, dataArray.get(0), ColorSpace.Linear);
```

The colorspace of the new created image is _always_ set to `ColorSpace.Linear` although the original colorspace of the image was `ColorSpace.sRGB`. When I adapt this line, and set the colorspace to the colorspace of the original image, the warning vanishes. The Image has now colorspace `ColorSpace.sRGB` and the MatParam still has colorspace `null`. But the warning is only logged when the colorspace of the images is Linear.

Looking further down the code, because the colorspace is not set on the MatParam, I noticed that in the `Material` class in `jme3-core` the colorspace is never set on the MatParam when a texture is specified:

com.jme3.material.Material:540:
```java
public void setTextureParam(String name, VarType type, Texture value) {
        if (value == null) {
            throw new IllegalArgumentException();
        }

        checkSetParam(type, name);
        MatParamTexture val = getTextureParam(name);
        if (val == null) {
[...]

PR #1110, opened 2019-06-03
fix for #1106

rvandoosselaer:
when the image object has a ColorSpace value we pass it on, else we leave it at the default value.

### DIFF
--- a/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/blending/TextureBlenderAWT.java
+++ b/jme3-blender/src/main/java/com/jme3/scene/plugins/blender/textures/blending/TextureBlenderAWT.java
@@ -143,7 +143,8 @@ public class TextureBlenderAWT extends AbstractTextureBlender {
             dataArray.add(newData);
         }
 
-        Image result = depth > 1 ? new Image(Format.RGBA8, width, height, depth, dataArray, ColorSpace.Linear) : new Image(Format.RGBA8, width, height, dataArray.get(0), ColorSpace.Linear);
+        ColorSpace colorSpace = baseImage.getColorSpace() != null ? baseImage.getColorSpace() : ColorSpace.Linear;
+        Image result = depth > 1 ? new Image(Format.RGBA8, width, height, depth, dataArray, colorSpace) : new Image(Format.RGBA8, width, height, dataArray.get(0), colorSpace);
         if (image.getMipMapSizes() != null) {
             result.setMipMapSizes(image.getMipMapSizes().clone());
         }
--- a/jme3-core/src/main/java/com/jme3/material/Material.java
+++ b/jme3-core/src/main/java/com/jme3/material/Material.java
@@ -537,9 +537,10 @@ public class Material implements CloneableSmartAsset, Cloneable, Savable {
         MatParamTexture val = getTextureParam(name);
         if (val == null) {
             checkTextureParamColorSpace(name, value);
-            paramValues.put(name, new MatParamTexture(type, name, value, null));
+            paramValues.put(name, new MatParamTexture(type, name, value, value.getImage() != null ? value.getImage().getColorSpace() : null));
         } else {
             val.setTextureValue(value);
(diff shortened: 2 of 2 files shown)


stephengold:
I had hoped for some review by a more senior monkey, but seem to know what you're doing.

rvandoosselaer:
I did a few tests with my own models and some free one’s I found online. I didn’t encountered any errors or strange behaviour, it seems to work.
However if people do find issues with it, I’ll be happy to revert the change or look for a solution.

riccardobl:
Ok, sorry i didn't see this (it might be a good idea for the future to use an actual title for PRs, because when you start having several "Fix *ISSUE NUMBER*" it's easy to miss some).

My question is, what will happen if i have a matdef that defines a -LINEAR param, but the image is sRGB? Will this pr break the linearization by forcing the mat param to become -LINEAR ?

rvandoosselaer:
> Ok, sorry i didn't see this (it might be a good idea for the future to use an actual title for PRs, because when you start having several "Fix _ISSUE NUMBER_" it's easy to miss some).
> 
> My question is, what will happen if i have a matdef that defines a -LINEAR param, but the image is sRGB? Will this pr break the linearization by forcing the mat param to become sRGB ?

Hi @riccardobl, I wasn't aware of the -LINEAR option in a material definition file.
I just tested it, and if you specify -LINEAR in the materialDef, the colorspace _will_ be set to linear.

This is my test case:
Gamma correction is enabled, the texture in the material is an sRGB one, the colorspace in the materialDef is set to '-LINEAR':
Screenshot 2019-06-26 at 21 00 46
I also added a log statement printing out the colorspace of the material that is set on the model:
From the log:
Colorspace: Linear

Same test as above, but without the -LINEAR flag:
Screenshot 2019-06-26 at 21 02 57
log:
Colorspace: sRGB

The first image is clearly gamma corrected (too bright) because of the colorspace, so all should still be working as expected.

riccardobl:
Can you share the code?

rvandoosselaer:
test:

```java
public class TestMaterialLinearColorspace extends SimpleApplication {

        new TestMaterialLinearColorspace().start();
    }

    @Override
    public void simpleInitApp() {
        AmbientLight ambientLight = new AmbientLight(ColorRGBA.White.mult(0.2f));
        DirectionalLight directionalLight = new DirectionalLight(new Vector3f(-0.2f, -1, -0.2f).normalizeLocal(), ColorRGBA.White);

        Spatial model = assetManager.loadModel("staff-0.j3o");
        Material material = assetManager.loadMaterial("colorspace-test.j3m");
        model.depthFirstTraversal(new SceneGraphVisitorAdapter() {
            @Override
                geom.setMaterial(material);
                System.out.println("Colorspace: " + material.getTextureParam("DiffuseMap").getColorSpace());
            }
        });

        rootNode.addLight(ambientLight);
        rootNode.addLight(directionalLight);
        rootNode.attachChild(model);
    }
}
```

material:
```
Material colorsapace-test : Common/MatDefs/Light/Lighting.j3md {
    MaterialParameters {
        Shininess: 0
        UseMaterialColors: true
        DiffuseMap: Flip texture.png
        Ambient: 1.0 1.0 1.0 1.0
        Diffuse: 1.0 1.0 1.0 1.0
        Specular: 0.0 0.0 0.0 1.0
    }
}
```

When running the testcase, I adapted this line in `Lighting.j3md`:
```
// Diffuse map
Texture2D DiffuseMap
```

to:

```
// Diffuse map
Texture2D DiffuseMap -LINEAR
```

If you want I can create a proper testcase for it, and put it in the jme3-examples module.

riccardobl:
Oh ok, i see.  I was mislead by the fact that both Materials and MatDefs keep a list of MatParamTexture. But the material will always look at the matdef list to see if it needs to switch the colorspace.

rvandoosselaer:
> Oh ok, i see. I was mislead by the fact that both Materials and MatDefs keep a list of MatParamTexture. But the material will always look at the matdef list to see if it needs to switch the colorspace.

Yep, indeed.

Issue #1107, opened 2019-06-03
Javadoc Warning Cleanup (OpenJDK 9+)


louhy:
Any interest in cleaning up Javadoc warnings? 

```
> Task :jme3-jbullet:javadoc
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:791: warning - invalid usage of tag >
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:807: warning - invalid usage of tag >
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:791: warning - invalid usage of tag >
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:807: warning - invalid usage of tag >
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:920: warning - invalid usage of tag &
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/objects/PhysicsVehicle.java:304: warning - invalid usage of tag &
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/objects/PhysicsVehicle.java:315: warning - invalid usage of tag &
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/objects/VehicleWheel.java:181: warning - invalid usage of tag &
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:791: warning - invalid usage of tag >
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:807: warning - invalid usage of tag >
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:807: warning - invalid usage of tag >
/home/lou/git/jmonkeyengine/jme3-jbullet/src/main/java/com/jme3/bullet/PhysicsSpace.java:791: warning - invalid usage of tag >
[...]

PR #1111, opened 2019-06-05
#1107-Fix Javadoc warnings in OpenJDK 9+

louhy:
Fixes for all subprojects excluding core & vr


stephengold:
Looks good. I'll integrate tomorrow---if no one else does so first.

PR #1115, opened 2019-06-07
Add support for wheelspin and damage on VehicleWheels

jayfella:
By giving the user the ability to add an external rotation to a vehicle wheel we can simulate wheel-spin and wheel damage.

https://i.ibb.co/wWsT8wX/image.png

https://youtu.be/tvrN8PDyqeQ

### DIFF
--- a/jme3-bullet/src/common/java/com/jme3/bullet/control/VehicleControl.java
+++ b/jme3-bullet/src/common/java/com/jme3/bullet/control/VehicleControl.java
@@ -199,6 +199,10 @@ public class VehicleControl extends PhysicsVehicle implements PhysicsControl, Jm
             // Copy the wheel spatial reference directly for now.  They'll
             // get fixed up in the cloneFields() method
             newWheel.setWheelSpatial(wheel.getWheelSpatial());
+
+            Quaternion extRotation = new Quaternion();
+            wheel.getExtRotation(extRotation);
+            newWheel.setExtRotation(extRotation);
         }
         control.setApplyPhysicsLocal(isApplyPhysicsLocal());
         control.setEnabled(isEnabled());
--- a/jme3-bullet/src/main/java/com/jme3/bullet/objects/VehicleWheel.java
+++ b/jme3-bullet/src/main/java/com/jme3/bullet/objects/VehicleWheel.java
@@ -133,6 +133,11 @@ public class VehicleWheel implements Savable {
     private boolean applyLocal = false;
 
     /**
+     * An external rotation, typically used to emulate wheelspin or damage.
+     */
+    private Quaternion extRotation = new Quaternion();
+
+    /**
      * No-argument constructor needed by SavableClassUtil. Do not invoke
      * directly!
      */
@@ -200,6 +205,18 @@ public class VehicleWheel implements Savable {
     private native void getWheelRotation(long vehicleId, int wheelId, Matrix3f location);
 
     /**
+     * Set an external rotation to the wheel. Typically used to emulate wheelspin or damage.
+     * @param quaternion
+     */
+    public void setExtRotation(Quaternion quaternion) {
+        extRotation.set(quaternion);
+    }
+
+    public void getExtRotation(Quaternion store) {
+        store.set(extRotation);
+    }
+
+    /**
      * Apply this wheel's physics location and orientation to its associated
      * spatial, if any.
      */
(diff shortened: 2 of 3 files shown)


riccardobl:
:thinking:  Is it wrong to say that this simulates loss of traction for the wheel?

jayfella:
That's exactly what it does - with the benefit of also being able to make the wheel wonky - as if it were broken or damaged.

riccardobl:
Wouldn't be possible to get the same effect by setting different frictionSlip for every wheel? Getting also the simulation of the traction loss for the vehicle as an added benefit?

jayfella:
Regardless of any current setting, the wheel will follow the movement of the vehicle. There is no current way of forcing the wheel geometry to rotate faster. The "data" behind the scenes can be manipulated, but it does not affect the wheel rotation.

riccardobl:
So it is not possible to get the actual rotation of the wheel from bullet?

jayfella:
Yes it's possible to get the rotation, but not set it. It returns a copy of it, and even if you use reflection to set it, it is calculated every frame, so it's ignored. The only way (so far as I can tell) is to add a rotation to the calculated rotation - which is what this PR does.

riccardobl:
I mean, if the wheel friction is different causing the wheel to spin faster, shouldn't be possible to get the current wheel rotation from bullet and so have the spatial match its speed?  (My premise might be totally wrong, i've never used vehicles in bullet. )

jayfella:
The wheel friction value of the wheel does not affect the rotation of the wheel. It just tells you how much it is slipping. No matter what you set it will always rotate in relation to its movement. It also doesn't know *why* it's slipping. It could be slipping down a hill, and so no additional rotation should be applied.

stephengold:
This seems like a useful feature. While we *could* provide a setter for the native `btWheelInfo.m_rotation`, applying an additional rotation in Java seems simpler.

As it stands, this is not ready to be integrated:
 + need to also implement the feature for `jme3-jbullet`
 + need to implement serialization and cloning for the new `extRotation` field

jayfella:
Done and done. Thank you for the review :)

stephengold:
Thanks for the quick action. I think you still need to clone `extRotation` in the `VehicleControl.jmeClone()` method.

jayfella:
Oh. I was looking in the wheel for the clone methods and couldn't find one. It's done.

stephengold:
I'm satisfied. I'll leave this open for a day or 2 in case others wish to review it.

riccardobl:
I'm still not entirely convinced by this.

Ali-RS:
I am not qualified to review this, but I agree that naming is confusing. 
Not sure, maybe better to separate it into two separate rotation, one for wheel spin rotation multiplier and one for wheel alignment or camber angle (https://en.wikipedia.org/wiki/Camber_angle)

jayfella:
There is no way currently to modify the rotation of the wheel. Which means there is no way to simulate wheelspin or "damage" of the wheel by altering the z-axis. There is no "faking" going on. When there is wheelspin the wheel rotates faster. When a wheel is "damaged" it is usually "bent" on the y and z axis.

Without providing this ability there is no way to alter the rotation of the wheel. It will always rotate as if the vehicle is rolling normally at full traction - even if the wheel is slipping.

I agree that wheelspin should be bound by physics, but it should be optional and dependant on the model of physics they want. The same for simulating damage. This should not be forced, but instead provide the bindings to do so. It would also require heavy extension of the vehicle model. If the wheel should wheelspin - how much? What tire is on the vehicle? Do I now make a tire object and force that model upon them? What material is the wheel on? Does that affect the wheelspin? This is far too much work that the user may not even want or will not be able to override if they don't want it to work the way I've implemented it.

I am implementing the pajecka "magic formula" - which is what led me to search for a way to do this. This PR has also been a requested feature on the forum over the years. The formula I am implementing should not be "forced" upon the user, but instead provide the bindings for the formula to be implemented. The pajecka formula is not the only "tire model" - and may be too much for mobile devices that want a simplified version or complete arcade style (such as crazy taxi) instead.

This PR simply provides the means to add wheel spin and simulate damage. The name of the method was chosen because that's exactly what it is - an external rotation. If you can think of another name I'd be more than happy to discuss it.

riccardobl:
Wouldn't be possible to do the same by using a Node for the wheel and a local rotation for the children?

jayfella:
The wheel geometry is controlled by the physics control, so any modification is immediately overridden by it.

riccardobl:
It updates the wheelSpatial, the wheelSpatial can be a node, and your actual wheel can be a child of said node, so you can rotate it however you want. Is this incorrect?

jayfella:
Sorry, I edited my previous post regarding parent-Child approaches.

riccardobl:
I mean the other way around, the parent moved by the physics control, the child moved by your code.

jayfella:
Hmm. Maybe. I’ll give it a try in the morning.

jayfella:
Actually your solution does work as expected. The only side effect is that I now have to calculate the delta rotation myself. I guess this PR is a non-issue.

Thank you for your review :)

stephengold:
Thank you for the follow-up @jayfella. If the added library complexity is not needed, let's avoid it.

Issue #1121, opened 2019-06-13
PQTorus Rendering Glitch


louhy:
PQTorus vanishes when you're not looking at it directly. Looks like world bounds aren't updated after the vertex buffer is set: 
https://vimeo.com/341919874

Suggest adding an updateBound() call at the end of updateGeometry() within PQTorus.java: 

```
        ...
        setBuffer(Type.Position, 3, fpb);
        setBuffer(Type.Normal,   3, fnb);
        setBuffer(Type.TexCoord, 2, ftb);
        setBuffer(Type.Index,    3, sib);
        
        updateBound();
    }
```

stephengold:
Good catch!

stephengold:
@louhy Your fix is integrated. Please verify that the glitch is gone from `master` branch.

louhy:
@stephengold Works here!

PR #1122, opened 2019-06-13
#1121-updateBound() call for PQTorus geometry updates

louhy:
Proposed fix for issue #1121. Since it's visual not sure if there's any good options for a test, but we'll have at least one which uses this shape when I update the pending physics GImpact shape test.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/shape/PQTorus.java
+++ b/jme3-core/src/main/java/com/jme3/scene/shape/PQTorus.java
@@ -210,6 +210,8 @@ public class PQTorus extends Mesh {
         setBuffer(Type.Normal,   3, fnb);
         setBuffer(Type.TexCoord, 2, ftb);
         setBuffer(Type.Index,    3, sib);
+        
+        updateBound();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


stephengold:
For testing, how about using it in a `Geometry`, reading the bounding volume, and comparing against the expected bounds?

louhy:
Went a bit overkill on this one... world bounds tests added for pretty much _all_ shapes (seemed weird to just do it for PQTorus). Test fails without fix, succeeds with fix.

stephengold:
Very nice. Thank you.

stephengold:
How about stricter tests of the bounding volumes? Either verify that all the vertex positions are within the bounding volume, or check the bounds against hand-calculated minima, or both.

louhy:
See how that one looks, I did make a few assumptions but I believe they're safe. Should be a more useful test like this I think. Put it together pretty quick, so double check that it makes sense. Reversing the vertex component order to 2,1,0 made it fail as I expected but I didn't check much else.

stephengold:
Looks good to me. Unless there's some objection in the next day or so, I'll integrate this PR.

PR #1132, opened 2019-06-26
Fix forced colorspace in matdefs working only once.

riccardobl:
When you have forced colorspace (-LINEAR) in the matdef and you set a texture param, it will work as expected, but if the value of the matparam is replaced with a different texture, the code that takes care of the colorspace switching will be skipped all together.

The unit test shows the issue in detail.


stephengold:
Thanks for documenting the issue. Do you have a fix for us?

PR #1136, opened 2019-07-04
jme-core: IndexBuffer fixes & API improvements

NetzwergX:
As discussed in:
https://hub.jmonkeyengine.org/t/indexbuffer-suggestions-improvements-consistency/42022/4

* [Fix] createIndexBuffer(int, int): return IndexByteBuffer as well for
	vertexCount < 128
* [Enhancement] Make put(int i, int value) fluid interface
* [Enhancement] Add relative put(int i) to allow easier chaining (parity
	with other buffer implementations)
* [Enhancement] Add getFormat() to allow setting an IndexBuffer to a
	Mesh directly without type inspection.
* [Fix] Fix WrappedIndexBuffer
	API changes make it possible that IndexByteBuffer is now a valid  type
for outBuf, leverage the new getFormat() method to set the buffer to the
mesh regardless of its type.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/mesh/IndexBuffer.java
+++ b/jme3-core/src/main/java/com/jme3/scene/mesh/IndexBuffer.java
@@ -31,11 +31,13 @@
  */
 package com.jme3.scene.mesh;
 
-import com.jme3.util.BufferUtils;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.IntBuffer;
 import java.nio.ShortBuffer;
+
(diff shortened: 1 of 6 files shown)


NetzwergX:
There are likely more places inside JME where IndexBuffer should be used. I remarked the above place in the forum thread, but I didn't want to clutter this PR also with updates of the whole codebase. If wanted I can include that change in this PR.

riccardobl:
I think it should be done. Mostly because with the current code,  mesh2.extractVertexData(meshWithByteIndex) will make a short index buffer instead of a byte one.

PR #1144, opened 2019-07-13
SSR Post Filter 

neph1:
SSR Filter based on: https://github.com/riccardobl/SimpleSSRShader/ 
Some liberties taken when restructuring it.

Consider skipping  testScene.j3o. Might just clutter an already cluttered testData project?

Some artifacts remain. Unsure which are due to SSR in general, the original shader and my implementation


riccardobl:
Thanks for taking the time to adapt this to jme filters.

I see that you have replaced some macros with uniforms, if there is no specific reason for doing that, i suggest to switch them back (using the Defines { } block in j3md) because they are generally faster (no memory access) and allow compile time optimization.

Also, i made this patch at some point, that fixed some artifacts that were caused by wrong depth linearization https://github.com/riccardobl/SimpleSSRShader/commit/f0e91a9f4453fcbf0944714ffe5f17b74c37a2e6 .

neph1:
I replaced the macros to allow as much runtime tweaking as possible (since the filter probably needs to be adapter to different use cases). But performance is everything, so I switched back (and added even more macros for the blur pass). If anyone wants to tweak in runtime they'll need to reinit the materials to see the changes).

jme3-effects/src/main/java/com/jme3/post/filters/SsrFilter.java
```diff
@@ -0,0 +1,305 @@
...
+        
+}
+    
+    @Override
+    protected void postQueue(RenderQueue queue) {
+        if(!approximateNormals) {
+            Renderer r = renderManager.getRenderer();
+            r.setFrameBuffer(normalPass.getRenderFrameBuffer());
+            renderManager.getRenderer().clearBuffers(true, true, true);
+            renderManager.setForcedTechnique("PreNormalPass");
```
riccardobl: Those normals are in view space, the shader wants world space normals.

jme3-effects/src/main/java/com/jme3/post/filters/SsrFilter.java
```diff
@@ -0,0 +1,305 @@
...
+
+            @Override
+            public boolean requiresSceneAsTexture() {
+                return true; //To change body of generated methods, choose Tools | Templates.
+            }
+            
+            
+        };
+
+        ssrPass.init(renderManager.getRenderer(), (int) (screenWidth / downSampleFactor), (int) (screenHeight / downSampleFactor), Image.Format.RGBA8, Image.Format.Depth, 1, ssrMaterial);
```
riccardobl: The format shouldn't be forced to RGBA8.

jme3-effects/src/main/java/com/jme3/post/filters/SsrFilter.java
neph1: I know. With the changes to the shader I can't see a difference between approximateNormals and the normalPass.

jme3-effects/src/main/java/com/jme3/post/filters/SsrFilter.java
neph1: How do you propose doing it? A setter for the Image.Format? Or is it just that RGBA8 isn't the supposed format?

neph1:
I noticed that the blur is meant to be run over several passes. I'll try to implement that as well

jme3-effects/src/main/java/com/jme3/post/filters/SsrFilter.java
riccardobl: Mhh.. yes a setter and maybe use RGBA16F as default, since the default format for filters is RGB111110F .

jme3-effects/src/main/java/com/jme3/post/filters/SsrFilter.java
riccardobl: Well, with this code, if you disable approximated normals, it will look very wrong. But the idea is that with the approximated normals you can only get an approximation of the surface normals, while with the normal pass you can get the real world space normals with the normal map applied. 
The shader can also reconstruct the normal Z and you can use the free Z channel of the normal map to store the glossiness (eg for PBR materials).

But the PreNormalPass doesn't do any of that, so that pass should be replaced with something else. Or we could decide to disable the normal pass entirely and allow only approximated normals, this could be a valid or maybe also preferable solution, since rerendering the scene only to extract the normals is not a good idea in any case, this should be done with MRT instead but here i think we are hitting the limitations of jme post processing.

What do you think?

neph1:
added new technique for rendering world normals + glossiness in pbrlighting

jme3-effects/src/main/resources/Common/MatDefs/SSR/normal_gloss.vert
```diff
@@ -0,0 +1,22 @@
...
+
+void main(void)
+{
+   texCoord=inTexCoord.xy;
+   vec4 modelSpacePos = vec4(inPosition, 1.0);
+   vec3 modelSpaceNormals = inNormal;
+   #ifdef NUM_BONES
+       Skinning_Compute(modelSpacePos,modelSpaceNormals);
+   #endif
+   normal = normalize((g_WorldMatrix * vec4(modelSpaceNormals, 1.0)).xyz);
```
riccardobl: In order to support instancing properly, this should use TransformWorldNormal from Instancing.glsllib

neph1:
The current pre pass only uses vertex normals. Would it make sense to read the normal map if available? Or make that a HQ option? Or is it simply not worth the cost since the roughness map will take care of the details?

riccardobl:
Yes it should, otherwise the result won't be much different from the approximated normals.

neph1:
All right, I think that's it? Thank's for reviewing this feature. The filter would have been crap otherwise.

riccardobl:
Do you have a test scene that uses the PBRLighting?

neph1:
Yeah. I removed the one i committed because a simple isnt really
representative. So i've been testing with a scene from work which is
complex.

neph1:
I noticed that approxNormals = false didn't work on the master branch, whereas it worked on my somewhat obsolete local branch. Some digging led to a difference in Instancing.glsllib
master:
```
vec3 TransformWorldNormal(vec3 normal) {
    return normalize(g_WorldNormalMatrix * normal);
}
```
mine:
```
vec3 TransformWorldNormal(vec3 normal) {
    return normalize((g_WorldMatrix * vec4(normal,0.0)).xyz);
}
```
the latter works just as expected, but the master version leads to the wrong perspective. PBRLighting and relevant shaders in effects are identical.

I've changed (not pushed) normal_gloss.vert to:
`wNormal = normalize(TransformWorld(vec4(modelSpaceNormals, 1.0)).xyz);`
rather than TransformWorldNormal. Convoluted, but works as expected now.

Thoughts?

I also noticed that the envmaps from a lightprobe doesn't really look that good together with SSR. The only thing I can think of is an option in the material to turn env maps off. I don't really like to propose changes to PBRLighting though

riccardobl:
Mh, actually the master version should be the correct one... i've also noticed some pretty bad artifacts when you move the camera away, i'll have to take a look at this more carefully and maybe move the raytracing to screenspace while i'm at it. I hope it is not a problem if i keep this unmerged for a while until i have the time to look into it.

neph1:
Any progress on this? Otherwise I suggest we merge it to avoid it rotting away. What's there works quite well on short to medium distances. Merging would allow anyone to refactor it in the future.

AlrikG:
@riccardobl What artifacts? Can you take a screenshot?

neph1:
FYI: I've pushed this to a separate repo: https://github.com/neph1/SsrFilter
Maybe that is a better way to move forward than pushing everything into master. Both from a maintainability and accessibility point of view.
Close?

PR #1145, opened 2019-07-14
Wrap potential IOExceptions during native file extraction

louhy:
When extracting native files (.dll), if an IOException of any kind is thrown, the Exception details were being lost. Instead this PR wraps the original Exception into the new (unchecked) one being created, so that a "Caused by" (IOException) is also reported.


empirephoenix:
Thanks for the fix, merged it

PR #1147, opened 2019-07-18
Android openGL ES 3 support

joliver82:
Added gles30 and newer functionality for android. All the information at the forum thread: https://hub.jmonkeyengine.org/t/opengl-es-3-0-and-newer-on-android/41913

Chage list:

- Initialize the proper opengl context (2.0, 3.0, 3.1…)
- Added Caps for all of the new supported features
- Added Instancing as core feature of GLES30 and newer
- Properly adding versioning to shaders (#version xxx es) to shaders
- Added GLSL320,GLSL310 and/or GLSL300 to most materials depending on the required level for them to work properly
- Added depth24 texture usage if hardware supports it
- Enabled Geometry shaders and tessellation (3.2 core feature and 3.1 as extension) and fixed compilation errors on the examples
- Enabled glBlitFramebuffer
- Added GLSL300 to filters not able to work with GLSL100.
- Created Glsl300ShaderGenerator to support GLSL300 in shadernodes
- Fixed most shader compilation issues (strict type checking, adding OpenGL ES specific extensions, function usage depending on version and some other)
- Set default precision to highp solving most filter issues although maybe a little slower. Added specific sampler precision definition for gles300 or newer
- Make use of GL_TEXTURE_MAX_LEVEL in GLES300 to avoid black rendering of non complete mipmapped textures
- Changed some assets from compressed formats not supported in android to RGB/RGBA depending on the needs so all effects and tests can run also on android
- updated precompiled android.jar to have access to newest android API
- enabled some other core features of GLES300, GLES310 and GLES320
- Fixed build issues in jme3-android-examples
- Added 3D texture and texture array support
- Added float texture formats, fixed rgb8 format if having cap rgb8 and fixed glTexImage2D call on AndroidGL
- PBRLighting works and any other effect requiring HDR texture formats should work also
- Effects and postprocessing
- Multisampling

Known issues/pending work:
- Fixed most issues with shadow rendering in android (GLES 2.0 and GLES 3.0 or better) although some glitches still apear randomly depending on ShadowMode of the nodes and the device
- Fixed water filter. Still some rendering glitches in my old redmi3 (gles3.0). The water renders properly but the rest of the scene is overwritten in red. Not reproduced in any other device I've tried


jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Environment/envMapping100.frag
```diff
@@ -4,5 +4,9 @@ void main(){
     //@input samplerCube cubeMap the cube map
     //@output vec4 color the output color
 
-    color = textureCubeLod(cubeMap, refVec, 0.0);
-}
\ No newline at end of file
+    #ifdef GL_ES
+        color = textureCube(cubeMap, refVec);
```
riccardobl: What is the reason for this?

jme3-core/src/main/resources/Common/ShaderLib/GLSLCompat.glsllib
```diff
@@ -19,18 +19,27 @@
...
 #    define varying out
 #    define attribute in
 #  elif defined FRAGMENT_SHADER
 #    define varying in
 #    define gl_FragColor outFragColor
 #  endif
-#endif
\ No newline at end of file
+#else
+#  define isnan(val) !(val<0.0||val>0.0||val==0.0)
```
riccardobl: Is this needed anywhere in this pr? If not, it's probably safer to keep it out.

jme3-examples/src/main/java/jme3test/texture/TestTexture3D.java
```diff
@@ -116,16 +116,17 @@ public void simpleInitApp() {
...
                 bb.put((byte) (255f*i/10f));
                 bb.put((byte) (255f*i/10f));
                 bb.put((byte) (255f));
+                bb.put((byte) (255f));
             }
         }
         bb.rewind();
         data.add(bb);
-        return new Texture3D(new Image(Format.RGB8, 10, 10, 10, data, null, ColorSpace.Linear));
+        return new Texture3D(new Image(Format.RGBA8, 10, 10, 10, data, null, ColorSpace.Linear));
```
riccardobl: Why?

jme3-examples/src/main/java/jme3test/texture/TestTextureArray.java
```diff
@@ -31,8 +31,8 @@ public void simpleInitApp()
        }
        
        
-       Texture tex1 = assetManager.loadTexture( "Textures/Terrain/Pond/Pond.jpg");
-       Texture tex2 = assetManager.loadTexture("Textures/Terrain/Rock2/rock.jpg");
+       Texture tex1 = assetManager.loadTexture( "Textures/Terrain/Pond/Pond.png");
+       Texture tex2 = assetManager.loadTexture("Textures/Terrain/Rock2/rock.png");
```
riccardobl: Why?

jme3-examples/src/main/java/jme3test/texture/TestTextureArrayCompressed.java
```diff
@@ -31,8 +31,8 @@ public void simpleInitApp()
        }
        
        
-       Texture tex1 = assetManager.loadTexture( "Textures/Terrain/Pond/Pond.dds");
-       Texture tex2 = assetManager.loadTexture("Textures/Terrain/BrickWall/BrickWall.dds");
+       Texture tex1 = assetManager.loadTexture( "Textures/Terrain/Pond/Pond_dxt5.dds");
+       Texture tex2 = assetManager.loadTexture("Textures/Terrain/BrickWall/BrickWall_dxt5.dds");
```
riccardobl: Why?

riccardobl:
Also, would it be possible to make the build script download lib/android.jar from an official source?
It would be preferable than uploading a mistery jar to the repo.

jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Environment/envMapping100.frag
joliver82: If using textureCubeLod the shader failed to compile because as stated in gles sl 100 specs "The built-ins suffixed with “Lod” are allowed only in a vertex shader" (page 71)

jme3-core/src/main/resources/Common/ShaderLib/GLSLCompat.glsllib
joliver82: isnan is used only in the water fragment shader. This is the best way I found to simulate isnan in glsl not having this function

jme3-examples/src/main/java/jme3test/texture/TestTexture3D.java
joliver82: 3D textures not having alpha channel weren't rendered at all

jme3-examples/src/main/java/jme3test/texture/TestTextureArray.java
joliver82: texture arays not having alpha channel weren't rendered at all, so I added png version of the textures including alpha to be able to run the test

jme3-examples/src/main/java/jme3test/texture/TestTextureArrayCompressed.java
joliver82: I created non-compressed dds files for most existing dds and renamed the previous to file_format.dds so if running any regular test non-compressed image is used being able to run in desktop and mobiles while this test which is related to compressed images I needed to change it to keep using the compressed image

joliver82:
> Also, would it be possible to make the build script download lib/android.jar from an official source?
> It would be preferable than uploading a mistery jar to the repo.

I agree but a previous file of this one was included this way before and honestly I don't know where I could get this jar directly from the net. Android.jar is extracted to platform path of android sdk when installing it. Maybe adding a copy action to the gradle scripts as the sdk dir is set in the local.properties file?

jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Environment/envMapping100.frag
riccardobl: Are we talking about GL ES 2.0 ? Because that's the minimum supported version in jme.
If that's the case, this extension should enable textureCubeLod https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_shader_texture_lod.txt

jme3-examples/src/main/java/jme3test/texture/TestTexture3D.java
riccardobl: Do you know why? We might need to investigate that...

jme3-examples/src/main/java/jme3test/texture/TestTextureArrayCompressed.java
riccardobl: Ok, can we keep only the non compressed version for every texture except the ones used in tests that require compressed textures?

jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Environment/envMapping100.frag
joliver82: Yes, I had a look at that extension but if the device don't support it, it'll fail, so I consider it safer to avoid using lod in glsl100

jme3-examples/src/main/java/jme3test/texture/TestTextureArrayCompressed.java
joliver82: Yes, compressed dds's not being used in tests could be removed. These two textures are the only compressed textures used

jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Environment/envMapping100.frag
riccardobl: The problem is that some env maps might not render properly because mipmaps levels could be used to store different informations, see for example pbr env maps.

jme3-examples/src/main/java/jme3test/texture/TestTexture3D.java
joliver82: At first I didn't know, but I've just checked RGB8 and RGBA8 and they both work properly now. Could be related with latest changes I did to GLImageFormats to correct it on gles platforms. If you want, this can be reverted to RGB

riccardobl: Yes, if possible revert also the changes in TestTextureArray.java, please.

jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Environment/envMapping100.frag
riccardobl: Yes, i think this is the best solution.

jme3-examples/src/main/java/jme3test/texture/TestTexture3D.java
joliver82: Just comited

jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Environment/envMapping100.frag
joliver82: Just comited

riccardobl:
> > Also, would it be possible to make the build script download lib/android.jar from an official source?
> > It would be preferable than uploading a mistery jar to the repo.
> 
> I agree but a previous file of this one was included this way before and honestly I don't know where I could get this jar directly from the net. Android.jar is extracted to platform path of android sdk when installing it. Maybe adding a copy action to the gradle scripts as the sdk dir is set in the local.properties file?

If this is a dependency, gradle can just load it from the sdk folder. I will open an issue for that, but for now i just want to verify the integrity of the android.jar provided in this pr.

This is the sha256 hash of android.jar from the last android studio release
31cca6a9203b04349e88ec65cc1a2b915058b699b28ab9259b98d30d5c3b26ac  android.jar

and this is the hash of the jar provided in this pr
d20853d289a5161d9cf1c1268d7923076487d4fa68c970b0733e2d5a0ccfc3c5  android.jar

Can you point me to the release from which this jar comes from?

riccardobl:
Ok, i was testing the wrong file.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -1379,7 +1423,17 @@ public void updateShaderSourceData(ShaderSource source) {
...
+            if(version>=310) {
+                stringBuf.insert(idx + 1, "precision highp sampler2DMS;\n");
+            }
+            if(version>=300) {
+                stringBuf.insert(idx + 1, "precision highp sampler2DArray;\n");
+                stringBuf.insert(idx + 1, "precision highp sampler2DShadow;\n");
+                stringBuf.insert(idx + 1, "precision highp sampler3D;\n");
+                stringBuf.insert(idx + 1, "precision highp sampler2D;\n");
+            }
+            stringBuf.insert(idx + 1, "precision highp float;\n");
```
riccardobl: Is it a good idea to force this on the developer?
I know the code already forced precision mediump float; , but this looks like bad practice to me.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -1853,23 +1908,39 @@ public void setReadDrawBuffers(FrameBuffer fb) {
         if (fb == null) {
             // Set Read/Draw buffers to initial value.
             if (context.boundDrawBuf != INITIAL) {
-                gl2.glDrawBuffer(context.initialDrawBuf);
+                if(gl2!=null) {
+                    gl2.glDrawBuffer(context.initialDrawBuf);
+                } else {
+                    gles30.glDrawBuffer(context.initialDrawBuf);
+                }
```
riccardobl: Would it be a bad idea to remove duplicate code like this by making GLES30 extend GL2 ? It's mostly a duplicate of that interface anyway, minus two methods... that can be overriden in the implementation to return unsupported operation exception.

joliver82: Yes, nowadays it's duplicate of GL2 plus some minor modifications although my idea is, for the long term, add everything from GLES30 and create more interfaces (GLES31, GLES32 and later when new versions are created) according to jme3 needs. 

Also there's an issue here which is the reason why I did it this way. There's plenty of code checking gl being an instance of GL2, and specially the loadCapabilities method at GLRenderer class:

`
....
        this.gl2 = gl instanceof GL2 ? (GL2)gl : null;
....
    private void loadCapabilities() {
        if (gl2 != null) {
            loadCapabilitiesGL2();
        } else {
            loadCapabilitiesES();
        }
        loadCapabilitiesCommon();
    }
`

So using GL2 interface would use GL2 caps not ES caps. Maybe we could look into a better way of checking if it's desktop gl2 or gles3 instead but may require lots of changes

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
joliver82: So you mean removing all these inserts and making the game/shader developer? This will lead to plenty of current jme3 shaders not to compile on gles at all.

riccardobl: Could it be implemented in GLSLCompat? So that devs can leave it out for custom shaders.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
riccardobl: > Yes, nowadays it's duplicate of GL2 plus some minor modifications although my idea is, for the long term, add everything from GLES30 and create more interfaces (GLES31, GLES32 and later when new versions are created) according to jme3 needs.

I don't mean to remove GLES30, but to make it extend GL2. GLES30 will add the missing features and GLES31 will extend GLES30

> Also there's an issue here which is the reason why I did it this way. There's plenty of code checking gl being an instance of GL2, and specially the loadCapabilities method at GLRenderer class:

Yes, this can be an issue, but using duplicate code every time a gl2 api is called is also pretty bad. :thinking:

joliver82: So, if I understood you correctly. GLES30 extends GL2 so gl2 == | != null checks are enough for newer GL functions and the duplicated code could be removed safely. 

In loadCapabilities, we would check (gl2 != null && ! gl2 instanceof GLES30) -> loadGL2 else -> loadES, safely checking desktop vs mobile

For gles missing functions (glAlphaFunc, glPointSize and glPolygonMode) we could throw an exception and in the glrenderer check gl instance of GLES30 or just leave empty functions. Those functions are only used in GLDebugDesktop and glPolygonMode in GLRenderer once

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
joliver82: Not really, precision directive must be after last #extension line. All shaders include GLSLCompat, then other requirements (that could have extension directives) and then the shader code, so having precision defined in GLSLCompat will make compilation fail

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
riccardobl: yes

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
riccardobl: Isn't the shader loader supposed to move the #extension directives to the top?

joliver82:
> Is jme3-effects/src/main/resources/Common/MatDefs/Water/Textures/water_normalmap_dxt1.dds used anywhere?

Not at all, I didn't realize that file when removing the others :man_facepalming: just removed and comited

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
riccardobl: > Although this may be clearer and avoid duplicated code, is not strict regarding gl releases. And for the long time could generate other issues. For example, a function that's being used in jme that is defined in the specs of opengl 4.1 and in opengl es 3.2 GLES32 should extend GLES31 (clearer solution but having duplicated codelike in here) or GL4 (avoiding duplicated code but weird stuff because it's just one function defined in gles32)?

We will deal with future problems in the future. But right now now we have several of this

```
if(){
   gl.function(a,b,c)
}else{
   gles.sameFunction(a,b,c);
}
```

that are pretty bad to maintain. 
But I'm open to other solutions that can fix that.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
joliver82: Yes, you're right and should work. I'll change and test it and get back to you

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
joliver82: Maybe a cleaner solution would be making AndroidGL implement GL2 and GLES30. GLES30 would just have aditional gl functions specific for gles30 instead of extending GL2. This would avoid duplicated code, allow checking instanceof GLES30 and avoid the future problems :thinking:

joliver82: I've changed the code to match my last comment which solves the duplicated code

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
joliver82: I've tried changed it and some shaders included by default in jme3 like Common/MatDefs/Terrain/TerrainLighting.frag don't include GLSLCompat.glsllib. We could either leave it as it's now or review all shaders to include GLSLCompat as you think is a better option :thinking:

riccardobl: Yeah, not a problem, we can add the GLSLCompat with a future PR,

stephengold:
I'd like to include this in the 3.3.0-alpha3 release. How soon can it be finished?

joliver82:
> I'd like to include this in the 3.3.0-alpha3 release. How soon can it be finished?
From my side it's finished. @riccardobl changed it to Testing but I don't know who and when should test it before merging

stephengold:
Have the `precision` instructions been moved to `GLSLCompat`?

joliver82:
No because plenty of shaders would stop working as they don't include GLSLCompat. So if I understood Riccardo correctly, we'll leave this for the future

riccardobl:
I meant to add the precision to GLSLCompat now and include GLSLCompat on all the shaders in a future pr, but if you think we should delay that and leave the pr as it is, then it's fine either way.

Since i'm not an android developer i'm waiting for the approvation of someone else who can validate and maybe test this pr on an existing application to see if we missed something.

kaloyanDEV:
Testing on existing application using https://github.com/joliver82/jmonkeyengine and android-gl3 branch
```
def jmerepo = 'org.jmonkeyengine'
def jmeversion = '3.3.0-alpha2'
android.defaultConfig.javaCompileOptions.annotationProcessorOptions.includeCompileClasspath = true
dependencies {
    compile(project(":game")) {
        exclude module: "jme3-lwjgl"
        exclude module: "jme3-desktop"
        exclude group: 'org.codehaus.groovy', module: 'groovy-all'
        exclude module: 'log4j-slf4j-impl'
        exclude module: "jme3-core"
        exclude module: "jme3-effects"
    }
    compile 'org.apache.logging.log4j:log4j-core:2.5'

//    compile jmerepo + ':jme3-android:' + jmeversion

    compile files('/home/kaloyan/Projects/jmonkeyengine/jme3-android/build/libs/jme3-android-3.3.0-android-gl3-SNAPSHOT.jar')
    compile files('/home/kaloyan/Projects/jmonkeyengine/jme3-core/build/libs/jme3-core-3.3.0-android-gl3-SNAPSHOT.jar')
    compile files('/home/kaloyan/Projects/jmonkeyengine/jme3-effects/build/libs/jme3-effects-3.3.0-android-gl3-SNAPSHOT.jar')




    compile(jmerepo + ':jme3-android-native:' + jmeversion) {
        exclude module: "jme3-android"
        exclude module: "jme3-core"
    }

    implementation 'com.google.guava:guava:27.1-android'
}

```

 public class MainActivity extends Activity {

    private AndroidHarnessFragment jmeFragment;
    //....
    public static class JmeFragment extends AndroidHarnessFragment {
        public JmeFragment() {
            // Set main project class (fully qualified path)
            appClass = Main.class.getName();

            // Set the desired EGL configuration
            eglBitsPerPixel = 24;
            eglAlphaBits = 0;
            eglDepthBits = 16;
            eglSamples = 0;
            eglStencilBits = 0;

            // Set the maximum framerate
            // (default = -1 for unlimited)
            frameRate = 60;

            // Set the maximum resolution dimension
            // (the smaller side, height or width, is set automatically
            // to maintain the original device screen aspect ratio)
            // (default = -1 to match device screen resolution)
            maxResolutionDimension = -1;

            // Set input configuration settings
            joystickEventsEnabled = false;
            keyEventsEnabled = true;
            mouseEventsEnabled = true;

            // Set application exit settings
            finishOnAppStop = true;
            handleExitHook = true;
            exitDialogTitle = "Do you want to exit?";
            exitDialogMessage = "Use your home key to bring this app into the background or exit to terminate it.";

            // Set splash screen resource id, if used
            // (default = 0, no splash screen)
            // For example, if the image file name is "splash"...
            //     splashPicID = R.drawable.splash;
            //splashPicID = 0;
            splashPicID = R.drawable.splash;
        }
    }

    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        //--------- JME Stuff ---------

        // Set window fullscreen and remove title bar
        requestWindowFeature(Window.FEATURE_NO_TITLE);
        getWindow().setFlags(
                WindowManager.LayoutParams.FLAG_FULLSCREEN,
                WindowManager.LayoutParams.FLAG_FULLSCREEN);
        setContentView(R.layout.main);

        // find the fragment
        FragmentManager fm = getFragmentManager();
        jmeFragment = (AndroidHarnessFragment) fm.findFragmentById(R.id.jmeFragment);
    }
    //...
}

```

```
 //...
    public void start() {
        super.start(); //To change body of generated methods, choose Tools | Templates.
        this.getContext().getSettings().setRenderer(AppSettings.LWJGL_OPENGL3);
    }
}

```
Using test data from same branch (sun is just  DirectionalLight)

```
public class TestPBR extends BaseAppState {

    private Node modelNode;
    private Geometry model;
    private Material pbrMat;

    private float roughness = 0.0f;
    private float frame;
    private AssetManager assets;

    private Node tex;

    private Container buttons;

    @Override
    protected void initialize(Application app) {

        assets = app.getAssetManager();

        modelNode = (Node) new Node("modelNode");
        modelNode.getLocalTranslation().y -= 2f;
        model = (Geometry) assets.loadModel("Models/Tank/tank.j3o");
        model.setLocalScale(0.5f);
        MikktspaceTangentGenerator.generate(model);
        modelNode.attachChild(model);
        pbrMat = assets.loadMaterial("Models/Tank/tank.j3m");
        pbrMat.setFloat("Roughness", roughness);
        model.setMaterial(pbrMat);

        final EnvironmentCamera envCam = new EnvironmentCamera(256, new Vector3f(0, 3f, 0));
        getStateManager().attach(envCam);

        getApplication().getInputManager().addListener(new ActionListener() {
            @Override

                if (name.equals("rup") && isPressed) {
                    roughness = FastMath.clamp(roughness + 0.1f, 0.0f, 1.0f);
                    pbrMat.setFloat("Roughness", roughness);
                }
                if (name.equals("rdown") && isPressed) {
                    roughness = FastMath.clamp(roughness - 0.1f, 0.0f, 1.0f);
                    pbrMat.setFloat("Roughness", roughness);
                }

                if (name.equals("light") && isPressed) {
                    getState(LightState.class).getSun().setDirection(getApplication().getCamera().getDirection().normalize());
                }
            }
        }, "light", "rup", "rdown");

        getApplication().getInputManager().addMapping("light", new KeyTrigger(KeyInput.KEY_J));
        getApplication().getInputManager().addMapping("rup", new KeyTrigger(KeyInput.KEY_F));
        getApplication().getInputManager().addMapping("rdown", new KeyTrigger(KeyInput.KEY_G));

        getState(LightState.class).getSun().setDirection(new Vector3f(0.8999188f, -0.34202015f, -0.2704965f));

        //buttons
        float width = app.getContext().getSettings().getWidth();
        float height = app.getContext().getSettings().getHeight();

        buttons = new Container();
        buttons.setLocalScale(1);
        buttons.setPreferredSize(new Vector3f(width * 0.24f, height * .3f, 0));

        ActionButton button = buttons.addChild(new ActionButton(new CallMethodAction("Roughness UP", this, "rup")));
        button.setTextVAlignment(VAlignment.Center);
        button.setFontSize(height * 0.03f);
        button = buttons.addChild(new ActionButton(new CallMethodAction("Roughness DOWN", this, "rdown")));
        button.setTextVAlignment(VAlignment.Center);
        button.setFontSize(height * 0.03f);
        button = buttons.addChild(new ActionButton(new CallMethodAction("LIGHT", this, "light")));
        button.setTextVAlignment(VAlignment.Center);
        button.setFontSize(height * 0.03f);

        buttons.setLocalTranslation(0, height * 0.625f, 0);
    }

    public void rup() {
        roughness = FastMath.clamp(roughness + 0.1f, 0.0f, 1.0f);
        pbrMat.setFloat("Roughness", roughness);
    }

    public void rdown() {
        roughness = FastMath.clamp(roughness - 0.1f, 0.0f, 1.0f);
        pbrMat.setFloat("Roughness", roughness);
    }

    public void light() {
        getState(LightState.class).getSun().setDirection(getApplication().getCamera().getDirection().normalize());
    }


    protected void cleanup(Application aplctn) {
    }

    protected void onEnable() {
        Node gui = ((Main) getApplication()).getGuiNode();
        gui.attachChild(buttons);
        GuiGlobals.getInstance().requestFocus(buttons);
    }

    protected void onDisable() {
    }

    public void update(float tpf) {
        frame++;

        if (frame == 2) {
            modelNode.removeFromParent();
            final LightProbe probe = LightProbeFactory.makeProbe(
                    getApplication().getStateManager().getState(EnvironmentCamera.class),
                    ((Main) getApplication()).getRootNode(),
                    new JobProgressAdapter<LightProbe>() {

                @Override
                public void done(LightProbe result) {
                    System.err.println("Done rendering env maps");
                    tex = EnvMapUtils.getCubeMapCrossDebugViewWithMipMaps(result.getPrefilteredEnvMap(), assets);
                }
            });
            ((SphereProbeArea) probe.getArea()).setRadius(100);
            ((Main) getApplication()).getRootNode().addLight(probe);
            //getStateManager().getState(EnvironmentManager.class).addEnvProbe(probe);

        }
        if (frame > 10 && modelNode.getParent() == null) {
            ((Main) getApplication()).getRootNode().attachChild(modelNode);
        }
    }

}
```

On Xiaomi Redmi Note 3 (android 6) reflection from LightProbe is rendered (it is taking long)

**Comment by kaloyanDEV** [comment]
Problem appears while trying to use running heightmap on 3.3.0-alpha2 

```
java.lang.NoSuchMethodError: No virtual method put(II)V in class Lcom/jme3/scene/mesh/IndexBuffer; or its super classes (declaration of 'com.jme3.scene.mesh.IndexBuffer' appears in /data/app/com.tech.gladiators-1/base.apk:classes5.dex)
        at com.jme3.terrain.geomipmap.LODGeomap$VerboseBuffer.put(LODGeomap.java:931)
        at com.jme3.terrain.geomipmap.LODGeomap.writeIndexArrayLodDiff(LODGeomap.java:175)
        at com.jme3.terrain.geomipmap.LODGeomap.createMesh(LODGeomap.java:85)
        at com.jme3.terrain.geomipmap.LODGeomap.createMesh(LODGeomap.java:77)
        at com.jme3.terrain.geomipmap.TerrainPatch.<init>(TerrainPatch.java:190)
        at com.jme3.terrain.geomipmap.TerrainQuad.createQuadPatch(TerrainQuad.java:722)
        at com.jme3.terrain.geomipmap.TerrainQuad.split(TerrainQuad.java:567)
        at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:223)
        at com.jme3.terrain.geomipmap.TerrainQuad.createQuad(TerrainQuad.java:611)
        at com.jme3.terrain.geomipmap.TerrainQuad.split(TerrainQuad.java:569)
        at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:223)
        at com.jme3.terrain.geomipmap.TerrainQuad.createQuad(TerrainQuad.java:611)
        at com.jme3.terrain.geomipmap.TerrainQuad.split(TerrainQuad.java:569)
        at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:223)
        at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:176)
        at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:157)
```

**Comment by joliver82** [comment]
> Problem appears while trying to use running heightmap on 3.3.0-alpha2
> 
> ```
> java.lang.NoSuchMethodError: No virtual method put(II)V in class Lcom/jme3/scene/mesh/IndexBuffer; or its super classes (declaration of 'com.jme3.scene.mesh.IndexBuffer' appears in /data/app/com.tech.gladiators-1/base.apk:classes5.dex)
>         at com.jme3.terrain.geomipmap.LODGeomap$VerboseBuffer.put(LODGeomap.java:931)
>         at com.jme3.terrain.geomipmap.LODGeomap.writeIndexArrayLodDiff(LODGeomap.java:175)
>         at com.jme3.terrain.geomipmap.LODGeomap.createMesh(LODGeomap.java:85)
>         at com.jme3.terrain.geomipmap.LODGeomap.createMesh(LODGeomap.java:77)
>         at com.jme3.terrain.geomipmap.TerrainPatch.<init>(TerrainPatch.java:190)
>         at com.jme3.terrain.geomipmap.TerrainQuad.createQuadPatch(TerrainQuad.java:722)
>         at com.jme3.terrain.geomipmap.TerrainQuad.split(TerrainQuad.java:567)
>         at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:223)
>         at com.jme3.terrain.geomipmap.TerrainQuad.createQuad(TerrainQuad.java:611)
>         at com.jme3.terrain.geomipmap.TerrainQuad.split(TerrainQuad.java:569)
>         at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:223)
>         at com.jme3.terrain.geomipmap.TerrainQuad.createQuad(TerrainQuad.java:611)
>         at com.jme3.terrain.geomipmap.TerrainQuad.split(TerrainQuad.java:569)
>         at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:223)
>         at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:176)
>         at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:157)
> ```

@kaloyanDEV  Did it work previously in official jme3? Did you run a sample from jme3-examples or any personal code? Could you send me a simple testcase if the later case?

**Comment by kaloyanDEV** [comment]
@joliver82 
```
    AssetManager assetManager = getApplication().getAssetManager();

        /**
         * 1. Create terrain material and load four textures into it.
         */
        Material mat_terrain = new Material(assetManager,
                "Common/MatDefs/Terrain/Terrain.j3md");

        /**
         * 1.1) Add ALPHA map (for red-blue-green coded splat textures)
         */
        mat_terrain.setTexture("Alpha", assetManager.loadTexture(
                "Textures/Terrain/alphamap.png"));

        /**
         * 1.2) Add ROCKS texture into the red layer (Tex1).
         */
        Texture grass = assetManager.loadTexture(
                "Textures/Terrain/rock.png");
        grass.setWrap(Texture.WrapMode.Repeat);
        mat_terrain.setTexture("Tex1", grass);
        mat_terrain.setFloat("Tex1Scale", 24f);

        /**
         * 1.3) Add GRASS texture into the green layer (Tex2)
         */
        Texture dirt = assetManager.loadTexture(
                "Textures/Terrain/grass.png");
        dirt.setWrap(Texture.WrapMode.Repeat);
        mat_terrain.setTexture("Tex2", dirt);
        mat_terrain.setFloat("Tex2Scale", 32f);

        /**
         * 1.4) Add SOMETHING texture into the blue layer (Tex3)
         */
        Texture rock = assetManager.loadTexture("Textures/Terrain/black_patch.png");
        rock.setWrap(Texture.WrapMode.Repeat);
        mat_terrain.setTexture("Tex3", rock);
        mat_terrain.setFloat("Tex3Scale", 28f);
        /**
         * 2. Create the height map
         */
        AbstractHeightMap heightmap = null;
        Texture heightMapImage = assetManager.loadTexture(
                "Textures/Terrain/untitled.png");
        heightmap = new ImageBasedHeightMap(heightMapImage.getImage());
        heightmap.load();

        /**
         * 3. We have prepared material and heightmap. Now we create the actual
         * terrain: 3.1) Create a TerrainQuad and name it "my terrain". 3.2) A
         * good value for terrain tiles is 64x64 -- so we supply 64+1=65. 3.3)
         * We prepared a heightmap of size 512x512 -- so we supply 512+1=513.
         * 3.4) As LOD step scale we supply Vector3f(1,1,1). 3.5) We supply the
         * prepared heightmap itself.
         */
        TerrainQuad terrain = new TerrainQuad("my terrain", 65, 513, heightmap.getHeightMap());

        /**
         * 4. We give the terrain its material, position & scale it, and attach
         * it.
         */
        terrain.setMaterial(mat_terrain);
        terrain.setLocalScale(.15f, .1f, .2f);
```

if you can no reproduce it will send you the assets also

riccardobl:
What is the status for this?

Ali-RS:
The issue that @kaloyanDEV  mentioned seems related to terrain stuff, not sure what it has to do with this PR :roll_eyes:

kaloyanDEV:
yes Terrain.j3md is not part of changes but might be new post in hub if someone try using it. I run all tests from JME3 Tests Android apk successfully from this joliver repo.

joliver82:
Hi, yes, there's nothing explicitly related to terraing in this PR. From my point of view, if @kaloyanDEV  tested every other tests and worked, we could merge and leave this new issue for a future PR. I'll get into it also once I have some more spare time

Ali-RS:
Okay, I am letting this be opened for one more day, going to merge it tomorrow if no objection is raised.

PR #1155, opened 2019-07-30
Modified TransformTrack.jmeClone() return type to TransformTrack

Ali-RS:


empirephoenix:
Sounds about right for me

PR #1156, opened 2019-08-02
Added support for null values in Vector3Serializer

Ali-RS:


Ali-RS:
Not sure how bad this will impact the message size by adding an extra byte, so @pspeed42  if you think this won't worth to add an extra byte to message size feel free to close this.

pspeed42:
In the past I haven't fixed this because it breaks compatibility for anything that updates the client and not the server or vice-versa.  I don't know how relevant that is now but I spent a lot of time back then trying to avoid it.

(Frankly, it's dumb that null wasn't built into the protocol and any new bit-based protocol I write will certainly include null support... and only take 1 bit to do it.)

Ali-RS:
Okay, I see. Looking forward to see a new bit-based protocol :)

pspeed42:
I wasn't saying your change is necessarily bad.  Just why I never made the change myself.

It could be that the situation has changed and no one will mind.

Ali-RS:
My use case for a null vector was a rare case which now I am convinced it won't worth to add an extra byte for it so I am sending a Vector3f.ZERO instead.

Issue #1158, opened 2019-08-10
BitmapText right to left line wrapping not work


Ali-RS:
Line wrapping of a right to left text should be handled differently than left to right text. BitmapText wraps the right to left text in the same way it does for the left to right text.

See the forum for more info: 
https://hub.jmonkeyengine.org/t/how-to-display-a-right-to-left-text-with-wrap-mode-with-bitmaptext/42148

Ali-RS:
Added a simple test case into jme3-examples.

Aufricer:
I will push the changes asap as I feel I am finished. 
If the shape of a letter is changing as well me realy need a new BitmapFont...

Aufricer:
Short feedback: Did working on alignment. Wordwrap is done. A lot of interrupted work. 
Open: 
- Need to check for letterwrap.
- Need to check for the kerning
Kerning - as you see there is the issue (still) as pspeed42 has mentioned. Kerning is calculated to "0" always but I need to dive deeper to the issue. Have already some ideas. 
As soon as this one is done I will pull the changes to my JMonkey (github) repo. We might need a test with the persian or an arabic font. Might be possible, that Kerning is correct in the font so - adjusting for left to right text might be optional. 
I will be more detailed. 

![grafik](https://user-images.githubusercontent.com/38795159/137462732-ce025440-0aa5-430a-b5be-8aa0fb67c14d.png)

Blue Color =  right to left text (text) 
White Color = left to right text (text2) - to compare for position/alignement etc. 
 private String text = ".text left to right test a is This";
 private String  text2 = "to right test a is This text left.";

Aufricer:
I have pushed my changes to my local JME3 repositiory. 

Changed files are: 
- Letters.Java
- LetterQuad.java
- BitmapText.java

It should work. But I have not yet tested it with a font that has tabs or offset or kernamount.
See the ToDos. 
![grafik](https://user-images.githubusercontent.com/38795159/137626149-fe5ebd4f-a03e-4a96-8a65-7fcb5a4b9d18.png)


What would be the next steps ?

Ali-RS:
Thanks @Aufricer , going to test it soon.

Ali-RS:
Tried it, I noticed I need to reverse the text to properly display it. It seems the `reverseText()` method in BitmapText is extra. If I remove it, then the text displays properly.

[...]

PR #1159, opened 2019-08-11
Added TestRTLBitmapText for issue #1158

Ali-RS:
The expected result  should be: 

```
                                                                   to right test a is This
                                                                                .text left 
```

but it displays: 

```
                                                                  .text left to right test 
                                                                                 a is This 
```

### DIFF
--- a/jme3-examples/src/main/java/jme3test/gui/TestRTLBitmapText.java
+++ b/jme3-examples/src/main/java/jme3test/gui/TestRTLBitmapText.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2009-2019 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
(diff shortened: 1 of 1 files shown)


pspeed42:
Thanks.  The "RTL" instead of "Rtl" makes me a little sad... but I'm happy there is a test for this.

Ali-RS:
>  The "RTL" instead of "Rtl" makes me a little sad...

Sorry, fixed it in this commit: https://github.com/jMonkeyEngine/jmonkeyengine/commit/16ccd36d56ec6af52923115b189e7f7359f4af4d

pspeed42:
Hahah.  Thanks. :)

Aufricer:
I have looked a bit into that matter. 
As @Ali-RS said in https://hub.jmonkeyengine.org/t/issue-with-line-wrapping-of-right-to-left-text-in-bitmaptext/42148/11 
It seems BitmapFont and BitmapText classes are only holding alignements of the text. They dont do something on the position of characters in the text. 

As a note, I try to show the workflow of how characters are applied and its position is determined: 

Bitmaptext.updateLogicalState() -> calls assemble() --> calls Letter.class update() 
-> available in that class are letterquads (created when the class was created or via settext) those letterquads are like a line called "head to tail" 
In Letter.update -> Letterquad.update(Stringblock) is called. All letters are iterated and position of x and y is determined. 
A variable incrScale is used in that calculation its created when Letters are created what is happening when Bitmaptext is created. Depending on the boolean "rightToLeft"  is either 1 or - 1. As far as I understood its only multiplied with an bitmapChar.getXOffset() * sizeScale (in the testcase above it resulted always in "0") . So it had no influence on x or y at all. 

In Letter.update also linewraps or wordraps are determined. Finally Letter.align() is called from Letter.update.
Here all letters/Letterquad (using the chain again) have there alignx and aligny calculated and set. 
After returning to Bitmaptext.assemble -->    textPages[i].assemble(letters); Here Buffers etc. are set. 

To sum it up: For me Alignement only has an influence to the alignement of the text not to the wordflow. Parameter rightToLeft may only have an influence if the bitmapchar do have an offset.

Thinking of the complexity of the current Bitmaptext there is only two solutions to apply right to left text
a) rewrite bigger parts of the font package
b) try to adjust the text! before it is assembled 

I will need to think about it later, but I think adding a new variable/parameter for righttoleft text should be added
Determining on that parameter the given text should be transformed before even the letter chain is created and the x + y position is calculated

BitMapText.java 
 private void assemble() {
      **ADDTEXTCONVERSION**   
// first generate quadlist
        letters.update();
        for (int i = 0; i < textPages.length; i++) {
            textPages[i].assemble(letters);
        }
        needRefresh = false;
    }

Not sure if it is as "easy" as transforming input from ABCD --> DCBA or if there is more calculation (as linewidth etc. necessary  )

Aufricer:
Letter. The wrapping (word, letter, ) happens in said letters.update as well. It had to be ajusted as well when coming from the right side. 
For lemur I have my own wordwrap functionality. But even if you only implement it to lemur only, you need to calculate the width of each letter/word/ line to check if it is > width of the given textblock. 
I would say its best done in letters/letterquad and Bitmaptext. Most of the logic is already there.

Ali-RS:
> Not sure if it is as "easy" as transforming input from ABCD --> DCBA

>if we do this in BitmapText, it will ruin the static texts (i.e labels) where the provided text is already in RTL

@Aufricer, sorry I was wrong about this. You are right that we need to inverse the letters in BitmapText. I just realized that char arrays returned by `String.toCharArray()` on a RTL text contains the characters from right to left (index 0 refrence to the first letter from the right)

```
String s = "کتاب"; 
for (int i = 0; i < s.length(); i++) {
    System.out.println("char at (" + i + ") is => " + s.charAt(i));
}
```
output:

```
char at (0) is => ک
char at (1) is => ت
char at (2) is => ا
char at (3) is => ب

PR #1160, opened 2019-08-12
Update the link of the "Lightspeed Frontier".

dmyroniuk:
The old link leads to the site for adults.


Ali-RS:
Thanks @dmyroniuk

Ali-RS:
I am pinging @MoffKalast in case he is not aware of this.

MoffKalast:
Thanks for the ping, though I did find out yesterday by chance actually. I've removed the links on Steam and other places I could remember but there may be some remaining.

Thing is, we didn't renew the domain so it ran out a while ago, but for some reason kept forwarding to the site. Did not expect it to get hijacked for this sort of thing to be honest and have no idea how long it's been like this. Sorry guys.

PR #1163, opened 2019-08-26
Some fixes in jme-vr

grizeldi:
This PR:

- Adds Valve Index to the list of recognized HMDs
- Brings the overall lwjgl version of the jme-vr module up to speed with the rest of the engine. This fixes OpenVR lwjgl backend.
- Removes Linux from the list of blacklisted OSes when lwjgl backend is used. In my testing, I got a picture inside the HMD on Linux, but it's cropped to be only the top half of the rendered view for some reason. Debugging this issue led me to believe that the version of OpenVR lwjgl 3.2.1 ships with is bugged on Linux, so this is unlikely to be fixed until we update lwjgl. On Windows it works fine.

_Also, is it just me or does git really think I deleted the whole files and then created them again?_

### DIFF
--- a/jme3-vr/build.gradle
+++ b/jme3-vr/build.gradle
@@ -2,7 +2,7 @@ if (!hasProperty('mainClass')) {
     ext.mainClass = ''
 }
 
-def lwjglVersion = '3.2.0'
+def lwjglVersion = '3.2.1'
 
 sourceCompatibility = '1.8'
 
--- a/jme3-vr/src/main/java/com/jme3/app/VREnvironment.java
+++ b/jme3-vr/src/main/java/com/jme3/app/VREnvironment.java
@@ -433,9 +433,14 @@ public class VREnvironment {
         // we are going to use OpenVR now, not the Oculus Rift
         // OpenVR does support the Rift
         String OS     = System.getProperty("os.name", "generic").toLowerCase(Locale.ENGLISH);
-        vrSupportedOS = !OS.contains("nux") && System.getProperty("sun.arch.data.model").equalsIgnoreCase("64"); //for the moment, linux/unix causes crashes, 64-bit only
-        compositorOS  = OS.contains("indows");
-        
+        vrSupportedOS = System.getProperty("sun.arch.data.model").equalsIgnoreCase("64"); //64-bit only
+        compositorOS  = OS.contains("indows") || OS.contains("nux");
+
+        if (OS.contains("nux") && vrBinding != VRConstants.SETTING_VRAPI_OPENVR_LWJGL_VALUE){
+        	logger.severe("Only LWJGL VR backend is currently (partially) supported on Linux.");
+        	vrSupportedOS = false;
+        }
+
         if( vrSupportedOS) {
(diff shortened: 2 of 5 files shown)


Ali-RS:
> Also, is it just me or does git really think I deleted the whole files and then created them again?
Yes it displays you removed −1,209 and added +1,223. It makes the code rewing hard. Any posibility to open another one with only the changed lines?

grizeldi:
Yeah, thought so. Go home git, you're drunk.

Not really sure what to change to not get this result as all I did was regular commits, push to a new branch on github and open a PR. No idea how this happened. Maybe I can retry.

MeFisto94:
Just click on the settings wheel and then select "hide whitespace changes" and it looks good.
I guess grizeldi did some reformatting changing tabs to spaces or something

grizeldi:
> Are there problems when the lwjgl versio doesn't match the one in core?

Yes, I got crashes about some abstract void not returning the correct type.

```
Exception in thread "main" java.lang.AbstractMethodError: org.lwjgl.system.StructBuffer.getElementFactory()Lorg/lwjgl/system/Struct;
    at org.lwjgl.system.StructBuffer.get(StructBuffer.java:90)
    at com.jme3.input.vr.lwjgl_openvr.LWJGLOpenVR.initialize(LWJGLOpenVR.java:180)
    at com.jme3.app.VREnvironment.initialize(VREnvironment.java:477)
    at com.grizeldi.vrtests.Main.main(Main.java:29)
```

jseinturier:
@MeFisto94 Yes the OS.startsWith("indows") || OS.startsWith("nux") is from the very first implementation and could be updated. The initial concern was to process "Windows" and "windows" but as now we use lower case OS, this is no more revelant.

Regarding this PR, my concern is that the files seems to be completely rewritten and a cannot see what are the changes provided. I'm not agree with validating this PR without seeing the real changes.

MeFisto94:
@jseinturier take a look at what I described above, there you can see the changes. Essentially it adds a constant to the enum for valve index, bumps lwjgl version and "supports linux".

> Yes, I got crashes about some abstract void not returning the correct type.
So we need to change jme3-lwjgl3 as well? So this should/could've been in this pr as well?

grizeldi:
> So we need to change jme3-lwjgl3 as well? So this should/could've been in this pr as well?

No, since jme-lwjgl3 is already on lwjgl version 3.2.1. Only jme-vr was on 3.2.0. Fixing the inconsistency fixed the crash, since I assume it was native related.

jseinturier:
Be also careful if you re using both jme3-lwjgl and jme3-lwjgl3 within your application. This can lead to unstability.

grizeldi:
As far as I'm aware lwjgl 2 doesn't have OpenVR bindings, so you can't really use it with jme-vr.

jseinturier:
Of course but if you're integrating a jme3-vr based module within large JME application with other jme3-lwjgl modules that it can lead to inconsistencies between lwjgl2 and lwjgl3 are some classes have same names but different implemntations.

PR #1165, opened 2019-09-01
Accelerated env baking on the GPU

riccardobl:
This PR improves the performances and user friendliness of environment baking. 

# Accelerated Baking
This PR adds 3 accelerated bakers that can be used in place of the current CPU baker.
The enhancements brought by these bakers is very remarkable, as they allow the environment baking process to be completed in <1s, effectively rendering the need for prebaked environments obsolete.

## IBLGLEnvBaker
This is the most high-performing baker.  It run entirely on the GPU, and can generate prefiltered, irradiance, and BRDF maps. 
This is the classic env baker implementation that is currently not used in jme.

## IBLHybridEnvBakerLight and IBLGLEnvBakerLight
These bakers generate the prefiltered map and spherical harmonics currently used in the latest PBR shader in jme.
Although this method reduces the VRAM consumption, this approach doesn't allow for as much optimization as the IBLGLEnvBaker. Nonetheless, the following accelerated baking methods are available:
- **IBLHybridEnvBakerLight**: A hybrid baker that creates the prefiltered environmental map using GPU while performing spheric harmonics calculations on the CPU.
- **IBLGLEnvBakerLight**: Similar to the IBLHybridEnvBakerLight, but it leverages the GPU to handle spheric harmonics generation as well.

# New APIs
The LightProbeFactory has now been deprecated in favor of the LightProbeFactory2, which seamlessly integrates the accelerated baking methods. This transition serves as a dedicated migration path for legacy code.

The new API `EnvironmentProbeControl` is designed to streamline and simplify the entire baking and rebaking process. 
 Here's how it works:

- *Integration*: Developers are only required to instantiate an EnvironmentProbeControl and attach it to the root node of their targeted scene (e.g., the rootNode).

- *Geometry Tagging*: By employing the EnvironmentProbeControl.tag(Spatial) method, you can mark which geometries are part of the environment and should be baked (e.g., the skybox).

- *Profit*: Once the tagging is in place, the control takes care of the rest. On the next frame, it automatically initiates baking for the tagged geometries within its node and completes it before the scene is rendered. No more waiting for the processes to conclude or resorting to convoluted culling tricks.

- *Update the baking*: Every time the baking needs to be updated (eg. day-night cycles) the rebake() method of the control can be invoked.

This new system operates directly within the same scene. It uses  RenderManager.setRenderFilter() (added by this PR) to filter out non tagged geometries, avoiding cloning, extra viewports, or other convoluted approaces. 

## Example
This is a fully functional remake of TestPBRLighting using the new API
```java
package jme3test.light.pbr;


import com.jme3.environment.EnvironmentProbeControl;
import com.jme3.input.ChaseCamera;
import com.jme3.math.FastMath;
import com.jme3.util.SkyFactory;
import com.jme3.util.mikktspace.MikktspaceTangentGenerator;

/**
 * TestPBRSimple
 */
public class TestPBRSimple extends SimpleApplication {

        new TestPBRSimple().start();
    }

    @Override
    public void simpleInitApp() {

        Geometry model = (Geometry) assetManager.loadModel("Models/Tank/tank.j3o");
        MikktspaceTangentGenerator.generate(model);

        Material pbrMat = assetManager.loadMaterial("Models/Tank/tank.j3m");
        model.setMaterial(pbrMat);
        rootNode.attachChild(model);

        ChaseCamera chaseCam = new ChaseCamera(cam, model, inputManager);
        chaseCam.setDragToRotate(true);
        chaseCam.setMinVerticalRotation(-FastMath.HALF_PI);
        chaseCam.setMaxDistance(1000);
        chaseCam.setSmoothMotion(true);
        chaseCam.setRotationSensitivity(10);
        chaseCam.setZoomSensitivity(5);
        flyCam.setEnabled(false);

        Spatial sky = SkyFactory.createSky(assetManager, "Textures/Sky/Path.hdr", SkyFactory.EnvMapType.EquirectMap);
        rootNode.attachChild(sky);

        // Create baker control
        EnvironmentProbeControl envProbe = new EnvironmentProbeControl( assetManager, 256);
        rootNode.addControl(envProbe);

        // Tag the sky, only the tagged spatials will be rendered in the env map
        EnvironmentProbeControl.tag(sky);

    }

}

```

As you can see, much simpler and cleaner.

## Knowing issues
The difference in roughness levels in the prefiltered envmap generated by the accelerated baker is less prominent than the current one, i am not sure why, you can test it out by toggling USE_ACCELERATED_BAKING in the TestPBRLighting class included in this PR.
I am not sure if this is an issue or which one is the correct behavior.

## Minimum requirements
OpenGL 3.1 or GL ES 3.0

### DIFF
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -90,6 +90,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     private boolean serializable = false;
     private float frustumNear = 0.001f, frustumFar = 1000f;
     private String uuid = "none";
+    private boolean enabled = true;
 
     private Function<Geometry, Boolean> filter = (s) -> {
         return s.getUserData("tags.env") != null || s.getUserData("tags.env.env" + uuid) != null;
@@ -215,8 +216,11 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     @Override
     public void setSpatial(Spatial spatial) {
-        spatial.addLight(this);
+        if (this.spatial != null && spatial != null && spatial != this.spatial) {
+            throw new IllegalStateException("This control has already been added to a Spatial");
+        }
         this.spatial = spatial;
+        if (spatial != null) spatial.addLight(this);
     }
 
     @Override
@@ -226,6 +230,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     @Override
     public void render(RenderManager rm, ViewPort vp) {
+        if (!isEnabled()) return;
         if (bakeNeeded) {
             bakeNeeded = false;
             rebakeNow(rm);
@@ -284,14 +289,17 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     void rebakeNow(RenderManager renderManager) {
         if (assetManager == null) {
-            LOG.log(Level.SEVERE, "AssetManager is null, cannot bake environment. Please use setAssetManager() to set it.");
+            LOG.log(Level.SEVERE,
+                    "AssetManager is null, cannot bake environment. Please use setAssetManager() to set it.");
             return;
         }
         IBLHybridEnvBakerLight baker;
         if (!USE_GL_IR) {
-            baker = new IBLHybridEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
+            baker = new IBLHybridEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
+                    envMapSize, envMapSize);
         } else {
-            baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
+            baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
+                    envMapSize, envMapSize);
         }
         baker.setTexturePulling(isRequiredSavableResults());
 
@@ -310,11 +318,25 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
         baker.clean();
     }
+    
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+    }
+
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public Spatial getSpatial() {
+        return spatial;
+    }
 
     @Override
     public void write(JmeExporter ex) throws IOException {
         super.write(ex);
         OutputCapsule oc = ex.getCapsule(this);
+        oc.write(enabled, "enabled", true);
+        oc.write(spatial, "spatial", null);
         oc.write(envMapSize, "size", 256);
         oc.write(serializable, "serializable", false);
         oc.write(bakeNeeded, "bakeNeeded", true);
@@ -327,6 +349,8 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     public void read(JmeImporter im) throws IOException {
         super.read(im);
         InputCapsule ic = im.getCapsule(this);
+        enabled = ic.readBoolean("enabled", true);
+        spatial = (Spatial) ic.readSavable("spatial", null);
         envMapSize = ic.readInt("size", 256);
         serializable = ic.readBoolean("serializable", false);
         bakeNeeded = ic.readBoolean("bakeNeeded", true);
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -34,8 +34,6 @@ package com.jme3.environment;
 import java.io.IOException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 import com.jme3.asset.AssetManager;
 import com.jme3.environment.baker.IBLGLEnvBakerLight;
 import com.jme3.environment.baker.IBLHybridEnvBakerLight;
@@ -79,8 +77,6 @@ import com.jme3.texture.Image.Format;
  * @author Riccardo Balbo
  */
 public class EnvironmentProbeControl extends LightProbe implements Control {
-    private final boolean USE_GL_IR = true;
-    private static final Logger LOG = Logger.getLogger(EnvironmentProbeControl.class.getName());
     private static AtomicInteger instanceCounter = new AtomicInteger(0);
 
     private AssetManager assetManager;
@@ -97,7 +93,10 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     };
 
     protected EnvironmentProbeControl() {
+        super();
         uuid = System.currentTimeMillis() + "_" + instanceCounter.getAndIncrement();
+        this.setAreaType(AreaType.Spherical);
+        this.getArea().setRadius(Float.MAX_VALUE);
     }
 
     /**
@@ -112,9 +111,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     public EnvironmentProbeControl(AssetManager assetManager, int size) {
         this();
         this.envMapSize = size;
-        this.assetManager = assetManager;
-        this.setAreaType(AreaType.Spherical);
-        this.getArea().setRadius(Float.MAX_VALUE);
+        this.assetManager = assetManager;        
     }
 
     /**
@@ -175,7 +172,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
      * Untag spatial as part of the environment for every
      * EnvironmentProbeControl.
      * 
-     * @param s
+     * @param s the spatial
      */
     public static void untagGlobal(Spatial s) {
         if (s instanceof Node) {
@@ -247,7 +244,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     /**
      * Set the minimum distance to render
      * 
-     * @param frustumNear
+     * @param frustumNear the minimum distance to render
      */
     public void setFrustumNear(float frustumNear) {
         this.frustumNear = frustumNear;
@@ -255,6 +252,8 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     /**
      * Set the maximum distance to render
+     * 
+     * @param frustumFar the maximum distance to render
      */
     public void setFrustumFar(float frustumFar) {
         this.frustumFar = frustumFar;
@@ -281,28 +280,17 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     /**
      * Set the asset manager used to load the shaders needed for the baking
      * 
-     * @param assetManager
+     * @param assetManager the asset manager
      */
     public void setAssetManager(AssetManager assetManager) {
         this.assetManager = assetManager;
     }
 
     void rebakeNow(RenderManager renderManager) {
-        if (assetManager == null) {
-            LOG.log(Level.SEVERE,
-                    "AssetManager is null, cannot bake environment. Please use setAssetManager() to set it.");
-            return;
-        }
-        IBLHybridEnvBakerLight baker;
-        if (!USE_GL_IR) {
-            baker = new IBLHybridEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
-                    envMapSize, envMapSize);
-        } else {
-            baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
-                    envMapSize, envMapSize);
-        }
+        IBLHybridEnvBakerLight baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
+                envMapSize, envMapSize);
+                    
         baker.setTexturePulling(isRequiredSavableResults());
-
         baker.bakeEnvironment(spatial, getPosition(), frustumNear, frustumFar, filter);
         baker.bakeSpecularIBL();
         baker.bakeSphericalHarmonicsCoefficients();
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -1,0 +1,104 @@
+package com.jme3.environment;
+
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import com.jme3.asset.AssetManager;
+import com.jme3.environment.baker.IBLGLEnvBakerLight;
+import com.jme3.light.LightProbe;
+import com.jme3.math.Vector3f;
+import com.jme3.renderer.RenderManager;
+import com.jme3.renderer.ViewPort;
+import com.jme3.scene.Geometry;
+import com.jme3.scene.Node;
+import com.jme3.scene.Spatial;
+import com.jme3.scene.control.Control;
+import com.jme3.texture.Image.Format;
+
+/**
+ * SmartLightProbe
+ */
+public class EnvironmentProbeControl extends LightProbe implements Control {
+
+
+
+    RenderManager renderManager;
+    AssetManager assetManager;
+    int envMapSize;
+    Spatial spatial;
+    boolean BAKE_NEEDED=true;
+    Function<Geometry,Boolean> filter=(s)->{
+        return s.getUserData("tags.env")!=null;
+    };
+
+    public static void tag(Spatial s){
+        if(s instanceof Node){
+            Node n=(Node)s;
+            for(Spatial sx:n.getChildren()){
+                tag(sx);
+            }
+        }else if(s instanceof Geometry){
+            s.setUserData("tags.env", true);
+        }
+    }
+
+    public EnvironmentProbeControl(RenderManager rm,AssetManager am, int size){
+        renderManager=rm;
+        assetManager=am;
+        envMapSize=size;
+    }
+
+    @Override
+    public Control cloneForSpatial(Spatial spatial) {
+        return null;
+    }
+
+    @Override
+    public void setSpatial(Spatial spatial) {
+        
+        spatial.addLight(this);
+        this.spatial=spatial;
+
+    }
+
+    @Override
+    public void update(float tpf) {
+
+    }
+
+    @Override
+	public void render(RenderManager rm, ViewPort vp) {
+        if(BAKE_NEEDED){
+            BAKE_NEEDED=false;
+            rebakeNow();
+        }
+    }
+
+    public void rebake(){
+        BAKE_NEEDED=true;
+    }
+    
+    void rebakeNow() {
+        System.out.println("BAKE");
+
+        IBLGLEnvBakerLight baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
+                envMapSize, envMapSize);
+        
+       
+        baker.bakeEnvironment(spatial, Vector3f.ZERO, 0.001f, 1000f,filter);
+        baker.bakeSpecularIBL();
+        baker.bakeSphericalHarmonicsCoefficients();
+
+
+        // probe.setPosition(Vector3f.ZERO);
+        setPrefilteredMap(baker.getSpecularIBL());
+        setNbMipMaps(getPrefilteredEnvMap().getImage().getMipMapSizes().length);
+        setShCoeffs(baker.getSphericalHarmonicsCoefficients());
+        setPosition(Vector3f.ZERO);
+        setReady(true);
+
+        baker.clean();
+
+    }
+    
+}
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -260,7 +260,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Get the minimum distance to render
+     * Gets the minimum distance to render.
      * 
      * @return frustum near
      */
@@ -269,7 +269,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Gets the maximum distance to render
+     * Gets the maximum distance to render.
      * 
      * @return frustum far
      */
@@ -278,7 +278,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Sets the asset manager used to load the shaders needed for the baking
+     * Sets the asset manager used to load the shaders needed for the baking.
      * 
      * @param assetManager the asset manager
      */
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -100,7 +100,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Create a new environment probe control.
+     * Creates a new environment probe control.
      * 
      * @param assetManager
      *            the asset manager used to load the shaders needed for the
@@ -115,7 +115,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Tag spatial as part of the environment for this EnvironmentProbeControl.
+     * Tags the specified spatial as part of the environment for this EnvironmentProbeControl.
      * Only tagged spatials will be rendered in the environment map.
      * 
      * @param s
@@ -133,7 +133,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Untag spatial as part of the environment for this
+     * Untags the specified spatial as part of the environment for this
      * EnvironmentProbeControl.
      * 
      * @param s
@@ -151,7 +151,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Tag spatial as part of the environment for every EnvironmentProbeControl.
+     * Tags the specified spatial as part of the environment for every EnvironmentProbeControl.
      * Only tagged spatials will be rendered in the environment map.
      * 
      * @param s
@@ -169,7 +169,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Untag spatial as part of the environment for every
+     * Untags the specified spatial as part of the environment for every
      * EnvironmentProbeControl.
      * 
      * @param s the spatial
@@ -191,7 +191,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Request savable results from the baking process. This will make the
+     * Requests savable results from the baking process. This will make the
      * baking process slower and more memory intensive but will allow to
      * serialize the results with the control.
      * 
@@ -203,7 +203,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Return true if savable results are required by this control.
+     * Returns true if savable results are required by this control.
      * 
      * @return true if savable results are required.
      */
@@ -235,14 +235,14 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Schedule a rebake of the environment map.
+     * Schedules a rebake of the environment map.
      */
     public void rebake() {
         bakeNeeded = true;
     }
 
     /**
-     * Set the minimum distance to render
+     * Sets the minimum distance to render.
      * 
      * @param frustumNear the minimum distance to render
      */
@@ -251,7 +251,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Set the maximum distance to render
+     * Sets the maximum distance to render.
      * 
      * @param frustumFar the maximum distance to render
      */
@@ -269,7 +269,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Get the maximum distance to render
+     * Gets the maximum distance to render
      * 
      * @return frustum far
      */
@@ -278,7 +278,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     /**
-     * Set the asset manager used to load the shaders needed for the baking
+     * Sets the asset manager used to load the shaders needed for the baking
      * 
      * @param assetManager the asset manager
      */
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -33,7 +33,7 @@ package com.jme3.environment;
 
 import java.io.IOException;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import com.jme3.asset.AssetManager;
@@ -92,7 +92,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     private String uuid = "none";
     private boolean enabled = true;
 
-    private Function<Geometry, Boolean> filter = (s) -> {
+    private Predicate<Geometry> filter = (s) -> {
         return s.getUserData("tags.env") != null || s.getUserData("tags.env.env" + uuid) != null;
     };
 
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -67,6 +67,8 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     public EnvironmentProbeControl(AssetManager assetManager,int size) {     
         this.envMapSize = size;
         this.assetManager = assetManager;
+        this.setAreaType(AreaType.Spherical);
+        this.getArea().setRadius(Float.MAX_VALUE);
     }
 
     @Override
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -136,7 +136,10 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         baker.bakeSphericalHarmonicsCoefficients();
 
         setPrefilteredMap(baker.getSpecularIBL());
-        setNbMipMaps(getPrefilteredEnvMap().getImage().getMipMapSizes().length);
+        
+        int[] mipSizes = getPrefilteredEnvMap().getImage().getMipMapSizes();
+        setNbMipMaps(mipSizes != null ? mipSizes.length : 1);
+        
         setShCoeffs(baker.getSphericalHarmonicsCoefficients());
         setPosition(Vector3f.ZERO);
         setReady(true);
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -87,7 +87,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     private boolean bakeNeeded = true;
     private int envMapSize = 256;
     private Spatial spatial;
-    private boolean serializable = false;
+    private boolean requiredSavableResults = false;
     private float frustumNear = 0.001f, frustumFar = 1000f;
     private String uuid = "none";
     private boolean enabled = true;
@@ -202,7 +202,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
      *            true to enable (default: false)
      */
     public void setRequiredSavableResults(boolean v) {
-        serializable = v;
+        requiredSavableResults = v;
     }
 
     /**
@@ -211,7 +211,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
      * @return true if savable results are required.
      */
     public boolean isRequiredSavableResults() {
-        return serializable;
+        return requiredSavableResults;
     }
 
     @Override
@@ -338,7 +338,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         oc.write(enabled, "enabled", true);
         oc.write(spatial, "spatial", null);
         oc.write(envMapSize, "size", 256);
-        oc.write(serializable, "serializable", false);
+        oc.write(requiredSavableResults, "requiredSavableResults", false);
         oc.write(bakeNeeded, "bakeNeeded", true);
         oc.write(frustumFar, "frustumFar", 1000f);
         oc.write(frustumNear, "frustumNear", 0.001f);
@@ -352,7 +352,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         enabled = ic.readBoolean("enabled", true);
         spatial = (Spatial) ic.readSavable("spatial", null);
         envMapSize = ic.readInt("size", 256);
-        serializable = ic.readBoolean("serializable", false);
+        requiredSavableResults = ic.readBoolean("requiredSavableResults", false);
         bakeNeeded = ic.readBoolean("bakeNeeded", true);
         assetManager = im.getAssetManager();
         frustumFar = ic.readFloat("frustumFar", 1000f);
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -1,10 +1,10 @@
 package com.jme3.environment;
 
 import java.util.function.Function;
-import java.util.function.Predicate;
 
 import com.jme3.asset.AssetManager;
 import com.jme3.environment.baker.IBLGLEnvBakerLight;
+import com.jme3.environment.baker.IBLHybridEnvBakerLight;
 import com.jme3.light.LightProbe;
 import com.jme3.math.Vector3f;
 import com.jme3.renderer.RenderManager;
@@ -28,6 +28,8 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     private int envMapSize;
     private Spatial spatial;
     private boolean BAKE_NEEDED = true;
+    private boolean USE_GL_IR = true;
+    private boolean serializable = false;
 
     private Function<Geometry, Boolean> filter = (s) -> {
         return s.getUserData("tags.env") != null;
@@ -62,6 +64,14 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         return null;
     }
 
+    public void setSerializeBakeResults(boolean v) {
+        serializable = v;
+    }
+
+    public boolean isSerializeBakeResults() {
+        return serializable;
+    }
+
     @Override
     public void setSpatial(Spatial spatial) {
 
@@ -92,7 +102,13 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     void rebakeNow() {
 
-        IBLGLEnvBakerLight baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
+        IBLHybridEnvBakerLight baker;
+        if(!USE_GL_IR){
+            baker = new IBLHybridEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
+        } else {
+            baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
+        }
+        baker.setTexturePulling(isSerializeBakeResults());
 
         baker.bakeEnvironment(spatial, Vector3f.ZERO, 0.001f, 1000f, filter);
         baker.bakeSpecularIBL();
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -2,6 +2,8 @@ package com.jme3.environment;
 
 import java.io.IOException;
 import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import com.jme3.asset.AssetManager;
 import com.jme3.environment.baker.IBLGLEnvBakerLight;
@@ -28,6 +30,8 @@ import com.jme3.texture.Image.Format;
  */
 public class EnvironmentProbeControl extends LightProbe implements Control {
     private final boolean USE_GL_IR = true;
+    private static final Logger LOG = Logger.getLogger(EnvironmentProbeControl.class.getName());
+
     
     private AssetManager assetManager;
     private boolean bakeNeeded = true;
@@ -57,6 +61,9 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         }
     }
 
+    protected EnvironmentProbeControl() {
+    }
+    
     public EnvironmentProbeControl(AssetManager assetManager,int size) {     
         this.envMapSize = size;
         this.assetManager = assetManager;
@@ -64,7 +71,7 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     @Override
     public Control cloneForSpatial(Spatial spatial) {
-        return null;
+        throw new UnsupportedOperationException();
     }
 
     public void setSerializeBakeResults(boolean v) {
@@ -79,7 +86,6 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     public void setSpatial(Spatial spatial) {
         spatial.addLight(this);
         this.spatial = spatial;
-
     }
 
     @Override
@@ -102,7 +108,19 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         bakeNeeded = true;
     }
 
+    /**
+     * Set the asset manager used to load the shaders needed for the baking
+     * @param assetManager
+     */
+    public void setAssetManager(AssetManager assetManager) {
+        this.assetManager = assetManager;
+    }
+
     void rebakeNow(RenderManager renderManager) {
+        if (assetManager == null) {
+            LOG.log(Level.SEVERE, "AssetManager is null, cannot bake environment. Please use setAssetManager() to set it.");
+            return;            
+        }
         IBLHybridEnvBakerLight baker;
         if(!USE_GL_IR){
             baker = new IBLHybridEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -1,10 +1,15 @@
 package com.jme3.environment;
 
+import java.io.IOException;
 import java.util.function.Function;
 
 import com.jme3.asset.AssetManager;
 import com.jme3.environment.baker.IBLGLEnvBakerLight;
 import com.jme3.environment.baker.IBLHybridEnvBakerLight;
+import com.jme3.export.InputCapsule;
+import com.jme3.export.JmeExporter;
+import com.jme3.export.JmeImporter;
+import com.jme3.export.OutputCapsule;
 import com.jme3.light.LightProbe;
 import com.jme3.math.Vector3f;
 import com.jme3.renderer.RenderManager;
@@ -22,13 +27,12 @@ import com.jme3.texture.Image.Format;
  * @author Riccardo Balbo
  */
 public class EnvironmentProbeControl extends LightProbe implements Control {
-
-    private RenderManager renderManager;
+    private final boolean USE_GL_IR = true;
+    
     private AssetManager assetManager;
-    private int envMapSize;
+    private boolean bakeNeeded = true;
+    private int envMapSize=256;
     private Spatial spatial;
-    private boolean BAKE_NEEDED = true;
-    private boolean USE_GL_IR = true;
     private boolean serializable = false;
 
     private Function<Geometry, Boolean> filter = (s) -> {
@@ -53,10 +57,9 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         }
     }
 
-    public EnvironmentProbeControl(RenderManager rm, AssetManager am, int size) {
-        renderManager = rm;
-        assetManager = am;
-        envMapSize = size;
+    public EnvironmentProbeControl(AssetManager assetManager,int size) {     
+        this.envMapSize = size;
+        this.assetManager = assetManager;
     }
 
     @Override
@@ -74,7 +77,6 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     @Override
     public void setSpatial(Spatial spatial) {
-
         spatial.addLight(this);
         this.spatial = spatial;
 
@@ -87,9 +89,9 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     @Override
     public void render(RenderManager rm, ViewPort vp) {
-        if (BAKE_NEEDED) {
-            BAKE_NEEDED = false;
-            rebakeNow();
+        if (bakeNeeded) {
+            bakeNeeded = false;
+            rebakeNow(rm);
         }
     }
 
@@ -97,11 +99,10 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
      * Schedule a rebake of the environment map.
      */
     public void rebake() {
-        BAKE_NEEDED = true;
+        bakeNeeded = true;
     }
 
-    void rebakeNow() {
-
+    void rebakeNow(RenderManager renderManager) {
         IBLHybridEnvBakerLight baker;
         if(!USE_GL_IR){
             baker = new IBLHybridEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
@@ -121,7 +122,26 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         setReady(true);
 
         baker.clean();
+    }
 
+
+    @Override
+    public void write(JmeExporter ex) throws IOException {
+        super.write(ex);
+        OutputCapsule oc = ex.getCapsule(this);
+        oc.write(envMapSize, "size", 256);
+        oc.write(serializable, "serializable", false);
+        oc.write(bakeNeeded, "bakeNeeded", true);
+    }
+    
+    @Override
+    public void read(JmeImporter im) throws IOException {
+        super.read(im);
+        InputCapsule ic = im.getCapsule(this);
+        envMapSize = ic.readInt("size", 256);
+        serializable = ic.readBoolean("serializable", false);
+        bakeNeeded = ic.readBoolean("bakeNeeded", true);
+        assetManager = im.getAssetManager();
     }
 
 }
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -16,36 +16,45 @@ import com.jme3.scene.control.Control;
 import com.jme3.texture.Image.Format;
 
 /**
- * SmartLightProbe
+ * A control that automatically handles environment bake and rebake including
+ * only tagged spatials.
+ * 
+ * @author Riccardo Balbo
  */
 public class EnvironmentProbeControl extends LightProbe implements Control {
 
+    private RenderManager renderManager;
+    private AssetManager assetManager;
+    private int envMapSize;
+    private Spatial spatial;
+    private boolean BAKE_NEEDED = true;
 
-
-    RenderManager renderManager;
-    AssetManager assetManager;
-    int envMapSize;
-    Spatial spatial;
-    boolean BAKE_NEEDED=true;
-    Function<Geometry,Boolean> filter=(s)->{
-        return s.getUserData("tags.env")!=null;
+    private Function<Geometry, Boolean> filter = (s) -> {
+        return s.getUserData("tags.env") != null;
     };
 
-    public static void tag(Spatial s){
-        if(s instanceof Node){
-            Node n=(Node)s;
-            for(Spatial sx:n.getChildren()){
+    /**
+     * Tag spatial as part of the environment. Only tagged spatials will be
+     * rendered in the environment map.
+     * 
+     * @param s
+     *            the spatial
+     */
+    public static void tag(Spatial s) {
+        if (s instanceof Node) {
+            Node n = (Node) s;
+            for (Spatial sx : n.getChildren()) {
                 tag(sx);
             }
-        }else if(s instanceof Geometry){
+        } else if (s instanceof Geometry) {
             s.setUserData("tags.env", true);
         }
     }
 
-    public EnvironmentProbeControl(RenderManager rm,AssetManager am, int size){
-        renderManager=rm;
-        assetManager=am;
-        envMapSize=size;
+    public EnvironmentProbeControl(RenderManager rm, AssetManager am, int size) {
+        renderManager = rm;
+        assetManager = am;
+        envMapSize = size;
     }
 
     @Override
@@ -55,9 +64,9 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
 
     @Override
     public void setSpatial(Spatial spatial) {
-        
+
         spatial.addLight(this);
-        this.spatial=spatial;
+        this.spatial = spatial;
 
     }
 
@@ -67,30 +76,28 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     @Override
-	public void render(RenderManager rm, ViewPort vp) {
-        if(BAKE_NEEDED){
-            BAKE_NEEDED=false;
+    public void render(RenderManager rm, ViewPort vp) {
+        if (BAKE_NEEDED) {
+            BAKE_NEEDED = false;
             rebakeNow();
         }
     }
 
-    public void rebake(){
-        BAKE_NEEDED=true;
+    /**
+     * Schedule a rebake of the environment map.
+     */
+    public void rebake() {
+        BAKE_NEEDED = true;
     }
-    
-    void rebakeNow() {
-        System.out.println("BAKE");
 
-        IBLGLEnvBakerLight baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
-                envMapSize, envMapSize);
-        
-       
-        baker.bakeEnvironment(spatial, Vector3f.ZERO, 0.001f, 1000f,filter);
+    void rebakeNow() {
+
+        IBLGLEnvBakerLight baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth, envMapSize, envMapSize);
+
+        baker.bakeEnvironment(spatial, Vector3f.ZERO, 0.001f, 1000f, filter);
         baker.bakeSpecularIBL();
         baker.bakeSphericalHarmonicsCoefficients();
 
-
-        // probe.setPosition(Vector3f.ZERO);
         setPrefilteredMap(baker.getSpecularIBL());
         setNbMipMaps(getPrefilteredEnvMap().getImage().getMipMapSizes().length);
         setShCoeffs(baker.getSphericalHarmonicsCoefficients());
@@ -100,5 +107,5 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
         baker.clean();
 
     }
-    
+
 }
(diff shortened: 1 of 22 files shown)


tlf30:
@riccardobl this looks really cool. Is there more info on this?

riccardobl:
You can see it working in
TestPBRLighting.java 
and
TestPBRSimple.java

The code is based on this https://learnopengl.com/PBR/IBL/Specular-IBL (and following articles).

The PR works but as you can see in the two examples, it exposes two different apis, one that is similar to what is used now in the master branch (LightProbeFactory2) and one that use a control. I'm not sure which one should go in the final pr. Also it requires modifications to the framebuffer class, that i will have to extract to a dedicated PR.

riccardobl:
I've updated the PR and added a description. This is ready to be reviewed

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,127 @@
...
+ * 
+ * @author Riccardo Balbo
+ */
+public class EnvironmentProbeControl extends LightProbe implements Control {
+
+    private RenderManager renderManager;
+    private AssetManager assetManager;
+    private int envMapSize;
+    private Spatial spatial;
+    private boolean BAKE_NEEDED = true;
```
danielperano, line 30: When I see an `ALL_CAPS` name I expect that I'm looking at a constant. I find it slightly confusing that `BAKE_NEEDED` and `USE_GL_IR` are named this way but are variables.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,254 @@
...
+    public GenericEnvBaker(RenderManager rm, AssetManager am, Format colorFormat, Format depthFormat, int env_size) {
+        this.depthFormat = depthFormat;
+
+        renderManager = rm;
+        assetManager = am;
+
+        cam = new Camera(128, 128);
+
+        env = new TextureCubeMap(env_size, env_size, colorFormat);
+        env.setMagFilter(MagFilter.Bilinear);
```
danielperano, line 121: Should min/mag filters be configurable, or are these effectively irrelevant to the rendered result (either the map itself or the rendered result)?

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,254 @@
...
+        cam.setLocation(position);
+        cam.setFrustumPerspective(90.0F, 1F, frustumNear, frustumFar);
+        cam.setRotation(new Quaternion().fromAxes(axisX[id], axisY[id], axisZ[id]));
+        return cam;
+    }
+
+    @Override
+    public void clean() {
+        env.getImage().dispose();
+        System.gc();
```
danielperano, line 152: Is there a specific reason for two `System.gc()` calls in a row, and do we absolutely need these? I generally consider `System.gc()` something to be avoided except when testing/debugging certain behaviors (it can negatively impact performance, for one thing, and it's not a reliable way to force GC since it may be a non-op or ignored at the GC's whim).

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,236 @@
...
+                        Format format,
+                        Format depthFormat,
+                        int env_size,int specular_size,
+                        int irradiance_size,
+                        int brtf_size
+    ){
+        super(rm,am,format,depthFormat,env_size);  
+
+        irradiance=new TextureCubeMap(irradiance_size,irradiance_size,format);
+        irradiance.setMagFilter(MagFilter.Bilinear);
```
danielperano, line 75: As per earlier comment, is there a specific expectation for these filters? If not, could these parameters be configurable if they may impact final render quality?

jme3-core/src/main/java/com/jme3/environment/baker/IBLHybridEnvBakerLight.java
```diff
@@ -0,0 +1,148 @@
...
+    protected TextureCubeMap specular;
+    protected Vector3f[] shCoef;
+
+    public IBLHybridEnvBakerLight(RenderManager rm, AssetManager am, Format format, Format depthFormat, int env_size, int specular_size
+
+    ) {
+        super(rm, am, format, depthFormat, env_size);
+
+        specular = new TextureCubeMap(specular_size, specular_size, format);
+        specular.setMagFilter(MagFilter.Bilinear);
```
danielperano, line 70: Same question as with other texture filters. ;)

jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
```diff
@@ -1298,4 +1301,25 @@ public void render(float tpf, boolean mainFrameBufferActive) {
...
         }
     }
+
+    /**
+     * Set a render filter. Every geometry will be tested against this filter
+     * before rendering and will only be rendered if the filter returns true.
+     * 
+     * @param filter
+     */
+    public void setRenderFilter(Function<Geometry, Boolean> filter) {
```
danielperano, line 1311: In general I don't love temporary state like this that can (if mishandled) persist across operations (rendering env -> rendering main scene).

I find the following pattern for situations like this intuitive and easy to use correctly because it doesn't alter state:
```
render(<params>, filter) {
  ...
  if(filter != null && filter.test(n))
  ...
}

render(<params>){
  render(<params>, null);
}
```

riccardobl: I generally agree with you, but the render manager was developed like some sort of state machine... goblin, there are too many render related methods to make passing a filter around efficient, also the same approach is used by many other methods (eg.   setForcedMaterial ).
I consider the RenderManager to be an internal api, so i wouldn't worry about this too much as long as code calling it does it correctly. Also, with this approach you can make interest things, like adapting existing code and libraries .

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
riccardobl: They are required here

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
riccardobl: Fixed

jme3-core/src/main/java/com/jme3/environment/baker/IBLHybridEnvBakerLight.java
riccardobl: They are required here

danielperano (approved):

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
riccardobl: I've removed this method entirely, since the generated env map can be used after the baker is disposed, having a clean method like that might be counterintuitive as it would destroy the env map unexpectedly

riccardobl:
Thank you for your review @danielperano

riccardobl:
Rebased to current master

riccardobl:
Are there any objections to merging this? It's worth noting that this PR doesn't break the current baking process.
@jMonkeyEngine/core

jme3-core/src/main/java/com/jme3/environment/LightProbeFactory2.java
```diff
@@ -0,0 +1,114 @@
...
+import com.jme3.environment.util.EnvMapUtils;
+import com.jme3.light.LightProbe;
+import com.jme3.math.Vector3f;
+import com.jme3.renderer.RenderManager;
+import com.jme3.scene.Node;
+import com.jme3.scene.Spatial;
+import com.jme3.texture.Image.Format;
+
+/**
+ * A faster version of LightProbeFactory that uses accelerated Baking.
```
stephengold, line 46: There's a lot of info in the PR that could be added here so it's not lost.

The key information would be whether this replaces `LightProbeFactory` and why?

jme3-core/src/main/java/com/jme3/environment/LightProbeFactory2.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param am The AssetManager
+     * @param size The size of the probe
+     * @param pos The position of the probe
+     * @param frustumNear The near frustum of the probe
+     * @param frustumFar The far frustum of the probe
+     * @param scene The scene to bake
+     * @return The baked LightProbe
+     */
+    public static LightProbe makeProbe(RenderManager rm,
+    AssetManager am, int size,Vector3f pos, float frustumNear,float frustumFar,Spatial scene) {
```
stephengold, line 64: This line needs at least 8 spaces additional indentation.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,170 @@
+package com.jme3.environment;
+
+import java.io.IOException;
+import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
```
stephengold, line 7: No blank line here per Google style guide.

jme3-core/src/main/java/com/jme3/environment/baker/EnvBaker.java
```diff
@@ -0,0 +1,85 @@
...
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.jme3.environment.baker;
+
+import java.util.function.Function;
+
```
stephengold, line 36: No blank line is allowed here per Google style.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+package com.jme3.environment.baker;
+
+import java.io.ByteArrayOutputStream;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
```
stephengold, line 42: No blank line is allowed here per Google style.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+    protected final RenderManager renderManager;
+    protected final AssetManager assetManager;
+    protected final Camera cam;
+    protected  boolean texturePulling=false;
+    protected List<ByteArrayOutputStream> bos = new ArrayList<>();
+    private static final Logger LOG=Logger.getLogger(GenericEnvBaker.class.getName());
+
+ 
+
+    public GenericEnvBaker(RenderManager rm, AssetManager am, Format colorFormat, Format depthFormat, int env_size) {
```
stephengold, line 112: Public constructor needs javadoc.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+    @Override
+    public boolean isTexturePulling() {
+        return texturePulling;
+    }
+
+    public TextureCubeMap getEnvMap() {
+        return env;
+    }
+
+    Camera getCam(int id, int w, int h, Vector3f position, float frustumNear, float frustumFar) {
```
stephengold, line 141: Public constructor needs javadoc.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+import com.jme3.util.BufferUtils;
+
+/**
+ * Render the environment into a cubemap
+ *
+ * @author Riccardo Balbo
+ */
+public abstract class GenericEnvBaker implements EnvBaker {
+
+    protected static Vector3f[] axisX = new Vector3f[6];
```
stephengold, line 70: Since 6 (the number of sides of a cube?) gets used so much, it should have a named constant.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+        for (int i = 0; i < 6; i++) {
+            envbakers[i] = new FrameBuffer(env.getImage().getWidth(), env.getImage().getHeight(), 1);
+            envbakers[i].setDepthTarget(FrameBufferTarget.newTarget(depthFormat));
+            envbakers[i].setSrgb(false);
+            envbakers[i].addColorTarget(FrameBufferTarget.newTarget(env).face(TextureCubeMap.Face.values()[i]));
+        }
+
+       
+
+        if(isTexturePulling())startPulling();
```
stephengold, line 166: Please add a blank after `if` and put the loop body in braces per the style guide.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+            scene.updateLogicalState(0);
+            scene.updateGeometricState();
+
+            Function<Geometry, Boolean> ofilter = renderManager.getRenderFilter();
+
+            renderManager.setRenderFilter(filter);
+            renderManager.renderViewPort(viewPort, 0.16f);
+            renderManager.setRenderFilter(ofilter);
+
+            if (isTexturePulling()) pull(envbaker, env, i);
```
stephengold, line 189: Loop body should be enclosed in braces.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+            
+        }
+
+        if (isTexturePulling()) endPulling(env);
```
stephengold, line 193: Loop body should be enclosed in braces.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+     * Starts pulling the data from the framebuffer into the texture
+     */
+    protected void startPulling() {
+        bos.clear();
+    }
+
+    /**
+     * Pulls the data from the framebuffer into the texture
+     * Nb. mipmaps must be pulled sequentially on the same faceId
+     * @param fb the framebuffer to pull from
```
stephengold, line 211: There should be a line with just "*" before the first @param in the javadoc.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+    protected ByteBuffer pull(FrameBuffer fb, Texture env, int faceId) {
+
+        if (fb.getColorTarget().getFormat() != env.getImage().getFormat())
+            throw new IllegalArgumentException("Format mismatch: " + fb.getColorTarget().getFormat() + "!=" + env.getImage().getFormat());
+
+        ByteBuffer face = BufferUtils.createByteBuffer(fb.getWidth() * fb.getHeight() * (fb.getColorTarget().getFormat().getBitsPerPixel() / 8));
+        renderManager.getRenderer().readFrameBufferWithFormat(fb, face, fb.getColorTarget().getFormat());
+        face.rewind();
+
+        while (bos.size() <= faceId) bos.add(null);
```
stephengold, line 225: Loop body should be enclosed in curly braces.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+        ByteArrayOutputStream bo = bos.get(faceId);
+        if (bo == null) bos.set(faceId, bo = new ByteArrayOutputStream());
```
stephengold, line 227: Body of the `if` statement should be enclosed in braces.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Ends pulling the data into the texture
+     * @param tx the texture to pull into
+     */
+    protected void endPulling(Texture tx) {
+        for (int i = 0; i < bos.size(); i++) {
+            ByteArrayOutputStream bo = bos.get(i);
+            if (bo == null) {
+                LOG.log(Level.SEVERE, "Missing face {0}. Pulling incomplete!", i);
+                continue;
```
stephengold, line 248: The code would be easier to read if you used `else` instead of `continue` here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+package com.jme3.environment.baker;
+
+import java.util.Arrays;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
```
stephengold, line 39: Blank line not allowed here per Google style.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+    protected TextureCubeMap irradiance;
+    protected TextureCubeMap specular;
+    public IBLGLEnvBaker(RenderManager rm,AssetManager am,
+                        Format format,
+                        Format depthFormat,
+                        int env_size,int specular_size,
+                        int irradiance_size,
+                        int brtf_size
+    ){
+        super(rm,am,format,depthFormat,env_size);
```
stephengold, line 78: Please add a blank after each comma.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+    public IBLGLEnvBaker(RenderManager rm,AssetManager am,
+                        Format format,
+                        Format depthFormat,
+                        int env_size,int specular_size,
+                        int irradiance_size,
+                        int brtf_size
+    ){
+        super(rm,am,format,depthFormat,env_size);  
+
+        irradiance=new TextureCubeMap(irradiance_size,irradiance_size,format);
```
stephengold, line 80: Please add a blank before/after `=` and also after each comma.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+        specular=new TextureCubeMap(specular_size,specular_size,format);
+        specular.setMagFilter(MagFilter.Bilinear);
+        specular.setMinFilter(MinFilter.Trilinear);
+        specular.setWrap(WrapMode.EdgeClamp);
+        specular.getImage().setColorSpace(ColorSpace.Linear);
+        
+        int nbMipMaps = (int) (Math.log(specular_size) / Math.log(2) + 1);
+        nbMipMaps = limitMips(nbMipMaps, specular.getImage().getWidth(), specular.getImage().getHeight(),rm);
+        
+        int[] sizes=new int[nbMipMaps];
```
stephengold, line 95: Please add a blank before/after `=`.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+        for(int i=0;i<nbMipMaps;i++){
```
stephengold, line 96: Needs blanks after `for`, before and after '=', after each ';', and before and after '<'.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+
+        brtf=new Texture2D(brtf_size,brtf_size,format);
+        brtf.setMagFilter(MagFilter.Bilinear);
+        brtf.setMinFilter(MinFilter.BilinearNoMipMaps);
+        brtf.setWrap(WrapMode.EdgeClamp);
+        brtf.getImage().setColorSpace(ColorSpace.Linear);
+    }
+
+
+    public TextureCubeMap getSpecularIBL(){
```
stephengold, line 110: Needs a blank before '{' curly brace.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+
+        }
+        for (int i = 0; i < 6; i++) {
+            specularbakers[i].dispose();
+        }
+    }
+
+    @Override
+    public void bakeSpecularIBL() {
+        Box boxm=new Box(1,1,1);
```
stephengold, line 153: Needs a blank before and after '=' and a blank after each ','.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+        Geometry screen=new Geometry("BakeBox",boxm);
```
stephengold, line 154: Needs a blank before and after '=' and a blank after each ','.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,275 @@
...
+    
+        Material mat=new Material(assetManager,"Common/IBL/IBLKernels.j3md");
+        mat.setBoolean("UseSpecularIBL",true);
```
stephengold, line 157: Please run the autoformatter over all new files. After that I will re-review.

stephengold (changes requested):

jme3-core/src/main/java/com/jme3/environment/LightProbeFactory.java
```diff
@@ -45,31 +45,36 @@
...
+ * Since this process can take a long time, you can provide a
+ * JobProgressListener that will be notified of the ongoing generation process
+ * when calling the makeProbe method.
  * 
- * The process is as follows: 
- * 1. Create an EnvironmentCamera
- * 2. give it a position in the scene
- * 3. call {@link LightProbeFactory#makeProbe(com.jme3.environment.EnvironmentCamera, com.jme3.scene.Spatial)}
- * 4. add the created LightProbe to a node with the {@link Node#addLight(com.jme3.light.Light) } method.
+ * The process is as follows: 1. Create an EnvironmentCamera 2. give it a
```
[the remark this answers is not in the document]
riccardobl: Reverted since this was reformatted by mistake during the rebase and is not modified by the pr

stephengold:
Please don't mark a conversation as "resolved" unless you've participated in it and/or acted on its advice.

riccardobl:
> Please don't mark a conversation as "resolved" unless you've participated in it and/or acted on its advice.

I was just tracking my progress...
The new commits should address all the requested changes, i've also improved the documentation and some methods and class names, and made some addition to support frustum, position and local tagging (eg. one probe can tag some spatials as environment and another probe can tag others) for probes.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
riccardobl: I disagree, since the premise of this code is to deal with cubemaps that always have 6 faces, i think using a named constant is redundant

jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
```diff
@@ -104,6 +105,7 @@ public class RenderManager {
     private TechniqueDef.LightMode preferredLightMode = TechniqueDef.LightMode.MultiPass;
     private int singlePassLightBatchSize = 1;
     private MatParamOverride boundDrawBufferId=new MatParamOverride(VarType.Int,"BoundDrawBuffer",0);
+    private Function<Geometry, Boolean> renderFilter;
```
capdevon, line 108: Since the filter always returns a boolean value, you can use a `Predicate` instead of a `Function`. 

<html>
<body>


Sr. No. | Key | Function | Predicate
-- | -- | -- | --
1 | Basic | It can take 2 type parameters First one represents input type argument type and second one represents return type. | It can take one type parameter which represents input type or argument type.
2 | Return Type | It can return any type of value. | It can only return boolean value
3 | Method | It has abstract method apply(). | It has abstract method test().
4 | Use Case | It can be used to implement conditional checks | It can be used for the transformation and to the return values.


</body>
</html>

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,339 @@
...
+    /**
+     * Request savable results from the baking process. This will make the
+     * baking process slower and more memory intensive but will allow to
+     * serialize the results with the control.
+     * 
+     * @param v
+     *            true to enable (default: false)
+     */
+    public void setRequiredSavableResults(boolean v) {
```
capdevon, line 203: Please rename the methods set/is `RequiredSavableResults` -> set/is `Serializable` or rename the variable `serializable` -> `requiredSavableResults`. Keep variable names consistent with their getter/setter methods in accordance with java standards and to facilitate automatic GUI generation via the SDK using `Reflection` to access `Control` fields.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,339 @@
...
+        serializable = v;
+    }
+
+    /**
+     * Return true if savable results are required by this control.
+     * 
+     * @return true if savable results are required.
+     */
+    public boolean isRequiredSavableResults() {
```
capdevon, line 212: Please rename the methods set/is `RequiredSavableResults` -> set/is `Serializable` or rename the variable `serializable` -> `requiredSavableResults`. Keep variable names consistent with their getter/setter methods in accordance with java standards and to facilitate automatic GUI generation via the **SDK** using `Reflection` to access `Control` fields.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,339 @@
...
+
+        setShCoeffs(baker.getSphericalHarmonicsCoefficients());
+        setPosition(Vector3f.ZERO);
+        setReady(true);
+
+        baker.clean();
+    }
+
+    @Override
+    public void write(JmeExporter ex) throws IOException {
```
capdevon, line 315: All `AbstractControl` read and write their own `spatial` to which they belong in read/write methods. The EnvironmentProbeControl.java class is missing this detail. Perhaps it should be added.

```java
    @Override
        oc.write(enabled, "enabled", true);
        oc.write(spatial, "spatial", null);
    }
```

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,339 @@
...
+        oc.write(envMapSize, "size", 256);
+        oc.write(serializable, "serializable", false);
+        oc.write(bakeNeeded, "bakeNeeded", true);
+        oc.write(frustumFar, "frustumFar", 1000f);
+        oc.write(frustumNear, "frustumNear", 0.001f);
+        oc.write(uuid, "envProbeControlUUID", "none");
+    }
+
+    @Override
+    public void read(JmeImporter im) throws IOException {
```
capdevon, line 327: All `AbstractControl` read and write their own `spatial` to which they belong in read/write methods. The EnvironmentProbeControl.java class is missing this detail. Perhaps it should be added.

```java
    @Override
        InputCapsule ic = im.getCapsule(this);
        enabled = ic.readBoolean("enabled", true);
        spatial = (Spatial) ic.readSavable("spatial", null);
    }
```

capdevon:
I would suggest adding to the `EnvironmentProbeControl.java` class also the `enabled` variable and all the functionality of the `AbstractControl` class lost with the implementation of the `Control` class.

* getSpatial()
* setEnabled()
* isEnabled()
* write/read `spatial` and `enabled` fields.
* execution of the` update()` and `render()` methods only if the Controller is enabled.

```java
public abstract class AbstractControl implements Control, JmeCloneable {

    protected boolean enabled = true;
    protected Spatial spatial;

    public AbstractControl(){
    }

    @Override
        if (this.spatial != null && spatial != null && spatial != this.spatial) {
            throw new IllegalStateException("This control has already been added to a Spatial");
        }
        this.spatial = spatial;
    }

    public Spatial getSpatial(){
        return spatial;
    }

    public void setEnabled(boolean enabled) {
        this.enabled = enabled;
    }

    public boolean isEnabled() {
        return enabled;
    }

    /**
     * To be implemented in subclass.
     *
     * @param tpf time per frame (in seconds)
     */
    protected abstract void controlUpdate(float tpf);

    /**
     * To be implemented in subclass.
     *
     * @param rm the RenderManager rendering the controlled Spatial (not null)
     * @param vp the ViewPort being rendered (not null)
     */
    protected abstract void controlRender(RenderManager rm, ViewPort vp);

    @Deprecated
    @Override
    public Control cloneForSpatial(Spatial spatial) {
        throw new UnsupportedOperationException();
    }

    @Override
    public Object jmeClone() {
        try {
            return super.clone();
        } catch (CloneNotSupportedException e) {
            throw new RuntimeException("Can't clone control for spatial", e);
        }
    }

    @Override
    public void cloneFields(Cloner cloner, Object original) {
        this.spatial = cloner.clone(spatial);
    }

    @Override
    public void update(float tpf) {
        if (!enabled)
            return;

        controlUpdate(tpf);
    }

    @Override
    public void render(RenderManager rm, ViewPort vp) {
        if (!enabled)
            return;

        controlRender(rm, vp);
    }

    @Override
        oc.write(enabled, "enabled", true);
        oc.write(spatial, "spatial", null);
    }

    @Override
        InputCapsule ic = im.getCapsule(this);
        enabled = ic.readBoolean("enabled", true);
        spatial = (Spatial) ic.readSavable("spatial", null);
    }
}
```

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
riccardobl: Done, I renamed the variable, as 'set/isSerializable' is misleading since the control is serializable anyway. It will simply rebuild the environment on load if the results are not savable.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,339 @@
...
+     * Return true if savable results are required by this control.
+     * 
+     * @return true if savable results are required.
+     */
+    public boolean isRequiredSavableResults() {
+        return serializable;
+    }
+
+    @Override
+    public void setSpatial(Spatial spatial) {
```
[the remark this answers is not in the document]
riccardobl: fixed

jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
riccardobl: Done :+1:

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
riccardobl: fixed

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     * Set the asset manager used to load the shaders needed for the baking
+     * 
+     * @param assetManager
+     */
+    public void setAssetManager(AssetManager assetManager) {
+        this.assetManager = assetManager;
+    }
+
+    void rebakeNow(RenderManager renderManager) {
+        if (assetManager == null) {
```
capdevon, line 291: This check is not necessary; the `AssetManager` should always be not null here. I would suggest removing this whole conditional if block.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+    public float getFrustumFar() {
+        return frustumFar;
+    }
+
+    /**
+     * Set the asset manager used to load the shaders needed for the baking
+     * 
+     * @param assetManager
+     */
+    public void setAssetManager(AssetManager assetManager) {
```
capdevon, line 286: The `setAssetManager()` method is not necessary, since it is a mandatory parameter of the public constructor 
```java
public EnvironmentProbeControl(AssetManager assetManager, int size) {}
```
I would suggest removing it.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Create a new environment probe control.
+     * 
+     * @param assetManager
+     *            the asset manager used to load the shaders needed for the
+     *            baking
+     * @param size
+     *            the size of side of the resulting cube map (eg. 1024)
+     */
+    public EnvironmentProbeControl(AssetManager assetManager, int size) {
```
capdevon, line 112: For consistency, move the light `AreaType` & `AreaRadius` configuration instructions to the **protected** constructor to have them in one place please.

```java
protected EnvironmentProbeControl() {
    this.uuid = System.currentTimeMillis() + "_" + instanceCounter.getAndIncrement();
    this.setAreaType(AreaType.Spherical);
    this.getArea().setRadius(Float.MAX_VALUE);
}
```

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    void rebakeNow(RenderManager renderManager) {
+        if (assetManager == null) {
+            LOG.log(Level.SEVERE,
+                    "AssetManager is null, cannot bake environment. Please use setAssetManager() to set it.");
+            return;
+        }
+        IBLHybridEnvBakerLight baker;
+        if (!USE_GL_IR) {
```
capdevon, line 297: Because the variable `USE_GL_IR` is **private**, **final**, and set to **true**, the conditional `if(! USE_GL_IR )` block will never be executed. Please rename the variable USE_GL_IR -> useGLIR or something more intuitive (it is not clear exactly what this acronym means) and add related set/is methods to allow users to modify this parameter.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,283 @@
...
+        texturePulling = v;
+    }
+
+    @Override
+    public boolean isTexturePulling() {
+        return texturePulling;
+    }
+
+    public TextureCubeMap getEnvMap() {
+        return env;
```
capdevon, line 135: Please rename the `env` -> `envMap` variable in accordance with the get method signature.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+import com.jme3.texture.image.ColorSpace;
+import com.jme3.ui.Picture;
+
+/**
+ * Fully accelerated env baker for IBL that runs entirely on the GPU
+ * 
+ * @author Riccardo Balbo
+ */
+public class IBLGLEnvBaker extends GenericEnvBaker implements IBLEnvBaker {
+    private final Logger LOGGER = Logger.getLogger(IBLHybridEnvBakerLight.class.getName());
```
capdevon, line 63: please make the `LOGGER` variable **static**

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBakerLight.java
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * Fully accelerated env baker for IBL that bakes the specular map and spherical
+ * harmonics on the GPU.
+ * 
+ * This is lighter on VRAM but it is not as parallelized as IBLGLEnvBaker
+ * 
+ * @author Riccardo Balbo
+ */
+public class IBLGLEnvBakerLight extends IBLHybridEnvBakerLight {
+    public final static int NUM_SH_COEFFICIENT = 9;
```
capdevon, line 64: Reverse `final static` -> `static final` according to java language conventions, please. Make it protected or private if it is visible only in this class.
See how you did below for the LOG variable.

jme3-core/src/main/java/com/jme3/environment/baker/IBLHybridEnvBakerLight.java
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * An env baker for IBL that bakes the specular map on the GPU and uses
+ * spherical harmonics generated on the CPU for the irradiance map.
+ * 
+ * This is lighter on VRAM but uses the CPU to compute the irradiance map.
+ * 
+ * @author Riccardo Balbo
+ */
+public class IBLHybridEnvBakerLight extends GenericEnvBaker implements IBLEnvBakerLight {
+    private final Logger LOGGER = Logger.getLogger(IBLHybridEnvBakerLight.class.getName());
```
capdevon, line 64: please make the `LOGGER` variable **static**

capdevon (changes requested):

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     * Schedule a rebake of the environment map.
+     */
+    public void rebake() {
+        bakeNeeded = true;
+    }
+
+    /**
+     * Set the minimum distance to render
+     * 
+     * @param frustumNear
```
stephengold, line 250: The @param tag should not appear with an empty description.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+    private Predicate<Geometry> filter = (s) -> {
+        return s.getUserData("tags.env") != null || s.getUserData("tags.env.env" + uuid) != null;
+    };
+
+    protected EnvironmentProbeControl() {
+        uuid = System.currentTimeMillis() + "_" + instanceCounter.getAndIncrement();
+    }
+
+    /**
+     * Create a new environment probe control.
```
stephengold, line 104: The Javadoc summary fragment should not be a complete sentence.
I suggest "Create" -> "Creates"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+    public EnvironmentProbeControl(AssetManager assetManager, int size) {
+        this();
+        this.envMapSize = size;
+        this.assetManager = assetManager;
+        this.setAreaType(AreaType.Spherical);
+        this.getArea().setRadius(Float.MAX_VALUE);
+    }
+
+    /**
+     * Tag spatial as part of the environment for this EnvironmentProbeControl.
```
stephengold, line 121: The Javadoc summary fragment should not be a complete sentence.
I suggest "Tag" -> "Tags the specified"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+            for (Spatial sx : n.getChildren()) {
+                tag(sx);
+            }
+        } else if (s instanceof Geometry) {
+            s.setUserData("tags.env.env" + uuid, true);
+        }
+    }
+
+    /**
+     * Untag spatial as part of the environment for this
```
stephengold, line 139: The Javadoc summary fragment should not be a complete sentence.
I suggest "Untag" -> "Untags the specified"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+            for (Spatial sx : n.getChildren()) {
+                untag(sx);
+            }
+        } else if (s instanceof Geometry) {
+            s.setUserData("tags.env.env" + uuid, null);
+        }
+    }
+
+    /**
+     * Tag spatial as part of the environment for every EnvironmentProbeControl.
```
stephengold, line 157: The Javadoc summary fragment should not be a complete sentence.
I suggest "Tag" -> "Tags the specified"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+            for (Spatial sx : n.getChildren()) {
+                tagGlobal(sx);
+            }
+        } else if (s instanceof Geometry) {
+            s.setUserData("tags.env", true);
+        }
+    }
+
+    /**
+     * Untag spatial as part of the environment for every
```
stephengold, line 175: The Javadoc summary fragment should not be a complete sentence.
I suggest "Untag" -> "Untags the specified"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+        }
+    }
+
+    @Override
+    public Control cloneForSpatial(Spatial spatial) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Request savable results from the baking process. This will make the
```
stephengold, line 197: The Javadoc summary fragment should not be a complete sentence.
I suggest "Request" -> "Requests"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     * 
+     * @param v
+     *            true to enable (default: false)
+     */
+    public void setRequiredSavableResults(boolean v) {
+        requiredSavableResults = v;
+    }
+
+    /**
+     * Return true if savable results are required by this control.
```
stephengold, line 209: The Javadoc summary fragment should not be a complete sentence.
I suggest "Return" -> "Returns"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+    public void render(RenderManager rm, ViewPort vp) {
+        if (!isEnabled()) return;
+        if (bakeNeeded) {
+            bakeNeeded = false;
+            rebakeNow(rm);
+        }
+    }
+
+    /**
+     * Schedule a rebake of the environment map.
```
stephengold, line 241: The Javadoc summary fragment should not be a complete sentence.
I suggest "Schedule" -> "Schedules"

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     */
+    public void rebake() {
+        bakeNeeded = true;
+    }
+
+    /**
+     * Set the minimum distance to render
```
stephengold, line 248: The Javadoc summary fragment should not be a complete sentence.
I suggest "Set" -> "Sets"

Also, the Javadoc summary fragment should end with a period (".")

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     * 
+     * @param frustumNear
+     */
+    public void setFrustumNear(float frustumNear) {
+        this.frustumNear = frustumNear;
+    }
+
+    /**
+     * Set the maximum distance to render
```
stephengold, line 257: The Javadoc summary fragment should not be a complete sentence.
I suggest "Set" -> "Sets"

Also, the Javadoc summary fragment should end with a period (".")

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     */
+    public void setFrustumFar(float frustumFar) {
+        this.frustumFar = frustumFar;
+    }
+
+    /**
+     * Get the minimum distance to render
```
stephengold, line 264: The Javadoc summary fragment should not be a complete sentence.
I suggest "Get" -> "Gets"

Also, the Javadoc summary fragment should end with a period (".")

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     * 
+     * @return frustum near
+     */
+    public float getFrustumNear() {
+        return frustumNear;
+    }
+
+    /**
+     * Get the maximum distance to render
```
stephengold, line 273: The Javadoc summary fragment should not be a complete sentence.
I suggest "Get" -> "Gets"

Also, the Javadoc summary fragment should end with a period (".")

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     * 
+     * @return frustum far
+     */
+    public float getFrustumFar() {
+        return frustumFar;
+    }
+
+    /**
+     * Set the asset manager used to load the shaders needed for the baking
```
stephengold, line 282: The Javadoc summary fragment should not be a complete sentence.
I suggest "Set" -> "Sets"

Also, the Javadoc summary fragment should end with a period (".")

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
```diff
@@ -0,0 +1,363 @@
...
+     * 
+     * @param assetManager
```
stephengold, line 284: The @param tag should not appear with an empty description.

jme3-core/src/main/java/com/jme3/environment/baker/EnvBaker.java
```diff
@@ -0,0 +1,89 @@
...
+package com.jme3.environment.baker;
+
+import java.util.function.Predicate;
+import com.jme3.math.Vector3f;
+import com.jme3.scene.Geometry;
+import com.jme3.scene.Spatial;
+import com.jme3.texture.TextureCubeMap;
+
+/**
+ * An environment baker. It bakes the environment.
```
stephengold, line 42: Out of context, this description is vague. Perhaps clarify that it bakes a lighting environment?
If this is specifically for physically based rendering, mention that.

jme3-core/src/main/java/com/jme3/environment/baker/EnvBaker.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @author Riccardo Balbo
+ */
+public interface EnvBaker {
+    /**
+     * Bake the environment
```
stephengold, line 48: The Javadoc summary fragment should not be a complete sentence.

jme3-core/src/main/java/com/jme3/environment/baker/EnvBaker.java
```diff
@@ -0,0 +1,89 @@
...
+     *            The near frustum
+     * @param frustumFar
+     *            The far frustum
+     * @param filter
+     *            A filter to select which geometries to bake
+     */
+    public void bakeEnvironment(Spatial scene, Vector3f position, float frustumNear, float frustumFar, Predicate<Geometry> filter);
+
+    /**
+     * Get the environment map
```
stephengold, line 64: The Javadoc summary fragment should not be a complete sentence. Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/EnvBaker.java
```diff
@@ -0,0 +1,89 @@
...
+     * 
+     * @return The environment map
+     */
+    public TextureCubeMap getEnvMap();
+
+    /**
+     * Clean the environment baker This method should be called when the baker
```
stephengold, line 71: The Javadoc summary fragment should not be a complete sentence. Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/EnvBaker.java
```diff
@@ -0,0 +1,89 @@
...
+     * is no longer needed It will clean up all the resources
+     */
+    public void clean();
+
+    /**
+     * Set if textures should be pulled from the GPU
```
stephengold, line 77: The Javadoc summary fragment should not be a complete sentence. Also, it should end with a period (".").
Suggestion:  "Specifies whether textures should be pulled from the GPU."

jme3-core/src/main/java/com/jme3/environment/baker/EnvBaker.java
```diff
@@ -0,0 +1,89 @@
...
+     * 
+     * @param v
+     */
+    public void setTexturePulling(boolean v);
+
+    /**
+     * Get if textures should be pulled from the GPU
```
stephengold, line 84: The Javadoc summary fragment should not be a complete sentence. Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    /**
+     * Update the internal camera to face the given cubemap face
```
stephengold, line 139: The Javadoc summary fragment should not be a complete sentence. Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,283 @@
...
+
+        env.getImage().clearUpdateNeeded();
+
+        for (int i = 0; i < 6; i++) {
+            envbakers[i].dispose();
+        }
+    }
+
+    /**
+     * Starts pulling the data from the framebuffer into the texture
```
stephengold, line 215: The Javadoc summary fragment should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,283 @@
...
+     */
+    protected void startPulling() {
+        bos.clear();
+    }
+
+    /**
+     * Pulls the data from the framebuffer into the texture Nb. mipmaps must be
```
stephengold, line 222: There seems to be a period (".") missing after the word "texture".

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,283 @@
...
+     * pulled sequentially on the same faceId
+     * 
+     * @param fb
+     *            the framebuffer to pull from
+     * @param env
+     *            the texture to pull into
+     * @param faceId
+     *            id of face if cubemap or 0 otherwise
+     * @return
```
stephengold, line 231: The "@return" tag should not appear with an empty description.

jme3-core/src/main/java/com/jme3/environment/baker/GenericEnvBaker.java
```diff
@@ -0,0 +1,283 @@
...
+            } else {
+                LOG.log(Level.SEVERE, "Missing face {0}. Pulling incomplete!", i);
+            }
+        }
+        bos.clear();
+        tx.getImage().clearUpdateNeeded();
+    }
+
+    protected int limitMips(int nbMipMaps, int baseW, int baseH, RenderManager rm) {
+        if (nbMipMaps > 6) nbMipMaps = 6;
```
stephengold, line 279: Body of the `if` statement should be enclosed in braces.

jme3-core/src/main/java/com/jme3/environment/baker/IBLEnvBaker.java
```diff
@@ -0,0 +1,74 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.jme3.environment.baker;
+
+import com.jme3.texture.Texture2D;
+import com.jme3.texture.TextureCubeMap;
+
+/**
+ * An environment baker, but this one is for Imaged Base Lighting
```
stephengold, line 39: The Javadoc summary fragment should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/IBLEnvBaker.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @author Riccardo Balbo
+ */
+public interface IBLEnvBaker extends EnvBaker {
+    /**
+     * Generate the BRDF texture
```
stephengold, line 45: The Javadoc summary fragment should not be a complete sentence.
Suggestion:  "Generate" -> "Generates"
Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/IBLEnvBaker.java
```diff
@@ -0,0 +1,74 @@
...
+     * 
+     * @return The BRDF texture
+     */
+    public Texture2D genBRTF();
+
+    /**
+     * Bake the irradiance map
```
stephengold, line 52: The Javadoc summary fragment should not be a complete sentence.
Suggestion: "Bake" -> "Bakes"
Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/IBLEnvBaker.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public void bakeIrradiance();
+
+    /**
+     * Bake the specular IBL map
```
stephengold, line 57: The Javadoc summary fragment should not be a complete sentence.
Suggestion: "Bake" -> "Bakes"
Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/IBLEnvBaker.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public void bakeSpecularIBL();
+
+    /**
+     * Get the specular IBL map
```
stephengold, line 62: The Javadoc summary fragment should not be a complete sentence.
Suggestion: "Get" -> "Gets"
Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/IBLEnvBaker.java
```diff
@@ -0,0 +1,74 @@
...
+     * 
+     * @return The specular IBL map
+     */
+    public TextureCubeMap getSpecularIBL();
+
+    /**
+     * Get the irradiance map
```
stephengold, line 69: The Javadoc summary fragment should not be a complete sentence.
Suggestion: "Get" -> "Gets"
Also, it should end with a period (".").

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+    public void bakeSpecularIBL() {
+        Box boxm = new Box(1, 1, 1);
+        Geometry screen = new Geometry("BakeBox", boxm);
+
+        Material mat = new Material(assetManager, "Common/IBL/IBLKernels.j3md");
+        mat.setBoolean("UseSpecularIBL", true);
+        mat.setTexture("EnvMap", env);
+        screen.setMaterial(mat);
+
+        if (isTexturePulling()) startPulling();
```
stephengold, line 163: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+                try {
+                    LOGGER.log(Level.WARNING, "Workaround driver BUG: only one mip level available, regenerate it with higher roughness (shiny fix)");
+                    bakeSpecularIBL(0, 1f, mat, screen);
+                } catch (Exception e) {
+                    LOGGER.log(Level.FINE, "Error while recomputing mip level 0", e);
+                }
+            }
+        }
+
+        if (isTexturePulling()) endPulling(specular);
```
stephengold, line 192: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+
+        Picture screen = new Picture("BakeScreen", true);
+        screen.setWidth(1);
+        screen.setHeight(1);
+
+        FrameBuffer brtfbaker = new FrameBuffer(brtf.getImage().getWidth(), brtf.getImage().getHeight(), 1);
+        brtfbaker.setSrgb(false);
+        brtfbaker.addColorTarget(FrameBufferTarget.newTarget(brtf));
+
+        if (isTexturePulling()) startPulling();
```
stephengold, line 208: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+        screen.setMaterial(mat);
+
+        renderManager.getRenderer().setFrameBuffer(brtfbaker);
+        renderManager.setCamera(envcam, false);
+
+        screen.updateLogicalState(0);
+        screen.updateGeometricState();
+        renderManager.renderGeometry(screen);
+
+        if (isTexturePulling()) pull(brtfbaker, brtf, 0);
```
stephengold, line 223: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+
+        brtfbaker.dispose();
+
+        if (isTexturePulling()) endPulling(brtf);
```
stephengold, line 227: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+    @Override
+    public void bakeIrradiance() {
+
+        Box boxm = new Box(1, 1, 1);
+        Geometry screen = new Geometry("BakeBox", boxm);
+
+        FrameBuffer irradiancebaker = new FrameBuffer(irradiance.getImage().getWidth(), irradiance.getImage().getHeight(), 1);
+        irradiancebaker.setSrgb(false);
+
+        if (isTexturePulling()) startPulling();
```
stephengold, line 242: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+
+        for (int i = 0; i < 6; i++) irradiancebaker.addColorTarget(FrameBufferTarget.newTarget(irradiance).face(TextureCubeMap.Face.values()[i]));
```
stephengold, line 244: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLGLEnvBaker.java
```diff
@@ -0,0 +1,273 @@
...
+            mat.setInt("FaceId", i);
+
+            screen.updateLogicalState(0);
+            screen.updateGeometricState();
+
+            renderManager.setCamera(updateAndGetInternalCamera(i, irradiancebaker.getWidth(), irradiancebaker.getHeight(), Vector3f.ZERO, 1, 1000), false);
+            renderManager.getRenderer().setFrameBuffer(irradiancebaker);
+            renderManager.renderGeometry(screen);
+
+            if (isTexturePulling()) pull(irradiancebaker, irradiance, i);
```
stephengold, line 263: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLHybridEnvBakerLight.java
```diff
@@ -0,0 +1,203 @@
...
+            mat.setInt("FaceId", i);
+
+            screen.updateLogicalState(0);
+            screen.updateGeometricState();
+
+            renderManager.setCamera(updateAndGetInternalCamera(i, specularbaker.getWidth(), specularbaker.getHeight(), Vector3f.ZERO, 1, 1000), false);
+            renderManager.getRenderer().setFrameBuffer(specularbaker);
+            renderManager.renderGeometry(screen);
+
+            if (isTexturePulling()) pull(specularbaker, specular, i);
```
stephengold, line 137: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLHybridEnvBakerLight.java
```diff
@@ -0,0 +1,203 @@
...
+    public void bakeSpecularIBL() {
+        Box boxm = new Box(1, 1, 1);
+        Geometry screen = new Geometry("BakeBox", boxm);
+
+        Material mat = new Material(assetManager, "Common/IBL/IBLKernels.j3md");
+        mat.setBoolean("UseSpecularIBL", true);
+        mat.setTexture("EnvMap", env);
+        screen.setMaterial(mat);
+
+        if (isTexturePulling()) startPulling();
```
stephengold, line 155: Braces are required here.

jme3-core/src/main/java/com/jme3/environment/baker/IBLHybridEnvBakerLight.java
```diff
@@ -0,0 +1,203 @@
...
+                try {
+                    LOGGER.log(Level.WARNING, "Workaround driver BUG: only one mip level available, regenerate it with higher roughness (shiny fix)");
+                    bakeSpecularIBL(0, 1f, mat, screen);
+                } catch (Exception e) {
+                    LOGGER.log(Level.FINE, "Error while recomputing mip level 0", e);
+                }
+            }
+        }
+
+        if (isTexturePulling()) endPulling(specular);
```
stephengold, line 184: Braces are required here.

jme3-examples/src/main/java/jme3test/light/pbr/TestPBRSimple.java
```diff
@@ -0,0 +1,101 @@
...
+import com.jme3.input.ChaseCamera;
+import com.jme3.material.Material;
+import com.jme3.math.FastMath;
+import com.jme3.scene.Geometry;
+import com.jme3.scene.Spatial;
+import com.jme3.util.SkyFactory;
+import com.jme3.util.mikktspace.MikktspaceTangentGenerator;
+
+/**
+ * TestPBRSimple
```
stephengold, line 46: What does this test do? What's its purpose? How does one determine whether it passed or failed?

stephengold (changes requested):

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
riccardobl: i can be used to change the asset manager used internally by the control

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
riccardobl: removed, it was used to toggle the cpu irradiance baker

riccardobl:
Apologies, but going through all the Javadoc to insert 's' and '.' seems a bit like nitpicking at this stage. If you check any file in the JME repo, you'll notice that these 'rules' were never enforced before. 
I hope you can still appreciate this PR, since it introduces an important feature.

tonihele:
> Apologies, but going through all the Javadoc to insert 's' and '.' seems a bit like nitpicking at this stage. If you check any file in the JME repo, you'll notice that these 'rules' were never enforced before. I hope you can still appreciate this PR, since it introduces an important feature. @stephengold

I kinda agree that with Javadoc, good enough is good enough. Enforcing strict rules on Javadoc doesn't serve much purpose to me at least. If it is understandable English and gives good enough information, that would be good enough. Good enough, did I mention this term? Difficult to measure, subjective, maybe you understand what I mean.

And thanks @riccardobl for the feature and taking all the feedback maturely!

capdevon (commented):
Thank you for your patience @riccardobl . I am curious to try this new feature! Just a couple of finishing touches and I think it's good enough for me ;)

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
pspeed42: protected fields are evil.  They become part of the public API and provide no recourse for changing.
Always always always better to provide protected methods because at least then you can change implementations, provide backwards compatibility, etc..  With protected fields you are stuck forever.

stephengold:
> Apologies, but going through all the Javadoc to insert 's' and '.' seems a bit like nitpicking at this stage. If you check any file in the JME repo, you'll notice that these 'rules' were never enforced before. I hope you can still appreciate this PR, since it introduces an important feature. @stephengold

@riccardobl I agree it's nitpicking, and I do appreciate the importance of your work.

Are you implying you don't want me to review yourPRs? If so, just say so. In this instance, however, you specifically requested my review, so I felt obligated to do my best, poor though it is.

Prior to PR #1655 (November 2021) this project had no official coding style. With nothing to enforce, there was of course no enforcement. Since then, I've tried to follow and enforce our preferred style, at least on new Java files. This involves plenty of nitpicking. I'm sure I sometimes make mistakes.

If you don't like the "rules", you're in a good position to change them.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
riccardobl: JME technically supports multiple instances of AssetManager, so setAssetManager was there to allow to change the internal asset manager after the control is deserialized, but on second thought, maybe it  only adds confusion.

jme3-core/src/main/java/com/jme3/environment/FastLightProbeFactory.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * For debuging purposes only Will return a Node meant to be added to a GUI
+     * presenting the 2 cube maps in a cross pattern with all the mip maps.
+     *
+     * @param manager
+     *            the asset manager
+     * @return a debug node
+     */
+    public static Node getDebugGui(AssetManager manager, LightProbe probe) {
```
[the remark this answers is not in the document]
riccardobl: This was copied from LightProbeFactory, so while i agree with you, the purpose of this class is to migrate the code using LightProbeFactory to the new baker, so it should expose the same api

riccardobl:
> Are you implying you don't want me to review yourPRs? 

Not at all. I appreciate your dedication to the project, and I value your input. However, I'd like to suggest that we focus more on the content during the reviews, rather than getting caught up in minor styling issues or comments. I think this adjustment also helps prevent contributors from feeling discouraged due to the strict rules.

jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
capdevon: @riccardobl Ok, I suggest removing it.

jme3-core/src/main/java/com/jme3/environment/FastLightProbeFactory.java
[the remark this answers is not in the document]
capdevon: @riccardobl Ok, I don't particularly like it, but I can accept it.

scenemax3d:
> > Apologies, but going through all the Javadoc to insert 's' and '.' seems a bit like nitpicking at this stage. If you check any file in the JME repo, you'll notice that these 'rules' were never enforced before. I hope you can still appreciate this PR, since it introduces an important feature. @stephengold
> 
> @riccardobl I agree it's nitpicking, and I do appreciate the importance of your work.
> 
> Are you implying you don't want me to review yourPRs? If so, just say so. In this instance, however, you specifically requested my review, so I felt obligated to do my best, poor though it is.
> 
> Prior to PR #1655 (November 2021) this project had no official coding style. With nothing to enforce, there was of course no enforcement. Since then, I've tried to follow and enforce our preferred style, at least on new Java files. This involves plenty of nitpicking. I'm sure I sometimes make mistakes.
> 
> If you don't like the "rules", you're in a good position to change them.

@stephengold , 
Thanks for posting the detailed JavaDocs comments. I fixed all of them (I hope). Can you please re-review?  
10X

stephengold:
I'm not interested in re-reviewing this PR.

scenemax3d:
> I'm not interested in re-reviewing this PR.

You submitted changes request, therefore I asked

PR #1166, opened 2019-09-01
Add .glsl extension to be used as a replacement for glsllib. 

riccardobl:
glsllib is generally not recognized as proper glsl extension by editors.


pspeed42:
I'm ok with this but I guess it's a commentary on how crappy programmer's editors are today that this is not a one or two click change in the editor settings.

PR #1168, opened 2019-09-09
Parallel Builds

tlf30:
This allows for builds to be much faster, I always set this to true on my local builds when working with jmonkey. 

I do not see a reason to not enable parallel builds. Any thoughts?

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -39,3 +39,6 @@ POM_INCEPTION_YEAR=2009
 bintray_user=
 bintray_api_key=
 
+#Speed up builds
+org.gradle.parallel=true
+


stephengold:
Have you tested native builds with this option?

tlf30:
I have not had any issues building natives, but it would be a good idea for someone to also test it just to be sure, as I do not normally work with my own natives and am not confident enough to say that they work for sure, but I do not get any errors out of it.

Ali-RS:
I tested parallel build on `jme3-bullet-native` with Gradle 4.10 and it worked Ok.

stephengold:
Approved for integration. Leaving open for now in case anyone else has concerns.

louhy:
gradle clean build (with parallel)
1st) 1m 1s
2nd) 49s

gradle clean build (without)
1st) 1m 12s
2nd) 1m 10s

Given a 6700K was spiked 100% on all cores during the parallel (and overall CPU between 30% and 70% when off), interesting it's so close. Didn't seem to bust anything though.

riccardobl:
> Parallel builds require projects to be decoupled at execution time, i.e. tasks in different projects must not modify shared state. Read more about that topic in the User Manual before using --parallel extensively. Also be aware that Gradle versions before 4.0 could run clean and build tasks in parallel, resulting in failures. On these older versions it is best to call clean separately. 

Is this true for us?

louhy:
Dunno, but I think there's a good chance given the kind of stuff the build does? 

> Two projects are said to be decoupled if they do not directly access each other’s project model. Decoupled projects may only interact in terms of declared dependencies: project dependencies and/or task dependencies. Any other form of project interaction (i.e. by modifying another project object or by reading a value from another project object) causes the projects to be coupled. 

It goes on to say... 

> It may not be immediately apparent, but using key Gradle features like the allprojects and subprojects keywords automatically cause your projects to be coupled. 

What I don't understand is whether it's an "execution time" coupling, or before during configuration. 

Main build.gradle has: 
``` groovy
// This is applied to all sub projects
subprojects {
    if(!project.name.equals('jme3-android-examples')) {
        apply from: rootProject.file('common.gradle')
        if (!project.name.equals('jme3-testdata')) {
            apply from: rootProject.file('bintray.gradle')
        }
    } else {
        apply from: rootProject.file('common-android-app.gradle')
    }
}
```
(and quite a few other usages of the word 'subprojects'). 

https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:parallel_execution

They also say that no coupling warning is given if you've messed this up. :cold_sweat: 

I guess there would be a more significant speed difference on the CI server since it's like 15 times slower than mine was, but it seems that some significant testing would be needed.

riccardobl:
I'm not sure if it is safe to force this for everyone in the gradle properties, but i think we can try it in the CI if this makes the build faster.
There is also the --parallel arg that you can just pass to gradle cli without the need to change the gradle properties.

riccardobl:
Here: https://github.com/jMonkeyEngine/jmonkeyengine/pull/1190
the android build fails when parallel is enabled... 
https://github.com/RiccardoblSandbox/jmonkeyengine/runs/230152674

louhy:
I usually have to mess with stuff to get Android stuff working, so I didn't here. 

Wouldn't be too surprising if that were the part to break with parallel, especially given what I pasted above (if I interpret the Gradlespeak correctly, 50/50 chance on that one...).

tlf30:
@riccardobl I have been using parallel for all of my builds, none of which use android, so I am not sure if the android build has an issue with this. 

Is there anyone who does regular builds on android that can test this?

riccardobl:
So, since we cannot confirm if this is safe or not, and it had issues in the CI. I propose to close it.
People will still be able to toggle parallel build for their own builds using the --parallel arg if they chose to.

stephengold:
Sounds fine to me.

PR #1171, opened 2019-09-10
Build Android Natives (OpenAL, Vorbis, TextureLoading) on Travis

MeFisto94:
So I looked into building android natives and made them work on travis despite the jdk-11 changes which broke them.

I added another build to the matrix, which is android. That way the Android NDK is just easily available.

The header generation has been moved into jme3-android, which generates headers during the compilation which happens anyway (though now headers are generated if noone wants to build natives).

OpenJDK 10 is critical here, because the JDK Version 11 does not support Java EE Modules (XML) anymore, which are required by Android Tools.

What is left is the integration, there are two options:
1. Do it like Bullet-Native: Commit the binaries to the repository. These are already here, because when someone wants to build jme without having the Android NDK the existing binaries are just copied. The downside is that a release tag must not be an android-commit because the binaries are only updated the commit after.
2. Do it like one would do: Ensure jme3-android.jar is published to maven by this task and not by all the other envs (two linuxes, one osx).

Personally I am in for a hybrid approach: 1 + 2, usually 2 is relevant for most users but we need number 1 to allow users to build the engine (including the jme3-android) module without any issues.


MeFisto94:
3: Couldn't solve that entirely because I had a whiteline at the end and now git refuses to commit because the file only has a whitespace change.
2. has been changed.
1. will be solved when working on 4.

MeFisto94:
so 4 will be a separate pull request later, because it affects jme3-bullet(-native), as it looks like bullet-native-android doesn't contain any native code at all but seems to copy it from jme3-bullet.

.travis.yml
```diff
@@ -20,6 +20,20 @@ matrix:
...
+    dist: trusty
+    jdk: openjdk10
+    install:
+      - export JAVA_OPTS='-XX:+IgnoreUnrecognizedVMOptions --add-modules java.se.ee'
+      - echo y | sdkmanager "ndk-bundle" &> /dev/null
+    before_script:
+      - export ANDROID_NDK=$ANDROID_HOME/ndk-bundle
+    script:
+      - ./gradlew -PbuildNativeProjects=true jme3-android-native:assemble
+    after_success:
```
MeFisto94: Here we might need to think about this block still. I have not thought about how to commit / do built android natives. That is out of my scope

MeFisto94:
Thanks for the merge, just one thing: The deployment still has to be thought of. Currently it would try to deploy but I guess it depends on which OS has compiled first. It will work fine until the first tagged commit.

That means excluding the tasks from the other build matrices. Then deployment works. The second step would be to modify the upload script from bullet to also upload the android natives though, because otherwise while the artifacts are updated, everyone building the engine manually would be left of with the ancient files in this repository (prebuilt libraries)

stephengold:
I'm not sure I understand. The way I see it, deployment should happen in 2 stages, as it has for `jme3-bullet-native` in recent years:
  1. For each commit that changes a native library, a CI process builds the new `.so` and commits it to a `libs` directory in the repository, tagging it with `[ci skip]`.
  2. For each tagged commit, a CI process builds new `.jar`, `.pom`, and `.asc` files and deploys them to JFrog.

Unless commits come close together, or a tagged commit changes a native library, there should be no race between these processes. The main downside I see to this approach is that the repo grows rapidly.

MeFisto94:
Exactly. Even when a tagged commit changes a native library there isn't much of a problem because the .jar will contain the new libraries. Repo Size Growth can't be avoided without changing this completely, that's true as well.

What has to be done:
1. We have a script for that, it currently only checks for bullet-native, the two android builds have to be added there
2. For this, the regular build could exclude the native modules from being deployed, because then they get deployed from the android build. I mean: the main build excludes jme3-android-{native, bullet-native} and they get deployed by the build I added with this PR. Otherwise there would have to be yet another roundtrip with `[ci skip]`.

stephengold:
Are you working on this?

MeFisto94:
I wasn't so far but yeah I should do these changes

MeFisto94:
I am done. The problem is: we can only see if it works on the next tagged release and/or as soon as some natives change. Unless 21d0854 already commits something because things have changed.
The only thing which could be changed is that all 3 natives are updated in one commit, but that way it's more clean I think

stephengold:
Luckily I have some native change planned, along with a tagged release!

stephengold:
I just committed 30d1ecaec2ea995ddcc39cdcfca015c903a369e1 which should trigger recompilation of all the Bullet natives.

stephengold:
Hmm. The OSX/Windows native libraries got updated, but so far I haven't seen any evidence that the Android/Linux ones did. Please check.

stephengold:
The [Android build](https://travis-ci.org/jMonkeyEngine/jmonkeyengine/jobs/586240198) failed to commit the native libs because they weren't staged.

MeFisto94:
I see two oddities there. https://travis-ci.org/jMonkeyEngine/jmonkeyengine/jobs/586240198#L2291 Android Natives seem to have changed as well, why didn't they do that before? Or did I just not look closely enough?

The second one, you are right, is not staging them, however there is https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/private/upload_native.sh#L31
Why doesn't that work? Do I need a star `*` so that it picks up the subfolders? 

Also given that android-natives changed as well, I was expecting to see that output twice.

Since you are more experienced with the git command line: Do you spot some issues there?

stephengold:
Actually, since I use GUI tools for most of my gitwork, I'm unfortunately very ignorant of the git command line.

Yesterday I looked carefully at `upload_native.sh`, experimented with the `git add` command on my desktop, and didn't find the issue. I will keep looking, since I don't want to cut the alpha3 release with incorrect native libraries.

If we're both stuck, perhaps we should get others involved in the debugging.

stephengold:
I'm about to add some debugging commands to `upload_native.sh` and trigger another build.

stephengold:
I just noticed this failure in the Linux build:  https://travis-ci.org/jMonkeyEngine/jmonkeyengine/jobs/586791631#L386

As for the Android build, it looks like the `upload_native.sh` script terminated silently somewhere before `Push changes in jme3-bullet-native`. Non-zero error status from the `git commit` command?

Also, I may have figured out why the Android build is trying to commit jme3-bullet-native libraries: a race condition. `TRAVIS_COMMIT` refers to the commit that triggered the CI build, but during the `after_success` stage, this might not be the most recent commit. The `git diff-tree --name-only "$TRAVIS_COMMIT" -- jme3-bullet-native` is probably detecting changes that were committed by another CI process such as the Windows process at Appveyor. I'll research the Git command line to see if I can solve the race.

stephengold:
I've confirmed that `git commit` sets exit code 1 when there is nothing to commit. That, combined with the `set -euo pipefail` at the top of `upload_native.sh`, explains why the Android build aborted halfway through that script.

stephengold:
Using `git diff` got us closer to success, but now the pull-with-rebase fails:
https://travis-ci.org/jMonkeyEngine/jmonkeyengine/jobs/586809734#L2341

stephengold:
Android side is working. Looking at the Linux bulletjme error.

stephengold:
Resembles #915

stephengold:
I foolishly tested a .travis.yml change in the main repo. Should've done that in my own fork. Now the build is broken, but hopefully that'll be temporary.

PR #1175, opened 2019-09-13
Build android bullet-native on Travis

MeFisto94:
See #1171 for details. Note that #1171 has to be merged first (as this PR is based on it).
Also note that the deployment has to be thought of much like with #1171.
This also touches jme3-bullet-native, which should now compile a bit faster and cleaner (no intermediary files and two-pass-compilation).

Actually I wonder If I should re-open this PR once #1171 is merged into core, so this PR only contains one Commit?


stephengold:
The conflict resolution look trivial, but I'll leave you to decide whether to resolve it or create a new PR.

MeFisto94:
Odd. Now AppVeyor is run and fails for a good reason. I'll look into it.

stephengold:
Is this ready for integration?

MeFisto94:
It's ready, the deployment changes can then be done together with #1171.

Issue #1119, opened 2019-06-11
JME serialization does not work with protected/private constructors


Ali-RS:
We should fix JME serialization to work with protected/private constructors. Of course, that probably requires special dispensation in Java 11.

We just have to use reflection to grab the no-arg constructor and make setAccessible(true) or whatever.

Links to the forum page in case: 
https://hub.jmonkeyengine.org/t/jme-serialization-does-not-work-with-protected-private-constructor/41946
https://hub.jmonkeyengine.org/t/solved-need-help-with-triplanar-texturing-on-a-box/41940/24?u=ali_rs

Sailsman63:
What, in general, would a fix here look like?

The serialisation schemes that I've run across generally do require a public, no-arg constructor. Otherwise, your playing monkey-games with Reflection.

Or are you just running into cases where the class was inappropriately built with private constructors?

pspeed42:
The fix would look like playing games with reflection.  Games are already being played with reflection to access private fields, etc..

We require no-arg constructors but often it makes the classes themselves very fragile to have public no-arg constructors.  Any time you see "Provided for serialization, do not call" as a javadoc comment on a public constructor we should feel sad.... there's no reason for it.

pspeed42:
This PR came from another discussion and it's too bad none of the conversation was copied over or a link... because I also don't have the link handy anymore.

Ali-RS:
Description and links to the forum page added in the first post.

stephengold:
Many times I've been bitten by no-arg constructors that I invoked when I wasn't supposed to. Playing monkey games with reflection would be a small price to pay to avoid such errors.

Also: imho, monkey games are very much in the spirit of jMonkeyEngine.

stephengold:
Resolved in 'master' branch at 5eaf653de91210adf6263ebf99a92b09bd1cf28e

It'd be great if someone would sort out which no-arg constructors need protection and update the codebase accordingly. For instance, the no-arg constructor for `Geometry` is used in many places and works fine, but its javadoc says not to use it.

PR #1180, opened 2019-09-17
address issue #1119 (serialization with protected constructors)

stephengold:


pspeed42:
Seems like maybe the TransformTrack changes were committed by mistake.  No real code changed, though.

pspeed42:
As for the real changes, isn't it odd to set accessible to true for ALL of the constructors only to then also specifically set it for the retrieved constructor?

Unless I'm missing something:
noArgConstructor.setAccessible(true);

...was enough.  And:
AccessibleObject.setAccessible(allConstructors, true);

...is unnecessary.

stephengold:
Sorry for the confusion. I tried to re-use the fork from my previous PR, and it resulted in a mess, so I closed this PR and opened #1081. I'll address comments there.

PR #1181, opened 2019-09-17
address issue #1119 (serialization with protected/private constructor)

stephengold:


Ali-RS:
Transferred @pspeed42  comment from the previous issue (https://github.com/jMonkeyEngine/jmonkeyengine/pull/1180#issuecomment-532295427)

> As for the real changes, isn't it odd to set accessible to true for ALL of the constructors only to then also specifically set it for the retrieved constructor?
> 
> Unless I'm missing something:
> noArgConstructor.setAccessible(true);
> 
> ...was enough. And:
> AccessibleObject.setAccessible(allConstructors, true);
> 
> ...is unnecessary.

stephengold:
@pspeed42 is correct; that `setAccessible()` wasn't necessary. Thank you. I'll remove it.

pspeed42:
Well, I mean setAccessible() on the constructor you will actually use might be better than calling it on every constructor in the whole class.  It also means the find constructor method is doing what it says and not extra stuff.

jme3-core/src/main/java/com/jme3/export/SavableClassUtil.java
```diff
@@ -208,4 +223,27 @@ public static Savable fromName(String className, List<ClassLoader> loaders) thro
...
+     * Use reflection to gain access to the no-arg constructor of the named
+     * class.
+     *
+     * @return the pre-existing constructor, or null if the class lacks a no-arg
+     * constructor
+     */
+    private static Constructor findNoArgConstructor(String className)
+            throws ClassNotFoundException {
+        Class clazz = Class.forName(className);
+        Constructor[] allConstructors = clazz.getDeclaredConstructors();
```
Ali-RS: You could use `getDeclaredConstructor()` this instead of getting all constructors `getDeclaredConstructors()`

```
Constructor<?> constructor;
  try {
    constructor = type.getDeclaredConstructor();
  } catch (NoSuchMethodException e) {
    throw new Exception("Please ensure that the type '" + type.getSimpleName() + "' has a no-arg constructor.");
  }
  return constructor;
```

stephengold: OK, I'll use `getDeclaredConstructor()` instead of the for-loop.

jme3-core/src/main/java/com/jme3/export/SavableClassUtil.java
```diff
@@ -234,14 +233,11 @@ public static Savable fromName(String className, List<ClassLoader> loaders) thro
...
-        for (Constructor constructor : allConstructors) {
-            if (constructor.getParameterTypes().length == 0) {
-                result = constructor;
-                break;
-            }
+        Constructor result;
+        try {
+            result = clazz.getDeclaredConstructor();
+        } catch (NoSuchMethodException e) {
+            result = null;
```
Ali-RS: Sorry @stephengold, I mean instead of returning null result you can just throw 

```
throw new InstantiationException(
                    "Loading requires a no-arg constructor, but class "
                    + className + " lacks one.");
```

inside catch block.

stephengold: Done.

Ali-RS: Thanks :)

pspeed42:
Much better.

PR #1182, opened 2019-09-18
define 2 new Platform values for Linux ARM, register BulletJme lib paths

stephengold:
Add 2 new values to the `com.jme3.system.Platform` enum (and logic to `com.jme3.system.JmeSystemDelegate`) in order to distinguish Linux-on-ARM platforms. (LWJGL3 already supports such platforms.) Define paths to hypothetical BulletJme native libraries built for such platforms.

This is just a first step toward supporting Linux-on-ARM.


MeFisto94:
I am working on bullet compilation right now, actually it's just travis being annoying (and the fact that we use an ancient ubuntu distribution)

PR #1186, opened 2019-09-19
Build Bullet for Linux ARM (Raspberry Pi)

### LINKED ISSUE #1183: Building Bullet for Raspberry Pi
Related to #1101 
I have the build working on my fork, however I am having some problems with the CI, which is why I am opening this issue to maybe discuss further progress and/or someone willing to try a few things out.

So the problem essentially manifests in https://bugs.launchpad.net/ubuntu/+source/gcc-defaults/+bug/1300211 -> We can EITHER Compile Linux 32bit OR ARM, but not both. Oddly enough it works on my machine and in a docker container just fine, however I don't know if the produced binaries are valid then.

So, clean solutions would be:
1. Using multiple docker containers and building the natives from within travis
2. Splitting of the ARM builds into a separate artifact, as that is what is preventing me from using yet another travis instance/matrix entry.

Workarounds would be:
3. Fixing Travis, so that everything builds on it (locally and in a docker container it works, so the task here would be to spot which dependency might have been removed from travis to save space and add it, though it will be hard. Hint: `dpkg -l > file.txt` lists all installed packages into a file which would then have to be compared with the docker container (or use `dpkg -l` directly to have it print out everything).

4. Using the docker container which executes the following steps:
```
apt update && apt install git openjdk-11-jdk
git clone https://github.com/MeFisto94/jmonkeyengine && cd jmonkeyengine && git checkout feature/bullet-arm
apt install gcc-multilib g++-multilib
apt install gcc-7-arm-linux-gnueabihf gcc-7-arm-linux-gnueabi gcc-8-aarch64-linux-gnu g++-7-arm-linux-gnueabihf g++-7-arm-linux-gnueabi g++-8-aarch64-linux-gnu
./gradlew -PbuildNativeProjects=true jme3-bullet-native:assemble
```
This works as well. Note that the trick here is installing multilib  first and then the cross-compilers which uninstall multilib again but somehow the system keeps what is required to build multilib.

Why are 3 and 4 unclean? Simple: When I check my include path, I see that some 64bit headers are pulled in, which is what causes the build to work. Usually it's multilibs job to somehow manage these problems. It could be that the resulting library is 32bit but crashes because some contracts are broken.

TLDR: ATM it works, but prevents 32 bit Linux from building, so the easiest way would be to separate them

MeFisto94:
Fixes #1183 
This PR needs to have it's 32bit linux build tested and unfortunately requires sudo (for now), as the packages need to be installed in a two step progress, which is what travis wouldn't do.

I also move the native build to our more new openjdk11 environment and also updated the dist to be the newest available. This is just because otherwise the compiler versions would be 4 instead of 7. Also precise is in the process of being decomissioned and basically this PR has to be redone whenever the dist is upgraded, as the packages might change (have a higher version number).

So the second part of this PR is like: Move the build towards bionic and keep java 8 for testing / things not ready for java 11.

Also note the draft state of this PR, the binaries first have to be tested, some changes might need to be reverted (branchers: feature/bullet-arm).
Also the gradle wrapper couldn't be used locally as it lead to a NullPointerException. So we use the installed gradle here, we might consider upping the wrapper version a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,50 +14,48 @@ matrix:
     jdk: oraclejdk8
     dist: trusty
     env: UPLOAD=true UPLOAD_NATIVE=true
-    addons:
-        apt:
-            packages:
-                #- lib32gcc-7-dev
-                #- gcc-7-multilib
-                #- g++-7-multilib
-                #- gcc-7-arm-linux-gnueabihf
-                #- gcc-7-arm-linux-gnueabi
-                #- gcc-8-aarch64-linux-gnu
-                #- g++-7-arm-linux-gnueabihf
-                #- g++-7-arm-linux-gnueabi
-                #- g++-8-aarch64-linux-gnu
-                #- libc6-i386
-                #- libc6-dev-i386
-                #- libstdc++-7-dev-i386
-                #- libstdc++6-i386-cross
-                #- lib32stdc++-7-dev
-                #- libx32stdc++-7-dev
-                #- linux-libc-dev-arm64-cross
-                #- libubsan1-arm64-cross
-                #- libstdc++6-arm64-cross
-                #- libstdc++-8-dev-arm64-cross
-                #- libgcc1-arm64-cross
-                #- libgcc-8-dev-arm64-cross 
-                #- libc6-x32 
-                #- libc6-dev-arm64-cross
-                #- libatomic1-arm64-cross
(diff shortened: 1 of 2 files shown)


MeFisto94:
Gradle Wrapper works oddly enough.
So we need to discuss the move to the java 11 box and have someone test the built binaries.

Issue #1178, opened 2019-09-16
BulletAppState violates the AppState.isInitialized() contract


stephengold:
"`BulletAppState` is breaking the `isInitialized()` contract and creating a potential for weird inconsistencies in what is reported and what `AppStateManager` thinks."

https://hub.jmonkeyengine.org/t/appstate-interface-usage-poll/42265/28

stephengold:
Fixed in `master` branch at 0425c61dd4b34e44a5263cb8b7b350a3ef3d23d7 .

PR #1187, opened 2019-09-19
fix for issue #1178 (BulletAppState violates AppState contract)

stephengold:
Changed `BulletAppState` so it extends `AbstractAppState`.

### DIFF
--- a/jme3-bullet/src/common/java/com/jme3/bullet/BulletAppState.java
+++ b/jme3-bullet/src/common/java/com/jme3/bullet/BulletAppState.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2018 jMonkeyEngine
+ * Copyright (c) 2009-2019 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


pspeed42:
You asked me to review so I will jot down some thoughts. I'm not totally qualified to comment on the original intent of what the state was doing so these things may just be random things to think about.

stateAttached() is a little bit of a weird place to do anything other than debug logging.  It can technically be run from any thread and therefore the non-volatile isRunning could be stale with respect to some other thread that set it... but I'm not really clear on the lifecycle here.  Maybe it can never happen.

This is why I worried about the original abuse of 'initialized'... I'm not entirely sure the original intent was served by that variable, either.

That said, I'm fine with these changes.  It seems the worst that could happen is that some newly-failing use-case will be clarified.

However, at some point, it would be nice if someone more familiar with the BulletAppState integration take a step back and make sure it's really meeting the goals/use-cases it's trying to satisfy.  I think at best, it's a little weird now.

stephengold:
Thanks for the review, @pspeed42. With Normen gone, the original intent is difficult to reconstruct. I'll make `isRunning` volatile just to be safe.

PR #1196, opened 2019-10-04
Added new particle system

Jeddic:
This is the particle system I have been discussing on the forums and includes all the feedback i've received thus far.

The particle system includes a number of different influencers with attributes that can be animated for greater artistic control over the particle system's function. I've also added a new test file which demonstrates much of the functionality of the new particle system as well as several particle tweens for use in the new animation system.


stephengold:
I started looking at this, but didn't get very far. The first thing I noticed was that many of the added files don't have copyright notices.

Jeddic:
So based on feedback on the forums, I decided to put a pause on trying to get this into core for the time being and break it off as a separate plugin for now.

Issue #1197, opened 2019-10-07
ClassCastException in LODGeomap.createMesh()


stephengold:
In recent master branch, jme3test.bullet.TestIssue1125 crashes with the following stack trace:
```
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.ClassCastException: java.nio.DirectByteBuffer cannot be cast to java.nio.ShortBuffer
	at com.jme3.terrain.geomipmap.LODGeomap.createMesh(LODGeomap.java:89)
	at com.jme3.terrain.geomipmap.LODGeomap.createMesh(LODGeomap.java:77)
	at com.jme3.terrain.geomipmap.TerrainPatch.<init>(TerrainPatch.java:190)
	at com.jme3.terrain.geomipmap.TerrainQuad.createQuadPatch(TerrainQuad.java:709)
	at com.jme3.terrain.geomipmap.TerrainQuad.split(TerrainQuad.java:554)
	at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:223)
	at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:176)
	at com.jme3.terrain.geomipmap.TerrainQuad.<init>(TerrainQuad.java:157)
	at jme3test.bullet.TestIssue1125.addTerrain(TestIssue1125.java:122)
	at jme3test.bullet.TestIssue1125.simpleInitApp(TestIssue1125.java:110)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:237)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:130)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:211)
	at java.lang.Thread.run(Thread.java:748)
```

A quick glance at the source code suggests this has to do with PR #1136.

stephengold:
The issue doesn't occur with recent v3.2 branch, so targeting v3.3 only.

stephengold:
Fixed in `master` branch by PR #1198.

PR #1198, opened 2019-10-07
Fix issue #1197: ClassCastException in LODGeomap.createMesh()

Ali-RS:


stephengold:
Thank you. Looks good to me.

PR #1201, opened 2019-10-12
AnimComposer: added getTime() and setTime() methods

### LINKED ISSUE #1200: provide access to the time of an animation layer
I want to synchronize some anim clips with other events occuring in my app.

The old system provides `AnimChannel.getTime()` and `.setTime()` to access the time of the animation that’s playing. In the new system, time is stored in the `Layer`; there’s no `getTime()` or `setTime()`.

Public getters and setters should be added, probably to the `AnimComposer` class.

For more information, see the forum topic: https://hub.jmonkeyengine.org/t/synchronizing-an-animclip/42393

Ali-RS:
Resolves #1200

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -104,6 +104,36 @@ public class AnimComposer extends AbstractControl {
         l.time = 0;
         l.currentAction = null;
     }
+    
+    /**
+     * Returns current time of the specified layer.
+     */
+    public double getTime(String layerName) {
+        Layer l = layers.get(layerName);
+        if (l == null) {
+            throw new IllegalArgumentException("Unknown layer " + layerName);
+        }
+        return l.time;
+    }
+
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -104,6 +104,36 @@ public void removeCurrentAction(String layerName) {
...
+        }
+        if (l.currentAction == null) {
+            throw new RuntimeException("There is no action running in layer " + layerName);
+        }
+        double length = l.currentAction.getLength();
+        if (time >= 0) {
+            l.time = time % length;
+        } else {
+            l.time = time % length + length;
+        }
```
stephengold: This part looks strange to me. I think it should be `l.time = (time % length + length) % length;`

Is there some advantage to using the `if`?

pspeed42: less math and it's clearer what's going on.

PR #1209, opened 2019-11-01
Build: Fix the task 'getPrebuiltNatives'

theosotr:
Hello

This commit fixes the task `getPrebuiltNatives`.
Currently, this task is manually implemented as incremental. However, this implementation has some faults, because when the contents of one of the natives is changed, the task does not update the contents of the files properly.

So this commit fixes this task and splits it into two incremental tasks (i.e., `getNativesZipFile`, and `extractPrebuiltNatives`) that use the inputs and ouputs properties provided by gradle.

This commit removes a lot of custom logic (as it is already implemented by gradle).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -195,9 +195,12 @@ if(skipPrebuildLibraries!="true"&&buildNativeProjects!="true"){
                     String p2=srcRel.substring(j);
                     if(!p1.equals("android")&&!p2.startsWith("/native"))  srcRel=p1+"/native"+p2;
                     File dest=new File(nativesPath+File.separator+srcRel);
-                    println("Copy "+src+" "+dest);
-                    dest.getParentFile().mkdirs();
-                    Files.copy(src.toPath(), dest.toPath(), StandardCopyOption.REPLACE_EXISTING);
+                    boolean doCopy = !(dest.exists() && dest.lastModified() > src.lastModified())
+                    if (doCopy) {
+                        println("Copy "+src+" "+dest);
+                        dest.getParentFile().mkdirs();
+                        Files.copy(src.toPath(), dest.toPath(), StandardCopyOption.REPLACE_EXISTING);
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -189,29 +194,14 @@ if(skipPrebuildLibraries!="true"&&buildNativeProjects!="true"){
...
-
                     File dest=new File(nativesPath+File.separator+srcRel);
-
-                    boolean include=false;
-                    if(!dest.exists()){
-                        include=true;
-                        println("Copy "+src+" "+dest);
-                    }else if(dest.lastModified()<src.lastModified()){
-                        include=true;
-                        println("Copy "+src+" "+dest+  ". Source is newer. src "+src.lastModified()+ " dest "+dest.lastModified());
```
MeFisto94: Are you certain, that the copy mode (especially REPLACE_EXISTING) honors this behavior, that means: Can I have a locally built version of the natives which WONT be overridden then with the files from the zip?

build.gradle
```diff
@@ -168,18 +168,23 @@ if(skipPrebuildLibraries!="true"&&buildNativeProjects!="true"){
...
-                if (!target.exists()) {    
-                    println("Download natives from "+nativesUrl+" to "+nativesZipFile);
-                    target.getParentFile().mkdirs();
-                    ant.get(src: nativesUrl, dest: target);
-                }
+                println("Download natives from "+nativesUrl+" to "+nativesZipFile);
+                target.getParentFile().mkdirs();
+                ant.get(src: nativesUrl, dest: target);
+            }
+        }
```
riccardobl: This part is ok

build.gradle
```diff
@@ -189,29 +194,14 @@ if(skipPrebuildLibraries!="true"&&buildNativeProjects!="true"){
...
-                        println(""+dest+" Up to date. Skip.");
-                    }
-
-                    if(include){
-                        dest.getParentFile().mkdirs();
-                        Files.copy(src.toPath(), dest.toPath(), StandardCopyOption.REPLACE_EXISTING);
-                    }
+                    println("Copy "+src+" "+dest);
+                    dest.getParentFile().mkdirs();
+                    Files.copy(src.toPath(), dest.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
riccardobl: This part changes the behaviour, since it will always replace the files even if the local files are newer. We want this to work also for people that are compiling natives locally.

theosotr: Yes, you are right! I changed the code to respect this requirement.

riccardobl:
Merged. Thanks for the contribution :+1:

PR #1211, opened 2019-11-09
Restore -SNAPSHOT for local builds. Add custom version name param.

### LINKED ISSUE #1207: Buildscript issue
Recent cleanup in @riccardobl commit https://github.com/jMonkeyEngine/jmonkeyengine/commit/8a6bd1947cca1814e0139353ca629dd88faafc00  (on "configureVersionInfo" task)  which is part of PR #1190 makes local installing of maven artifacts from master to use the commit hash name for pom versioning instead of "-SNAPSHOT". 

This would be painful for those who are building the engine from source :(

riccardobl:
Fixes #1207


Ali-RS:
Thanks :)

PR #1218, opened 2019-11-18
Improve support for subimage copying.

riccardobl:
This pr overrides modifyTexture with a new method that allows to specify which part to copy from the source image. A workaround is implemented for gles, that doesn't provide GL_UNPACK_ROW_LENGTH.
The old method is kept for backward compatibility but deprecated in favor of the new one.


stephengold:
I don't foresee any problems with this. However, I ask that you reformat the added code to include more spaces, so that it will match the style of the pre-existing code.

riccardobl:
done

PR #1223, opened 2019-11-21
Closes #1210 Quaternion.lookAt now has a return value

### LINKED ISSUE #1210: Quaternion.lookAt should return "this"
Almost every method in the Quaternion class (from<whatever>) return this.
This allows for code like
```java
Quaternion q = new Quaternion().fromAngles(0f, FastMath.PI, 0f);
```

Quaternion.lookAt() doesn't yet, so the code would look like
```java
Quaternion q = new Quaternion();
q.lookAt(v, Vector3f.UNIT_Y)
```

Which especially in a case where you'd just want to pass a lookAt to a method, makes your code look worse. It's also not consistent with the other code (lookAt even uses fromAxes, which returns `this`), but looking at it's position in the file, it has been added later and just overlooked this convention.

Also changing this wouldn't break any user-code.

What do you think?

qs-dragon:


stephengold:
Looks good to me!

PR #1224, opened 2019-11-22
Add MavenLocal() to version.gradle

tlf30:
This is a PR to add the `MavenLocal()` command to the `version.gradle` file. 
This enables builds to be stored in the local `.m2` repo.

This is one of the first things I do when I clone jme, and I am going to guess that I am not the only one. Is there any reason this is not already (or should not be) in the `version.gradle`?

Thanks,
Trevor

### DIFF
--- a/version.gradle
+++ b/version.gradle
@@ -4,7 +4,7 @@ import org.ajoberstar.grgit.*
 
 buildscript {
     repositories {
-		mavenLocal()
+        mavenLocal()
         mavenCentral()
     }
     dependencies {
--- a/version.gradle
+++ b/version.gradle
@@ -4,6 +4,7 @@ import org.ajoberstar.grgit.*
 
 buildscript {
     repositories {
+		mavenLocal()
         mavenCentral()
     }
     dependencies {


pspeed42:
Did this break recently?  I gradle install my JME builds all the time.

oxplay2:
for me, cloning and build JME were working like month ago, so also had no issue.

riccardobl:
This pr adds mavenLocal to the repos used by the buildscript. It doesn't change the behavior of install.

tlf30:
@riccardobl by adding mavenLocal, it publishes the jme build to my `~\.m2` folder, without it, the build does not go into my .m2 folder.

tlf30:
I have always had issues with this, but it seems to be working fine now, so I am going to close this for the time being.

Ali-RS:
>by adding mavenLocal, it publishes the jme build to my ~\.m2 folder, without it, the build does not go into my .m2 folder.

As mentioned by others the proper way to add artifacts to local maven repository is by using the "install" task:
https://docs.gradle.org/current/userguide/maven_plugin.html#sec:maven_tasks

tlf30:
@Ali-RS I has having issues with that when I was making jme3 apps with maven. I had so many issues with it I actually wrote a script to insert that line into the version.gradle file for my jme builds. I am not sure what changed, but now it works without any issues. I always use the install task, not sure what I was doing wrong....

pspeed42:
mavenLocal() will never affect where JME installs its binaries.  So if it looked like that fixed it for you then it might as well have been the way you were holding your nose.  Just a random coincidence.

mavenLocal() only affects where the build looks for external non-project dependencies.  (And I don't think JME has any of these.)

PR #1228, opened 2019-11-30
protect 47 no-arg constructors in the jme3-core library

stephengold:
Now that issue #1119 is closed, protect some of the public no-arg constructors that are intended only for use by JME's serialization system (`SavableClassUtil`).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimClip.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimClip.java
@@ -16,7 +16,7 @@ public class AnimClip implements JmeCloneable, Savable {
 
     private AnimTrack[] tracks;
 
-    public AnimClip() {
+    protected AnimClip() {
     }
 
     public AnimClip(String name) {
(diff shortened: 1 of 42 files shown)


pspeed42:
Some of these I'm not so sure about.  In the cases like InstancedGeometry the only difference between the no-arg and String-arg constructor is that the geometry won't have a name.

Furthermore, things like LightNode and CameraNode, it's sometimes nice to be able to set the light and camera later after construction... but especially for CameraNode.  (Though an argument can be made for simply deprecating these classes in favor of the controls.)

stephengold:
The javadocs for `InstancedGeometry`, `LightNode`, and `CameraNode` state that the no-arg constructors should not be used.

pspeed42:
But at least in the case of InstancedGeometry, it's wrong.  The name should never be required.  I suspect that many of these comments were just cut-pasted or typed without thought.

pspeed42:
Note: it 100% does make sense to make protected the constructors that will make "bad objects"... so I'm completely on board with 90% of these changes as there is no possible way that a user could be using these correctly.  But the three cases I mention could be working fine for them... and definitely InstancedGeometry.  For example, I almost never use the name-based constructors myself because I couldn't care less about the names of my spatials.

stephengold:
I'll do a partial revert, then, publicizing those 3 constructors and removing the misleading javadocs.

PR #1230, opened 2019-12-03
re-publicize InstancedGeometry no-arg constructor, correct erroneous javadoc

stephengold:
Per post-merge discussion of PR #1228, I propose re-publicizing the no-arg constructors for `InstancedGeometry`, `LightNode`, and `CameraNode`. Also removing/correcting the javadoc for these classes (plus `Geometry` and `Node`) since the no-arg constructors are apparently NOT serialization-only.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/CameraNode.java
+++ b/jme3-core/src/main/java/com/jme3/scene/CameraNode.java
@@ -49,7 +49,10 @@ public class CameraNode extends Node {
 
     private CameraControl camControl;
 
-    public CameraNode() {
+    /**
+     * Serialization only. Do not use.
+     */
+    protected CameraNode() {
         super();
     }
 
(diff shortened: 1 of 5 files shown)


pspeed42:
Ok... so I looked into this deeper and I was wrong about CameraNode and LightNode.  The way they are written, they are totally broken if you use the no-arg constructor so it should be protected.  They internally use a control that is never created if you use the no-arg constructor.  I'd argue that we should write them to be more flexible but I still think they are useless classes.

Sorry I didn't look at the whole class before commenting on them.

stephengold:
Thanks for looking. I'll amend the PR.

pspeed42:
Thanks.  I was waiting on this to make the 3.3 branch.  :)

stephengold:
Huh. None of the other PRs seemed worth waiting for?

pspeed42:
I mean, this one was influx.  I wasn't going to cut a release until Sunday night and then couldn't find my notes on how to do it.  When I found it on Monday, this was still going back and forth so I thought I'd wait.

PR #1255, opened 2020-01-12
Fixes Bullet-Native Artifacts not containing natives when not building from cpp source and not using the build target.

MeFisto94:
So we need to depend on the assemble target, as build only happens when the project is additionally tested, which it might be for CI, but not when including as a subproject (like the SDK did) and also not when calling `gradlew run`, which one would do when executing the jme tests.

I've taken the freedom to run an auto formatter on that snippet as well to fix whitespaces, I hope you don't mind. It just looked like minified js to me ;)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -158,29 +158,29 @@ task configureAndroidNDK {
 
 gradle.rootProject.ext.set("usePrebuildNatives", buildNativeProjects!="true");
 
-if(skipPrebuildLibraries!="true"&&buildNativeProjects!="true"){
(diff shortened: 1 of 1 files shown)


MeFisto94:
Note: there is still a flaw in this build: the -sources jar also contains the built natives ALONG with the native sources.

riccardobl:
> Note: there is still a flaw in this build: the -sources jar also contains the built natives ALONG with the native sources.

Mhh, can you open an issue for that?

MeFisto94:
I've taken the freedom to self-merge this, given that the review was positive and that it's urgent as the SDK and the jme tests needed this commit to work :)

PR #1259, opened 2020-01-13
LWJGL2 early create issue

MeFisto94:
Fixes an issue where LWJGL2 was returning created too early, so that when having issues during initialization, it would return as sucessfully created.

Note: I am quite uncertain on this issue, but in my unit testing I discovered, that when trying to access OpenGL Contexts with Graphics Debug (where I know initialization should fail with a RendererException), and it was just freezing in start().

Quick debugging showed me that created.set(true) is called twice and thus the first time is too early in any case.
Also I've put it _after_ the openGL Context Init in _initContextFirstTime_.

I have yet to investigate the real case for the hang up in my case and whether this PR fixes it. Specifically I wonder why I am not notified about the exception which might be the root cause for my problem. Still this looks wrong, so please test it and if you think it's confident, merge it.

### DIFF
--- a/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java
+++ b/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java
@@ -113,7 +113,6 @@ public abstract class LwjglAbstractDisplay extends LwjglContext implements Runna
             createContext(settings);
             printContextInitInfo();
 
-            created.set(true);
             super.internalCreate();
         } catch (Exception ex){
             try {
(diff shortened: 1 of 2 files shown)


MeFisto94:
So while it is still true, that create(true) shouldn't return in that case, I figured that it is no freeze, it's just that update() is never called and the Unit Test Suite failed at propagating the handleError event from the Uncaught Exception Handler.

stephengold:
@MeFisto94 how would you like to proceed from here?

MeFisto94:
Good question, it apparently didn't cause my issue, but I _think_ it is still a bug to fix, so maybe someone of you guys can follow my logic/the callstack and confirm and merge this

PR #1262, opened 2020-01-18
Fixes #1261 - Clone the Terrain Picker, so that loading a terrain from file still works (after cloning, the picker would have the wrong terrain quad instance)

### LINKED ISSUE #1261: Terrain could not picked.
I test the problem on jme3.3 alpha5

1.First edit the terrain and then save is as j3o.
2.Use assetManager.loadModel(...terrain.j3o);
3.Use ray to test collide but not success in some terrain patch. 
		  	   
	terrain.collideWith(ray, results);
	
I think maybe is the problem of the code of TerrainQuad.

	private TerrainPicker picker = new BresenhamTerrainPicker(this);

When assetManager load the terrain. It will clone the TerrainQuad.

	if (obj instanceof CloneableSmartAsset) {
		clone = registerAndCloneSmartAsset(key, clone, proc, cache);
	}
	
But the TerrainPicker was not cloned, so the picker will link to anthor old TerrainQuad.
I use the old code can fix my pick problem. But I'm not sure is the right way.

    private int collideWithRay(Ray ray, CollisionResults results) {
        if (picker == null) {
            picker = new BresenhamTerrainPicker(this);
        }
        return picker.getTerrainIntersection(ray, results);
    }

MeFisto94:
Actually mind the comment that has been removed.
Earlier versions had the picker init in collideWith(), to bypass this problem partially, but I think this is the best solution (apart from making Picker cloneable, but it doesn't preserve "much" state anyway).

So: This would be a fast fix for 3.3 (cherry picking) and we can think about cloning later.
Actually the code has been written like that to allow for multiple terrain picker algorithms, where currently we only have Bresenham.

What would _need_ cloning, is the field `multipleCollisions` of `BresenhamTerrainPicker`, but since most cloning happens when loading, this isn't critical.

Unless you prefer that we'd be doing things right right away and make the Picker Cloneable, for this I'd need guidance, though.

### DIFF
--- a/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainQuad.java
+++ b/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainQuad.java
@@ -1793,9 +1793,8 @@ public class TerrainQuad extends Node implements Terrain {
         // This was not cloned before... I think that's a mistake.
         this.affectedAreaBBox = cloner.clone(affectedAreaBBox);
 
-        // picker is not cloneable and not cloned.  This also seems like
-        // a mistake if you ever load the same terrain twice.
-        // this.picker = cloner.clone(picker);
+        // Otherwise, picker would be cloned by reference and thus "this" would be wrong
+        this.picker = new BresenhamTerrainPicker(this);
 
         // neighbourFinder is also not cloned.  Maybe that's ok.
     }


pspeed42:
We can make it cloneable if anyone ever implements another picker type... I don't see that happening, though.  At least now they will figure it out sooner.

Issue #1013, opened 2019-02-03
Nifty GUI doesn't display correctly after context restart


ghost:
It seems JME's context restart feature is flawed because with the sample code below, the Nifty GUI refuses to display at all:
```
import com.jme3.app.SimpleApplication;
import com.jme3.material.Material;
import com.jme3.niftygui.NiftyJmeDisplay;
import com.jme3.scene.Geometry;
import com.jme3.scene.shape.Box;

import de.lessvoid.nifty.Nifty;
import de.lessvoid.nifty.builder.LayerBuilder;
import de.lessvoid.nifty.builder.PanelBuilder;
import de.lessvoid.nifty.builder.ScreenBuilder;
import de.lessvoid.nifty.controls.button.builder.ButtonBuilder;
import de.lessvoid.nifty.screen.Screen;
import de.lessvoid.nifty.screen.ScreenController;

public class NiftyJme3RestartTest extends SimpleApplication implements ScreenController {

    public static void main(String[] args) {
        new NiftyJme3RestartTest().start();
    }

    private NiftyJmeDisplay niftyDisplay;

    @Override
    public void simpleInitApp() {

        // this box here always renders
        Box b = new Box(1, 1, 1);
        Geometry geom = new Geometry("Box", b);
        Material mat = new Material(assetManager, "Common/MatDefs/Misc/Unshaded.j3md");
        mat.setTexture("ColorMap", assetManager.loadTexture("/com/jme3/app/Monkey.png"));
        geom.setMaterial(mat);
        rootNode.attachChild(geom);

        niftyDisplay = NiftyJmeDisplay.newNiftyJmeDisplay(assetManager, inputManager, audioRenderer, guiViewPort);

        Nifty nifty = niftyDisplay.getNifty();
        nifty.loadStyleFile("nifty-default-styles.xml");
        nifty.loadControlFile("nifty-default-controls.xml");

        ScreenController ctrl = this;

        new ScreenBuilder("start") {
            {
                controller(ctrl);
                layer(new LayerBuilder() {
                    {
                        childLayoutVertical();
                        panel(new PanelBuilder() {
                            {
                                childLayoutCenter();
                                width("100%");
                                height("50%");
                                backgroundColor("#ff0000");
                            }
                        });
                        control(new ButtonBuilder("RestartButton", "Restart Context") {
                            {
                                alignCenter();
                                valignCenter();
                                height("32px");
                                width("480px");
[...]

PR #1268, opened 2020-01-26
Lwjgl3 restart input handle

tonihele:
Fixes the input issue mentioned in issue #1013 . It of course happens on LWJGL3 universally. Not a Nifty thing.

The inputs are very much tied to the window context.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyInput.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyInput.java
@@ -83,11 +83,33 @@ public class GlfwKeyInput implements KeyInput {
 
     @Override
     public void initialize() {
+        if (!context.isRenderable()) {
+            return;
+        }
+        initCallbacks();
 
+        initialized = true;
+        logger.fine("Keyboard created.");
+    }
+
+    /**
+     * Re-initializes the key input context window specific callbacks
+     */
+    public void resetContext() {
         if (!context.isRenderable()) {
             return;
         }
 
+        closeCallbacks();
+        initCallbacks();
+    }
+
+    private void closeCallbacks() {
+        keyCallback.close();
+        charCallback.close();
+    }
+
+    private void initCallbacks() {
         glfwSetKeyCallback(context.getWindowHandle(), keyCallback = new GLFWKeyCallback() {
             @Override
             public void invoke(final long window, final int key, final int scancode, final int action, final int mods) {
@@ -122,9 +144,6 @@ public class GlfwKeyInput implements KeyInput {
                 keyInputEvents.add(released);
             }
         });
-
-        initialized = true;
-        logger.fine("Keyboard created.");
     }
 
     public int getKeyCount() {
@@ -149,8 +168,7 @@ public class GlfwKeyInput implements KeyInput {
             return;
         }
 
-        keyCallback.close();
-        charCallback.close();
+        closeCallbacks();
         logger.fine("Keyboard destroyed.");
     }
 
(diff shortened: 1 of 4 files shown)


tonihele:
Anything needed for this? Clarifications, better coding..? :)

stephengold:
As far as I'm concerned, all this needs is a reviewer, someone (other than you) who understands LWJGL.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -295,6 +298,11 @@ public void invoke(final long window, final boolean focus) {
         showWindow();
 
         allowSwapBuffers = settings.isSwapBuffers();
+
+        // Create OpenCL
+        if (settings.isOpenCLSupport()) {
+            initOpenCL(window);
+        }
```
MeFisto94: Any specific reason why this should now only happen when restarting and not as part of the general init process?  Wouldn't this prevent openCL init when regularly starting?

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -562,6 +578,7 @@ protected void deinitInThread() {
 
         destroyContext();
         super.internalDestroy();
+        glfwTerminate();
```
MeFisto94: I just wonder here if we can't/shouldn't arrange it that the restart could call deinitInThread() and initInThread instead of manually re-creating the context.

I think all that would be required is moving the glfwTerminate post deinitInThread, right?
I don't know if that works by the API.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
tonihele: It is part of the general init process. This is called in createContext which in turn is used by the restart and the initial context.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
tonihele: Well, I tried it just now. Seems to actually fix #1013 . But also causes some side effects. Cursor is somehow gone as well as some other elements and all inputs... Need to investigate this a bit more...

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyInput.java
```diff
@@ -83,11 +83,33 @@ public GlfwKeyInput(final LwjglWindow context) {
 
     @Override
     public void initialize() {
+        if (!context.isRenderable()) {
+            return;
+        }
+        initCallbacks();
+
+        initialized = true;
```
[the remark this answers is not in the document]
tonihele: I use the flag to see if I need to re-create the callbacks. I'm unsure what the flag otherwise symbolizes, it was there already.

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
```diff
@@ -176,7 +176,30 @@ private void onMouseButton(final long window, final int button, final int action
...
     public void initialize() {
+        initCallbacks();
 
+        if (listener != null) {
+            sendFirstMouseEvent();
+        }
+
+        setCursorVisible(cursorVisible);
+        logger.fine("Mouse created.");
+        initialized = true;
```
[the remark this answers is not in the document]
tonihele: I guess this all just mimics LWJGL 2. It doesn't seem to send either after restart.

PR #1269, opened 2020-01-26
Issue 801

### LINKED ISSUE #801: changing BitsPerPixel in AppSettings interferes with StatsAppState
In windowed mode, my attempts to change the display's bits per pixel cause the StatsAppState bitmap text to disappear. But some geometries (such as the blue cube and the darkeners) are unaffected. I see this issue both with JME 3.1 and 3.2 . Interestingly, I don't see it in full-screen mode.

Here is my test app:
```
package mygame;

import com.jme3.input.KeyInput;
import com.jme3.input.controls.ActionListener;
import com.jme3.input.controls.KeyTrigger;

        AppSettings initialSettings = new AppSettings(true);
        initialSettings.setBitsPerPixel(24);

        Main app = new Main();
        app.setSettings(initialSettings);
        app.start();
    }

    @Override
    public void simpleInitApp() {
        viewPort.setBackgroundColor(new ColorRGBA(0.3f, 0.3f, 0.3f, 1f));

        Box b = new Box(1, 1, 1);
        Geometry geom = new Geometry("Box", b);

        Material mat = new Material(assetManager, "Common/MatDefs/Misc/Unshaded.j3md");
        mat.setColor("Color", ColorRGBA.Blue);
        geom.setMaterial(mat);

        rootNode.attachChild(geom);
        inputManager.addMapping("changeBpp", new KeyTrigger(KeyInput.KEY_P));
        ActionListener listener = new ActionListener() {
            @Override
            public void onAction(String name, boolean keyPressed, float tpf) {
                if (name.equals("changeBpp") && keyPressed) {
                    goWindowed();
                }
            }
        };
        inputManager.addListener(listener, "changeBpp");
    }

    void goWindowed() {
        AppSettings newSettings = new AppSettings(false);
        newSettings.copyFrom(settings);
        newSettings.setBitsPerPixel(16);

        setSettings(newSettings);
        restart();
    }
}
```

tonihele:
This fixes #801. Fix is like suggested in the forum post referred in the issue. Fixes for both LWJGL 2 & 3. Basically now all the initial values are in the init function so the same are always used. That seems to do the trick.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/RenderContext.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/RenderContext.java
@@ -32,7 +32,6 @@
 package com.jme3.renderer;
 
 import com.jme3.material.RenderState;
-import com.jme3.material.RenderState.BlendFunc;
 import com.jme3.math.ColorRGBA;
 import com.jme3.scene.Mesh;
 import com.jme3.scene.VertexBuffer;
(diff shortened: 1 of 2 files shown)


stephengold:
I tested this and it worked for me. Sweet!
I'll leave this PR for a day or 2 before integrating, in case anyone else wishes to review it.

stephengold:
A possibly related issue was reported at the Forum:

    https://hub.jmonkeyengine.org/t/guinodes-z-order-is-reversed-in-canvas-mode-statsview-failed-to-display/42914

Ali-RS:
This issue was still happening for me on Linux. Running TestIssue801, after pressing the P key on the keyboard it shows a blank window.

Fixed in #1988

PR #1271, opened 2020-01-27
Added Override annotations, finals, some cleanup and better null checks

tonihele:
One file was reformatted completely, sorry about that. It was so weirdly indented. Better to ignore white spaces.

Added Override annotations, finals, some cleanup and better null checks (getCurrentScreen in couple of files caused a warning).

### DIFF
--- a/jme3-niftygui/src/main/java/com/jme3/cinematic/events/GuiEvent.java
+++ b/jme3-niftygui/src/main/java/com/jme3/cinematic/events/GuiEvent.java
@@ -37,6 +37,7 @@ import com.jme3.export.JmeExporter;
 import com.jme3.export.JmeImporter;
 import com.jme3.export.OutputCapsule;
 import de.lessvoid.nifty.Nifty;
+import de.lessvoid.nifty.screen.Screen;
 import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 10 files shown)


stephengold:
Why did you rename the logger?

tonihele:
I can change them back if you like. JME uses both, actually more the lowercase one. But the uppercase one is correct. https://www.oracle.com/technetwork/articles/javaee/codeconventions-135099.html. I don't exactly know do you follow the standards on this. But since I saw both, I thought that it would be ok.

stephengold:
A logger is not a constant. Please change the name back to the lowercase form.

tonihele:
I'll change them back when I get home.

You are right that it is not a constant. But I still think that it should be uppercase.

I'll also try to investigate can I add suppression to IDE warning to the project so that others wouldn't be so compelled to change those to upper case. This being Gradle project... but there seems to be now already some Netbeans specific configurations.

pspeed42:
LOGGER = ugly code and extra typing
log = pretty code and less typing

tonihele:
There you go fine sirs.

At least for Netbeans there didn't seem to be any project specific muting of the warnings.

stephengold:
Thank you for your contribution to JMonkeyEngine.

My copy of the SDK (which based on NetBeans) is configured to ignore constant naming conventions, both in the editor hints and in the inspector. I'm not 100% certain how I configured the editor. Probably it was a matter of clicking on the yellow lightbulb and selecting "disable hint" in the popup menu.

If you have a custom inspector configuration, disable the "Constant naming conventions" checkbox under the "Naming" category of NetBeans Java Hints.

PR #1274, opened 2020-01-30
jme3-blender: remove TextureGeneratorWood and its dependencies

stephengold:
The source code for `TextureGeneratorWood`  indicates that its license is the GNU General Public License. Compliance with the license would compel JME and software derived from JME to offer the same license. Some game developers would find that unacceptable.

The source code lists "Marcin Roguski (Kaelthas)" as the author. On 27 January 2020, I attempted to communicate with Kaelthas (using both e-mail and the JME Forum) to ask why the file was GPL'd. I've received no response.

jme-blender is no longer actively maintained, and even when it was, I believe the texture generator was rarely used. Removing the contaminated source code seems a prudent course of action.


pspeed42:
Fair enough to just remove the file.  It should be a high priority at this point to cut this project loose into jme-contrib.  They can resurrect the file then if they want to and change the project license or whatever.

Once I get 3.3 finally out the door and am willing to go back to operating on 'master', I may do this myself if it hasn't already been done.  I think there may be a way to do it any keep history, too... but I'm not going to go through too much extra effort for that.

MeFisto94:
Wouldn't that fall under the library clause?
Also maybe one could say that this code in theory has been donated to the project and as such falls under the BSD License, since that is our "requriement".
I wonder what the file actually does but yeah, I second moving this into a dedicated repository or something.

I planned to look into making it 2.8 compatible, though, if the effort is low, but that has nothing to do with where the repo lives.

stephengold:
Since the text of the license isn't provided, it's unclear whether the license has a library clause or not.

Also, the file doesn't describe what changes were made in the source code when it was (apparently) translated from C to Java. That would be another license violation.

PR #1276, opened 2020-02-02
Fixes #1272 - Improve GLDebugDesktop by reducing manual overrides

### LINKED ISSUE #1272: RendererException: Image format 'ABGR8' is unsupported
As I mentioned at PR #1256, `HelloJME3` with `setGraphicsDebug(true)` crashes quickly on my Linux:

```
Jan 26, 2020 11:58:09 PM com.jme3.app.LegacyApplication handleError
com.jme3.renderer.RendererException: Image format 'ABGR8' is unsupported by the video hardware.
	at com.jme3.renderer.opengl.TextureUtil.getImageFormatWithError(TextureUtil.java:102)
	at com.jme3.renderer.opengl.TextureUtil.uploadTexture(TextureUtil.java:252)
	at com.jme3.renderer.opengl.GLRenderer.updateTexImageData(GLRenderer.java:2498)
	at com.jme3.renderer.opengl.GLRenderer.setTexture(GLRenderer.java:2536)
	at com.jme3.material.Material.updateShaderMaterialParameters(Material.java:845)
	at com.jme3.material.Material.render(Material.java:1020)
	at com.jme3.renderer.RenderManager.renderGeometry(RenderManager.java:614)
	at com.jme3.renderer.queue.RenderQueue.renderGeometryList(RenderQueue.java:266)
	at com.jme3.renderer.queue.RenderQueue.renderQueue(RenderQueue.java:302)
	at com.jme3.renderer.RenderManager.renderViewPortQueues(RenderManager.java:905)
	at com.jme3.renderer.RenderManager.flushQueue(RenderManager.java:779)
	at com.jme3.renderer.RenderManager.renderViewPort(RenderManager.java:1108)
	at com.jme3.renderer.RenderManager.render(RenderManager.java:1166)
	at com.jme3.app.SimpleApplication.update(SimpleApplication.java:272)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.runLoop(LwjglAbstractDisplay.java:151)
	at com.jme3.system.lwjgl.LwjglDisplay.runLoop(LwjglDisplay.java:197)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:232)
```

MeFisto94:
Use a reflection-based approach to call checkError() after every call to the openGL API to reduce Code Duplication and increase Maintainability, while also fixing the regression caused by GLDebugDesktop extending from GLDebugES and thus making the Renderer think it is on mobile.

As discussed, if I should make this version more dynamic to support GLFbo and GLExt as separate GLDebugDesktop instances, I can do that, but having the if statement probably isn't _that_ bad, but I can see that it is better and more flexible and especially gets rid of the exception in line 73


MeFisto94:
I forgot to add the changes to LWJGL3, will force-push shortly.

stephengold:
Before we can integrate this, we need to upgrade the NDK version.

MeFisto94:
Do you talk about the build failure stemming from #1281 ?
If so I can rebase this to master, then the Android build will succeed.

stephengold:
I was assuming the build failure was something more serious. Go ahead and rebase then, and we'll see.

pspeed42:
Sorry it's taken me so long to look at this and thanks for being patient.  It mostly looks good.

A concern I have is that code that used to create on gl object and then cast it to different fields is now creating three separate proxies.
            gl = (GL) GLDebug.createProxy(gl, gl, GL.class, GL2.class, GL3.class, GL4.class);
            glext = (GLExt) GLDebug.createProxy(gl, glext, GLExt.class);
            glfbo = (GLFbo) GLDebug.createProxy(gl, glfbo, GLFbo.class);

I don't really know  how/where these fields are used and haven't had time to look, but any code expecting them to be the same object (or expecting to be able to case glfbo to GL4 or whatever) is going to fail where it would have succeeded before.

Perhaps we are correct to enforce this and I guess it would only break if they turn debugging on... and then they could see their error and correct it.

Otherwise, the changes look good to me.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLDebug.java
```diff
@@ -46,4 +97,33 @@ protected void checkError() {
             throw new RendererException("An OpenGL error occurred - " + decodeError(err));
         }
     }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+        Object result = method.invoke(obj, args);
+
+        if (method.getName().equals("glGetError")) {
```
MeFisto94: @pspeed42 Actually I just stumbled over this:
Do you think it would be more wise to lookup the Method object and use equals there instead of comparing the string all the time?

pspeed42: Sure it would be slightly faster, I guess, if the method is looked up at proxy creation time.  I wasn't overly concerned about it since this is a debugging class.  

I'm ok if that's a follow-on improvement, too.

riccardobl:
What is the status for this?

MeFisto94:
I need to change the if to be Method based and think about the separate proxies first, so this is mostly a "draft"

MeFisto94:
I looked through the concern of having multiple proxies and:
In Release Mode, GLEXT and GLFbo are different classes already, so it's just that the old Debug State was mixing them up. The old Debug state still provided separate variables for this, so there is no real reason to assume the above.

That means, I'm fine with breaking code that relied on glfbo = gl = glext only on debug.
I will fix what I commited eariler this day, though.

MeFisto94:
@pspeed42 Any objections based on the fact that I throw an unchecked exception when glGetError cannot be found?

pspeed42:
IllegalStateException isn't really the right exception here.  If you want to avoid a naked RuntimeException (and that's probably a good idea), IllegalArgumentException is probably better.  ie: the caller passed a GL context that is incompatible with this debugger.

stephengold:
Fixed in v3.3 branch at 4c2c3fa3.

PR #1281, opened 2020-02-03
Fix openal-soft dead link

riccardobl:


stephengold:
Thanks for addressing the issue!

Issue #1001, opened 2019-01-08
Java source files with missing or non-standard copyright notices


stephengold:
The authors of these classes need to be contacted so that we can make sure they've actually contributed them to the jMonkeyEngine project. Then the source files should be updated to the standard JME copyright notice.

checkers.quals.*
com.jme3.anim.AnimClip
com.jme3.anim.AnimComposer
com.jme3.anim.AnimTrack
com.jme3.anim.AnimationMask
com.jme3.anim.Armature
com.jme3.anim.ArmatureMask
com.jme3.anim.Joint
com.jme3.anim.MatrixJointModelTransform
com.jme3.anim.MorphControl
com.jme3.anim.SeparateJointModelTransform
com.jme3.anim.Weights
com.jme3.anim.interpolator.*
com.jme3.anim.tween.*
com.jme3.anim.tween.action.*
com.jme3.anim.util.*
com.jme3.app.DetailedProfiler
com.jme3.app.DetailedProfilerState
com.jme3.audio.AudioSource
com.jme3.audio.openal.AL

etcetera

Ali-RS:
> com.jme3.anim.AnimClip
> com.jme3.anim.AnimComposer
> com.jme3.anim.AnimTrack
> com.jme3.anim.AnimationMask
> com.jme3.anim.Armature
> com.jme3.anim.ArmatureMask
> com.jme3.anim.Joint
> com.jme3.anim.MatrixJointModelTransform
> com.jme3.anim.MorphControl
> com.jme3.anim.SeparateJointModelTransform
> com.jme3.anim.Weights
> com.jme3.anim.interpolator.*
> com.jme3.anim.tween.*
> com.jme3.anim.tween.action.*
> com.jme3.anim.util.*
> com.jme3.app.DetailedProfiler
> com.jme3.app.DetailedProfilerState

These are created by @Nehon so I believe we can updated them to standard JME copyright.

stephengold:
Many of them were created by @Nehon, but a few have SimSilica copyrights (meaning @pspeed42 created them.)

There are also files (not listed above) with other authorship. For example:
com.jme3.app.VRApplication

Ali-RS:
@pspeed42  now that jme core utilise Tweens, do you think it would be good idea to deprecate them from Lemur and put them in jme core (while keeping your copyrights on the files) ? 
Lemur already has dependency on jme core so this wont be an issue for Lemur.

pspeed42:
[...]

PR #1282, opened 2020-02-03
re-assign Simsilica LLC copyrights with Paul's permission (issue #1001)

stephengold:
Back in December, @pspeed42 gave permission to re-license the files @Nehon copied from Simsilica LLC, so we can address another tiny sliver of issue #1001.


stephengold:
Thank you!

PR #1296, opened 2020-02-17
RFC: Network bandwidth counters

tlf30:
Hello monkeys,
I have implemented a simple way to monitor network traffic from within spider monkey. 

Example usage on client:
```
@Override
        oneSec += tpf;
        if (oneSec > 1) {
            oneSec = 0;
            System.out.println("b/s rx=" + myClient.getCounters().getRx() + " tx=" + myClient.getCounters().getTx());
            myClient.getCounters().reset();
        }
```

Example on server:
```
@Override
        oneSec += tpf;
        if (oneSec > 1) {
            oneSec = 0;
            for (HostedConnection conn : clients) {
                System.out.println("b/s " + conn.getAddress() + " rx=" + conn.getCounters().getRx() + " tx=" + conn.getCounters().getTx());
                conn.getCounters().reset();
            }
        }
    }
```

I was wondering your thoughts on this implementation.


stephengold:
Looks OK, but why so many added blank lines?

pspeed42:
I always meant to go back and comment on this so I'm glad someone has.  I always thought it was like an add-on but I see that it's not.

It seems like such an invasive (and looks like not-thread-safe) change for something relatively unimportant.  And I think making it thread safe will probably affect performance for everyone even if they never look at these values.

It's also a little bit meaningless for the tx on UDP endpoints... as the packet can be dropped anywhere and never even make it to the network card.

I would have to look at this in a lot more detail to confirm the threading thing but given that sends can be from any thread and there was no thread-awareness anywhere that I saw, I'm pretty sure this won't actually work in a real application.

tlf30:
@stephengold IDEA IDE likes to insert blank lines everywhere... Probably some setting I have unintentionally set.

@pspeed42 I agree that there are some places that need to be more thread safe, for testing I have implemented this into a mmo engine I wrote that is heavily multi-threaded, and it worked as excepted. I have also tested this in the PhysicsSync project I built. 

As for UDP packets, while I agree that a packet may even be dropped before being sent, in my experience this is very unusual. I think having bandwidth for UDP is just as important as TCP, even if some packets get dropped, for many applications it is still important to know how much traffic was being sent and received. One could correlate the amount of data the client send/received to the server to determine how much data was dropped (although I doubt many would).

Is there a less invasive way to do this? Perhaps if some of the private/protected fields were made public the traffic information could be access externally.

pspeed42:
Public fields like that are the death of APIs.  Just rm -rf  jme-network right then, I guess. ;)

As to non-invasive, I think you could wrap all of your endpoints in a wrapper that calculates this and delegates calls to the real endpoints.

As to utility, you mention comparing server and client numbers.  This is only relevant for UDP as TCP will already match or the connection dropped.  And for UDP, if your protocol is not already dealing with dropped packets somehow then I doubt the tx and rx numbers are going to have much meaning.

Missing UDP is WAAAAAAAAY easier to count at the application layer.

stephengold:
This doesn't seem to be going anywhere. Closing...

PR #1298, opened 2020-02-18
Added a SpotBugs Implementation, which does not cause tasks to fail …

MeFisto94:
…and so is purely optional

So this is a controverse topic, specifically because the opinions on static analysis tools vary.
Let me outline the reason behind this change:
In the future we might have a bot comparing the spotbugs results of a PR with the "expected" results, highlighting additions or removals. This should be purely supportive, just a hint when reviewing a PR.

Since SpotBugs needs to be run on .class files and since some folks might want to run it locally, I've decided to add it as a gradle plugin, which adds the `spotbugsMain` task to the root project and all (expect native) subprojects.

One Cautious Word: I had to upgrade the android build tools, because they supplied an outdated Guava Version, which caused spotbugs to fail. I straight upgraded to the latest version released in december and compatible with our current gradle version (it requires our current gradle version as minimum).
I did _NOT_ check the android build after that, but the android compilation succeeded so at worst it's runtime errors which we might have to check in the future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,11 +5,15 @@ buildscript {
     repositories {
         google()
         jcenter()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.4'
+        classpath 'com.android.tools.build:gradle:3.5.3'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4'
         classpath 'me.tatarka:gradle-retrolambda:3.7.1'
+        classpath "com.github.spotbugs:spotbugs-gradle-plugin:3.0.0"
     }
 }
 
@@ -21,6 +25,7 @@ allprojects {
 }
 
 apply plugin: 'base'
+apply plugin: 'com.github.spotbugs'
 apply from: file('version.gradle')
 
 apply plugin: 'me.tatarka.retrolambda'
@@ -34,6 +39,22 @@ subprojects {
         }
     } else {
         apply from: rootProject.file('common-android-app.gradle')
+    }
+
+    if (!project.name.endsWith("-native") && project.name != "jme3-bullet-native-android") {
+        apply plugin: 'com.github.spotbugs'
+
+        // Currently we only warn about issues and try to fix them as we go, but those aren't mission critical.
+        spotbugs {
+            ignoreFailures = true
+        }
+
+        tasks.withType(com.github.spotbugs.SpotBugsTask) {
+            reports {
+                html.enabled = !project.hasProperty("xml-reports")
+                xml.enabled = project.hasProperty("xml-reports")
+            }
+        }
     }
 }
 


stephengold:
SpotBugs generates reports in XML. What are developers supposed to do with those reports?

In my development environment, the default application for opening XML files is XED. In XED, these files are cryptic to the point of being unreadable. Is there a better tool for browsing SpotBugs reports?

MeFisto94:
FindBugs could be configured to generate a HTML Report, but that conflicts with the CI, so I wonder if there is a command line flag to configure this behavior?

Reading the source at https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/src/main/java/com/github/spotbugs/internal/SpotBugsReportsImpl.java it looks like one needs to set the output file to `blah.html` to have an html report versus `blah.xml`.

Still looking on how to make this solveable with the command line

stephengold:
I tried viewing the XML reports in Firefox, but they're still too cryptic to be useful.

MeFisto94:
Something like
```
tasks.withType(com.github.spotbugs.SpotBugsTask) {
    reports {
      html {
        enabled = true
      }
      xml {
        enabled = false
      }
    }
  }
```
seems to work, I'll update the PR.

MeFisto94:
Should export HTML by default now.
When passing -Pxml-reports (or `-Pxml-reports=true`) then XML will be generated in favor of HTML

stephengold:
That's better.

Some of the issues reported by SpotBugs cannot be fixed without API changes. For instance, "EI_EXPOSE_REP: May expose internal representation by returning reference to mutable object". We should probably suppress those.

pspeed42:
This is the issue with static code analysis... from experience, we'll eventually suppress almost everything.

JME does some known bad practices ON PURPOSE... so we're going to get dinged for that all the time.

MeFisto94:
I guess we can easily backlist them, specifically for EL_EXPOSE_REP I consent to that.
It will probably list that existing failures/API "designs" in the logs of the XMLs and HTMLs, but that is nothing to consider, since we would only trigger differences on Pull Requests, so we can discourage bad behavior.

So at least those on purpose elements can/will be ignored.

stephengold:
Where's the mechanism trigger differences?

MeFisto94:
That mechanism isn't yet there and won't be part of that PR, for now all that happens is that `gradlew check` will inform you that SpotBugs found some issues and that you can read about them in html.
I will work on this mechanism tomorrow (basically it's only parsing XML and executing gradlew)

stephengold:
That's fine. I'd just like to see the mechanism before we integrate this PR.

MeFisto94:
Well, even without that mechanism this is useful, but here you can see an example output:
```
Great Work! jme3-jogg doesn't contain any bugs!
jme3-vr didn't contain any bugs!
Now analyzing additional bugs for jme3-core
Warning: Found new bug OBL_UNSATISFIED_OBLIGATION_EXCEPTION_EDGE => jme3tools.shadercheck.GpuAnalyzerValidator
Warning: Previously bugless module jme3-vr now has bugs!
Now analyzing solved bugs for jme3-core
Congratulations! Solved bug RV_CUSTOM_ERROR => jme3tools.shadercheck.GpuAnalyzerValidator
```

Note that: I only added jme3-vr and jme3-core and that the new bugs have been produced by me and that the output can be changed anytime (a grouping by module would be desireable, however that is non trivial).

The code for it is unfortunately written in nodeJS, ~150 LOC, I can PM you the code if you want, I'll create a public repository once I've got everything working. 

To explain the output:
jme3-jogg doesn't contain any bugs in the new revision, but this didn't change.
jme3-vr didn't contain any bugs! in an old revision.
Warning: Previously bugless module jme3-vr now has bugs! (they aren't printed here as I commented it out, since there was no "old" file, it would've printed every bug).
And the other: it found an OBL_UNSATISFIED_ ... error and I just copied that and added it as `RV_CUSTOM_ERROR`  to the "old" file.

stephengold:
Without the differ, I don't consider SpotBugs very useful. 99% of developers aren't going to read through 1,700 "bugs" in 21 files looking for ones that are worth fixing. For those 99%, all you'd be doing is slowing down their builds.

For this PR to be successful, the differ must run automatically on every push and every pull request, its output must be easy to interpret, and it must direct developers to possible issues with code that **they** modified. No extra effort, no cryptic codes. From what I've seen so far, I don't think we're there yet.

MeFisto94:
Still not complete, but a status update:
It's running on every Commit and hence looks like this.
I think only on PRs we can add inline comments.
The cryptic codes probably have to be turned into Hyperlinks into the Spotbugs Manual.

grafik

stephengold:
Exciting!

MeFisto94:
grafik
Unfortunately you cannot direct link the errors or extract an english sentence very easily (probably there is some xml file in the sourcecode?) and also Markdown isn't supported to link it, unfortunately.

It actually even works for non PRs.
Tomorrow I will see if linking to the file works or if something like `src/main` is missing (could be broken for tests then), currently I manipulated xml files and as such the commit didn't contain any change to java files and thus the links are broken anyway.

After that all that would be left is integration into this organization and probably UI considerations. Should that code be hosted at the organization then?
I would love to test this at the SDK but I can't since we don't use github actions yet.

MeFisto94:
grafik
Works, at least if we keep the `src/main/java`.
I need to fix a few more corner cases for the annotations and then we only have one more "problem":

By Design we cannot see if someone introduced another bug of an already present type in the same method or another method overload (i.e. the second missing try-with-resource). This is because we cannot rely on the sourcecode lines to stay the same, so we cannot effectively say if a bug on different lines is the same as before.

stephengold:
I can live with that last problem you pointed out.

By the way, I would prefer that we don't refer to the issues raised by SpotBugs as "bugs". In my mind, the word "bug" strongly implies incorrect behavior. Most of the issues raised by SpotBugs aren't actual bugs: they're style issues, poor coding practices, or minor defects. To me, even calling them "potential bugs" feels like a stretch. A small number (like the infinite loop you noted above) are actual bugs.

pspeed42:
There is an infinite loop somewhere?  I didn't see the note for that.

stephengold:
> Edit: SpotBugs just found this infinite loop, e.g.
> https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/renderer/opengl/GLDebugES.java#L262
> This will be fixed with my PR regarding Debugmode, but just as a heads up on why it's useful.

pspeed42:
Thanks, I was looking to see what it would look like graphically.  Didn't realize it was just an aside in another comment.

tlf30:
Any status on this?

MeFisto94:
So I FINALLY fixed this by converting the code at https://github.com/MeFisto94/spotbugs-github-integration into a github bot at https://github.com/MeFisto94/jaime-ci-bot.
I've already talked to @jayfella and we could host the bot at our store vps.

If there is no objection, I'd like to work on adding the bot tomorrow, that is:
 1. Merging this PR (it won't change the behavior, only add an additional `spotbugsMain` task, which can also be used manually for inspection)
 2. Creating a Github App. I'd call it `Jaime` or rather `Jaime CI Bot`. Objections? I guess we can always rename it later and I will add it belonging to the Organization
 3. Setting up the Bot together with Jay, it will start annotating Pull Requests right away.
 4. We need to add a github actions code to commit the "expectations file" and also commit a settings file to the repo (https://github.com/MeFisto94/jaime-ci-bot/blob/master/.github/ci-bot/config.yaml)

I've a couple of future plans, one is re-adding spotless support (which is code formatting), but we need to agree on that one and I'll also work on jme3-testing (I have a working version already, but I need to do some more explanations and maybe a few more tests before we'll merge it in).

stephengold:
There's a lot I don't understand yet, so I'll ask some dumb questions...

1. Why will the bot run on store vps instead of one of the standard GitHub Actions runners?
2. Why name the app "Jaime CI Bot"? Jaime is the name of the project mascot, but as I understand it the bot will only check PRs to the Engine libraries, which are just a piece of the project. I'd prefer a more self-explanatory name.
3. How will this change affect existing PRs?
4. What does the bot do if it doesn't find an expectations file?
5. If there's ever a need to disable the bot, what would be the easiest way to accomplish that?

MeFisto94:
Those are actually good questions and after deployment I will write an informative post on the hub (but this is only step one of three on our way to proper testing/quality assurance).

1. Because of the GH Actions Limitations, see https://github.com/MeFisto94/spotbugs-github-integration/issues/2, in theory we had this running on GH Actions already, but foreign PRs don't have the permission to add checks and secrets are also forbidden (as a security measure). While I don't agree with this policy, there is nothing we can do about it, we need to do it on our hardware where we securely store the Access Token.

2. What self-explanatory name would you imagine? "PR Bot"? I've prepended "Jaime" to give the bot a more personal identity (and would add the monkey as profile picture as well). And ultimatively it would also check PRs for the SDK and maybe others as well as providing other features we deem important

3. Existing PRs will _only_ be affected when they trigger a new check run, such as adding a commit. In theory re-requesting a check run using the UI could also work, but we're not aiming for that.

4. It will print an error on the host and continue, that means it would report _every_ bug present. That is no problem, though, because when setting up everything, we will ensure that the expectation files are committed. The expectation files should be auto-commited after every commit that changes a bug (adds or solves one).

Now that I write that, the bot will not only trigger for PRs, although that is it's main purpose. When someone is commiting directly, the bot will still analyze the commit and annotate it, much like travis. 

5. The best way would be to contact jay (or me), so we can simply shut down the bot. A faster but more disruptive approach would be to revoke the bot's token on the github organization. Chances are we'd need to re-create a new token and update the private cert.

stephengold:
>     2. What self-explanatory name would you imagine? "PR Bot"? I've prepended "Jaime" to give the bot a more personal identity (and would add the monkey as profile picture as well). And ultimatively it would also check PRs for the SDK and maybe others as well as providing other features we deem important

I'm not suggesting a specific name. There are currently 29 GitHub repos in the project. As I understand it, the bot will monitor only the "jmonkeyengine" repo. None of the names proposed so far indicate which repo the bot monitors.

MeFisto94:
Actually I might add the bot to the SDK first, to see if everything works out. But in general it could even be used for foreign projects, not only for us. 
That being said: names and descriptions can always be changed using the github web ui:
grafik
Not sure if you can see that, but: https://github.com/apps/jaime-jmonkeyengine-ci-dev that'd be the installation I've used during dev, it's probably private, though.

stephengold:
Lets give this a try. Approved!

stephengold (approved):

MeFisto94:
@stephengold One question: Due to security, we currently reject PRs based on an inclusive pattern (https://github.com/MeFisto94/jaime-ci-bot/blob/fd3ca205edc174c04059bfbe028132b3bf6d2773/src/index.js#L99), but I then figured that different projects might have different requirements (for the SDK that'd be all Ant files and properties (i.e. the netbeans download location)).

So I wonder if a blacklist wouldn't be better, i.e. as soon as any file is touched, that is NOT a `*.java` file, will be ignored. I don't know if that is secure enough, but for the 99% it should be.

Explanation: We checkout the PR and then run gradle. That means, in it's current form, someone could submit a PR with changes to build.gradle and others.

stephengold:
1. Does rejecting a PR mean that it can't be integrated?
2. What is the security issue that motivates rejecting PRs that aren't 100% java?

MeFisto94:
1. Rejecting is probably the wrong word here, the bot will just skip it's analysis and not provide any insight. It will be a blind flight for the reviewer
2. Essentially we don't operate on a "copy" of the sourcecode, but we clone the PR to jay's server, where we invoke gradle. If we let the PR modify build.gradle, the attacker can execute arbitrary code on our servers.

This is done mostly to allow us the flexibility of updating gradle, the tasks, settings etc via git, instead of hardcoding it in the bot. But that also means everyone could do it, unless we add such a protection.

stephengold:
ok

MeFisto94:
So would you recommend a whitelist or a blacklist in that case?

stephengold:
Tough call. Whitelist seems safest, blacklist more precise.

PR #1300, opened 2020-02-22
Add base classes to jme-network serializer

tlf30:
Adds ColorRGBA to the serializer. 
Adds Quaternion to the serializer.
I was surprised that neither of these were in there.

Edit: If there are others that people would like me to add in pr, I will do so.

### DIFF
--- a/jme3-networking/src/main/java/com/jme3/network/serializing/Serializer.java
+++ b/jme3-networking/src/main/java/com/jme3/network/serializing/Serializer.java
@@ -32,6 +32,7 @@
 package com.jme3.network.serializing;
 
 import com.jme3.math.ColorRGBA;
+import com.jme3.math.Quaternion;
 import com.jme3.math.Vector3f;
 import com.jme3.network.message.ChannelInfoMessage;
 import com.jme3.network.message.ClientRegistrationMessage;
@@ -123,6 +124,7 @@ public abstract class Serializer {
 
         registerClass(Vector3f.class,  new Vector3Serializer());
         registerClass(ColorRGBA.class,  new ColorRGBASerializer());
+        registerClass(Quaternion.class, new QuaternionSerializer());
 
         registerClass(Date.class,      new DateSerializer());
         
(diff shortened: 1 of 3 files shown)


stephengold:
I'm not surprised `ColorRGBA` was omitted, but I share your surprise concerning `Quaternion`. @pspeed42 care to comment?

pspeed42:
Because it's trivial to register them in user code and so they were just never added.  As it stands, Vector3f support is implemented in a bad way because it throws exceptions for null fields... but I guess there was some reason the author didn't want to use the default support.

Registering them is no big deal for us but it does silently break any clients that use a different version of JME than the server does.  I try to avoid that for "no good reason" and this didn't seem like a good enough reason.  The automatic serialization at least mitigates this issue for apps that use it.  So it's probably not as big a deal as it used to be to add them.

tlf30:
Perhaps a better solution then is to annotate many of the base classes with @Serializable so they can be registered like any other class. 
```
java.lang.IllegalArgumentException: Class is not marked @Serializable:class com.jme3.math.Quaternion
	at com.jme3.network.serializing.Serializer.registerClass(Serializer.java:254)
	at com.jme3.network.serializing.Serializer.registerClass(Serializer.java:165)
```
Otherwise users have to build their own serializers in order to register them, which I think makes it not trivial for many users who have not dug into how the networking system works.

pspeed42:
Nooooo... core should not depend on SpiderMonkey.  SpiderMonkey already depends on core.  These should be separate things.

I'm not sure what the issue is.  It's one line of user code:
Serializer.registerClass(Quaternion.class, new FieldSerializer());

tlf30:
I'm not sure it is documented anywhere to register non `@Serializable` classes with the `FieldSerializer`.
I agree, this is a solution. Perhaps it needs documented, even after all of my diggings through the serializer, I did not run across this until you pointed it out just now.

At some point I think it would be a good idea to get rid of the annotation all together, it is not doing anyone any good, especially since it can be bypassed.

pspeed42:
Probably... though every single one of my network examples does this at one point or another.  That's how I found that line of code to cut-paste.

PR #1302, opened 2020-02-24
Fix #1289 Matrix3f/Matrix4f equals() should require an exact class match

### LINKED ISSUE #1289: normalize the idiom for implementing equals()
@pspeed42 pointed out that the normal idiom for implementing `equals()` would be:
```
if( o == null || o.getClass() != getClass() ) return false;
```
since `.equals()` should require an exact class match.

Classes to be normalized include `Matrix3f` and `Matrix4f`.

louhy:


stephengold:
Looks good to me!

PR #1312, opened 2020-03-27
More informative exception for unsupported images.

tohidemyname:
When I call jme3tools.converters.ImageToAwt.DecodeParams#convert with an illgeal image format, it throws UnsupportedOperationException. So in my code, I catch this exception to handle unsupported images. However, I find that some other methods do not throw this exception when image is unsupported. For example, com.jme3.renderer.opengl.TextureUtil#getImageFormatWithError throws RendererException when image is unsupported:

`public GLImageFormat getImageFormatWithError(Format fmt, boolean isSrgb) {

        GLImageFormat glFmt = getImageFormat(fmt, isSrgb);

...

        if (glFmt == null) { 
  
          throw new RendererException("Image format '" + fmt + "' is unsupported by the video hardware.");
  
      }

        return glFmt;

    }
    ` 

Can the thrown exception of getImageFormatWithError be modified, so my code is cleaner?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 ######################################################################################
-# JME CI/CD 
+# JME CI/CD
 ######################################################################################
 # Quick overview of what is going on in this script:
 #   - Build natives for android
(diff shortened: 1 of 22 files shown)


riccardobl:
I don't understand... is this a PR? If so it has to be rebased on the current master.
If not, you should probably open an Issue

tohidemyname:
Sorry, I submit the bug to a wrong place. I have opened a new issue. I do not know how to delete this pull request. Please delete, if you can.

PR #1320, opened 2020-04-03
Fixes #1319 - LWJGL2: start(true) should not freeze when context could not be initialized

### LINKED ISSUE #1319: LWJGL2: Freeze when Context couldn't be initialized
"Testcase":
Call `app.start(true)` when the Renderer is set to OpenGL3.0, which will be impossible. Start will block.

Why?
https://github.com/jMonkeyEngine/jmonkeyengine/blob/728a05c4f3b432d2d75bde27bf0e8c731d30e8c2/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglDisplay.java#L173-L175
Init happens in another thread and waitFor(true) will wait for the createdLock to be set.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/728a05c4f3b432d2d75bde27bf0e8c731d30e8c2/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglContext.java#L443-L452

https://github.com/jMonkeyEngine/jmonkeyengine/blob/728a05c4f3b432d2d75bde27bf0e8c731d30e8c2/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java#L117-L132
^ This does not set created in every circumstance, it depends on where the exception happens.

But more importantly:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/728a05c4f3b432d2d75bde27bf0e8c731d30e8c2/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglContext.java#L422-L433
Here the createdLock is notified. What if this internalCreate is not called due to an Exception? Right! the lock is waiting infinitely.

Also note that we do set created = true from two different places at two different times.

MeFisto94:
Sorry, I couldn't resist but to format the nearby code a bit, specifically the missing braces around Display.isCreated.

Sorry for the Review Request, Paul, but together with the issue, it should just be a matter of a few minutes.

### DIFF
--- a/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java
+++ b/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java
@@ -93,15 +93,15 @@ public abstract class LwjglAbstractDisplay extends LwjglContext implements Runna
     /**
      * Does LWJGL display initialization in the OpenGL thread
      */
-    protected boolean initInThread(){
+    protected boolean initInThread() {
         try {
-            if (!JmeSystem.isLowPermissions()){
+            if (!JmeSystem.isLowPermissions()) {
                 // Enable uncaught exception handler only for current thread
                 Thread.currentThread().setUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
                     @Override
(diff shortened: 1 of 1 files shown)


stephengold:
Included in the v3.3 branch at 7e3ae5e .

PR #1321, opened 2020-04-04
Fixing PBR on <= GLSL 130:

MeFisto94:
- GLSLCompat now implements mat3(mat4) and determinant() and inverse() for mat2 and mat3
 - PBR.glslib uses this functions again instead of one-off workarounds

### DIFF
--- a/jme3-core/src/main/resources/Common/ShaderLib/GLSLCompat.glsllib
+++ b/jme3-core/src/main/resources/Common/ShaderLib/GLSLCompat.glsllib
@@ -43,9 +43,11 @@ out vec4 outFragColor;
 #endif
 
 #if __VERSION__ == 110
-mat3 mat3(mat4 m) {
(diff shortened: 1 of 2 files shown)


pspeed42:
Code looks good to me.

MeFisto94:
https://github.com/MeFisto94/test-spotbugs/runs/560300340#step:5:1972 which refers to https://github.com/MeFisto94/test-spotbugs/runs/560300340#step:5:1280
Any idea? To me it reads like it didn't expect the function name, as in I cannot use `mat3` as function return type? (Notice the column 6), but maybe it's some basic GLSL Syntax fault?

MeFisto94:
Green lights, https://github.com/MeFisto94/test-spotbugs/runs/560327969?check_suite_focus=true#step:5:1174, going to merge

stephengold:
Included in v3.3 branch at f68d889.

PR #1322, opened 2020-04-04
Fixes SSBO causing GL_INVALID_ENUM in Debug Mode

MeFisto94:
Background: https://github.com/MeFisto94/test-spotbugs/runs/560327969?check_suite_focus=true#step:5:1490

On MESA with Debug Mode or Assertions on, we get a GL_INVALID_ENUM at Line 510 (`getInteger(GL_MAX_GEOMETRY_SHADER_...)`).

I assume this is because this device supports SSBOs as an extension while still being on OpenGL 3.1. For most non-strict-real-gpus, SSBOs are only available with a much higher GL Version.

Anyway, this is my blind poke at this, I will re-run the above test, but I'd appreciate if someone finds some official source about when the Integers can be `getted`. 

Also note the `GL_ARB_geometry_shader_4`, I hope I did it right, it's not a `core extension` but an `extension` and I've read some stuff that those `_4` usually had a completely different API.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
@@ -506,12 +506,19 @@ public final class GLRenderer implements Renderer {
         if (hasExtension("GL_ARB_shader_storage_buffer_object")) {
             caps.add(Caps.ShaderStorageBufferObject);
             limits.put(Limits.ShaderStorageBufferObjectMaxBlockSize, getInteger(GL4.GL_MAX_SHADER_STORAGE_BLOCK_SIZE));
-            limits.put(Limits.ShaderStorageBufferObjectMaxComputeBlocks, getInteger(GL4.GL_MAX_COMPUTE_SHADER_STORAGE_BLOCKS));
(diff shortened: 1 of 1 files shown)


MeFisto94:
I just realized that Geometry and Compute Shaders are checked just below my code, so I can probably reorder that. 

I only wonder if the extensions are always present, even when the shader type should already be in core?

i.e. we miss a bunch of `if version >=`s there, don't we?
Also I wonder if this is a MESA only thing, because the UBO Extension should expose all those limits, even when the specific shaders aren't available. The good thing is this patch doesn't hurt other implementations.

MeFisto94:
Green light: https://github.com/MeFisto94/test-spotbugs/runs/560492126?check_suite_focus=true#step:5:1485

So @pspeed42 and @riccardobl let's talk about the specific implementation, especially given that if you expand the diff down you see that geometry shader detection is already there. When I move that up I could use Caps.contains(GeometryShader), but I think the Geometry Shader detection misses out implementations where the extension isn't present but the GL Version supports Geom Shaders already, or is this done somewhere else already?

MeFisto94:
When the test at https://github.com/MeFisto94/test-spotbugs/runs/560573233?check_suite_focus=true is green, I will merge this.

MeFisto94:
Debug Mode finally works, green lights: https://github.com/MeFisto94/test-spotbugs/runs/560573233?check_suite_focus=true#step:5:1483

stephengold:
Included in v3.3 branch at 828ca92.

PR #1324, opened 2020-04-05
Remove Bullet Physics

ItsMike54:
THIS PR IS INCOMPLETE:

To make way for Minie physics to become the default physics library of jME, I've removed all Bullet/jBullet libs from the engine.


riccardobl:
Dude. STOP.
Not only you are spamming our emails but you are also firing hundreds of builds in our CI.
Clone the repo and make ONE single commit or submit the PR for review after it is ready.
Screenshot from 2020-04-05 15-19-48

tlf30:
Along with that. While many of us want to see jBullet gone for good, native bullet is a different matter. Most of us do agree that Minie is a better option, and the goal should be to make it the official physics, but I think it would be wise to split that into two PRs, as there will be much more resistance on removing native bullet.

Lastly, but most importantly, those tests are not just tests, they are examples for people to see how to use physics in jme. The tests should not be removed, but migrated to use Minie instead. It would be a wast to throw them away when 99% of them are still perfectly valid for Minie as they are with native bullet.

oxplay2:
But why tests are removed, since **most of them work with Minie without any changes** 

........................................................

and why, just why, who decided to remove it :ear:  (because i dont feel like Paul approved)

what is going on here?

Ali-RS:
Do not worry, PR is already closed.

oxplay2:
yes, but i think because of what Ricc said:

"Clone the repo and make ONE single commit or submit the PR for review after it is ready."

pspeed42:
Yes, there are two levels here.  One, submit proper PRs if you are going to do it at all.  (And if you aren't going to locally clone/fork the repo then maybe just don't bother.)  The other is specifically whether the merits of this PR are worthy.  Which I also think is "no".

stephengold:
Note: native Bullet was removed from "master" branch on 3 March 2021.

PR #1325, opened 2020-04-05
Migrate TestSpatialAnim to the new Anim class

syluna:
Remove all deprecated class and migrate to the new animation system.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/model/anim/TestSpatialAnim.java
+++ b/jme3-examples/src/main/java/jme3test/model/anim/TestSpatialAnim.java
@@ -1,8 +1,9 @@
 package jme3test.model.anim;
 
-import com.jme3.animation.AnimControl;
-import com.jme3.animation.Animation;
-import com.jme3.animation.SpatialTrack;
+import com.jme3.anim.AnimClip;
+import com.jme3.anim.AnimComposer;
+import com.jme3.anim.AnimTrack;
+import com.jme3.anim.TransformTrack;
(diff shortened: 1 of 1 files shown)


Ali-RS:
Thanks for PR.

It probably would be better to have both. You can separate them in two methods, `testUsingOldAnimationSystem`, `testUsingNewAnimationSystem` and by default set it to use the new animation system.

oxplay2:
are you sure, that we should have Deprecated animation system tests in 3.3+?

Ali-RS:
Yes, I think until the old animation system is not removed we should keep it's tests also. When we removed it the we can remove it's tests as well. This is just my opinion, so let's see what others think.

oxplay2:
Paul and Darkchaos agreed on discord to keep only 3.3 related tests in 3.3+

reasons:
- encourage people to use new due to find all issues that left.(bug finding)
- people can anyway browse old tests

Ali-RS:
Okay, fine then.

pspeed42:
Yeah, specifically if the old tests start to disappear maybe folks reluctant to move for "some reason" will start panicking and making those reasons well known.

And from experience in my day job, if you leave bad code laying around, someone will cut and paste it into new code.  It propagates like a virus sometimes. :)

stephengold:
Not providing tests/examples for deprecated methods is fine by me.

Thank you for your contribution, @syluna!

stephengold:
Included in v3.3 branch at 68c6f71

PR #1330, opened 2020-04-06
Fixes error of loading assets

Grabli66:
It fixes error of loading assets in examples.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/water/TestSceneWater.java
+++ b/jme3-examples/src/main/java/jme3test/water/TestSceneWater.java
@@ -51,7 +51,7 @@ import java.io.File;
 public class TestSceneWater extends SimpleApplication {
 
     // set default for applets
-    private static boolean useHttp = true;
+    private static boolean useHttp = false;
 
     public static void main(String[] args) {
       
@@ -61,6 +61,11 @@ public class TestSceneWater extends SimpleApplication {
 
     @Override
     public void simpleInitApp() {
+        File file = new File("wildhouse.zip");
+        if (!file.exists()) {
+            useHttp = true;
+        }
+        
         this.flyCam.setMoveSpeed(10);
         Node mainScene=new Node();
         cam.setLocation(new Vector3f(-27.0f, 1.0f, 75.0f));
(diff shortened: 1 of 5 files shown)


stephengold:
Please explain how moving the `file.exists()` into `simpleInitApp()` fixes an error. What error? Under what circumstances does the error manifest?

Grabli66:
Static method main does not calling from the launcher. And that is why "useHttp" is always false. Because of that assets can't be loaded from the internet. And i am getting error that "wildhouse.zip" or "quake3level.zip" not found.

MeFisto94:
See for example the `TestMultipleFilters`.
I got the report on Discord and asked them to do the PR here.

Debugging showed that the Test Chooser does not call the main method, but just instantiates the Test as Application and starts that. I mean it's obvious, it cannot guess for any potential main methods etc.

The only thing we discussed was that we might want a redesign for those tests to use test-data and/or to have some caching for HTTPable resources. Also when google shuts down it's google code archive, we'd potentially loose those assets. But that's the topic of a future PR/Issue, this fixes the test because main() is not called from the TestChooser. We might want to double check every Test for that.

jme3-examples/src/main/java/jme3test/water/TestSceneWater.java
```diff
@@ -51,7 +51,7 @@
 public class TestSceneWater extends SimpleApplication {
 
     // set default for applets
-    private static boolean useHttp = true;
```
MeFisto94, line 54: However this change is interesting.
Did all other tests have http = false? Also the comment about applets is outdated and probably inappropriate anyway, so probably good to set this to false.

Actually I guess Applets are the reason we had http/zip locators.

Grabli66: Other tests have useHttp = false
It's the cause of that change.

stephengold: `TestQ3` has `useHttp = false`.
`TestMultipleFilters` has `useHttp = false`.
`TestSceneLoading` has `useHttp = false`.
`TestPostWaterLake` has `useHttp = false`.

So the answer is yes, all the other tests affected by this PR had `useHttp = false`.

stephengold:
Included in the v3.3 branch at 21cbdb5

PR #1333, opened 2020-04-07
Migrate TestOgreAnim to new anim system

syluna:
I applied some changes to have the same rendering as the original test (Add static inner class LoopAction)

### DIFF
--- a/jme3-examples/src/main/java/jme3test/model/anim/TestOgreAnim.java
+++ b/jme3-examples/src/main/java/jme3test/model/anim/TestOgreAnim.java
@@ -54,94 +54,94 @@ import com.jme3.scene.shape.Box;
 
 public class TestOgreAnim extends SimpleApplication implements ActionListener {
 
-        private AnimComposer animComposer;
-        private static LoopAction currentAction;
+    private AnimComposer animComposer;
+    private static LoopAction currentAction;
 
-        public static void main(String[] args) {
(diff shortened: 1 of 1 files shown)


stephengold:
Reformatting the code makes this PR difficult to review. Please limit your PRs to just the necessary changes.

syluna:
Sorry, I'll be careful next time

stephengold:
As it stands, you've changed the indentation from 4 spaces to 8 spaces. Please undo that also.

stephengold:
Also, you re-ordered the methods. Please move `onAction()` back to the bottom of the file where it was before.

pspeed42:
Looks like onAction() is at the bottom of the file.  There is an inner class below it now.

stephengold:
Given the inner class, the new arrangement is reasonable.

syluna:
Yes the inner class is only to keep the same result like old test (no loop anim). I can't find any way to do this.

stephengold:
You're not the first person to be puzzled by this:
    https://hub.jmonkeyengine.org/t/solved-when-animcomposer-reaches-the-end-of-an-action/41652

syluna:
Oh yes :) I didn't think about the Tweens sequence and callMethod ! Thanks Stephen

stephengold:
In order to commit to integrating a PR, I'd need to see the PR. However, your proposal sounds promising, and I don't foresee any objections.

PR #1335, opened 2020-04-07
Migrate TestOgreComplexAnim to new anim system

syluna:
Lots of code that did nothing in old test (on feet, hand). Only eft hand anim with the bone anim in simpleUpdate works. So I rework the test to do how actionSquence work and manipulate Armature and Joint with ArmatureMask apply.
So I think there is a bug in ArmatureDebugger when apply PointSize in material. In the render view all points on armature debug have same size.


pspeed42:
Hard tabs.

Absolutely no hard tabs in JME source code, please.

stephengold:
Integrated into the v3.3 branch at 1acd1d2 .

Issue #1313, opened 2020-03-28
More informative exception for unsupported images


tohidemyname:
When I call jme3tools.converters.ImageToAwt.DecodeParams#convert with an illgeal image format, it throws UnsupportedOperationException. So in my code, I catch this exception to handle unsupported images. However, I find that some other methods do not throw this exception when image is unsupported. For example, com.jme3.renderer.opengl.TextureUtil#getImageFormatWithError throws RendererException when image is unsupported:

`public GLImageFormat getImageFormatWithError(Format fmt, boolean isSrgb) {

    GLImageFormat glFmt = getImageFormat(fmt, isSrgb);
...

    if (glFmt == null) { 

      throw new RendererException("Image format '" + fmt + "' is unsupported by the video hardware.");

  }

    return glFmt;

}
` 
Can the thrown exception of getImageFormatWithError be modified, so my code is cleaner?

pspeed42:
What is the use case for calling `public GLImageFormat getImageFormatWithError(Format fmt, boolean isSrgb)' directly?

stephengold:
If there's no plausible use-case, then this issue (and PR #1337) should be closed.

PR #1337, opened 2020-04-07
Fix for #1313

Grabli66:
UnsupportedImageFormatException added.
It's inhereted from RendererException, nothing will breaks.


stephengold:
I'll re-run the checks. Integration also depends on clarification of issue #1313.

stephengold:
Closing due to no plausible use case.

PR #1338, opened 2020-04-08
Fixed an issue with AndroidNativeImageLoader which was opening stream…

Ali-RS:
… twice.

Relevent forum topic: 
https://hub.jmonkeyengine.org/t/solved-gltf-error-with-textured-mesh/43079


stephengold:
Looks good.

Ali-RS:
Thanks for the review @stephengold

stephengold:
Included in v3.3 branch at 0bdb015 .

PR #1342, opened 2020-04-10
Fixes #1341: AudioBuffer.updateData only allows direct buffers

### LINKED ISSUE #1341: AudioBuffer.update accepts any type of ByteBuffer
There is no check whether the provided buffer is a nativeBuffer(as required), probalby not easy to implement, but a hint in the javadoc would imho already be a great improvement...

https://hub.jmonkeyengine.org/t/solved-playing-audio-from-audiobuffer-in-audionode-causes-jre-to-die/43091/9

plassma:
See forum topic: 
https://hub.jmonkeyengine.org/t/solved-playing-audio-from-audiobuffer-in-audionode-causes-jre-to-die/43091/12

### DIFF
--- a/jme3-core/src/main/java/com/jme3/audio/AudioBuffer.java
+++ b/jme3-core/src/main/java/com/jme3/audio/AudioBuffer.java
@@ -31,9 +31,9 @@
  */
 package com.jme3.audio;
 
-import com.jme3.audio.AudioData.DataType;
 import com.jme3.util.BufferUtils;
 import com.jme3.util.NativeObject;
+
 import java.nio.ByteBuffer;
 
 /**
@@ -87,8 +87,12 @@ public class AudioBuffer extends AudioData {
     /**
      * Update the data in the buffer with new data.
      * @param data
+     * @throws IllegalArgumentException if the provided buffer is not a direct buffer
      */
     public void updateData(ByteBuffer data){
+        if (!data.isDirect())
+            throw new IllegalArgumentException("Currently only direct buffers are allowed");
+
         this.audioData = data;
         updateNeeded = true;
     }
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/audio/AudioBuffer.java
```diff
@@ -87,8 +87,12 @@ public String toString(){
     /**
      * Update the data in the buffer with new data.
      * @param data
+     * @throws IllegalArgumentException if the provided buffer is not a direct buffer
      */
     public void updateData(ByteBuffer data){
+        if (!data.isDirect())
```
MeFisto94, line 93: My personal opinion/style would want `{}s` even for one-line statements, but I don't know:
- What our jme desired standard would be and
- how our current code looks right now

pspeed42: Yeah, especially in this case where the line is pretty long.

Ali-RS: Yeah, `{}` is preferred.

MeFisto94:
Also for future PRs having a commit message like `Fixes #1341 AudioBuffer.update: Only allow direct buffers` or something would be perfect :)

stephengold:
It's not too late to change the commit message using 'push --force'!

jme3-core/src/main/java/com/jme3/audio/AudioBuffer.java
stephengold: I got impatient and standardized the coding style.

stephengold:
Integrated into the v3.3 branch at c85e2bc .

PR #1344, opened 2020-04-11
Using Tweens methods instead of the inner class

syluna:
Simplification using Tweens methods instead of NoLoopAction inner class


stephengold:
Looks good.

PR #1348, opened 2020-04-16
Remove Blender

ItsMike54:
Now that pspeed has moved blender importer to its own separate project; it seems fitting to remove the jme3-blender directory from the engine.
And hopefully no more spamming commits this time.


pspeed42:
I was going to do this but I was going to give someone a (small) chance to fix the JME-contrib blender loader build first... since they could still easily reference this one.  Not a big deal, though, I guess.

Off the top of me head, it looks like you missed General.cfg that includes the blender loader in the default asset manager config.  I haven't checked beyond that.

Also, kind of weird to do this as 8 commits.  Your git workflow must be really strange.

pspeed42:
Though I guess it sounds like Mefisto already fixed the blender loader build.

ItsMike54:
We can definitely wait for someone to use this directory as a reference in their work on the new contribution project. 

I did check, and actually at least in my engine fork, the General.cfg still has blender loader, so I'll fix that. And I'll keep looking for anymore blender references.

ItsMike54:
Since the Blender importer was already removed, I figure that I close this PR

PR #1352, opened 2020-04-20
Remove jBullet

ItsMike54:
As jBullet is about 10 years out of date, and jMonkey has a continually updated physics contribution, Minie, it is about time to part ways.
Farewell jBullet!


Ali-RS:
I do not think we can remove jBullet at the moment because: 
 

1. For iOS people yet need to use jBullet
2. For anyone who wants to use Java version over native version

But I vote for removing the native bullet in favor of Minie.

But first, you should ask core developers about their idea on the forum.

normen:
For 1) I was able to just throw in the bullet sources into an XCode project to get native physics. But I don't see a reason to remove jbullet, especially as jbullet/bullet backend can be easily switched without changes to the user code.

If you mean removing bullet based physics altogether in favor of Minie then thats another story but then it could simply be put in a separate project including the native and java backends.

Anyway, just posting because of the native bullet on iOS thing.

ItsMike54:
It is fine with me to move jBullet into it the jME contributions section, and if someone wants to maintain it, that's great. But since nobody is really working on it anymore, bugs sneak their way in quickly. Now the end goal I believe is to make Minie the official physics engine of jME, and that would require some discussion. But I think now that Minie supports Android, that it will have support for iOS soon, meaning that soon it will be a complete replacement for Bullet/jBullet. This PR can wait I guess until Minie does support iOS.

pspeed42:
I still prefer jbullet because it works and bullet has a show-stopping bug for me... but when I use it, I only barely use JME's boilerplate around it anyway.

But as Ali says, this is a discussion for the forum where it will get more than 4-5 sets of eyes.

oxplay2:
@normen I fully agree that this operation could be made only if Minie would become default one, and best if Minie works on iOS (because i dont know).

Minie is based on same C native lib as bullet-native, it just use more updated version and much better wrapper that allow use more features like cloth physics / DAC / etc. It also fix some old issues with bullet-native and allow for example auto-using V-HACD if i remember correctly.

Fact is that jbullet and bullet-native use same wrapper API, but Minie dont. so again, just like @normen said, it could be moved out of JME repo if Minie really gonna be new standard.

If not, then i dont see reason to remove it.

MeFisto94:
FYI I re-vived our jbullet here: https://github.com/jMonkeyEngine/jbullet
It's like with a few of my projects that just lack testing and close inspection (i.e. did something become worse or better) and in theory we could fix stuff there

ItsMike54:
We can wait, but again my point is that maybe we can move it out of core into contribution project. That way someone can pick up on it and work. 
Sgold has said for a while now that he expects Minie to at some point replace Bullet/jBullet 
image

stephengold:
1. Minie doesn't support IOS yet.
2. If the Blender importer can be decoupled from the core, then so can JBullet. I'm unconvinced that now is the time to do so, however.
3. I've tried to reproduce Paul's show-stopper bug and failed. I don't even know which version of Bullet it was observed in.

ItsMike54:
> 2\. If the Blender importer can be decoupled from the core, then so can JBullet. I'm unconvinced that now is the time to do so, however.

Thats part of my point. If Minie isn't ready yet to become the default physics engine, then we still keep Native-Bullet, and Normen has already said that it is possible to work iOS support into it. 
But I still would say remove jBullet to its own repo, @stephengold , what would prevent this from being reasonable now?

pspeed42:
Why does it matter?  What is it fixing?  Things are fine right now.  Until they aren't there is no solid reason to move it out.

There were many many solid reasons to move the blender loader out, chief among them is that it's broken and has questionable licensing.  Both were pretty serious.

ItsMike54:
I beleive its fixing the fact that it hasn't been updated in almost a decade.
And nobody on our end seems willing to work on it, and the time will come that bugs will work their way into a seeming abandoned project.

pspeed42:
But there is no issue right now and sometimes jbullet works when bullet doesn't (and vice versa).  While any bullet is still in the project then they both should still be in the project.

It lacks some features, yes.. but so does our bullet native.  The thing is that removing jbullet is not a simple thing, really.  Once it's gone, there is no reason for the wrapper layer that supported both and all of that could be collapsed and so on.  Or it just all goes away and JME becomes an "openGL scene graph engine" and not a game engine anymore.

ItsMike54:
So best to remove Bullet/jBullet together if we have an alternative good enough to allow that?
I'm fine with that, but if we are going to wait, we might as well close the PR for now.

tlf30:
@pspeed42 from the perspective of maintaining the physics wrapper, I completely agree. 
For now, it is probably best to keep both. 

Perhaps moving forward we might want to consider having an API for physics libraries to implement that lives in the engine to provide a uniform interface. But that is a conversation for the hub.

normen:
> Perhaps moving forward we might want to consider having an API for physics libraries to implement that lives in the engine to provide a uniform interface. But that is a conversation for the hub.

I really don't want to get into this discussion too much but that is basically where it ended in jME2 and it turned out to be a very bad idea, which is why we focused on bullet for jME3. Physics engines just differ too much and in the end you have a wrapper thats eating away at the performance of all engines and allows none to shine with their specific features.

ItsMike54:
Well until there is more support from the community on this decision, I say close the PR for now.

Issue #1357, opened 2020-04-21
Configure core profiles in jme3-examples


rvandoosselaer:
Setup the minimum required OpenGL core profiles for the Geometry and Tesselation shader examples.

stephengold:
Fixed in master branch at 3fa85bc and in v3.3 branch at 4b62b70 .

PR #1358, opened 2020-04-21
Set the minimum required OpenGL profiles for the geometry and tessellation shader test cases

rvandoosselaer:
See #1357

### DIFF
--- a/jme3-examples/src/main/java/jme3test/material/TestGeometryShader.java
+++ b/jme3-examples/src/main/java/jme3test/material/TestGeometryShader.java
@@ -8,6 +8,7 @@ import com.jme3.scene.Geometry;
 import com.jme3.scene.Mesh;
 import com.jme3.scene.VertexBuffer;
 import com.jme3.scene.shape.Sphere;
+import com.jme3.system.AppSettings;
 import com.jme3.util.BufferUtils;
 
 /**
@@ -37,6 +38,9 @@ public class TestGeometryShader extends SimpleApplication {
 
     public static void main(String[] args) {
         TestGeometryShader app = new TestGeometryShader();
+        AppSettings settings = new AppSettings(true);
+        settings.setRenderer(AppSettings.LWJGL_OPENGL33);
+        app.setSettings(settings);
         app.start();
     }
 }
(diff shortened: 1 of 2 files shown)


rvandoosselaer:
I went through the rest of the examples, and did not encounter any other test scenerio's that required a higher core profile.

You can close the PR.

stephengold:
Integrated into master branch at 3fa85bc and into v3.3 branch at 4b62b70 .

stephengold:
Thank you, @rvandoosselaer !

stephengold:
A couple caveats that just occurred to me:
 1. Setting the profile in the `main()` method doesn't help when tests are run from the test chooser, since the test chooser bypasses `main()`.
 2. Most of jme3-examples apps do not initialize the `AppSettings` or set the app title. In that case, the settings are saved from run to run, even between different apps. Unless you `setRenderer()` back to the default between runs, your testing may have missed cases where the default value is too low.

rvandoosselaer:
I didn't consider your first remark. That is indeed correct. I haven't checked how the TestChooser is launching the examples. You will have different behaviour when launching.

The second caveat is something I already bumped into a lot of times, given the strict behaviour of Macs. I did make sure to reset my preferences after running those 2 shader tests before testing other examples.
I'm making a habit of using `AppSettings settings = new AppSettings(true);` whenever I write a new (test) application, to make sure I always use the default settings. When I want to deviate from them, I need to explicitly set it. I think it would be good practice to do this for all the examples.

stephengold:
Regarding #2, I propose giving every example that modifies `appSettings` a title, so that it won't share the settings with the other examples.

stephengold:
My proposal doesn't work, so `AppSettings(true)` seems to be the best solution.

rvandoosselaer:
I’ll make it my ‘lockdown-project’ to systematically adapt all the test cases and force the default AppSettings.

I’ll first need to check how the TestChooser works so the settings are also applied when launching an application from the TestChooser.

stephengold:
I think you'll need to modify TestChooser. Other than that, I don't foresee any difficulties.

PR #1339, opened 2020-04-08
Remove jogl 

ItsMike54:
This PR removes support for JOGL from core lib as it is filled with bugs and developers using jME rarely use JOGL for rendering. LWJGL is rather built for games.

PR #1368, opened 2020-04-28
restore the JOGL/JOAL hooks

stephengold:
This PR restores the "hooks" for using JOGL/JOAL as JMonkeyEngine's backend (instead of, say, LWJGL).

Said hooks were deleted when jme3-jogl was removed from the Engine by PR #1339. However, the situation has changed. As discussed [at the Forum](https://hub.jmonkeyengine.org/t/merged-dropping-jogl/43046/16), jme3-jogl could be hosted in JogAmp’s repository and maintained by the JogAmp community. Restoring the hooks should enable them to do so without forking JMonkeyEngine.


stephengold:
This PR is ready to be reviewed. If I don't receive feedback in the coming week, I plan to integrate it without review.

PR #1372, opened 2020-05-11
Update license headers in jme3-core

ItsMike54:
Most files in the engine have headers that only extend the copyright through 2012. This PR updates those files in jme3-core which haven't yet been updated.
2009-2020


pspeed42:
The date range should only change if new updates were made to the file in question.  It's not right to just arbitrarily update the last date on all files.

ItsMike54:
As in this PR is not needed? 
Also I'm curious as to why that isn't right.

pspeed42:
As in this PR isn't needed.

The date range on a copyright means something.  You've decided that the end date doesn't mean anything.  Might as well leave it off then.

A file that hasn't been updated since 2012 does suddenly get to claim an extra 8 years of copyright protection just because we decide that it does.

ItsMike54:
Fine then.

PR #1373, opened 2020-05-27
Camera#contains javadoc updated

kkolyan:
I've caught unpredictable behavior, when tried to use `Camera.contains`, then checked javadoc and found the reason why.
But javadoc is a bit outdated. This PR updates javadoc according observable behavior.

Old javadoc:
image

New javadoc:
image

Discussion: https://hub.jmonkeyengine.org/t/camera-contains-javadoc-outdated/43320

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/Camera.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/Camera.java
@@ -1006,13 +1006,13 @@ public class Camera implements Savable, Cloneable {
      * on the negative side of the plane is can be culled out.
      *
      * NOTE: This method is used internally for culling, for public usage,
-     * the plane state of the bounding volume must be saved and restored, e.g:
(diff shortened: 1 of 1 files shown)


pspeed42:
The check plane is a small optimization.  It's used to make sure that whatever plane caused the BV to be outside last time is checked first the next time.  All of the planes will still be checked if the object passes that check.  So your added code is not necessary.

The only case where it would be of any benefit at all (and even then only a little) is if an application is constantly checking the BV of all of the spatials against a different camera every frame.  And an app like that probably already has a hundred worse performance problems than missing this small optimization.

kkolyan:
Thanks for explanation. I removed lines with `checkPlane`.

stephengold:
Is this ready to be integrated?

pspeed42:
Seems like it.

PR #1376, opened 2020-06-05
Adding getters/setters to AnimComposer

rkraneis:
- add getters and setters that should be there just for symmetry reasons
- add missing javadoc (also on existing methods)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -145,6 +145,10 @@ public class AnimComposer extends AbstractControl {
         return l.currentAction;
     }
     
+    public void removeCurrentAction() {
+        removeCurrentAction(DEFAULT_LAYER);
+    }
+    
     /**
      * Remove current action on specified layer.
      *
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -174,6 +174,10 @@ public class AnimComposer extends AbstractControl {
         }
         return l.time;
     }
+    
+    public void setTime(double time) {
+        setTime(DEFAULT_LAYER, time);
+    }
 
     /**
      * Sets current time on the specified layer. 
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -48,10 +48,16 @@ import java.io.IOException;
 import java.util.*;
 
 /**
- * Created by Nehon on 20/12/2017.
+ * AnimComposer is a Spatial control that allows manipulation of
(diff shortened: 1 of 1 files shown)


stephengold:
Looks good to me. Thank you for your contribution to JMonkeyEngine!

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -125,6 +125,30 @@ public Action setCurrentAction(String actionName, String layerName) {
         return currentAction;
     }
     
+    public Action getCurrentAction(String name) {
```
Ali-RS, line 128: `String name` is redundant!

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -125,6 +125,30 @@ public Action setCurrentAction(String actionName, String layerName) {
...
+    }
+    
+    /**
+     * Return current action on specified layer.
+     * 
+     * @param actionName The name of the action to run.
+     * @param layerName The layer on which action should run.
+     * @return The action corresponding to the given name.
+     */
+    public Action getCurrentAction(String actionName, String layerName) {
```
Ali-RS, line 139: `String actionName` is redundant!

MeFisto94:
Small note: A proper title (`Adding getters/setters to AnimComposer`) or something would've been more clear.

Ali-RS:
Can you also add documentation on those methods, please?

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
rkraneis: Classic brain fart :sweat_smile:

rkraneis:
> Can you also add Javadoc on those methods, please?

Yes, will do. What is the policy in regards to authorship in source files? I'd change the class comment
from
```java
/**
 * Created by Nehon on 20/12/2017.
 */
```
 to
```java
/**
 * {@code AnimComposer} is a Spatial control that allows manipulation of
 * armament (skeletal) animation.
 * 
 * @author Nehon
 * @since 20/12/2017.
 */
```

Ali-RS:
I have no idea about it, but it might be better do it in a new PR.

stephengold:
I'm not aware of any policy, only practice.

Adding an "@author" tag is fine. JMonkeyEngine usually does not use "@since".

Personally, I don't like seeing "@code" in javadoc. The context usually makes it redundant.

MeFisto94:
> Personally, I don't like seeing "@code" in javadoc. The context usually makes it redundant.
If something that'd be `@link` or `@see` though and would make more sense like that.

Ali-RS:
Looks fine to me, thanks

Ali-RS:
Unless there is an objection, I am going to merge it in the next 24 hours.

stephengold:
OK. Just remember to use "Squash and merge" instead of "Merge commit" please.

Issue #1275, opened 2020-02-01
Create Javadoc hint for Application#start() not returning on lwjgl3


MeFisto94:
So, since we probably won't be able to fix this problem, at least not relatively short timed, it might be worth to add some knowledge to the wiki (@mitm001) and/or the javadoc for the start() methods.

Specifically on lwjgl2 one can choose: Return immediately or return after the context has been created.
On lwjgl3, start() will return when the application has requested to quit (e.g. stop() has been called).

Another problem, I think, is that calling app.stop(true) will deadlock the application on lwjgl3, as it waits for context disposal, but this disposal happens after the update loop, which is locked due to passing true there.

That being said, lwjgl2 has the same problem there, I think.

Edit: The second part is a duplicate of #1257

Ali-RS:
>Specifically on lwjgl2 one can choose: Return immediately or return after the context has been created.
On lwjgl3, start() will return when the application has requested to quit (e.g. stop() has been called).

Note, this has been fixed for a while now. Only on Linux and Windows.
https://github.com/jMonkeyEngine/jmonkeyengine/pull/1690/files

stephengold:
In my opinion, the actual fix was here:

https://github.com/jMonkeyEngine/jmonkeyengine/commit/40bb6483601c0aa26ae4f0d77bc965914a24528b

Ali-RS:
Should I update the Javadoc and mention that is only happening on mac os?

Something like this

```
 A bug occurring when using the LWJGL3 prevents the method from returning until after #stop() is called on mac os.
```

PR #1386, opened 2020-08-12
Fix for #1275

ItsMike54:


stephengold:
@MeFisto94: since you opened issue #1275, I'd like you to please review this.

MeFisto94:
Maybe we should phrase it like:
- "the LWJGL3 backend" (or only LWJGL3 without "the").
- "this method" instead of "the method"
- until after the application is stopped

The problem with stop() is that this actually is illegal as well, see the second part of #1275

ItsMike54:
"The LWJGL3" was obviously a typo.
I meant to say "the LWJGL3 backend", would you prefer this instead, or keep what you suggested?

PR #1387, opened 2020-08-17
Add more license headers to Nehon's gltf loader

ItsMike54:


stephengold:
Refer to issue #1001.

pspeed42:
I'm 99.9999% sure that Nehon just forgot to put the headers.  There is a certain expectation when, as a formal team member, you commit files to a project that has an overall license and forget to put a license.  (Which is different than committing files with the wrong license, especially an incompatible one.)

But in an overabundance of caution, I've sent nehon an e-mail so we can clear this up once and for all.

In the future, I suggest we don't let anyone have commit access until they've agreed that code committed to the JME repository is licensed to the project.  (Which was already the informal agreement.)

pspeed42:
Got confirmation from Nehon that anything he checked into JME repo should be licensed with the JME license and leaving it off was just an oversight.

ItsMike54:
Thanks @pspeed42 for confirming. I'll send more PRs like this one soon to get the rest of Nehon's files fixed.

PR #1396, opened 2020-08-29
GltfLoader decodes file path uri

tlf30:
This is a fix for the GltfLoader to properly decode the percent-format uri references in the gltf model file. 
See https://hub.jmonkeyengine.org/t/jme-not-decoding-gltf-uri/43619/12 for discussion. 

Thanks,
Trevor


MeFisto94 (approved):
Looks good to me, given the mentioned caveats on the hub (problems with faulty exporters not escaping `+` or not escaping stuff at all).
Since this is caused by a faulty exporter, we shall fail there.
Requesting a review from paul as the final say as he's been into that discussion.

pspeed42:
Seems fine to me.

PR #1397, opened 2020-09-03
Actually use ease functions input to InOut when applying

8Keep:
The inputs supplied to InOut were unused, previously it always used the Quad ease functions declared earlier in the file.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Easing.java
+++ b/jme3-core/src/main/java/com/jme3/math/Easing.java
@@ -159,10 +159,10 @@ public class Easing {
         public float apply(float value) {
             if (value < 0.5) {
                 value = value * 2;
-                return inQuad.apply(value) / 2;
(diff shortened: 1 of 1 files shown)


MeFisto94 (approved):
Great spot! That's a big oversight.
This will affect https://github.com/jMonkeyEngine/jmonkeyengine/blob/f808187142b79fa6c82c38b58ba09b71c90aa2bb/jme3-core/src/main/java/com/jme3/anim/interpolator/AnimInterpolators.java#L134-L177

Which means the new animation system was kind of broken in that regard.
Is this worth backporting to 3.3? I guess we won't really do another 3.3 release, though.

stephengold:
A 3.3.3 release is still possible, though I don't think this PR would be eligible for inclusion in it. To me, this looks like a reason to cut 3.4-alpha sooner rather than later.

PR #1401, opened 2020-10-03
Refactoring of framebuffer: Unified way to set render targets, added level field.

riccardobl:
All the old methods are kept but deprecated.


stephengold:
There are many places in the codebase where the deprecated methods are still used.

For instance, `setDepthBuffer()` is used in jme3-core, jme3-desktop, jme3-examples, and jme3-effects. But `setDepthTarget()` (its replacement) has no javadoc and is only used in a single example (`TestMultiRenderTarget`).

The shortage of examples (and javadoc) makes it difficult to understand how the replacement methods are meant to be used.

PR #1403, opened 2020-10-10
Packing AO in the metallic roughness map

yaRnMcDonuts:
Updated frag shader to support reading the AmbientOcclusion value from the Red channel of the packed MetallicRoughness map. 
Adds a single boolean define that does this when set to true, so long as a metallic roughness map is also defined.

In response to this open issue: 
https://github.com/jMonkeyEngine/jmonkeyengine/issues/999

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -227,7 +227,7 @@ void main(){
     #endif
 
     #if defined(AO_PACKED_IN_MR_MAP) && defined(USE_PACKED_MR)
-       ao *= aoRoughnessMetallicValue.rrr;
+       ao = aoRoughnessMetallicValue.rrr;
     #endif
 
     float ndotv = max( dot( normal, viewDir ),0.0);
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -134,10 +134,12 @@ void main(){
         vec4 albedo = Color;
     #endif
 
+    //ao in r channel, roughness in green channel, metallic in blue channel!
+    vec3 metallicRoughnessAoValue = vec3(1.0, 1.0, 0.0);
     #ifdef USE_PACKED_MR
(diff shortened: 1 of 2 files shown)


jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
```diff
@@ -134,10 +134,12 @@ void main(){
         vec4 albedo = Color;
     #endif
 
+    //ao in r channel, roughness in green channel, metallic in blue channel!
+    vec3 metallicRoughnessAoValue = vec3(1.0, 1.0, 0.0);
```
riccardobl, line 138: could you call it something like aoRoughnessMetallicValue to respect the order of the channels?

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
```diff
@@ -208,19 +210,24 @@ void main(){
...
-         lightMapColor.gb = lightMapColor.rr;
-         ao = lightMapColor;
+       #if defined(AO_MAP) || defined(AO_PACKED_IN_MR_MAP)
+            lightMapColor.gb = lightMapColor.rr;
+            ao = lightMapColor;
        #else
-         gl_FragColor.rgb += diffuseColor.rgb * lightMapColor;
+            gl_FragColor.rgb += diffuseColor.rgb * lightMapColor;
        #endif
+
```
riccardobl, line 230: I think the new code should go here:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag#L227

And just replace (or multiply) the value of `vec3 ao`, so that it doesn't interfere with the lightmap code that can be used together with the ao.

riccardobl (changes requested):
Can you merge your other PR ( #1404  ) to this one?

yaRnMcDonuts:
I updated the name of the vec3 to aoRoughnessMetallicValue, and also switched the code to work how you suggested, so that the normal light map code is unaltered.

yaRnMcDonuts:
I could not find how to merge my other PR - but I managed to update this PR's branch to include the changes to PBRLighting.j3md that were in the closed pull request

riccardobl:
Perfect! Thank you.

PR #1404, opened 2020-10-10
Packing AO in the metallic roughness map

yaRnMcDonuts:
Update j3md file to contain a boolean and corresponding define that tells the fragment shader if the ao map is packed in the metallic roughness map


riccardobl:
Please commit this change to your `patch-1` branch to merge it with pr #1403

Issue #942, opened 2018-10-15
Keyboard Mapping inconsistency between lwjgl2 and 3


MeFisto94:
This issue might need further testing because I _think_ this was different on Windows than on the linux I am currently seeing this bug. It probably also only affects non-english people but certainly those with a german (QWERTZ) layout. The problem is that lwjgl3 doesn't seem to care about to local keyboard layout and treat it as if it was an us-layout causing confusion.

```java
package mygame;

import com.jme3.app.SimpleApplication;
import com.jme3.input.KeyInput;
import com.jme3.input.controls.ActionListener;
import com.jme3.input.controls.KeyTrigger;

/**
 * Small Testcase for (at least) german keyboards: Behavior on "y" is different from lwjgl2 to lwjgl3 (3 doesn't seem
 * to respect the local keyboard mapping)
 * @author normenhansen
 */
public class Main extends SimpleApplication implements ActionListener {

    public static void main(String[] args) {
        Main app = new Main();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        inputManager.addMapping("y", new KeyTrigger(KeyInput.KEY_Y));
        inputManager.addMapping("z", new KeyTrigger(KeyInput.KEY_Z));
        inputManager.addListener(this, "y", "z");
    }

    @Override
    public void onAction(String name, boolean isPressed, float tpf) {
        System.out.println(name + ": " + isPressed);
    }
}
```
Run this once with lwjgl2 and once with lwjgl3 and in 3 when you press y you actually see "z" printed and vice-versa.

For convenience I've added a slightly modified version of Stephens "BasicGame-on-Gradle" so that it works with the commandline, you just need to comment out lwjgl3 and execute ./gradlew run:
(Gradle doesn't automatically add run otherwise and the mainclass is also a different property)
```gradle
apply plugin: 'java'
apply plugin: 'application'

// select one source-code option
//sourceCompatibility = '1.6'
//sourceCompatibility = '1.7'
sourceCompatibility = '1.8'

gradle.projectsEvaluated {
    tasks.withType(JavaCompile) {
        options.compilerArgs << '-Xdiags:verbose'
        options.compilerArgs << '-Xlint:unchecked'
        options.deprecation = true
        options.encoding = 'UTF-8'
    }
    tasks.withType(JavaExec) {
        args = []
        classpath sourceSets.main.runtimeClasspath
        //debug true
        enableAssertions true
[...]

PR #1406, opened 2020-10-10
Lwjgl3 impl patches

riccardobl:
In this PR i am going to address the issues that are still stopping us from using lwjgl3 as default backend.

I'll report important changes below:

## Better FPS limiter
Ported the Sync class from lwjgl2 to provide a more accurate fps limiter
Fix for #1455

## InputManager.getKeyName
Workaround for #942

-------------------

The following will be addressed in a future PR
##~~Removed AWT calls from the core~~
~~The AWT context causes glfw to behave weirdly or break in some platforms. 
I moved all the classes that need an awt context into a dedicated package to be adapted for lwjgl3 or outright removed in the future. I removed the SettingsDialog calls from the core, and refactored it to be self contained in its own class. 
Now the splash screen needs to be launched manually after creating the AppSettings, to reproduce the old jme behaviour the code will look like this~~
```java
        
        if(!AWTSettingsDialog.showDialog(settings))return; // Splash screen

        app.setSettings(settings);
        app.start();
```

~~Some image loaders still depend on awt, i am not sure if this is a problem, but i guess they can be reworked to use a standalone solution.~~

## ~~STB based image loader~~
~~I added an alternative to AWTLoader using this pure java port of stb_image `https://github.com/StbJava/StbImageJava`. 
We could of course use the bindings for the official stb_image.h, but the pure java code is more portable. The library is pretty small with no dependencies and under public domain but there is only a private bintray repo for it, so i just copied it in the engine. It seems to be working fine.~~

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/KeyInput.java
+++ b/jme3-core/src/main/java/com/jme3/input/KeyInput.java
@@ -558,4 +558,6 @@ public interface KeyInput extends Input {
      * the last key.
      */
     public static final int KEY_LAST = 0xE0;
+
+    public String getKeyName(int key);
 }
--- a/jme3-core/src/main/java/com/jme3/input/KeyInput.java
+++ b/jme3-core/src/main/java/com/jme3/input/KeyInput.java
@@ -559,5 +559,10 @@ public interface KeyInput extends Input {
      */
     public static final int KEY_LAST = 0xE0;
 
+
+    /**
+    * Returns the name of the specified keycode in the current system language.
+    * @param key The keycode from {@link com.jme3.input.KeyInput}
+    */
     public String getKeyName(int key);
 }
--- a/jme3-core/src/main/java/com/jme3/input/dummy/DummyKeyInput.java
+++ b/jme3-core/src/main/java/com/jme3/input/dummy/DummyKeyInput.java
@@ -50,7 +50,7 @@ public class DummyKeyInput extends DummyInput implements KeyInput {
 
     
     public String getKeyName(int key){
-        return null;
+        return "Unknown";
     }
 
 }
--- a/jme3-core/src/main/java/com/jme3/input/dummy/DummyKeyInput.java
+++ b/jme3-core/src/main/java/com/jme3/input/dummy/DummyKeyInput.java
@@ -48,4 +48,9 @@ public class DummyKeyInput extends DummyInput implements KeyInput {
         return 0;
     }
 
+    
+    public String getKeyName(int key){
+        return null;
+    }
+
 }
--- a/jme3-desktop/src/main/java/com/jme3/input/AWTKeyInput.java
+++ b/jme3-desktop/src/main/java/com/jme3/input/AWTKeyInput.java
@@ -234,4 +234,8 @@ public class AWTKeyInput extends AWTInput implements KeyInput, KeyListener{
       System.out.println("Key released "+e.getKeyChar());
       onKeyEvent(e, false);
     }
+
+    public String getKeyName(int key){
+      return null;
+  }
 }
--- a/jme3-desktop/src/main/java/com/jme3/input/AWTKeyInput.java
+++ b/jme3-desktop/src/main/java/com/jme3/input/AWTKeyInput.java
@@ -236,6 +236,6 @@ public class AWTKeyInput extends AWTInput implements KeyInput, KeyListener{
     }
 
     public String getKeyName(int key){
-      return null;
-  }
+      throw new UnsupportedOperationException("getKeyName is not implemented in AWTKeyInput");
+    }
 }
--- a/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyMap.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyMap.java
@@ -168,4 +168,14 @@ public class GlfwKeyMap {
     public static int toJmeKeyCode(final int glfwKey) {
         return GLFW_TO_JME_KEY_MAP[glfwKey];
     }
+
+    public static int fromJmeKeyCode(final int jmeKey) {
+        for(int i=0;i<GLFW_TO_JME_KEY_MAP.length;i++){
+            if(GLFW_TO_JME_KEY_MAP[i]==jmeKey){
+                return i;
+            }
+        }
+        return GLFW_KEY_UNKNOWN;
+    }
+    
 }
--- a/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyMap.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyMap.java
@@ -165,15 +165,24 @@ public class GlfwKeyMap {
         reg(KEY_RMETA, GLFW_KEY_RIGHT_SUPER);
     }
 
+    /**
+     * Returns the jme keycode that matches the specified glfw keycode
+     * @param glfwKey the glfw keycode
+     */
     public static int toJmeKeyCode(final int glfwKey) {
         return GLFW_TO_JME_KEY_MAP[glfwKey];
     }
 
+
+    /**
+     * Returns the glfw keycode that matches the specified jme keycode or
+     * GLFW_KEY_UNKNOWN if there isn't any match.
+     * 
+     * @param jmeKey the jme keycode
+     */
     public static int fromJmeKeyCode(final int jmeKey) {
-        for(int i=0;i<GLFW_TO_JME_KEY_MAP.length;i++){
-            if(GLFW_TO_JME_KEY_MAP[i]==jmeKey){
-                return i;
-            }
+        for (int i = 0; i < GLFW_TO_JME_KEY_MAP.length; i++) {
+            if (GLFW_TO_JME_KEY_MAP[i] == jmeKey) return i;
         }
         return GLFW_KEY_UNKNOWN;
     }
--- a/jme3-core/src/main/java/com/jme3/input/InputManager.java
+++ b/jme3-core/src/main/java/com/jme3/input/InputManager.java
@@ -154,6 +154,10 @@ public class InputManager implements RawInputListener {
         firstTime = keys.getInputTimeNanos();
     }
 
+    public String getKeyName(int key){
+        return keys.getKeyName(key);
+    }
+
     private void invokeActions(int hash, boolean pressed) {
         ArrayList<Mapping> maps = bindings.get(hash);
         if (maps == null) {
--- a/jme3-desktop/src/main/java/com/jme3/input/awt/AwtKeyInput.java
+++ b/jme3-desktop/src/main/java/com/jme3/input/awt/AwtKeyInput.java
@@ -623,4 +623,8 @@ public class AwtKeyInput implements KeyInput, KeyListener {
         return 0;
     }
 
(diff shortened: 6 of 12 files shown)


stephengold:
This PR seems like something that would be of broad interest. Please initiate some discussion at the Forum...

riccardobl:
We've been discussing about those issue for a long time. 
There are still some issues in the tracker that i need to look into, that's why this is a draft.

tlf30:
@riccardobl perhaps you can address #1048 in this as well. Regardless of the solution, as it is right now jme is not consistent between lwjgl 2 and 3.

riccardobl:
I wouldn't care too much about the consistency since lwjgl2 is dead anyway, however the lwjgl3 behaviour is the correct one IMO, the app should start in the thread from where start is called and lock it. And if the developer needs to run the app into a different thread, it should be created and started from there. We can deprecate the other start methods and say that start() is locking. 
The fact that macos doesn't allow opengl apps to run on anything but the main thread is another unrelated issue, I don't know if there is a workaround for it :man_shrugging:

tlf30:
I agree, that would be the best approach. Considering that the other start methods do not work as documented, it would be good to also fix the javadocs on them as well as deprecating them.

MeFisto94:
> The fact that macos doesn't allow opengl apps to run on anything but the main thread is another unrelated issue, I don't know if there is a workaround for it

It's not about openGL Apps, but the glFW Spec and the Mac OS Windows syscalls. The reason it works on lwjgl2 is that it probably avoids certain methods that require to be happening from the MainThread.

> It is a problem. Most GLFW functions are required to be called from the main thread, which also has to be the first thread on macOS (see -XstartOnFirstThread). The javadoc of each function mentions if it's callable from any thread or only from main. This is neither an LWJGL nor a GLFW limitation; it's the only way to write a usable cross-platform windowing system. Your application may work on Windows, but will certainly fail on macOS. Note that this applies to GLFW functions only; OpenGL contexts can be made current in other threads and those threads can be used for rendering.
https://github.com/LWJGL/lwjgl3/issues/311#issuecomment-301849854

> Note that the following files use the glfw api and thus potentially require the MainThread. This can be seen in the docs, see https://www.glfw.org/docs/latest/group__input.html#ga1caf18159767e761185e49a3be019f8d
```
./src/main/java/com/jme3/input/lwjgl/GlfwJoystickInput.java
./src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
./src/main/java/com/jme3/input/lwjgl/GlfwKeyInput.java
./src/main/java/com/jme3/input/lwjgl/GlfwKeyMap.java
./src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
./src/main/java/com/jme3/system/lwjgl/LwjglContext.java
```

The problem is that this kills lwjgl3 for Mac OS X for the SDK and others (in cases where you just CANNOT get hold of the java main thread)

MeFisto94:
Would it be possible to use stb_image directly (i.e. host it on maven or using git submodules)? 
Because in the current way, any changes will just get lost, because even if we are aware about upstream changes, I am not sure someone would incooporate them.

riccardobl:
I am having problems loading some images with that library, so i am not even sure i will keep it atm. I need to look into it.

stephengold:
I'd like to include this PR in JME v3.4. What needs to happen to get it out of "draft" status?

Would it help to break it up into several smaller PRs?

riccardobl:
What is currently working and tested (i've been testing lwjgl3 on linux for a long time) :
- Fix for fps synching code
- InputManager.getKeyName api

What is left to figure out is how to remove the awt stuff without breaking too much code.

I will commit the portion that we can merge

riccardobl:
Done

stephengold:
Thanks, Riccardo. Where can I find the original of "Sync.java"?

riccardobl:
> Thanks, Riccardo. Where can I find the original of "Sync.java"?

LWJGL2 source code: https://github.com/LWJGL/lwjgl/blob/master/src/java/org/lwjgl/opengl/Sync.java

tlf30:
Since `Sync` will be added in, can you also make the `NullContext` use it as well? This change would require `Sync` to be added to the `jme-core` module. Thoughts?

jme3-core/src/main/java/com/jme3/input/KeyInput.java
```diff
@@ -558,4 +558,6 @@
      * the last key.
      */
     public static final int KEY_LAST = 0xE0;
+
+    public String getKeyName(int key);
```
stephengold, line 562: please add javadoc here

jme3-core/src/main/java/com/jme3/input/dummy/DummyKeyInput.java
```diff
@@ -48,4 +48,9 @@ public int getKeyCount() {
         return 0;
     }
 
+    
+    public String getKeyName(int key){
+        return null;
```
stephengold, line 53: might be better to throw an UnsupportedOperationException

jme3-desktop/src/main/java/com/jme3/input/AWTKeyInput.java
```diff
@@ -234,4 +234,8 @@ public void keyReleased(KeyEvent e) {
       System.out.println("Key released "+e.getKeyChar());
       onKeyEvent(e, false);
     }
+
+    public String getKeyName(int key){
+      return null;
```
stephengold, line 239: might be better to throw an UnsupportedOperationException

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwKeyMap.java
```diff
@@ -168,4 +168,14 @@ private static void reg(final int jmeKey, final int glfwKey) {
     public static int toJmeKeyCode(final int glfwKey) {
         return GLFW_TO_JME_KEY_MAP[glfwKey];
     }
+
+    public static int fromJmeKeyCode(final int jmeKey) {
```
stephengold, line 172: please add javadoc

stephengold:
What's the motivation for renaming threads?

riccardobl:
> What's the motivation for renaming threads?

Because for debugging purposes the threads should have short descriptive names and the main thread should be called the same way across different renderers

jme3-core/src/main/java/com/jme3/input/dummy/DummyKeyInput.java
riccardobl: Not sure. In this case it might be better to return a constant string eg "Unknown" to not break user code with npe or other exceptions in case the code is designed to work with both real and dummy input

stephengold: "Unknown" sounds good.

stephengold:
This seems very close to being done. Just a couple review comments to resolve...

riccardobl:
Should be good to go

riccardobl:
I have also an example usage for this, but it is written with https://github.com/riccardobl/jme-igui

```java
package jme3test.input;

import com.jme.igui.*;

import com.jme3.input.KeyInput;
import com.jme3.input.controls.ActionListener;
import com.jme3.input.controls.KeyTrigger;

public class TestKeyNames extends SimpleApplication implements ActionListener {

    private IGui iGui;
    
    private boolean forward,backward,left,right,launch;
    
        TestKeyNames app=new TestKeyNames();
        app.start();
    }


    @Override
    public void simpleInitApp() {

        // Positional mapping
        int jmeW=KeyInput.KEY_W;
        int jmeA=KeyInput.KEY_A;
        int jmeS=KeyInput.KEY_S;
        int jmeD=KeyInput.KEY_D;

        System.out.println("KeyInput.KEY_W = " + inputManager.getKeyName(jmeW));
        System.out.println("KeyInput.KEY_A = " + inputManager.getKeyName(jmeA));
        System.out.println("KeyInput.KEY_S = " + inputManager.getKeyName(jmeS));
        System.out.println("KeyInput.KEY_D = " + inputManager.getKeyName(jmeD));

        // mnemonic mapping
        int launch=KeyInput.KEY_L;
        for(int i=0;i<KeyInput.KEY_LAST;i++){
            String layoutName=inputManager.getKeyName(i);
            if(layoutName!=null){
                layoutName=layoutName.toUpperCase();
                if(layoutName.equals("L")||layoutName.equals("Б")){
                    launch=i;
                    break;
                }
            }
        }
        System.out.println("L-aunch Key (mnemonic)" + inputManager.getKeyName(launch));

        inputManager.addMapping("Launch",new KeyTrigger(launch));
        inputManager.addMapping("Forward",new KeyTrigger(jmeW));
        inputManager.addMapping("Backward",new KeyTrigger(jmeS));
        inputManager.addMapping("Left",new KeyTrigger(jmeA));
        inputManager.addMapping("Right",new KeyTrigger(jmeD));

        inputManager.addListener(this, "Launch","Forward","Backward","Left","Right");
        
        iGui=IGuiAppState.newRelative(assetManager,stateManager,guiNode,cam.getWidth(),cam.getHeight());
        

        iGui.textFont("Interface/Fonts/Default.fnt").textSize(0.04f).textColor(ColorRGBA.White);
        iGui.text("Press "+ inputManager.getKeyName(jmeW).toUpperCase(), 0f,1f,true);
        iGui.text("Press "+ inputManager.getKeyName(jmeS).toUpperCase(), 0,1f-0.05f,true);
        iGui.text("Press "+ inputManager.getKeyName(jmeA).toUpperCase(), 0,1f-0.10f,true);
        iGui.text("Press "+ inputManager.getKeyName(jmeD).toUpperCase(), 0,1f-0.15f,true);
        iGui.text("Press "+ inputManager.getKeyName(launch).toUpperCase(), 0,1f-0.20f,true);        


    }

    

    public void simpleUpdate(float tpf){
        iGui.textColor(forward?ColorRGBA.Green:ColorRGBA.Red);
        iGui.text(forward?"PRESSED":"NOT PRESSED", 0.4f,1f);

        iGui.textColor(backward?ColorRGBA.Green:ColorRGBA.Red);
        iGui.text(backward?"PRESSED":"NOT PRESSED", 0.4f,1f-0.05f);

        iGui.textColor(left?ColorRGBA.Green:ColorRGBA.Red);
        iGui.text(left?"PRESSED":"NOT PRESSED", 0.4f,1f-0.10f);

        iGui.textColor(right?ColorRGBA.Green:ColorRGBA.Red);
        iGui.text(right?"PRESSED":"NOT PRESSED", 0.4f,1f-0.15f);

        iGui.textColor(launch?ColorRGBA.Green:ColorRGBA.Red);
        iGui.text(launch?"PRESSED":"NOT PRESSED", 0.4f,1f-0.20f);
    }


    @Override
        switch(name){
            case "Forward":
                forward=isPressed;
                break;
            case "Backward":
                backward=isPressed;
                break;
            case "Left":
                left=isPressed;
                break;
            case "Right":
                right=isPressed;
                break;
            case "Launch":
                launch=isPressed;
                break;
        }
    }

    

}```

jme3-desktop/src/main/java/com/jme3/input/AWTKeyInput.java
stephengold: Still needs resolution.

stephengold:
There's still the unresolved conversation regarding `AWTKeyInput`: should `getKeyName()` throw an exception or not?

riccardobl:
Right, i missed that one. Done

Issue #1098, opened 2019-05-27
CONTRIBUTION.md should be updated to specify what code style should be applied to PRs


riccardobl:
Following : https://github.com/jMonkeyEngine/jmonkeyengine/issues/1056
It seems that the common agreement is that we should use netbeans style, if that is the case, we should write that down on CONTRIBUTION.md and provide settings for at least the most common ides.

This link contains the configurations for eclipse and other compatible editors (eg vscode) https://we.tl/t-jcnutcbB4U

Since these days, many people use idea, someone should also check if the default idea style is ok for us or if the configuration provided above can be applied (with extensions or whatever) or if there is another way to define and export the style.

louhy:
Some other things to consider adding there if adopting standards, with reasoning below. 

1. Default NetBeans defaults to an **80 character right margin**. I have mixed feelings about this. Many existing comments will be wrapped as a result (creating more Git chaos). It seems outdated with today's larger monitors. But, it may allow you to snap your IDE to one half of the monitor without cutting off code. And I guess it helps when we're forced to use craptops. :stuck_out_tongue: 

2. Should there be (or is there) a guideline on commented out code? Git has the history for things removed, and "WIP" stuff should be documented. If something is being removed but there's doubt that it's correct to do so, at least _add a comment explaining why it was disabled_. 
**TLDR: If you check in (new) commented out code, explain why.** Otherwise others are wasting time wondering "why is this here?", and it just detracts from code readability/quality. Experimental code under very heavy development working a tricky problem can be an exception here... 

3. **Remove Unused Imports** - NB and Eclipse can do it with a single key or "On Save" (for NB). I don't know why this isn't enabled by default when it's usually also a single keypress to re-add them if needed. 

4. Here's where I really get annoying... **Try to address warnings/suggestions as much as possible.** Your IDE will often offer suggestions to clean up warnings or improve code with a single click. Sometimes it's not always possible, and not all IDEs will agree on these, but warnings are there for a reason so they shouldn't be ignored totally. If you ignore 100 "not important" warnings, you'll miss the one that would have helped you catch a mistake... which is the whole point of warnings. Also, it annoys the heck out of people who want to use the right-gutter to navigate a file, find references, and so on when there's dozens of yellow flag warnings cluttering it up. I go as far as using suppression annotations if necessary (rarely), ex. `@SuppressWarnings("unchecked")`... should be as narrow as possible, don't suppress all warnings. 

[...]

PR #1410, opened 2020-10-10
Add vscode config & formatter

riccardobl:
This PR adds the formatter settings discussed here https://github.com/jMonkeyEngine/jmonkeyengine/issues/1098 for vscode.
The xml file is also compatible with eclipse. I think we could add configurations for more common IDEs  (idea?) as long as they are self contained in identifiable folders (eg no .settings)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 **/.classpath
 **/.settings
 **/.project
-**/.vscode
+**/.vscode/**
 **/out/
 /.gradle/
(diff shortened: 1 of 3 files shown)


stephengold:
I believe the "/**" suffix added in the .gitignore file has no effect. Ignoring a directory ignores everything therein.

riccardobl:
I don't want to ignore the directory, but the content minus those two files

PR #1413, opened 2020-10-18
Deleted conditional break statement inside the infinite loop and while loop converted to do-while loop

abdullahcevik:
Hi, I saw a conditional break statement inside the infinite loop. I converted a  while loop to do-while loop

### DIFF
--- a/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java
+++ b/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java
@@ -219,8 +219,8 @@ public abstract class LwjglAbstractDisplay extends LwjglContext implements Runna
             logger.log(Level.SEVERE, "Display initialization failed. Cannot continue.");
             return;
         }
-        while (true){
-            if (renderable.get()){
+        do {
+            if (renderable.get()) {
                 if (Display.isCloseRequested())
                     listener.requestClose(false);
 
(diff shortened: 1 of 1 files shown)


pspeed42:
I have mixed feelings about this.  I didn't write this code so I don't have a personal stake but I generally find do/while loops to be awkward and ugly.  I think I can count the times I've used them on one hand.  Usually I find it a sign I've done something else wrong.  Likewise for while(true) loops, though.

...in this case, I have to wonder why it's not just a while(!needClose.get()) loop.  If by luck of fate, something tries to destroy the context before the run method gets to the loop... is it really desirable to still run everything once before stopping?

It's weird.

abdullahcevik:
> I have mixed feelings about this. I didn't write this code so I don't have a personal stake but I generally find do/while loops to be awkward and ugly. I think I can count the times I've used them on one hand. Usually I find it a sign I've done something else wrong. Likewise for while(true) loops, though.
> 
> ...in this case, I have to wonder why it's not just a while(!needClose.get()) loop. If by luck of fate, something tries to destroy the context before the run method gets to the loop... is it really desirable to still run everything once before stopping?
> 
> It's weird.

I disagree with your views on the do-while loop, on the contrary, I believe it makes the code more readable and clean.

tlf30:
I agree with Paul on this. As someone who owns a small software company and manages a team of JavaEE programmers, do-while loops cause much more unnecessary confusion and bugs than any other type of loop.

riccardobl:
I also agree that the current code is better than the proposed one.
This is the render loop, the while(true) put emphasis on the fact that this code is supposed to always loop until there is a special event.

stephengold:
I, too, prefer the current code. Does anyone see a reason *not* to close this PR?

riccardobl:
I am closing

PR #1416, opened 2020-11-09
Add project license header information

tonihele:
Since you have been adding those IDE specific settings... here is one from the home team :)

This re-adds the once deleted license.txt and adds it as the project license in Gradle properties. This works in Netbeans (jME SDK). Probably not in any other IDE. Whenever you add a new class file to any of the jME projects in NB, this license header will be added on top of the class.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,9 @@ allprojects {
     }
 }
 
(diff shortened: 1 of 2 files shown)


stephengold:
Looks good to me!

Issue #1382, opened 2020-07-24
LODGeomap creates index buffers with elements too small 


stephengold:
In the jme3-terrain library, the `LODGeomap` class allocates index buffers for each `TerrainPatch` using `IndexBuffer.createIndexBuffer()`. It does so assuming that the number of vertices in the mesh is less than or equal to the number of indices, but this erroneous.

Consider the `TerrainTest` application in jme3-examples. With the current lod multiplier of 2.7 (specified in the `DistanceLodCalculator` constructor) only 3 LODs are actually used. If we change the multiplier to a lower value, such as 1.0, then all 6 LODs will be used, and some of the more distant terrain patches disappear entirely. If we lower it further, to 0.5, then large swaths of terrain vanish.

`LODGeomap` is allocating `ByteBuffer`s for indices, then trying to write indices greater than 255 to them. I added range checking to `IndexByteBuffer.put(int i, int value)` and saw values such as 2112 being passed down from `LODGeomap.writeIndexArrayLodDiff()`. If I replace the `createIndexBuffer(numIndexes, numIndexes)` calls in `LODGeomap` with `createIndexBuffer(66666, numIndexes)`, `IntBuffer`s are allocated, and the missing terrain patches re-appear.

1. It's very bad that `IndexByteBuffer` and `ShortByteBuffer` don't do any range checking on their inputs.
2. `LODGeomap` needs to correctly calculate the number of vertices, or else it should always use `IntBuffer`s.

pspeed42:
They shouldn't have to do range checking because the underlying jni buffers should be.  It's a bit surprising that ByteBuffer.put() is allowing out of range values.  It should be throwing IndexOutOfBoundsException.

pspeed42:
re: " It does so assuming that the number of vertices in the mesh is less than or equal to the number of indices, but this erroneous."
I'm not sure I see that code.  The index buffer size seems to be created based on the LOD level, etc. in private int calculateNumIndexesLodDiff(int lod)

When calling createIndeBuffer() normally the vertex count and index count are passed separately and LODGeomap is indeed incorrectly passing index count for both... but that's the right value for the size of the buffer.  The reason they are normally both passed is so that the method can decide which type of buffer to use.  If you only have 255 vertexes then it can use a ByteBuffer for the values.  If you have 100,000 vertexes then it must use and int buffer.

[...]

PR #1418, opened 2020-11-09
fix issue #1382 (LODGeomap index buffers with elements too small)

stephengold:
This resolves the immediate issue. I will submit additional PRs to add range-checking assertions.


MeFisto94 (approved):
LGTM, you can merge it when/if it fits for you.

Issue #1421, opened 2020-11-12
Attempted to add mapping "ScreenShot" twice to trigger


stephengold:
If you attach a `ScreenshotAppState`, detach it, and attach it again, you will get a warning:
```text
WARNING: Attempted to add mapping "ScreenShot" twice to trigger.
```

Here's what I think is happening: `ScreenshotAppState.initialize()` adds a mapping to the `InputManager`. Its `cleanup()` method should remove the mapping, but that method isn't implemented anywhere.

MeFisto94:
You are right but the issue is even worse. If I a not mistaken there is no cleanup method at all, meaning not only will this AppState instance leak due to being registered as a Listener, it also doesn't cleanup the Processor, causing dirty behavior.

pspeed42:
Some of these earlier app states were very bad with life cycle.  Not all of them have been fixed.

stephengold:
Solved by PR #1421 in master at a3c3678

stephengold:
I confirmed that the issue is resolved in v3.4.0-beta3.

PR #1424, opened 2020-11-15
solution and testcase for issue #1421 (ScreenshotAppState never cleans up)

stephengold:


jme3-core/src/main/java/com/jme3/app/state/ScreenshotAppState.java
```diff
@@ -187,6 +195,26 @@ public void initialize(AppStateManager stateManager, Application app) {
...
+     */
+    @Override
+    public void cleanup() {
+        if (inputManager != null) {
+            inputManager.deleteMapping("ScreenShot");
+            inputManager.removeListener(this);
+            inputManager = null;
+        }
+
+        ViewPort viewPort = last;
```
MeFisto94, line 209: Just so I get this, you use `viewPort` because calling `viewPort.removeProcessor` triggers a cleanup call on this AppState?
Because otherwise this wouldn't be needed, but I am not 100% certain what you mean with indirect recursion, might need some more commenting.

Also is there a reason why super.cleanup() is called last and not first? I guess it doesn't make a difference, but I am used to it being the first statement.

stephengold: Yes, `removeProcessor()` calls `cleanup()`. So if we weren't careful, we'd get an infinite recursion. I should document that fact explicitly, I guess.

My view is that subclasses are built upon (and depend on) superclasses. When I'm initializing, I initialize the superclass before the subclass. Then, when I'm cleaning up, I clean up the subclass before the superclass.

BTW, due to the recursion, it's likely that `super.cleanup()` gets invoked twice, but that doesn't seem to harm anything.

MeFisto94:
I couldn't connect how a SceneProcessor could cleanup the AbstractAppState, but apparently there is a name clash between `SceneProcessor#cleanup` and `AbstractAppState#cleanup`, where from AAS we call SP#cleanup, which in turn invokes on our AppState.

Your workaround _will_ work, `super.cleanup` is no problem for the current case (because it only sets a boolean), but there might be a more clever way.
I am sure @pspeed42 has had this case a few times already and knows something clever?

What I could imagine is that making the Processor an internal class (could even be anonymous, I guess) would make this code more clean and avoid such a hack, then.
Essentially those are two classes anyway, the appstate only does the lifecycle management of the Processor.

pspeed42:
Since I was tagged: Yeah, my own personal idiomatic approach is to avoid excessive multiclassing just for (fake) convenience.  So that app state has to be an app state because that's what it is.  It should not also be a scene processor... because it's an app state.  The issue described is one of a half-dozen pitfalls to the 'quick and dirty' approach uses in this app state... but probably too long to go into here.

jme3-core/src/main/java/com/jme3/app/state/ScreenshotAppState.java
stephengold: I've added some javadoc explaining the recursion hazard in detail.

stephengold:
If there's no further discussion, I'd like to integrate this PR today.

PR #1426, opened 2020-11-15
Codequality Pass 1

MeFisto94:
Info: This PR can be squashed, I mostly did over precise commits for review.
The most common change was adding a `break` to the last `case` statement of the `switch`, that's bad code that could lead to a problem, when someone adds another `case` statement overlooking the `break`.


riccardobl (approved):

stephengold (approved):
Looks good to me!

Issue #1383, opened 2020-07-27
LWJGLException: Could not locate OpenAL library


stephengold:
Forum thread: https://hub.jmonkeyengine.org/t/solved-exception-could-not-locate-openal-library/43423

It appears that sometimes the OpenAL initialization fails, causing this exception to be thrown. In that case, initialization should be retried. Retry code should be added for both jme3-lwjgl and jme3-lwjgl3.

stephengold:
This issue remains open because: in order to address it for LWJGL v3, I'd need to see a stack trace.

stephengold:
No stack trace has been reported. Closing the issue.

PR #1432, opened 2020-11-22
mitigate issue #1383 (Could not locate OpenAL library) for LWJGL v2

stephengold:
This adds a series of retries (with increasing delays) in case the first attempt to `AL.create()` throws an `LWJGLException`.

### DIFF
--- a/jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
+++ b/jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
@@ -20,7 +20,7 @@ public class LwjglALC implements ALC {
     @Override
     public void createALC() {
         int numRetriesRemaining = 4;
-        int retryDelayMsec = 100; // start with an 0.1-second delay
+        int retryDelayMsec = 100; // 0.1-second delay between retries
 
         while (true) {
             try {
@@ -47,9 +47,6 @@ public class LwjglALC implements ALC {
                     Thread.sleep(retryDelayMsec);
                 } catch (InterruptedException exception2) {
                 }
-
-                // Triple the wait time after each failure.
-                retryDelayMsec *= 3;
             }
         }
     }
--- a/jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
+++ b/jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
@@ -2,6 +2,8 @@ package com.jme3.audio.lwjgl;
 
 import com.jme3.audio.openal.ALC;
 import java.nio.IntBuffer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.lwjgl.LWJGLException;
 import org.lwjgl.openal.AL;
 import org.lwjgl.openal.ALC10;
@@ -9,6 +11,11 @@ import org.lwjgl.openal.ALCcontext;
 import org.lwjgl.openal.ALCdevice;
 
 public class LwjglALC implements ALC {
+    /**
+     * message logger for this class
+     */
+    private static final Logger logger
+            = Logger.getLogger(LwjglALC.class.getName());
 
     @Override
     public void createALC() {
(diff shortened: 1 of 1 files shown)


jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
```diff
@@ -12,10 +12,36 @@
...
+                break;
+
+            } catch (LWJGLException exception1) {
+                if (numRetriesRemaining < 1) {
+                    throw new RuntimeException(exception1);
+                }
+
+                // Retry to mitigate JME Issue 1383.
+                --numRetriesRemaining;
+                System.out.printf("Caught an LWJGLException from AL.create(). "
```
riccardobl, line 30: Can this be printed as a warn by the logger?

jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
```diff
@@ -12,10 +12,36 @@
...
+                        numRetriesRemaining,
+                        (numRetriesRemaining == 1) ? "y" : "ies");
+
+                try {
+                    Thread.sleep(retryDelayMsec);
+                } catch (InterruptedException exception2) {
+                }
+
+                // Triple the wait time after each failure.
+                retryDelayMsec *= 3;
```
riccardobl, line 43: Not sure about this. If it fails a couple of times it might cause the app to lock too much and trigger the os to ask the user to kill it.
I think it would be better to always wait for a very short time (10ms or 100 ms?) and keep spamming it until the library is loaded, to get the app going as soon as possible.

pspeed42: Or increase it linearly instead of geometrically.

stephengold: The maximum number of retries is 4. If `create()` always fails, then the total sleep time would be 4 seconds (0.1 + 0.3 + 0.9 + 2.7). Could 4 seconds of sleeping (not busy waiting) during initialization really trigger the OS to intervene? I don't see how.

I implemented increasing sleeps because I don't know how much waiting is needed. What if 100 ms isn't enough time?
I limited the number of retries in case it's a situation where `create()` always fails. In which case, I want:

 + the app to terminate (so it's obvious that something has gone wrong) and
 + the software to generate a stack trace.

jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
stephengold: Thanks for the suggestion. Implemented in b546358

jme3-lwjgl/src/main/java/com/jme3/audio/lwjgl/LwjglALC.java
riccardobl: I am not sure, but i think 4 seconds of waiting on the main thread for graphical applications could be problematic.
What is the reason for increasing the sleeping time in the first place?

pspeed42: This is during initialization... not during "application running".  And it seems that sometimes OpenAL has to be hit a few times before it will actually start.

riccardobl: Does this happen before the lwjgl context is created?

stephengold: No. The GL context is created first.

stephengold: @riccardobl I'm looking for constructive feedback. What would be an acceptable amount of sleep time, in your opinion?

riccardobl: I am not sure about the reason for increasing the sleeping time. Since this has to be loaded as fast as possible, i wouldn't wait more than 100ms. That's more than enough to not hog the cpu.

stephengold: I've pushed a commit to remove the increases. I believe that resolves this conversation. Thank you both for your help!

riccardobl (approved):

PR #1434, opened 2020-11-24
Use BufferedOutputStream to achieve maximum throughput

tonihele:
Ah, sorry for the white space changes.

Well, I was heavily optimizing our OpenKeeper's file operations. Basically trying to convert it to NIO world instead of the old I/O libraries. The results are stunning. There was still this place where I was relying on jME to write out the converted assets. And here it is.

We are writing 2015 files (615 MB) of j3o models with this BinaryExporter. Files range from few kilobytes to anywhere around 8 megabytes.

**Before: 110 seconds
After: 23 seconds**

The time includes a lot of stuff really, but this is the effect of this little change. This is not NIO, instead the old I/O. Still the fastest way to write a file in Java. See for example https://www.happycoders.eu/java/how-to-write-files-quickly-and-easily/#Performance_overview_writing_files.

Dunno what you think about doing some NIO work. That means basically getting rid of File, replace it with Path, Paths and Files classes. That doesn't alone yield any performance improvements though, or at least measurable ones. Mainly API thing.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryExporter.java
+++ b/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryExporter.java
@@ -161,23 +161,23 @@ public class BinaryExporter implements JmeExporter {
 
     public static boolean debug = false;
     public static boolean useFastBufs = true;
-      
+
     public BinaryExporter() {
     }
 
     public static BinaryExporter getInstance() {
         return new BinaryExporter();
     }
-    
+
     /**
      * Saves the object into memory then loads it from memory.
-     * 
+     *
      * Used by tests to check if the persistence system is working.
-     * 
+     *
      * @param <T> The type of savable.
      * @param assetManager AssetManager to load assets from.
      * @param object The object to save and then load.
-     * @return A new instance that has been saved and loaded from the 
+     * @return A new instance that has been saved and loaded from the
      * original object.
      */
     public static <T extends Savable> T saveAndLoad(AssetManager assetManager, T object) {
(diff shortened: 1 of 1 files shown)


stephengold:
Thank you for your contribution.

oxplay2:
wow, simple change but so much difference. thanks!

tonihele:
> 
> 
> wow, simple change but so much difference. thanks!
> 
> looks like we should use new way.

They make a ton of difference. Also on the reading side. But a quick glimpse made me think that the read side was already maxed out :) Using at least BufferedReader.

Here is basically what I'm doing for OpenKeeper (if you are interested, also if you wanna bring jME to NIO world):
https://github.com/tonihele/OpenKeeper/compare/nio-stuff?expand=1

It is like night and day, using BufferedReaders/Writers. I feel stupid for not starting with this knowledge :D For us the whole process what this touches has come from 10 minutes to 30 seconds or so. Of course there is also parallelization etc. But this is the major thing, how to handle files. And basically it is Java's Achille's heel they say. Especially if one does it wrong it seems :)

pspeed42:
Yeah, the Buffered* classes are a must.  They will get you to 95% of the maximum performance possible.  Getting the Buffered* instances to use the right sized buffers is another 2-3%.  I found the last NIO steps to be negligible in the end and often not worth the trouble.

danielperano:
Looks great to me!

riccardobl (approved):

stephengold:
@tonihele Next time, without the whitespace changes, please!

tonihele:
> @tonihele Next time, without the whitespace changes, please!

Thanks. Need to figure out where they come from :( I do have automatic formatting set, but "only modified lines".

PR #1435, opened 2020-11-24
Use BufferedWriter

tonihele:
Related to: https://github.com/jMonkeyEngine/jmonkeyengine/pull/1434

These we write 1100 files (namely the Material files, j3m). The speedup is on the error margin of the measurement. The reason being that the usage really limits to these really small files (500 bytes...). It still should be less writes on the disk.

### DIFF
--- a/jme3-plugins/src/main/java/com/jme3/material/plugin/export/material/J3MExporter.java
+++ b/jme3-plugins/src/main/java/com/jme3/material/plugin/export/material/J3MExporter.java
@@ -10,13 +10,13 @@ import com.jme3.export.OutputCapsule;
 import com.jme3.export.Savable;
 import com.jme3.material.Material;
 import com.jme3.material.MaterialDef;
-
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
-import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Saves a Material to a j3m file with proper formatting.
@@ -50,13 +50,14 @@ public class J3MExporter implements JmeExporter {
             throw new IllegalArgumentException("J3MExporter can only save com.jme3.material.Material class");
         }
 
-        OutputStreamWriter out = new OutputStreamWriter(f, Charset.forName("UTF-8"));
+        try (OutputStreamWriter out = new OutputStreamWriter(f, StandardCharsets.UTF_8);
+                BufferedWriter writer = new BufferedWriter(out)) {
 
-        rootCapsule.clear();
-        object.write(this);
-        rootCapsule.writeToStream(out);
+            rootCapsule.clear();
+            object.write(this);
+            rootCapsule.writeToStream(writer);
 
-        out.flush();
+        }
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


pspeed42:
Wow.  After all this time, I still find JME code that surprises me.
public void writeToStream(OutputStreamWriter out)

:-/

Thanks for fixing these.

jme3-plugins/src/main/java/com/jme3/material/plugin/export/material/J3MExporter.java
```diff
@@ -50,13 +50,14 @@ public void save(Savable object, OutputStream f) throws IOException {
             throw new IllegalArgumentException("J3MExporter can only save com.jme3.material.Material class");
         }
 
-        OutputStreamWriter out = new OutputStreamWriter(f, Charset.forName("UTF-8"));
+        try (OutputStreamWriter out = new OutputStreamWriter(f, StandardCharsets.UTF_8);
+                BufferedWriter writer = new BufferedWriter(out)) {
```
riccardobl, line 54: I am not sure about this.
The method might be called to write on a stream that doesn't need (or have already) buffering cache.
I think this should be done like in the other PR, where the Buffered*Steam is used on the method that writes on files.

riccardobl (changes requested):

jme3-plugins/src/main/java/com/jme3/material/plugin/export/material/J3MExporter.java
pspeed42: That is true.  The method signature changes are really good, though.  Requiring an OutputStreamWriter was kind of silly.

tonihele: Yeah, I also believe that is how Java is generally wanting us to work. If you give input/output stream as param, you want to control how/when/where. The BinaryImporter does violate this where as the BinaryExporter doesn't.

This was a bit controversial change in that sense. I just thought that the worst thing that can happen is double buffering. And BufferedWriter differs from BufferedStream a bit, I think it also adds cache to the characters. And you can't do this anymore from outside then. Only give a stream that buffers.

I think all the usages of this are outside the engine. But I can check that and leave just the method signature changes.

riccardobl: > I just thought that the worst thing that can happen is double buffering.

It can reduce performances in common use cases and add extra delays that are unexpected.

>  And BufferedWriter differs from BufferedStream a bit, I think it also adds cache to the characters.

But if this is piped to a BufferedOutputStream, everything will be cached anyway.

> I think all the usages of this are outside the engine. But I can check that and leave just the method signature changes.

The signature changes are ok, and i think  a BufferedOutputStream should be used here

https://github.com/jMonkeyEngine/jmonkeyengine/blob/d25ab90bb99c95850d059791026d0a2428d09c3d/jme3-plugins/src/main/java/com/jme3/material/plugin/export/material/J3MExporter.java#L65-L67

Similarly to what you did for the importer, since i believe it is reasonable to assume that this code will always benefit from this change.

tonihele: Yep, changed those. Also did the trick to the MaterialDefinitionExporter.

tonihele:
I was tempted to add the license headers to those files. Seems they have nice Netbeans default template. But maybe better to do it separate to keep things clean.

riccardobl:
The pr is fine, but there are a lot of changes in whitespaces... I would merge this anyway, but if you could remove them it would be better.

riccardobl (approved):

PR #1437, opened 2020-11-27
[M]Fix issue#1436

jmecn:
Fix the code error when checking UTF-8 data

let content = [0xE4, 0x8A, 0xBC], let `b = 0xE4` (1110 0100)

See this part:
````java
                if (b < 0x80) {
                    // good
                }
                else if ((b & 0xC0) == 0xC0) {//   (0xE4 & 0xC0) == 0xC0     =====>  true
                    utf8State = UTF8_2BYTE;
                }
                else if ((b & 0xE0) == 0xE0) {//   (0xE4 & 0xE0) == 0xE0      =====>  true
                    utf8State = UTF8_3BYTE_1;
                }
                else {
                    utf8State = UTF8_ILLEGAL;
                }
````

3 bytes UTF-8 data while always be treated as 2 bytes UTF-8 data.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryInputCapsule.java
+++ b/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryInputCapsule.java
@@ -37,11 +37,11 @@ import com.jme3.export.SavableClassUtil;
 import com.jme3.util.BufferUtils;
 import com.jme3.util.IntMap;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
 import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 import java.nio.ShortBuffer;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


stephengold:
Please change "is" to "are" in 3 places. Other than that, this looks great.

jmecn:
It looks better that always treat String data as UTF-8.
write it UTF-8, read it UTF-8

see https://hub.jmonkeyengine.org/t/code-error-on-checking-utf-8-data/43909

riccardobl:
Can you use StandardCharsets.UTF_8 for the charset?

jmecn:
> Can you use StandardCharsets.UTF_8 for the charset?

OK, but it's better to use it in both input and output.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/2196e4c/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryOutputCapsule.java#L688

riccardobl (approved):
Perfect.

PR #1440, opened 2020-12-02
Fix spherical light probes always starting at 0,0,0

grizeldi:
When setting the type of a light probe to spherical, previously the center of the area wasn't set to the probe's position.

The code is untested, as I can't currently build the engine locally, so someone testing it would be welcome. The changes are minimal though, so chances of something breaking are small.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/LightProbe.java
+++ b/jme3-core/src/main/java/com/jme3/light/LightProbe.java
@@ -231,9 +231,9 @@ public class LightProbe extends Light implements Savable {
                 break;
             case OrientedBox:
                 area = new OrientedBoxProbeArea(new Transform());
-                area.setCenter(position);
                 break;
         }
+        area.setCenter(position);
     }
 
     public AreaType getAreaType(){


riccardobl:
If someone wants to test this quickly, it is always possible to download the maven artifacts for every commit from the CI artifacts, in this case from: https://github.com/jMonkeyEngine/jmonkeyengine/actions/runs/396272941

grizeldi:
Tested it using the following code and it seems to be working. The printed position stays the same.
```java
	private LightProbe probe;

	@Override
	public void simpleInitApp() {
		EnvironmentCamera envCam = new EnvironmentCamera(); //Make an env camera
		stateManager.attach(envCam);
		envCam.initialize(stateManager, this); //Manually initilaize so we can add a probe before the next update happens
		probe = LightProbeFactory.makeProbe(envCam, rootNode);
		probe.getArea().setRadius(10); //Set the probe's radius in world units
		probe.setPosition(new Vector3f(1, 0, -2)); //Position it in 3d space
		rootNode.addLight(probe);
	}

	@Override
		System.out.println("Probe position: " + probe.getPosition());
		switch (probe.getAreaType()){
			case Spherical:
				probe.setAreaType(LightProbe.AreaType.OrientedBox);
				break;
			case OrientedBox:
				probe.setAreaType(LightProbe.AreaType.Spherical);
				break;
		}
	}

	public static void main(String [] args){
		new Main().start();
	}
}
```

PR #1444, opened 2020-12-13
Allow use of a GlowColor with GlowMap

Ali-RS:
In the case of grayscale GlowMap, GlowColor can be used for changing color without manipulating glow texture.

Forum topic: https://hub.jmonkeyengine.org/t/glowcolor-unused-if-a-glowmap-is-specified/44002


Ali-RS:
Going to merge this in 24h if no objection.

riccardobl (approved):

PR #1449, opened 2020-12-19
Add desaturation functionality to Unshaded.j3md

yaRnMcDonuts:
I added a small amount of code to the Unshaded.j3md and Unshaded.frag files to allow for desaturating the material's output color based on a float value, specifically for use in GUIs. 

Nearly all our GUI libraries use the Unshaded matdef for elements that are attached to the GUI node, and any custom gui elements will typically start from extending Unshaded if they need to do something in the shader, so adding this to Unshaded would alow all of our existent gui libraries to have desaturation functionality.
This should also not break any existing code, as nothing happens with this code unless the float value for desaturation gets defined for the material.

Following up on this thread https://hub.jmonkeyengine.org/t/how-to-desaturate-gui-element/43963

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.frag
@@ -48,7 +48,7 @@ void main(){
         }
     #endif
     
-    #ifdef DESATURATE
+    #ifdef DESATURATION
         vec3 gray = vec3(dot(vec3(0.2126,0.7152,0.0722), color.rgb));
         color.rgb = vec3(mix(color.rgb, gray, m_DesaturationValue));       
     #endif
--- a/jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.frag
@@ -50,7 +50,7 @@ void main(){
     
     #ifdef DESATURATE
         vec3 gray = vec3(dot(vec3(0.2126,0.7152,0.0722), color.rgb));
-        color = vec4(mix(color.rgb, gray, m_DesaturationValue), 0.0);       
+        color.rgb = vec3(mix(color.rgb, gray, m_DesaturationValue));       
     #endif
 
     gl_FragColor = color;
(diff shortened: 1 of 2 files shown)


jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.frag
```diff
@@ -45,6 +47,11 @@ void main(){
            discard;
         }
     #endif
+    
+    #ifdef DESATURATE
```
Ali-RS, line 51: I am not a shader guy here, but shouldn't this be `#ifdef DESATURATION`?

yaRnMcDonuts: Yes you're right, thanks for the catch, I updated the name of the define so the spelling is consistent

riccardobl:
Instead of hard coding gray, i would prefer this to take the color from the material params and mix the alpha channel aswell.
This would expand the usecase to 
- Desaturate and disappear
- Blend to color

riccardobl (approved):

jme3-core/src/main/resources/Common/MatDefs/Misc/Unshaded.j3md
```diff
@@ -59,6 +59,8 @@ MaterialDef Unshaded {
         Float ShadowMapSize
 
         Boolean BackfaceShadows: true
+        
+        Float DesaturationValue
```
Ali-RS, line 63: Should this DesaturationValue be in [0, 1]? 
If so it might be nice to document this on the code.

yaRnMcDonuts: I could add a short comment saying "desaturation value should be in 0-1 range, inclusive" or just "1.0 indicates 100% desaturation" 

I thought of adding a min/max check, although I think it can also work to input a negative desaturation value if you want to increase the saturation, so I left it unbound.

Ali-RS: "1.0 indicates 100% desaturation" sounds good to me.

PR #1454, opened 2020-12-23
Updates The jme3-lwjgl module to use jinput 2.0.9 instead of the 2.0.…

Markil3:
…5 suggested by LWJGL2.

This fixes a gamepad bug discussed in https://hub.jmonkeyengine.org/t/linux-gamepad-input-on-jme3-lwjgl-splits-input-between-two-logical-gamepads


stephengold:
Thanks for your contribution, @Markil3 !

PR #1458, opened 2021-01-11
Fix for lighting vertex shader when used with Fog and InstancedNode.

duncanj:
Same fix as before, but in SPLighting.vert for SinglePassLighting (spotted by Ali_RS).
More info:  https://hub.jmonkeyengine.org/t/shader-compile-error-with-instancednode-lighting-material-and-linearfog/44124/2


Ali-RS:
Thanks

PR #1463, opened 2021-02-01
Gltf unlit

Markil3:
This adds a GLTF loader extention that will allow jMonkeyEngine to use the Common/MatDefs/Misc/Unshaded.j3md material definition whenever a GLTF scene has the "unlit" material on an object. The branch contains a test environment for demonstration.

Note that in some other testing, I've noticed that the supposedly "unlit" material will be affected by shadows, which I did not notice when composing the scene in Blender. This appears to be due to jMonkey's shadows in particular, and not related to GLTF.

I first introduced it [here](https://hub.jmonkeyengine.org/t/gltf-unlit-material-and-punctual-light-extensions/44132). The thread mostly focuses on punctual lights, but I originally started on it for the unlit lights.


stephengold:
Thanks for your contribution to JMonkeyEngine!

stephengold:
Do you understand the cause of the "AssetNotFoundException: gltf/scene/map.gltf"?

Markil3:
It appears that there is a file missing that one of the tests use. The only issue is that I have no idea where it is, or where it may have dropped off (I don't think it was one of my commits). I'll try to track it down.

Markil3:
Found it. The test is actually one of mine, it seems. I was probably using an asset from my own game, but didn't want to include it in the commit. 
Since I think that the TestGLTFUnlit example demonstrates things well enough, I'll probably just remove the test. Give me a few minutes...

Markil3:
And, the fix is green. Sorry I didn't catch that earlier.

jme3-examples/src/main/java/jme3test/light/TestGltfUnlit.java
```diff
@@ -0,0 +1,36 @@
+package jme3test.light;
```
stephengold, line 1: Please add JME's standard copyright notice.

Markil3: Done.

Issue #1459, opened 2021-01-12
merging an infinite BoundingSphere with itself yields invalid result


stephengold:
Issue discovered by the More Advanced Vehicles project, which occasionally caused an assert to fail:
```text
Jan 12, 2021 1:17:37 PM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[main,5,main]
java.lang.AssertionError
	at com.jme3.bounding.Intersection.intersect(Intersection.java:67)
	at com.jme3.bounding.BoundingSphere.intersectsBoundingBox(BoundingSphere.java:683)
	at com.jme3.bounding.BoundingBox.intersects(BoundingBox.java:595)
	at com.jme3.shadow.ShadowUtil$OccludersExtractor.process(ShadowUtil.java:428)
	at com.jme3.shadow.ShadowUtil$OccludersExtractor.addOccluders(ShadowUtil.java:368)
	at com.jme3.shadow.ShadowUtil.updateShadowCamera(ShadowUtil.java:501)
	at com.jme3.shadow.DirectionalLightShadowRenderer.getOccludersToRender(DirectionalLightShadowRenderer.java:194)
	at com.jme3.shadow.AbstractShadowRenderer.renderShadowMap(AbstractShadowRenderer.java:427)
	at com.jme3.shadow.AbstractShadowRenderer.postQueue(AbstractShadowRenderer.java:412)
	at com.jme3.shadow.AbstractShadowFilter.postQueue(AbstractShadowFilter.java:116)
	at com.jme3.post.FilterPostProcessor.postQueue(FilterPostProcessor.java:241)
	at com.jme3.renderer.RenderManager.renderViewPort(RenderManager.java:1103)
	at com.jme3.renderer.RenderManager.render(RenderManager.java:1158)
	at com.jme3.app.SimpleApplication.update(SimpleApplication.java:272)
	at com.jme3.system.lwjgl.LwjglWindow.runLoop(LwjglWindow.java:530)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:632)
	at com.jme3.system.lwjgl.LwjglWindow.create(LwjglWindow.java:466)
	at com.jme3.app.LegacyApplication.start(LegacyApplication.java:463)
	at com.jme3.app.LegacyApplication.start(LegacyApplication.java:424)
	at com.jme3.app.SimpleApplication.start(SimpleApplication.java:127)
[...]

PR #1465, opened 2021-02-02
testcase and fix for issue #1459 (BoundingSphere merge yields NaNs)

stephengold:


stephengold:
I decided that the overflow in `Vector3f.length()` is part of the issue, so fixing it should be part of this PR. Switching to "draft" mode.

PR #1443, opened 2020-12-13
Added KHR_lights_punctual gltf extension

tlf30:
This PR adds support for the GLTF KHR_lights_punctual extension which allows lights from gltf files to be imported.
https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual

Please feel free to give feedback. I am currently looking for advice on implementing the intensity field from the gltf spec into the jme lights. 

GLTF model taken from the KHR_light_punctual repo listed above.

PR #1466, opened 2021-02-02
Light control fix

Markil3:
As @tlf30, @pspeed42 and I were working on #1443, it became apparent that the LightControl class wasn't entirely functional, particularly when it came to spot lights. This PR not only fixes that (including filling out the missing spotlight code in the lightToSpatial method), but also replaces some legacy code relating to directional lights from before version 3.
https://hub.jmonkeyengine.org/t/gltf-unlit-material-and-punctual-light-extensions/44132/29?u=markil3

Note that, when dealing with "world" coordinates when moving the spatial to the light, the rotations aren't always entirely accurate (the distance between the vectors can be as much as 4 units during testing). However, they appear close enough in debugging that I'm not concerned (Line 141 of TestLightControl2Spot).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2021 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
@@ -39,6 +39,7 @@ import com.jme3.light.DirectionalLight;
 import com.jme3.light.Light;
 import com.jme3.light.PointLight;
 import com.jme3.light.SpotLight;
+import com.jme3.math.Quaternion;
 import com.jme3.math.Vector3f;
 import com.jme3.renderer.RenderManager;
 import com.jme3.renderer.ViewPort;
@@ -161,35 +162,38 @@ public class LightControl extends AbstractControl {
      */
     private void lightToSpatial(Light light) {
         TempVars vars = TempVars.get();
-        boolean rot = false, trans = false;
+        Vector3f translation = vars.vect1;
+        Vector3f direction = vars.vect2;
+        Quaternion rotation = vars.quat1;
+        boolean rotateSpatial = false, translateSpatial = false;
 
         if (light instanceof PointLight) {
             PointLight pLight = (PointLight) light;
-            vars.vect1.set(pLight.getPosition());
-            trans = true;
+            translation.set(pLight.getPosition());
+            translateSpatial = true;
         } else if (light instanceof DirectionalLight) {
             DirectionalLight dLight = (DirectionalLight) light;
-            vars.vect2.set(dLight.getDirection()).negateLocal();
-            rot = true;
+            direction.set(dLight.getDirection()).negateLocal();
+            rotateSpatial = true;
         } else if (light instanceof SpotLight) {
             SpotLight sLight = (SpotLight) light;
-            vars.vect1.set(sLight.getPosition());
-            vars.vect2.set(sLight.getDirection()).negateLocal();
-            trans = rot = true;
+            translation.set(sLight.getPosition());
+            direction.set(sLight.getDirection()).negateLocal();
+            translateSpatial = rotateSpatial = true;
         }
         if (spatial.getParent() != null) {
             spatial.getParent().getLocalToWorldMatrix(vars.tempMat4).invertLocal();
-            vars.tempMat4.rotateVect(vars.vect1);
-            vars.tempMat4.translateVect(vars.vect1);
-            vars.tempMat4.rotateVect(vars.vect2);
+            vars.tempMat4.rotateVect(translation);
+            vars.tempMat4.translateVect(translation);
+            vars.tempMat4.rotateVect(direction);
         }
 
-        if (rot) {
-            vars.quat1.lookAt(vars.vect2, Vector3f.UNIT_Y);
-            spatial.setLocalRotation(vars.quat1);
+        if (rotateSpatial) {
+            rotation.lookAt(direction, Vector3f.UNIT_Y).normalizeLocal();
+            spatial.setLocalRotation(rotation);
         }
-        if (trans) {
-            spatial.setLocalTranslation(vars.vect1);
+        if (translateSpatial) {
+            spatial.setLocalTranslation(translation);
         }
         vars.release();
     }
--- a/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
@@ -39,6 +39,8 @@ import com.jme3.light.DirectionalLight;
 import com.jme3.light.Light;
 import com.jme3.light.PointLight;
 import com.jme3.light.SpotLight;
+import com.jme3.math.Matrix4f;
+import com.jme3.math.Quaternion;
 import com.jme3.math.Vector3f;
 import com.jme3.renderer.RenderManager;
 import com.jme3.renderer.ViewPort;
@@ -70,7 +72,7 @@ public class LightControl extends AbstractControl {
          * Means, that the Spatial's transform is "copied"
          * to the Transform of the light.
          */
-        SpatialToLight;
+        SpatialToLight
     }
 
     private Light light;
@@ -128,10 +130,15 @@ public class LightControl extends AbstractControl {
         }
     }
 
+    /**
+     * Sets the light to adopt the spatial's world transformations.
+     *
+     * @author Markil 3
+     * @author pspeed42
+     */
     private void spatialToLight(Light light) {
-
         final Vector3f worldTranslation = spatial.getWorldTranslation();
-        Vector3f worldDirection = spatial.getWorldRotation().mult(Vector3f.UNIT_Z).negateLocal();
+        final Vector3f worldDirection = spatial.getWorldRotation().mult(Vector3f.UNIT_Z).negateLocal();
 
         if (light instanceof PointLight) {
             ((PointLight) light).setPosition(worldTranslation);
@@ -144,24 +151,37 @@ public class LightControl extends AbstractControl {
         }
     }
 
+    /**
+     * Sets the spatial to adopt the light's world transformations.
+     *
+     * @author Markil 3
+     */
     private void lightToSpatial(Light light) {
         TempVars vars = TempVars.get();
+
         if (light instanceof PointLight) {
-
             PointLight pLight = (PointLight) light;
-
-            Vector3f vecDiff = vars.vect1.set(pLight.getPosition()).subtractLocal(spatial.getWorldTranslation());
-            spatial.setLocalTranslation(vecDiff.addLocal(spatial.getLocalTranslation()));
-        }
-
-        if (light instanceof DirectionalLight) {
+            vars.vect1.set(pLight.getPosition());
+            spatial.getLocalRotation().mult(Vector3f.UNIT_Z, vars.vect2);
+        } else if (light instanceof DirectionalLight) {
             DirectionalLight dLight = (DirectionalLight) light;
-            vars.vect1.set(dLight.getDirection()).multLocal(-1.0f);
-            Vector3f vecDiff = vars.vect1.subtractLocal(spatial.getWorldTranslation());
-            spatial.setLocalTranslation(vecDiff.addLocal(spatial.getLocalTranslation()));
+            vars.vect1.set(spatial.getLocalTranslation());
+            vars.vect2.set(dLight.getDirection()).negateLocal();
+        } else if (light instanceof SpotLight) {
+            SpotLight sLight = (SpotLight) light;
+            vars.vect1.set(sLight.getPosition());
+            vars.vect2.set(sLight.getDirection()).negateLocal();
         }
+        if (spatial.getParent() != null) {
+            spatial.getParent().getLocalToWorldMatrix(vars.tempMat4).invertLocal();
+            vars.tempMat4.rotateVect(vars.vect1);
+            vars.tempMat4.translateVect(vars.vect1);
+            vars.tempMat4.rotateVect(vars.vect2);
+        }
+        vars.quat1.lookAt(vars.vect2, Vector3f.UNIT_Y);
+        spatial.setLocalRotation(vars.quat1);
+        spatial.setLocalTranslation(vars.vect1);
         vars.release();
-        //TODO add code for Spot light here when it's done
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


jme3-examples/src/main/java/jme3test/light/TestLightControl2Directional.java
```diff
@@ -0,0 +1,135 @@
...
+        angles[0] += rotAxis.x * ROT_SPEED * tpf;
+        angles[1] += rotAxis.y * ROT_SPEED * tpf;
+        angles[2] += rotAxis.z * ROT_SPEED * tpf;
+        direction.setDirection(new Quaternion().fromAngles(angles).mult(INIT_DIR));
+        super.simpleUpdate(tpf);
+        /*
+         * Make sure they are equal.
+         */
+        if (FastMath.abs(direction.getDirection().normalize().subtractLocal(lightNode.getWorldRotation().mult(Vector3f.UNIT_Z).negateLocal().normalizeLocal()).lengthSquared()) > .1F) {
+            System.err.printf("Rotation not equal: is %s (%s), needs to be %s (%f)\n", lightNode.getWorldRotation().mult(Vector3f.UNIT_Z).normalizeLocal(), lightNode.getLocalRotation().mult(Vector3f.UNIT_Z).normalizeLocal(), direction.getDirection().normalize(), FastMath.abs(direction.getDirection().normalize().subtract(lightNode.getWorldRotation().mult(Vector3f.UNIT_Z).normalizeLocal()).lengthSquared()));
```
aimiebell, line 117: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

jme3-examples/src/main/java/jme3test/light/TestLightControl2Spot.java
```diff
@@ -0,0 +1,142 @@
...
+        angles[0] += rotAxis.x * ROT_SPEED * tpf;
+        angles[1] += rotAxis.y * ROT_SPEED * tpf;
+        angles[2] += rotAxis.z * ROT_SPEED * tpf;
+        spot.setDirection(new Quaternion().fromAngles(angles).mult(INIT_DIR));
+        super.simpleUpdate(tpf);
+        /*
+         * Make sure they are equal.
+         */
+        if (spot.getPosition().subtract(lightNode.getWorldTranslation()).lengthSquared() > 0.1F) {
+            System.err.printf("Translation not equal: is %s (%s), needs to be %s\n", lightNode.getWorldTranslation(), lightNode.getLocalTranslation(), spot.getPosition());
```
aimiebell, line 121: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

jme3-examples/src/main/java/jme3test/light/TestLightControlDirectional.java
```diff
@@ -0,0 +1,127 @@
...
+        angles[0] += rotAxis.x * ROT_SPEED * tpf;
+        angles[1] += rotAxis.y * ROT_SPEED * tpf;
+        angles[2] += rotAxis.z * ROT_SPEED * tpf;
+        lightNode.setLocalRotation(new Quaternion().fromAngles(angles));
+        super.simpleUpdate(tpf);
+        /*
+         * Make sure they are equal.
+         */
+        if (FastMath.abs(direction.getDirection().normalize().subtractLocal(lightNode.getWorldRotation().mult(Vector3f.UNIT_Z).negateLocal().normalizeLocal()).lengthSquared()) > .1F) {
+            System.err.printf("Rotation not equal: is %s (%s), needs to be %s (%f)\n", lightNode.getWorldRotation().mult(Vector3f.UNIT_Z).normalizeLocal(), lightNode.getLocalRotation().mult(Vector3f.UNIT_Z).normalizeLocal(), direction.getDirection().normalize(), FastMath.abs(direction.getDirection().normalize().subtract(lightNode.getWorldRotation().mult(Vector3f.UNIT_Z).normalizeLocal()).lengthSquared()));
```
aimiebell, line 109: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

jme3-examples/src/main/java/jme3test/light/TestLightControlSpot.java
```diff
@@ -0,0 +1,134 @@
...
+        angles[0] += rotAxis.x * ROT_SPEED * tpf;
+        angles[1] += rotAxis.y * ROT_SPEED * tpf;
+        angles[2] += rotAxis.z * ROT_SPEED * tpf;
+        lightNode.setLocalRotation(new Quaternion().fromAngles(angles));
+        super.simpleUpdate(tpf);
+        /*
+         * Make sure they are equal.
+         */
+        if (spot.getPosition().subtract(lightNode.getWorldTranslation()).lengthSquared() > 0.1F) {
+            System.err.printf("Translation not equal: is %s (%s), needs to be %s\n", lightNode.getWorldTranslation(), lightNode.getLocalTranslation(), spot.getPosition());
```
aimiebell, line 113: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
```diff
@@ -127,49 +130,58 @@ protected void controlUpdate(float tpf) {
...
+    /**
+     * Sets the light to adopt the spatial's world transformations.
+     *
+     * @author Markil 3
+     * @author pspeed42
+     */
     private void spatialToLight(Light light) {
-
         final Vector3f worldTranslation = spatial.getWorldTranslation();
+        final Vector3f worldDirection = spatial.getWorldRotation().mult(Vector3f.UNIT_Z).negateLocal();
```
stephengold, line 141: The new `spatialToLight()` method generated a garbage `Vector3f` each time it is invoked. Please use `TempVars` to avoid this.

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
```diff
@@ -127,49 +130,58 @@ protected void controlUpdate(float tpf) {
...
+        if (light instanceof PointLight) {
             PointLight pLight = (PointLight) light;
-
-            Vector3f vecDiff = vars.vect1.set(pLight.getPosition()).subtractLocal(spatial.getWorldTranslation());
-            spatial.setLocalTranslation(vecDiff.addLocal(spatial.getLocalTranslation()));
-        }
-
-        if (light instanceof DirectionalLight) {
+            vars.vect1.set(pLight.getPosition());
+            spatial.getLocalRotation().mult(Vector3f.UNIT_Z, vars.vect2);
```
stephengold, line 165: Altering the spatial's transform in this manner doesn't update the refresh flags, which looks to me like a bug.

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
Markil3: Done.

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
stephengold: Sorry, I must've misread the code.

MeFisto94:
It is nit picking, but would you mind following https://app.codacy.com/gh/jMonkeyEngine/jmonkeyengine/pullRequest?prid=6891987?
Especially because some of those are examples and thus being over-correct style wise might not be bad.
Plus we improve a ficitional number :p

Markil3:
Right, Codacity is now happy with the PR. It does have a duplication rating of +28, but that is likely due to the fact that the four tests I contributed are technically just copy/pastes of each other with minor changes.

stephengold:
I'm not convinced that `lightToSpatial()` does what it ought to. For one thing, it tries to rotate the spatial for a point light (which has no direction) and translate the spatial for a directional light (which has no location). What's the intention here?

Markil3:
At lines 184 through 186, I just set the rotation and translation of the spatial based on vect1 (translation) and vect2 (direction). Since PointLight doesn't have a direction and DirectionalLight doesn't have a position, I just fill those fields with what the spatial already has (lines 168 and 171). This way, a PointLight won't change the rotation of a spatial and DirectionalLight won't change the position. The alternative would be to reset the rotation or translation of the spatial with a default, but I decided against it.

stephengold:
You fill those fields with the *local* translation and direction, relative to the parent. But that doesn't work if the parent happens to be translated and rotated. Wouldn't it be simpler and less error-prone to update only the translation OR rotation in those cases?

Markil3:
I must have forgotten about that. I primarily did it because it looked elegent. That said, trying to get the parent-offsets and such would be harder than its worth. I'll fix that.


Mar 7, 2021, 12:59 PM by notifications@github.com:

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
```diff
@@ -127,49 +128,70 @@ protected void controlUpdate(float tpf) {
...
+            vars.tempMat4.rotateVect(vars.vect2);
         }
 
-        if (light instanceof DirectionalLight) {
-            DirectionalLight dLight = (DirectionalLight) light;
-            vars.vect1.set(dLight.getDirection()).multLocal(-1.0f);
-            Vector3f vecDiff = vars.vect1.subtractLocal(spatial.getWorldTranslation());
-            spatial.setLocalTranslation(vecDiff.addLocal(spatial.getLocalTranslation()));
+        if (rot) {
+            vars.quat1.lookAt(vars.vect2, Vector3f.UNIT_Y);
```
stephengold, line 188: If `vect2` is (0, 1, 0) or (0, -1, 0) then `lookAt()` will produce an invalid rotation.

Markil3: That is a bit odd. I've tested TestLightControl2Spotlight with the spotlight set to multiples of UNIT_Y, but I'm not seeing anything odd. The spotlight does point in the right direction.
That said, if I try replacing the red spotlight model with something a little more detailed, such as Models/MonkeyHead/MonkeyHead.mesh.xml, rotating the spotlight along the X axis reveals some odd behavior where the model does not like to be upside down, and will flip 180 degrees on the local vertical axis so that the crown of the head points in a y direction greater than 0. This flip will happen at the aforementioned vectors. I suspect that this has to do with the limitations of using just directions for the lights rather than a full-blown rotation.
[Demonstration of the problem. Note that the back of the monkey head always points towards the light, even if the monkey did get rotated.](https://github.com/jMonkeyEngine/jmonkeyengine/files/6103391/monkeyrotdemo.zip)
At this point, though, I am a little stumped as to how to go about fixing this. I've had limited success negating the second parameter of lookAt at certain points, but it honestly doesn't feel like the right way to go about it.

stephengold: I think it's reasonable for a `LightControl` in `LightToSpatial` mode to flip a model that doesn't have rotational symmetry around its Z axis. That's a consequence of (non-point) lights having a direction but not a full-blown rotation. It's also a feature worth documenting, in my opinion.

What's puzzling me is that the invalid rotations don't cause visible distortion of the model, even though I can demonstrate distortion using `lightNode.getWorldTransform().transformVector()`. (Perhaps the renderer normalizes the quaternion at some point?) To avoid invalid rotations, I suggest normalizing `quat1` by adding `vars.quat1.normalizeLocal();` immediately after the `lookAt()`.

By the way, this code would be much more readable if you set up aliases for `vect1` and `vect2` in `lightToSpatial()`, as has been done in `spatialToLight()`. Similarly, `rot` and `trans` should have more descriptive names: perhaps `rotateSpatial` and `translateSpatial`.

stephengold: I just had an idea for avoiding sudden model flips. Instead of using `Vector3f.UNIT_Y` as the "up" direction in `lookAt()` you might use the model's current local "up" direction, that is, `spatial.getLocalRotation().mult(Vector3f.UNIT_Y, vars.vect3)`.

Markil3: I tried your last fix, but it doesn't seem to reset rotational changes appropriately. When I replace the cone model with Models/MonkeyHead/MonkeyHead.mesh.xml, the model goes through the X, Y, and Z rotations quite well. However, after it rotates on the random axis, it doesn't reset when it goes back to X, resulting in it rotating at an odd angle.

stephengold: Thanks for testing my idea. So either we have a consistent "up" direction or we avoid sudden flips. While it would be nice to have both features, I don't see a way to achieve that.

stephengold:
I appreciate your cooperation and patience. A few more items:
+ Please add JMonkeyEngine copyright notices to each of the 4 added files.
+ Please update the years in the "LightControl.java" copyright notice to include 2021.
+ When the test apps are run, it's difficult to tell what's happening because (1) the light moves too fast and (2) the initial camera position is too close to the light.
+ The test apps could be made more readable by (1) deleting commented-out code and (2) breaking complicated vector-math expressions across multiple lines in the "make sure they are equal" tests
+ Both "TestLightControl2Spot" and "TestLightControl2Direction" contain the following useless code:
```java
        /*
         * Offset the light node to check global stuff
         */
        Node axis = new Node();
        axis.setLocalTranslation(5, -3, 2.5F);
        axis.setLocalRotation(new Quaternion().fromAngles(FastMath.PI / -4F, FastMath.PI / 2F, 0));
        axis.attachChild(lightNode);
```
It seems to me that `axis` is never attached to the scene graph, and `lightNode` get reparented to the root node immediately afterward.

stephengold (approved):
Much improved.

PR #1467, opened 2021-02-04
Improve & bug fix InstancedGeometry and add new tests.

Ali-RS:
This should fix the issues with using InstancedGeometry and shadow filter reported on the forum: https://hub.jmonkeyengine.org/t/npe-with-instancednode-and-shadow-filter/44240

- Now if all instances of an InstancedGeometry are removed the InstancedGeometry will be properly detached as well and its internal buffer will be cleared.
- Instead of returning a null bound in InstancedGeometry (when it is empty) it will return a bounding box with zero extents. (following the same approach in Node class)
- Renamed `getActualNumInstances` to a more meaningful name `getNumVisibleInstances`
- Fixed a bug in visible instances count calculation that was causing `com.jme3.renderer.RendererException: Attempting to upload empty buffer (limit = 0), that's an error` when used with shadow filter.
- Two test classes are added that demonstrates using instancing with raycasting and shadow. (thanks @duncanj for providing tests)
- Fixed an issue where the LOD level was being ignored by InstancedGeometry!

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/logic/DefaultTechniqueDefLogic.java
+++ b/jme3-core/src/main/java/com/jme3/material/logic/DefaultTechniqueDefLogic.java
@@ -64,8 +64,10 @@ public class DefaultTechniqueDefLogic implements TechniqueDefLogic {
         int lodLevel = geom.getLodLevel();
         if (geom instanceof InstancedGeometry) {
             InstancedGeometry instGeom = (InstancedGeometry) geom;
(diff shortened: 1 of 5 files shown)


Ali-RS:
Going to merge this in 24h, in case someone wants to review this.

stephengold:
I don't understand instancing well enough to comment on the changes. I notice you've added some tests---thank you for that!
A small request: please update the copyright dates of the source files you modify, to at least include 2021.

Ali-RS:
> please update the copyright dates of the source files you modify, to at least include 2021.

Sure, will do.

PR #1471, opened 2021-02-22
avoid an NPE in GltfLoader.findChildren()

Xenokieser:
null pointer exception if the GLTF file references a node that doesn't exist.
I got this problem while tring to convert my GLTF file, i got 440 animations on the file with 3 animation that have joint index missmatch.
Maybe it's better to log the errors detail so the user have all the information to correct the error

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -1086,11 +1086,10 @@ public class GltfLoader implements AssetLoader {
 
     private void findChildren(int nodeIndex) throws IOException {
         JointWrapper jw = fetchFromCache("nodes", nodeIndex, JointWrapper.class);
-        
-        if(jw == null) {
+        if (jw == null) {
             return;
         }
-        
+
         JsonObject nodeData = nodes.get(nodeIndex).getAsJsonObject();
         JsonArray children = nodeData.getAsJsonArray("children");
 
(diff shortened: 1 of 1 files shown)


jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
```diff
@@ -1086,6 +1086,12 @@ public Joint readNodeAsBone(int nodeIndex, int jointIndex, int skinIndex, Matrix
 
     private void findChildren(int nodeIndex) throws IOException {
         JointWrapper jw = fetchFromCache("nodes", nodeIndex, JointWrapper.class);
+        
+        //Check if jw node exist or leave the function
+        if(jw == null){
```
tlf30, line 1091: Can you please add a space between `){` for proper formatting?

tlf30:
In this instance I would think it would be better to throw an IOException as it seems to me the gltf file is corrupt. I don't know what you used to convert your models to gltf, but it is not normal for nodes to be missing. I think even if the model loads with missing nodes in your case, it would just add the possibility for a corrupt gltf to get further in the load process before breaking, which would be even more difficult to troubleshoot.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
tlf30: Also, the comment is explaining the code, which is simple enough to read without the comment. A good comment should not describe the code when the code can easily be read, it should describe why the code is there or what the code is doing from a abstract perspective. 

In this case I see little value added from a comment. Just my opinion.

Xenokieser:
Its true that if there is any corruption in the file it need to be corrected; But from an end user of the plug-in he need to understand what is happening while trying to convert a file. A bulk null-pointer exception in console log is very hard to debug. A more friendly parsing log is better.
As for the throw exception proposal the calling methods don't have any proper exception error handling structure.
If i have more time, i am going to refactor the entire class and add some unit test.

stephengold:
No refactoring or units tests have been forthcoming. I'll finish this PR and integrate it.

stephengold:
I believe this is ready to be integrated now.

PR #1473, opened 2021-02-24
iOS gl3 and some improvements

joliver82:
This is the list of changes:

- Forced compilation using java 8 (options.release parameter) to avoid random NoSuchMethodException for some methods of java.nio.*Buffer classes.
- Modified GLRenderer to get default framebuffer on start. Most gl implementations use 0 but iOS use a random buffer depending on the execution (usually 1). This allows the usage of filters and postprocessors using an additional framebuffer on ios. Previously if you enabled any of those the screen went black.
- Implemented missing methods of gles2.
- Implemented gles3 methods.

Topic at hub: https://hub.jmonkeyengine.org/t/ios-improvements-including-gles3-0/44332

** This PR needs the related SDK PR: https://github.com/jMonkeyEngine/sdk/pull/295


jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -630,6 +631,16 @@ public void initialize() {
...
+
+	IntBuffer tmp=BufferUtils.createIntBuffer(16);
+	gl.glGetInteger(GL.GL_FRAMEBUFFER_BINDING, tmp);
+	tmp.rewind();
+	int fbOnLoad=tmp.get();
+	if(fbOnLoad>0)
+	{
+            // Overriding default FB to fbOnLoad. Mostly iOS fix for scene processors and filters
+	    defaultFBO=fbOnLoad;
+	}
```
Ali-RS, line 643: Can you please fix formatting issues here? 
(spacing before and after '=', '>' and in 'if('  )

joliver82: Done ;)

Ali-RS: Thanks

jme3-ios/src/main/java/com/jme3/renderer/ios/IosGL.java
```diff
@@ -717,6 +718,59 @@ public Object glFenceSync(int condition, int flags) {
     
     @Override
     public void glFramebufferTextureLayerEXT(int target, int attachment, int texture, int level, int layer) {
-        throw new UnsupportedOperationException("OpenGL ES 2 does not support texture arrays");
+        JmeIosGLES.glFramebufferTextureLayer(target, attachment, texture, level, layer);
+    }
+
+    // New methods from GL2 interface which are supported in GLES30
```
MeFisto94, line 724: Supported or unsupported? given that they are stubs atm

joliver82: The comment is not too clear... Methods below are those defined in GL2 which we need to implement because gles30 supports GL{Read|Write}Buffer and also 3D textures. The reason to create the stubs is that glAlphaFunc, glPointSize and glPolygonMode have no gles equivalent but are defined in GL2 interface.

Ali-RS:
Going to merge this in the next 24H if there is no objection.

PR #1474, opened 2021-02-27
remove native Bullet from the Engine

stephengold:
As discussed at [the Forum](https://hub.jmonkeyengine.org/t/suggestions-for-v3-4/43836/49).


tlf30:
The code in `jme3-bullet/src/common/java/com/jme3/bullet` still exists. Is this intentional/due to jme-jbullet? 
https://github.com/jMonkeyEngine/jmonkeyengine/tree/gsw/jme3-bullet/src/common/java/com/jme3/bullet

Just wondering as the gradle build script is now gone in that module.

stephengold:
Yeah, that's intentional. As long as I can remember, jme3-jbullet has used the "src/common" files from jme3-bullet, and that dependency is still apparent here: https://github.com/jMonkeyEngine/jmonkeyengine/blob/ace22d1e0f5bbaa157000d406b7a77d6f96a416b/jme3-jbullet/build.gradle#L9. (The "src/main" files from jme3-bullet are deleted by this PR.)

Issue #1239, opened 2019-12-31
Android sound is broken


Ali-RS:
It seems playing sound on Android has stopped working since we have updated the build script to build Android OpenAL natives by CI (since v3.3.0-alpha5).

Forum post:
https://hub.jmonkeyengine.org/t/sound-not-working-on-android-for-version-3-3-0-alpha5/42645/16

Unless someone is willing to go through the pain of debugging the issue on the native side we probably better revert the build script to use the good old precompiled OpenAL natives (which we were using in pre 3.3.0-alpha5 releases) instead of rebuilding them by CI.

pspeed42:
I concur.  I'm also willing to cut a beta2 at any point we are sure we've resolved some of the more immediate issues we've found this past week.

Ali-RS:
Yet another report
https://hub.jmonkeyengine.org/t/android-ogg-524/43233

Ali-RS:
Putting the workaround here so I won't repeat myself in possibly upcoming reports on the forum: 

> For now, one needs to use the 3.3.0-alpha-2 version for `jme3-android-native.jar`. 
Note that you can yet use JME 3.3.0-stable and higher for other jars and you should be fine.

Ali-RS:
Need to check if commit https://github.com/jMonkeyEngine/jmonkeyengine/commit/38a225cf8e766685ab090948033a0b3a0f822a25 is affecting this issue.

MeFisto94:
It should not, it shouldn't change anything in the bytecode at all. Also note I did a followup commit as it was broken.
It is something we should resolve and debugging that will be funny, helpful but also take a long time, so I'm more seeing that for 3.5 or at least not immediate as part of 3.4 as we've got a lot of other things ahead.

The ogg error seems to be unrelated but also interesting

Ali-RS:
>  take a long time, so I'm more seeing that for 3.5 or...

Could be 5 or more years...!!

I mean until then we can just revert it back to the good old prebuilt natives instead of shipping a broken API... :)

stephengold:
I looked into this. I don't believe we *ever* used pre-built natives in jme3-android-native. The most relevant change between 3.3.0-alpha5 and 3.3.0-beta1 was that the Android native build was moved from Travis to GitHub. Am I missing something?

Ali-RS:
[...]

PR #1475, opened 2021-03-10
Android native sound fix and update openal-soft to 1.21.1

joliver82:
This PR fixes issue https://github.com/jMonkeyEngine/jmonkeyengine/issues/1239 (both sound not working and ogg files not loaded).

Also this PR updates openal-soft to latest (1.21.1).

EDIT: I saw it's failing to compile. Maybe this is caused by an old release of the ndk, I'm using latest (22)


Ali-RS:
Thanks, @joliver82 

What is bsinc_inc.h needed for?

joliver82:
bsinc_inc.h is an autogenerated header file for openal-soft 1.19.1 (see commit https://github.com/joliver82/jmonkeyengine/commit/fc99412186fa3e43912a319d56542ff54b72686b) for 1.21.1 is useless and I've just removed it

MeFisto94:
Well, you bumped the Target Version from 9 to 19, so that certainly requires a more recent NDK. 
It may be worth asking @riccardobl to update https://github.com/jMonkeyEngine/jmonkeyengine/blob/fc8128c139ac6cbde7a8a8db665bb4e1ca6ecbc9/.github/workflows/main.yml#L64 and it's integrated NDK version.

This is something that may even fix the compiler issues.
The "optimal" solution here would be to look at the relevant bytecode and find out why optimization fails, maybe with the help of some clang/gcc guys. Maybe we can make it more clear from the source code to prevent that issue, if it's not fixed in a more recent version already.

I mean the compiler shouldn't like change values into -1 randomly.
That may however be done after this workaround, but having Optimizations, especially on mobile devices, is a good idea

joliver82:
Android api level 19 stands for kitkat 4.4 which was released on 2013 so I assume riccardobl's image is new enough to support it.

I agree that it would be better to properly fix this keeping optimizations enabled but I'm not that good at C/C++ and I didn't want to change code from external libraries (the issues are located at tremor and openal-soft)

About enabling C++14, if setting any -str=c++* flag ndk-build fails because there's plain C files (com_jme3_*.c) in the same library, that's why I didn't.

Ali-RS:
>it would be great if @riccardobl could just rebuild the riccardoblb/buildenv-jme3:android docker image so we would use latest ndk to build jme3 android natives (and also add file package to avoid the ndk line 148 error)

I am not familiar with JME's build env but can this be configured from https://github.com/jMonkeyEngine/buildenv-jme3?

then changing this
https://github.com/jMonkeyEngine/jmonkeyengine/blob/fc8128c139ac6cbde7a8a8db665bb4e1ca6ecbc9/.github/workflows/main.yml#L64

to use `jmonkeyengine/buildenv-jme3`

riccardobl:
> > it would be great if @riccardobl could just rebuild the riccardoblb/buildenv-jme3:android docker image so we would use latest ndk to build jme3 android natives (and also add file package to avoid the ndk line 148 error)
> 
> I am not familiar with JME's build env but can this be configured from https://github.com/jMonkeyEngine/buildenv-jme3?
> 
> And about this:
> 
> https://github.com/jMonkeyEngine/jmonkeyengine/blob/fc8128c139ac6cbde7a8a8db665bb4e1ca6ecbc9/.github/workflows/main.yml#L64
> 
> Should we change it to use `jmonkeyengine/buildenv-jme3:android`? (https://hub.docker.com/r/jmonkeyengine/buildenv-jme3)

Yes. It should be changed to `jmonkeyengine/buildenv-jme3:android` (from https://hub.docker.com/u/jmonkeyengine) . 
As you have noticed I mirrored my repo to the jme org, but i've never actually changed the references in the build script.

joliver82:
I just changed it. The image was built 12 days ago, so it should have latest ndk and compile properly

MeFisto94:
Well it is based on https://github.com/jMonkeyEngine/buildenv-jme3/blob/f6a80d1bf5ae39746b84b2391d9ebb28135b3168/android/Dockerfile, so it should be recent enough

Ali-RS:
Note that `/opt/android/ndk-bundle/build/ndk-build: line 151: file: command not found` is still exist so.

joliver82:
Not a real issue, but I'll add file package to the docker image later

Ali-RS:
Thanks

stephengold:
Is this PR ready to be integrated?

joliver82:
> Note that `/opt/android/ndk-bundle/build/ndk-build: line 151: file: command not found` is still happening. In case this is still an issue!

Created a PR for buildenv to fix this



> Is this PR ready to be integrated?

From my point of view, it is

PR #1478, opened 2021-03-14
Added PBR Terrain shaders

yaRnMcDonuts:
I've added all the shader files necessary for "AfflictedPBRTerrain.j3md" as well as "AfflictedAdvancedPBRTerrain.j3md" and I have also updated the paths to the vert/frag/glslib files accordingly.


stephengold:
Would it be possible to remove "Afflicted" from the filenames and add an example/testcase?

yaRnMcDonuts:
Yes I could remove that from the name. Although It might make it harder for me to maintain this copy of the shader on core since the versions in my game that they were pulled from will have to keep the prefix "afflicted" since they are part of a set of other non-terrain based shaders that share the decal and desaturation features that I titled as affliction when I made them a few years ago to differentiate them from the stock shader they were forked from. 

I can also upload the test cases soon. I am still planning to upload the required test-assets to jme-test-data like we discussed in another thread recently. But as for the .java test case files, where would you like me to locate them?

stephengold:
> as for the .java test case files, where would you like me to locate them?

Hmm. If they're applications (not JUnit tests) then the logical place would be either in jme3-examples/src/main/java/jme3test/material or else in jme3-examples/src/main/java/jme3test/terrain

I'm leaning toward the latter option.

jme3-terrain/src/main/resources/Common/MatDefs/Terrain/PBRTerrain.j3md
```diff
@@ -0,0 +1,366 @@
...
+        Float Metallic_5 : 0.0
+        Float Metallic_6 : 0.0
+        Float Metallic_7 : 0.0
+        Float Metallic_8 : 0.0
+        Float Metallic_9 : 0.0
+        Float Metallic_10 : 0.0
+        Float Metallic_11 : 0.0
+
+
+        Int DebugValuesMode
```
stephengold, line 72: Could we give this a debug value?

stephengold:
I note a bunch of code is commented out. For readability, it might be better to delete that code.
Also, I see some function names and variable names containing "Affliction". Do they have any use outside of your game?

jme3-terrain/src/main/resources/Common/MatDefs/Terrain/PBRTerrain.j3md
yaRnMcDonuts: I could give DebuValuesMode a default value of -1; this would have the same functionality as leaving it undefined. 

As is, DebugValuesMode is intended to be -1 or undefined until the user needs to inspect the albedo/normal/roughness/metallic/ao outputs. The albedo debug mode can also be used to render in an unshaded mode, but is best disabled by default.

yaRnMcDonuts:
Yes the variables with the word afflicted or plagued in front of them correspond to the decal capabilities. The shader basically has an extra alpha map that is named "afflictionTexture" and can be managed with java code using an image raster to splat decals and desaturation onto the terrain by editing the image's R and G channels, while each square pixel in the AfflictionMap represents ~5 square feet of terrain. 

But the functionality would still work for other games. It's been a long time since I've seen any requests on the forums, but I do recall seeing lots of older posts asking how to do this without getting much of a response while I was trying to figure this out myself. Although this functionality is a key feature of my game (which is why I used the word "afflicted" when I couldn't think of any appropriate short adjective to describe something that had a decal applied) would still be useful for other 3d projects that also need basic decals applied to a terrain.

yaRnMcDonuts:
And I will also take some time to go through and remove a lot of the unecessary code that's been left commented out when I get a chance.

A lot of it has to do with the OcclusionParallax feature that I was going to add at the request of other users a long time ago, but then I discovered it killed the framerate for a large scale 3d game, so I wasn't sure if I should leave it there commented out in case other users want to implement it eventually, or just remove it altogether.

jme3-terrain/src/main/resources/Common/MatDefs/Terrain/PBRTerrain.j3md
stephengold: Leave it as undefined, then. But DebugValuesMode is a feature we'll need to document at some point.

stephengold:
> I used the word "afflicted" when I couldn't think of any appropriate short adjective to describe something that had a decal applied

I couldn't think of an adjective either. But perhaps the code would be clearer if you simply replaced "afflicted" with "decal".

Ali-RS:
>replaced "afflicted" with "decal".
Yeah, "decal" is more clear.

yaRnMcDonuts:
I actually avoided the word "decal" on purpose because I have another traditional decal system that better fits the definition of a decal, and I would not consider this terrain's decal-like feature to be a true decal. It is more of a "creep" like effect  (creep is a single, spreading decal-like effect from Starcraft) that can only splat one texture, but it can do so dynamically at a large scale with no FPS drop (unlike standard decals which are more varied but cannot be used as liberally since each real decal would be a new geometry). I also thought the word afflicted was better for describing the part that can alter the saturation of the terrain per-pixel to make textures appear burnt/dead.

I was going to use the word "creep" based on starcraft, but ended up going with afflicted and affliction to avoid copying another game.

Currently my project is not in a position for me to easily refactor these, since I have 6 other forked shaders that share these variable names and a big part of my project would break if changed. So while I do not mind if someone chooses a consistent name for all of these variables, it would be a much larger task for me to refactor all of my shaders and java files that manage the affliction features in order to get myself back on par with core.

I hope it does not seem like I am trying to push the word "afflicted" into the shaders since that corresponds to the name of my game - that is not my intention at all.

If you'd definitely prefer the variables be named "decal" then I won't contest, since this is the version of the shaders for core and I can still go on using my own versions as is, even though that would make it harder to keep my local version equal to the version on core for bug fixes and any potential updates I may add. But I still think the word afflicted or corrupted makes more sense when I think of the parts of the code that use some of these variables for desaturation.

yaRnMcDonuts:
Thinking about it more, I can see that it would make sense to rename some variables to decal rather than affliction. I think I was worrying more about a few other variables that would make less with the word decal in place of afflicted/affliction. 

For example, I could definitely change the set of textures related to the decal-splatting to be named "DecalAlbedoMap" "DecalNormalMap" etc etc, instead of "AfflictionAlbedoMap" "AfflictionNormalMap"

But for other variables like "AfflictionMode_0" "AfflictionMode_1" etc etc then I think it is better to leave the word afflicted since these variables are related to the desaturation splatting and don't do anything to do with the decal

There's some more variable names I've been wanting to change for a while as well, but like I mentioned I unfortunately backed myself into a corner with my main project these came from, where I'll have to fix some unrelated stuff first before I can finally update and rename a lot of variables without breaking things in the project they were pulled from.

pspeed42:
Without having looked at the code to see what it's actually doing... another term used for things that use a texture to "splat" things onto another texture is... splat.

stephengold:
I still want to include this in JMonkeyEngine v3.4, but it looks like it may become the long pole that holds up v3.4 feature freeze.

@yaRnMcDonuts what are your thoughts?

yaRnMcDonuts:
Currently this is high on my to-do list and I'm currently working on the restructuring to my project that stopped me from doing all the renaming and follow-up testing on my device ( I guess I could have kept doing all the updates directly to this branch but figured it would be less total work and better for testing to make my version equal). 

Hopefully it won't take me longer than the next week but I also understand if you need to go ahead with the 3.4 release before this is ready, but I will keep working on it nonetheless.

stephengold:
Thanks for the update. I'm willing to wait another week for this. Please speak up if there's any snag.

yaRnMcDonuts:
I have the current test case set to let the user view the scene with 2 different light probes (one with blue-ish purple sky, and one with orangish sunset colors). One is my own so there is no ilcensing issues there

However the other one titled "fantasy-sky_Probe.j3o " is from from @Ali-RS

Did you make that light probe on your own or do you know if it is licensed under creative commons that I can include it in jme-test data? If not I will only upload the one I made, but since there are no other good light probes in jme-test-data I thought it would be good to include yours if we can (I don't like defaultProbe.j3o which is the only other probe in jme-test-data).

Ali-RS:
>Did you make that light probe on your own or do you know if it is licensed under creative commons that I can include it in jme-test data?

Yes, I generated the probe myself. 

Does the skybox texture that is used to generate the probe matter as well? If so I used [this free skybox texture](https://assetstore.unity.com/packages/2d/textures-materials/sky/fantasy-skybox-free-18353) from the Unity asset store.

yaRnMcDonuts:
I will likely leave the test case without a sky box, since I added some code to scale the DirecitonalLight and AmbientLight to show the shader with day/night cycles, and if I add a sky box then I do not know how to scale that for day/night as well.

That's great to hear you made the lightprobe, I will include both of ours so the test case can be viewed in more than one pbr lighting environment. I think that inspecting models with multiple probes is very helpful when learning to properly use the PBR workflow, so hopefully having the ability to toggle between 2 differing probes will lead new users in that direction.

yaRnMcDonuts:
Everything is uploaded and finished, aside from the variable renaming.

I also need to correct a few asset paths in the test cases to point to the asset's new path in jme-test-data, but first want to ensure that the locations I've uploaded the new textures and light probes are in the correct spot in jme-test-data (let me know if I should move anything)

yaRnMcDonuts (commented):
I'm not entirely happy with this test case I made for the advanced pbr terrain shader. I think I could have found better textures, as there is some unwanted granulation (especially at far distances and for textures with high scale values) that typically appears with bad roughness maps... However these were the best free CC textures I could find quickly.

 Additionally, the test case has an extremely low framerate, and I'm not sure why. I've had the same  AdancedPBRTerrain shader run with way higher FPS in my own game with 3x as many terrains, shadows enabled, x16 AntiAlisasing, plus a bunch of other non-terrain models in the scene. 

So I think something is wrong with the way the terrain in this test case is generated from an image-based heightmap, or something else that I've never encountered since I sculpt my terrains starting with a flat, blank terrain in an editor.

Regardless, this test case definitely still works to show how to use the shader so I will still include it... but if you don't turn on anti-aliasing at x16 then the sparkly granulation on certain texture slots will look much worse.

 It could be something to look into in the future when I (or someone else) has more time to figure out what is causing this test case to run the shader slower than it does in a larger scale game.

stephengold:
> Does the skybox texture that is used to generate the probe matter as well?
Yes, that matters. If you use someone else's skybox, then the probe is a derivative work, not entirely your own.

Ali-RS:
Thanks for the clarification. I am not an expert in this area so I am not sure under what license the light probe should be.

Anyway, this is the link to Unity [EULA](https://unity3d.com/legal/as_terms) in case anyone wants to take a look.



> 2. END-USER's Rights and Obligations
> 2.1
> 
> END-USER may use the licensed Assets only for their intended purpose.
> 2.2
> 
> 2.2.1 Non-Restricted Assets. The following concerns only Assets that are not Restricted Assets: Licensor grants to the END-USER a non-exclusive, worldwide, and perpetual license to the Asset to integrate it only as incorporated and embedded components of electronic applications and digital media and distribute such electronic application and digital media; reproduction and display in distributed physical advertising materials is permitted solely for marketing purposes in respect of such electronic applications or digital media. Except for services software development kits (“Services SDKs”), END-USERS may modify Assets. END-USER may otherwise not reproduce, publicly display, publicly perform, transmit, distribute, sublicense, rent, lease or lend the Assets. It is emphasized that the END-USERS shall not be entitled to distribute or transfer in any way (including, without, limitation by way of sublicense) the Assets in any other way than as integrated components of electronic applications and digital media or in supporting physical marketing materials. Without limitation of the foregoing it is emphasized that END-USER shall not be entitled to share the costs related to purchasing an Asset and then let any third party that has contributed to such purchase use such Asset (forum pooling).

yaRnMcDonuts:
Hmmm it looks like that sky box may not be a good idea to use then. The license type says "Extension Asset" and the short description says "One license required for each individual user." Although I don't know if this means the engine would count as the user or what exactly...

 In the past I got a sky box from the Unity Store but made sure to pick one that was creative commons, without the attribution requirement. If there is an attribution requirement, I still consider it but sometimes look for a better non-attribution asset because it can get tedious keeping track of a huge list of people who have contributed a single asset, unless the asset is really worth it.

I think I should be okay to include my own probe since the scene I baked it from has a skybox that was CC , as well as many custom models with cc textures as well as textures I've paid for. 
However this seems to be a very nuanced area of licensing, as my light probe is a derivative work of paid and cc assets that I have the rights to distribute in my game (which is also derivative art technically) for profit using my paid assets, then I am fairly certain I have the rights to distribute a light probe  - so long as I am not redistributing the original source assets in the scene it was created from.

However since I am no official expert in this field I'll leave the final decision to a core developer.

stephengold:
My reading of the Unity Store license indicates there are significant limitations on how the assets can be used, limitations that JME's license doesn't address.

How about using the "quarry_03" skybox and probe from the jme-vehicles project? It's based on a CC0 texture, its provenance fully documented:

https://github.com/stephengold/jme-vehicles/tree/master/MavCommon/src/main/resources/Textures/Sky/quarry_03

yaRnMcDonuts:
That sounds good, I will definitely remove the prove based on the Unity Sky Box. Mine should be fine based on the licensing from substance designer, but since yours has very detailed licensing documentation I will just remove mine to be safe. 

I have definitely avoided using assets from the Unity store in my game due to their detailed licensing - it sounds like some are free, but only up until your company/organization makes a certain amount per year. So even though my entire game is written in open source JME, it sounds like using a single free asset from the Unity Store could bind you to their paid licenses if you surpass their yearly revenue thresholds.

stephengold:
> it sounds like some are free, but only up until your company/organization makes a certain amount per year

Yet another reason to love JME (as if I needed one!)

yaRnMcDonuts:
I'm about to rename some variables. I definitely think the use of the word 'splat' is better than 'decal' in this situation. Here are my thoughts so far on the renaming we discussed above:

1. Rename `AfflictionTexture` to `AfflictionAlphaMap ` (this texture uses the red channel for desaturation, and the g channel for splatting the rest of the affliction/splat textures (might use b and a for other stuff in the future) so I coined the word "affliction" to clump these two together for naming purposes, but the only other decent thing I thought of would just be calling it DesaturationAndSplatAlphaMap )

2. Rename `AfflictionAlbedoMap`, `AfflictionNormalMap`, `AfflictionEmissiveMap`, etc etc to either of the following:
`SplatAlbedoMap`, `SplatNormalMap`, `SplatEmissiveMap`  or
`AfflictionSplatAlbedoMap`, `AfflictionSplatNormalMap`, `AfflictionSplatEmissiveMap` 

The only concern I'd have is that dropping the word Affliction from some variables may make them get confused with other variables related to the regular terrain texture slot splatting code. Which is why I chose to keep "AfflictionAlphaMap"  "SplatAlphaMap" for the first variable in question.

stephengold:
You've obviously given this some thought. I'll trust your judgment.

stephengold:
@yaRnMcDonuts, any progress to report?

yaRnMcDonuts:
Yes I think that I've finally finished everything required for this PR:

1. I just uploaded the finalized shader files with cleaned up code and some variable/method renaming 
2. The test cases for both pbr shaders are uploaded to jme-examples
3. And the test assets are all uploaded to jme-test-data

I do however need to do one final check of the asset paths to make sure that the test cases all include the "Common" part, as I was testing these test cases by placing the textures in the asset folders of a standard project, rather than rebuilding the engine from source.

It is getting late here, so I will do this final check of all the assetPaths tomorrow with a fresh mindstate, to ensure that I don't overlook any assetpaths or make any typos

Then, if everything else is looking good to you, the PR should be ready to merge.

yaRnMcDonuts:
I deleted the previous 2 light probes and updated it to use your quarry probe from MAV instead. 
I also copy/pasted your license file in the spot i put the quarry probe.

As for the textures I got from cc0textures.com, I added a text file that says the license and where they came from.

Their license page actually says that you don't have to give them credit, and can simply say "Contains assets from CC0Textures.com, licensed under CC0 1.0 Universal" if you want to, but I still left a bit of extra information to follow the format of your license.txt file.

It looks like everything is now done, so long as I did not make any mistakes or miss anything. Let me know if so - but for now I will mark this as ready for review.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,395 @@
+package jme3test.terrain;
```
stephengold, line 1: needs a JME copyright notice and license

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainTest.java
```diff
@@ -0,0 +1,314 @@
+package jme3test.terrain;
```
stephengold, line 1: needs a JME copyright notice and license

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,395 @@
...
+import com.jme3.terrain.heightmap.ImageBasedHeightMap;
+import com.jme3.texture.Image;
+import com.jme3.texture.Texture;
+import com.jme3.texture.Texture.WrapMode;
+import com.jme3.texture.TextureArray;
+import java.util.ArrayList;
+import java.util.List;
+
+
+// Contains assets from CC0Textures.com, licensed under CC0 1.0 Universal.
```
stephengold, line 33: This Java code *uses* assets, it does not *contain* them.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,395 @@
...
+    private AmbientLight ambientLight;
+    private DirectionalLight directionalLight;
+    private boolean isNight = false;
+    
+    private BitmapText keybindingsText;
+
+    public static void main(String[] args) {
+        PBRTerrainAdvancedTest app = new PBRTerrainAdvancedTest();
+        AppSettings s = new AppSettings(true);
+        s.put("FrameRate", 140);
```
stephengold, line 62: Why does this app set the frame rate? Shouldn't it synchronize with the graphics adapter/monitor?

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,395 @@
...
+        matTerrain.setTexture("AlphaMap", assetManager.loadTexture("Textures/Terrain/splat/alpha1.png"));
+        matTerrain.setTexture("AlphaMap_1", assetManager.loadTexture("Textures/Terrain/splat/alpha2.png"));
+        // this material also supports 'AlphaMap_2', so you can get up to 12 texture slots
+
+        // HEIGHTMAP image (for the terrain heightmap)
+        TextureKey hmKey = new TextureKey("Textures/Terrain/splat/mountains512.png", false);
+        Texture heightMapImage = assetManager.loadTexture(hmKey);
+
+        // load textures 
+        // it is IMPORTANT that these MUST all have the same dimensions and format in order to be put into a texture array
```
stephengold, line 91: Aren't the textures all 1K pixels on a side?

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainTest.java
```diff
@@ -0,0 +1,314 @@
...
+    private DirectionalLight directionalLight;
+    private boolean isNight = false;
+    
+    private BitmapText keybindingsText;
+
+    public static void main(String[] args) {
+        PBRTerrainTest app = new PBRTerrainTest();
+        AppSettings s = new AppSettings(true);
+        
+        s.put("FrameRate", 140);
```
stephengold, line 51: Why is the frame rate specified?

jme3-testdata/src/main/resources/Scenes/LightProbes/license.txt
```diff
@@ -0,0 +1,16 @@
+Licensing history for files in "src/main/resources/Textures/Sky/quarry_03"
```
stephengold, line 1: file path looks incorrect

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
yaRnMcDonuts: Calling that line sets 140 fps as the max, but it can still fall to any number below 140 fps.

I can remove it if you'd like, but my GPU makes weird noises when apps do not cap the framerate, so anytime I run a test case I add this line of code so my GPU isn't running at 9000 fps.

jme3-testdata/src/main/resources/Scenes/LightProbes/license.txt
```diff
@@ -0,0 +1,16 @@
...
+* download file:     quarry_03.jpg
+* downloaded from:   https://hdrihaven.com/hdri/?h=quarry_03
+* author:            Sergej Majboroda
+* author URL:        https://hdrihaven.com/hdris/?a=Sergej%20Majboroda
+* license type:      CC0
+* license URL:       https://creativecommons.org/publicdomain/zero/1.0/
+
+Derived files (also licensed CC0):
+* equirec_4k.jpg
+* probe.j3o
```
stephengold, line 13: filename looks incorrect

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
pspeed42: When you run the test, tick the box that says "frame sync" or v sync or whatever.  Setting the frame rate as you are will cause the application to sleep periodically.

jme3-testdata/src/main/resources/Textures/Terrain/PBR/license.txt
```diff
@@ -0,0 +1,9 @@
+Licensing history for files in "src/main/resources/Textures/Terrain/PBR"
+
+All textures were downloaded and derrived from:
+* downloaded from:   https://cc0textures.com/
+* license type:      CC0
+* license URL:       https://docs.cc0textures.com/licensing.html
+
```
stephengold, line 7: Please list names of the files that were downloaded.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
yaRnMcDonuts: Okay I will remove it from both test cases.

For reference for other projects: I see that settings VSync true will cap the app at 60fps, is there any way to increase this without setting the framerate as I've previously done?

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
yaRnMcDonuts: I added the copyright notice from another test case, let me know if it looks alright

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
pspeed42: Vsync is locking refresh to the display's refresh rate.  It prevents tearing and prevents drawing frames that are just going to be thrown away anyway.  Also, at least in my day, the app can begin drawing the next frame already while the driver waits for the vsync.

I'm not sure why you would want to "increase it" unless you think "FPS" is a more useful metric than "how long did my frame take to render"... but "how long did my frame take to render" is a Waaaaaaay more useful metric and already provided by the built in profiling.

jme3-testdata/src/main/resources/Scenes/LightProbes/license.txt
yaRnMcDonuts: I've updated it to match the proper location

jme3-testdata/src/main/resources/Scenes/LightProbes/license.txt
yaRnMcDonuts: I've removed "equirec_4k.jpg" from the list since I did not use it, and updated probe.j3o to match name I gave it of "quarry_Probe.j3o"

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
yaRnMcDonuts: I personally have no problem with my apps being capped at 60fps, however I know some younger kids who are into mainstream games nowadays like to brag about how their computer runs a game at 120 fps while their friends only run at 80, so it is really not important, but that is the only reason I asked. 
There are undoubtedly many kids who think a game is bad for running at 60 fps when they play fortnite at 120 fps, however irrational that thought may be.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
yaRnMcDonuts: Yes I left this comment to let the user know that the fact they share a size and format is important, since that may not be common knowledge when using texture arrays. 
Would you suggest removing the comment or changing it to something that sounds better?

yaRnMcDonuts:
Am I supposed to mark a conversation as "resolved" after I've done what you asked and am ready for you to review it again? Or am I supposed to wait for you to do that. I'm not an expert with github, so my apologies if I wrongly marked anything as resolved before you checked that I properly addressed your review.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
pspeed42: They are free to uncheck the box.

Making "vsync run faster" doesn't make any sense, though.  It's like "how can I make a day longer than a day?"

jme3-testdata/src/main/resources/Textures/Terrain/PBR/license.txt
yaRnMcDonuts: I've added a list of all the textures in that folder

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
yaRnMcDonuts: I changed the wording to reflect that. I probably don't need that line here anymore since I've copied your format for a more detailed license.txt file but I will leave the updated comment here for added documentation

stephengold:
> Am I supposed to mark a conversation as "resolved" after I've done what you asked and am ready for you to review it again?

Well, if you're *sure* the conversation is resolved, you can mark it as such, but it's safer to leave it for the reviewer to decide.

yaRnMcDonuts:
My last two minor commits seem to have both failed. I only updated one line of code with one commit, and the other was just to remove comments I missed in my last check, so I'm unsure what triggered the failure.

stephengold:
> My last two minor commits seem to have both failed.

Not due to anything you changed. Sometimes CI builds fail due to network issues. I'll re-run the CI jobs for the latest commit.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
stephengold: Delete the portions of the comment that say they're not currently the same size.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,428 @@
...
+        Texture metallicRoughnessAoEiMapGravel = assetManager.loadTexture("Textures/Terrain/PBR/Gravel_015_PackedMetallicRoughnessMap.png");        
+        Texture metallicRoughnessAoEiMapGrass = assetManager.loadTexture("Textures/Terrain/PBR/Ground037_PackedMetallicRoughnessMap.png");        
+        Texture metallicRoughnessAoEiMapMarble = assetManager.loadTexture("Textures/Terrain/PBR/Marble013_PackedMetallicRoughnessMap.png");        
+        Texture metallicRoughnessAoEiMapRoad = assetManager.loadTexture("Textures/Terrain/PBR/Tiles083_PackedMetallicRoughnessMap.png");
+        
+        
+        // put all images into lists to create texture arrays.
+        //IMPORTANT to note that the index of each image in its list will be sent to the material to tell the shader to choose that texture from the textureArray when setting up a texture slot's mat parans
+        
+        List<Image> albedoImages = new ArrayList();
```
stephengold, line 159: use diamond notation to avoid an unchecked conversion

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,428 @@
...
+        List<Image> normalMapImages = new ArrayList();
```
stephengold, line 160: use diamond notation to avoid an unchecked conversion

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,428 @@
...
+        List<Image> metallicRoughnessAoEiMapImages = new ArrayList();
```
stephengold, line 161: use diamond notation to avoid an unchecked conversion

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,428 @@
...
+        
+        keybindingsText = new BitmapText(assetManager.loadFont("Interface/Fonts/Default.fnt"));
+        keybindingsText.setText("Press 'N' to toggle day/night fade (takes a moment) \nPress 'P' to toggle tri-planar mode");
+        getGuiNode().attachChild(keybindingsText);
+        keybindingsText.move(new Vector3f(200,120,0));
+        
+    }
+    private ActionListener actionListener = new ActionListener() {
+
+        public void onAction(String name, boolean pressed, float tpf) {
```
stephengold, line 351: add @Override notation here

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,428 @@
...
+            if (name.equals("triPlanar") && !pressed) {
+                triPlanar = !triPlanar;
+                if (triPlanar) {
+                    matTerrain.setBoolean("useTriPlanarMapping", true);
+                    // planar textures don't use the mesh's texture coordinates but real world coordinates,
+                    // so we need to convert these texture coordinate scales into real world scales so it looks
+                    // the same when we switch to/from tr-planar mode (1024f is the alphamap size)
+                    matTerrain.setFloat("AlbedoMap_0_scale", 1f / (float) (1024f / dirtScale));
```
stephengold, line 359: Casting to float is unnecessary here and (in the following statements) because float divided-by float always yields a float. Please remove the casts.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,428 @@
...
+            e.printStackTrace();
+        }
+
+        
+        terrain = new TerrainQuad("terrain", 65, 513, heightmap.getHeightMap());//, new LodPerspectiveCalculatorFactory(getCamera(), 4)); // add this in to see it use entropy for LOD calculations
+   //     TerrainLodControl control = new TerrainLodControl(terrain, getCamera());
+  //      control.setLodCalculator(new DistanceLodCalculator(65, 2.7f)); // patch size, and a multiplier
+     //   terrain.addControl(control);
+        terrain.setMaterial(matTerrain);
+        terrain.setModelBound(new BoundingBox());
```
stephengold, line 303: Why do you set the model bound and then update it? Doesn't `new TerrainQuad` yield a valid bound?

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,428 @@
...
+            }
+            if (name.equals("toggleNight") && !pressed) {
+                isNight = !isNight;
+                //ambient and direcitonal light are faded smoothly in update loop below !
+                
+            }
+        }
+    };
+    
+        private float dayLightIntensity = 1.0f;
```
stephengold, line 387: Please declare fields *before* methods: `dayLightIntensity`, `nightLightIntensity`, and `actionListener`.
Also, please declare all 3 to be `final`.

stephengold:
> I'll re-run the CI jobs for the latest commit.
Re-running didn't resolve the failures. I'll take a closer look.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
yaRnMcDonuts: I'm not sure actually, but I think you are right that it is unnecessary. Some parts of this test case were not written by me (this part is one of them) and in this case I left this line of code there in case it was necessary for the way the terrain was generated from a heightmap. 'm fairly certain I should be able to remove that line without breaking anything, but I will double check in my local copy of the test case and then remove the setModelBound() call from both test cases

yaRnMcDonuts:
> Edit: `LightProbe.setBounds()` has been deleted from the "master" branch. It was a no-op anyway, so you should avoid using it in "PBRTerrainTest.java" and "PBRTerrainAdvancedTest.java".

After looking through the LightProbe code on master it is looking like the new way is to define a `new SphereProbeArea(location, radius); `and send that to the LightProbe with the .`setArea() `method.

Does this sound right? Just want to confirm, since my own projects are not running the version of jme with these changes yet for me to quickly confirm on my own.

stephengold:
Sorry, I'm not intimately familiar with the `LightProbe` class yet. I do know that the "quarry_03/probe.j3o" has a radius of 100. If that's sufficient for your test, then you don't need to override it.

yaRnMcDonuts:
The terrain is size 1024, but it looks like I managed to properly set the location and radius with the new code. I believe I have also updated the code for both test cases to account for everything you mentioned in your most recent reviews, let me know if I overlooked anything or if there is anything else.

stephengold:
When I tried running `PBRTerrainTest` I got a crash:
```text
Apr 10, 2021 9:02:37 AM com.jme3.app.LegacyApplication handleError
com.jme3.asset.AssetNotFoundException: Scenes/lightprobe/quarry_Probe.j3o
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:387)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:441)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:446)
	at jme3test.terrain.PBRTerrainTest.simpleInitApp(PBRTerrainTest.java:274)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:136)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:218)
```

When I tried running `PBRTerrainAdvancedTest` I got a warning and a crash:
```text
WARNING: In technique 'Default':
Define 'HORIZON_FADE' mapped to non-existent material parameter 'HorizonFade', ignoring.
Apr 10, 2021 8:59:22 AM com.jme3.app.LegacyApplication handleError
com.jme3.asset.AssetNotFoundException: Scenes/lightprobe/quarry_Probe.j3o
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:387)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:441)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:446)
	at jme3test.terrain.PBRTerrainAdvancedTest.simpleInitApp(PBRTerrainAdvancedTest.java:348)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:136)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:218)
```

stephengold:
Both test apps need class javadoc to explain what they test and how they differ from one another.

Both test apps have unused imports that should be deleted: "com.jme3.bounding.BoundingBox" and "com.jme3.bounding.BoundingSphere".

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainTest.java
```diff
@@ -0,0 +1,348 @@
...
+        @Override
+        public void onAction(String name, boolean pressed, float tpf) {
+            if (name.equals("triPlanar") && !pressed) {
+                triPlanar = !triPlanar;
+                if (triPlanar) {
+                    matTerrain.setBoolean("useTriPlanarMapping", true);
+                    // planar textures don't use the mesh's texture coordinates but real world coordinates,
+                    // so we need to convert these texture coordinate scales into real world scales so it looks
+                    // the same when we switch to/from tr-planar mode (1024f is the alphamap size)
+                    matTerrain.setFloat("AlbedoMap_0_scale", 1f / (1024f / dirtScale));
```
stephengold, line 106: The expression can be simplified to `dirtScale / 1024f'. Likewise in the following lines.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainTest.java
```diff
@@ -0,0 +1,348 @@
...
+        terrain = new TerrainQuad("terrain", 65, 513, heightmap.getHeightMap());//, new LodPerspectiveCalculatorFactory(getCamera(), 4)); // add this in to see it use entropy for LOD calculations
+        TerrainLodControl control = new TerrainLodControl(terrain, getCamera());
+        control.setLodCalculator(new DistanceLodCalculator(65, 2.7f)); // patch size, and a multiplier
+        terrain.addControl(control);
+        terrain.setMaterial(matTerrain);
+        terrain.setLocalTranslation(0, -100, 0);
+        terrain.setLocalScale(1f, 1f, 1f);
+        rootNode.attachChild(terrain);
+
+        Node probeNode = (Node) assetManager.loadModel("Scenes/lightprobe/quarry_Probe.j3o");
```
stephengold, line 274: These 2 lines don't work. Replace them with something like:
```java
LightProbe probe = (LightProbe) assetManager.loadAsset("Scenes/LightProbes/quarry_Probe.j3o");
```

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,429 @@
...
+        @Override
+        public void onAction(String name, boolean pressed, float tpf) {
+            if (name.equals("triPlanar") && !pressed) {
+                triPlanar = !triPlanar;
+                if (triPlanar) {
+                    matTerrain.setBoolean("useTriPlanarMapping", true);
+                    // planar textures don't use the mesh's texture coordinates but real world coordinates,
+                    // so we need to convert these texture coordinate scales into real world scales so it looks
+                    // the same when we switch to/from tr-planar mode (1024f is the alphamap size)
+                    matTerrain.setFloat("AlbedoMap_0_scale", 1f /  (1024f / dirtScale));
```
stephengold, line 117: The expression can be simplified to `dirtScale / 1024f'. The following statements can be similarly simplified.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,429 @@
...
+        terrain = new TerrainQuad("terrain", 65, 513, heightmap.getHeightMap());//, new LodPerspectiveCalculatorFactory(getCamera(), 4)); // add this in to see it use entropy for LOD calculations
+   //     TerrainLodControl control = new TerrainLodControl(terrain, getCamera());
+  //      control.setLodCalculator(new DistanceLodCalculator(65, 2.7f)); // patch size, and a multiplier
+     //   terrain.addControl(control);
+        terrain.setMaterial(matTerrain);
+        terrain.setLocalTranslation(0, -100, 0);
+        terrain.setLocalScale(1f, 1f, 1f);
+        rootNode.attachChild(terrain);
+
+        Node probeNode = (Node) assetManager.loadModel("Scenes/lightprobe/quarry_Probe.j3o");
```
stephengold, line 348: These 2 lines don't work. They should be something like:
```java
        LightProbe probe = (LightProbe) assetManager.loadAsset("Scenes/LightProbes/quarry_Probe.j3o");
```

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainAdvancedTest.java
```diff
@@ -0,0 +1,429 @@
...
+                //ambient and direcitonal light are faded smoothly in update loop below !
+                
+            }
+        }
+    };
+    
+    
+    
+    @Override
+    public void simpleInitApp() {
```
stephengold, line 148: The `simpleInitApp()` method is too long to be easily understood. Please break it up into smaller methods.

jme3-examples/src/main/java/jme3test/terrain/PBRTerrainTest.java
```diff
@@ -0,0 +1,348 @@
...
+            if (name.equals("toggleNight") && !pressed) {
+                isNight = !isNight;
+                //ambient and direcitonal light are faded smoothly in update loop below !
+                
+            }
+        }
+    };
+    
+    @Override
+    public void simpleInitApp() {
```
stephengold, line 135: `simpleInitApp()` is too long for easy comprehension. Please split it into smaller methods.

yaRnMcDonuts: > Both test apps need class javadoc to explain what they test and how they differ from one another.

Do you have a reference I could use to base this off of? I have never generated a javadoc for a testcase, so I'm unsure exactly how this should look.

And my apologies for leaving so much unorganized/unoptimized code. I was unaware that official test cases needed to be. I believe these test cases were based off of this one which is also not as great as it could be, but its also much older: https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-examples/src/main/java/jme3test/terrain/TerrainTestAdvanced.java

So I apologize for making this such a long review process for you, and I appreciate your patience. When I address your most recent reviews, I will also take a slightly more careful look through the test cases to make sure the code is easy to understand and organized.

stephengold: + Code is never perfect, and that's not the standard being applied here.
+ Since these tests are not automated, they are useless if the person running them doesn't know what they are supposed to do and what results/behaviors to expect. 
+ Test cases have to be maintained over time. Since I doubt you'll be the one maintaining them, it's important that they should be easy to understand and not brittle.
+ These test cases are also examples (hence the name jme3-examples) which will help developers figure out how the PBR terrain materials are supposed to be used. Official examples should ... set a good example.

For a good example of class javadoc in jme3-examples, see https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-examples/src/main/java/jme3test/bullet/shape/TestGimpactShape.java

yaRnMcDonuts: Yes I don't think I was right to use the word perfect, and I don't want it to sound like I am contesting your requests to make a good test case. I just was using that other test case mine were based off of as the standard, and in hindsight I should have just put in the effort from the get go to make the test cases better from the start to save you the time of having to review everything so much.

stephengold: The JME repo contains plenty of crappy code. If such code shaped your expectations, I can see why you might be surprised by some of my feedback.

Perhaps a more experienced reviewer than myself would've caught more issues on each review pass. Unfortunately, the project's more experienced reviewers seem to have other priorities these days, so I've had to fill their shoes. I know I've learned much from this process, and I hope you're learning also.

We're close to integrating this PR. I'm looking forward to it!

yaRnMcDonuts:
I've updated the advanced test case to be a lot cleaner and attempted to add the javadoc information following the same format as the testCase you linked, let me know if it looks like I've missed anything or should include more information. 

I tried my best to describe what the test case should look like, but I will also plan to upload some screenshots and possibly short videos once everything is finalized, and then I'll update the comments I left in the test cases to include a link to those as well. That way the users will know for sure if their results from the test case match what they are supposed to look like.

yaRnMcDonuts:
I've also figured out what is causing the extremely low framerate in the advanced test case. Initiating a new TerrainQuad object with a non-null heightMap (as is done in this test case currently) makes the framerate exponentially slower, and it is only noticeable with this test-case because of the large amount of texture-reads that TextureArrays allow for.

When I changed the line of code in the test case to send a null value for the new TerrainQuad's height map, the framerate and sparkly-pixelation problem went away. 

Once this PR is integrated into the engine, I will reference it in a new issue so it can be further investigated.  The easy work-around is to not use a heightmap when generating terrains. However a lot of JME's older terrain based test-cases generate the heightmaps this way, so it could still be an important issue to file.

stephengold:
Have you fixed the crashes in the test apps?

yaRnMcDonuts:
I accidentally overlooked your comment about the proper way to load the quarry probe, but I just fixed it and also tested it to confirm that the advanced test case should run without crashing.

Now I just need to apply the same changes to the standard test case later on tonight when I have some time, and then that one will be ready as well.

yaRnMcDonuts:
Both test cases should be error free and ready for review again now, let me know if anything looks incorrect or needs improved.

stephengold (approved):
Thank you for your contribution to JMonkeyEngine!

yaRnMcDonuts:
Glad to contribute! :smiley:

PR #1485, opened 2021-03-20
AppStateManager profiler for render and post render

tlf30:
Added profile tracking to AppStateManager for render and post render.

If the community thinks this is useful, feel free to merge, otherwise just close. I found it very helpful when tracing which app state was impacting my fps.


pspeed42:
I don't object to it... but it's interesting that of the thousands of app states that I've written in various apps, maybe half a dozen do anything significant on render() or postRender().

tlf30:
Yeah, this came up when troubleshooting Minie's BulletAppState which does blocking on render or post render depending if running in sequential or parallel. I would have never figured out it was bullet without this.

stephengold:
I'm okay with this. Ready for integration.

tlf30:
It is ready to merge when ever you are.

danielperano (approved):

Issue #1452, opened 2020-12-22
convert jme3-examples to the new animation system


stephengold:
A few of the classes in the jme-examples subproject have been converted to the new (v3.3) animation system, thanks to @syluna and others. The following remain to be converted:

 * `TestCameraMotionPath`
 * `TestCinematic`
 * `TestJaime`
 * `TestWalkingChar`
 * `HelloAnimation`
 * `TestAnimationFactory`
 * `TestAttachmentsNode`
 * `TestLodGeneration`

In my opinion, `HelloAnimation` should be a top priority!

stephengold:
Looks like `TestLodGeneration` has been converted.

Furthermore, `TestHWSkinningOld` can be deleted, since `TestHWSkinning` has been converted.

stephengold:
I'm working on TestWalkingChar right now.

stephengold:
I'm working on TestAttachmentsNode.

stephengold:
+ `TestCameraMotionPath` doesn't actually use the new animation system, just `LoopMode`. To resolve the deprecation warning, we need something like #1495.
+ `TestCinematic` and `TestAnimationFactory` use `AnimationFactory`.  It would be much easier to resolve those deprecations if we had resolution of #1453.
+ `TestJaime` uses `AnimationEvent` in cinematics, which has no equivalent in the new system. Another missing-feature issue!

stephengold:
#1508 will provide a new `CinematicEvent` to replace `AnimationEvent`

Issue #1303, opened 2020-02-27
Animation: BlendAction does not honor speed


MeFisto94:
I have to give this a check, chances are I am doing things wrong or there is some documentation about this already, this issue is more of a reminder to me, but:

```java
composer.action("Run").setSpeed(4f);
System.out.println(composer.action("Run").getSpeed()); // 4

BlendAction action = composer.actionBlended("Walking, Running", new LinearBlendSpace(1, 4), "Walk", "Run");
action.getBlendSpace().setValue(4f);

//composer.setCurrentAction("Walking, Running");
composer.setCurrentAction("Run");
```

This sets the speed of the Running Animation to 4.
If the BlendAction is used, the Running Animation clearly isn't up to speed.

Edit: I already found something: actionBlended calls "makeAction" instead of action, so it does not use the "cached" action for `Run` but creates a fresh one from the AnimClips.
Personally I have an odd feeling about using Strings as keys and the Controls being giant hash maps, specifically if all the methods return Actions.

It feels like I am supposed to pass them and keep multiple almost identical Actions, but this only works in some scenarios, depending on what the Control has in it's HashMap.

The problem is we probably cannot/shouldn't refactor the whole code, specifically as the old animation system was designed the same way.

Ali-RS:
>Edit: I already found something: actionBlended calls "makeAction" instead of action, so it does not use the "cached" action for Run but creates a fresh one from the AnimClips.

Even doing so won't change anything I guess. I guess that's not how BlendAction works. You can not set speed individually on each child of BlendAction. That's because AnimComposer reads the speed only from the parent action means BlendAction in this case. I guess you should directly set speed on BlendAction.("Walking, Running" action in your case).

For example, lets' say "Walk" action is 4s and "Run" action is 2s. If you blend them using BlendAction the the "Run" action will be stretched to 4s (max length). 

What you can do is to sync the speed of "Walking, Running" action by a factor of character move speed or the blend value.

MeFisto94:
You are right: https://github.com/jMonkeyEngine/jmonkeyengine/blob/0fd70b81c93adde86a3590db49725a0c68fc3618/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java#L39-L40

And no, I have a problem with that: Run is 1 second, Walk 3 or 4 seconds (there could be other more extreme examples) and no matter how I scale it, I get a relativly long stretched run animation that way.

[...]

PR #1487, opened 2021-03-24
convert HelloAnimation to the new animation system

stephengold:
This addresses a single piece of issue #1452.

To blend smoothly between animations, I used `actionBlended()`, but the resulting "halt" action is much longer than 0.5 seconds. To avoid ignoring keypresses during the "halt" action, I used `simpleUpdate()` to initiate the animation. If someone knows a better way to blend animations in the new system, let me know!


tlf30:
Perhaps the blending issue is related to #1303?

stephengold:
Not directly related, but both contributing to my impression that the new animation system was never quite completed.

Here's an idea: make `Action.setLength()` public. Then we could use `halt.setLength(0.5)` to truncate the `Action` to its desired duration.

Issue #1446, opened 2020-12-13
test whether an EnvironmentCamera is busy


stephengold:
`EnvironmentCamera` has the capability to queue up multiple snapshot jobs. However, it is of limited utility because the camera's position, background color, and image format aren't job-specific. Therefore, they shouldn't be modified while a job is running. To generate multiple snapshots from a single `EnvironmentCamera`, it's important to know whether there are active jobs. This is difficult because the `jobs` field is private.

This issue proposes adding an `isBusy()` method to `EnvironmentCamera` so that applications can avoid modifying the camera's position, background color, and image format if the EC is busy.

stephengold:
Implemented this, but it's not as useful as expected. Most of the latency for generating a `LightProbe` is in calculating spherical harmonics and such. The `EnvironmentCamera` is busy for just one frame.

PR #1490, opened 2021-03-24
resolve issue 1446 (test whether an EnvironmentCamera is busy)

stephengold:


tlf30:
Looks good to me.

jme3-core/src/main/java/com/jme3/environment/EnvironmentCamera.java
```diff
@@ -172,8 +172,7 @@ public Void call() throws Exception {
 
     @Override
     public void render(final RenderManager renderManager) {
-
-        if (jobs.isEmpty()) {
+        if (!isBusy()) {
```
daBlesr, line 175: I would personally prefer to remove negatives where possible,  hence `isIdle()` would be better in my opinion.

jme3-examples/src/main/java/jme3test/light/pbr/RefEnv.java
```diff
@@ -161,5 +162,9 @@ public void done(LightProbe result) {
             rootNode.addLight(probe);
 
         }
+
+        if (eCam.isBusy()) {
+            System.out.println("EnvironmentCamera busy as of frame " + frame);
```
daBlesr, line 167: Question: can it be possible that logging is done here every frame? Because that can slow things down.

jme3-core/src/main/java/com/jme3/environment/EnvironmentCamera.java
pspeed42: It's not really about how it's called.  It's usually more about the default state.  Usually, you try to arrange things so that the default is false.  In which case, isBusy() can make sense.  The steady state is that usually isBusy() returns false.

jme3-examples/src/main/java/jme3test/light/pbr/RefEnv.java
Ali-RS: Nope, based on https://github.com/jMonkeyEngine/jmonkeyengine/issues/1446#issuecomment-806053365 it will be busy just for one frame!

jme3-core/src/main/java/com/jme3/environment/EnvironmentCamera.java
stephengold: I think it'd be okay either way. If the negative is a major concern, I'd prefer to reverse the sense of the `if` statement:
```java
    if (isBusy()) {
       // process jobs
    }
```
instead of
```java
   if (isIdle()) {
     return;
   }
   // process jobs
```

Bear in mind that the motivation of this PR is to add a public method that might be used many times (in applications) not just in this one place.

daBlesr: Makes sense, all good👍

Issue #1491, opened 2021-03-24
Add Getter/Setter methods for attribute fieldOfView in the class Camera.java


capdevon:
As discussed in the forum, would it be possible to add the getter and setter methods for the attribute fieldOfView in the Camera.java class to make it easier to read and write in the next release ?

https://hub.jmonkeyengine.org/t/getter-setter-methods-for-attribute-fieldofview-in-the-class-camera-java-in-jme-3-4/44426

stephengold:
Yes, it's possible. Are you going to write the code and submit a pull request? Or would you prefer someone else did that?

capdevon:
I'm not very familiar with github yet. It would be better if someone else did it, who knows the project well

PR #1492, opened 2021-03-25
Adds getters/setters for Camera field of view and aspect ratio

Markil3:
As discussed in https://hub.jmonkeyengine.org/t/getter-setter-methods-for-attribute-fieldofview-in-the-class-camera-java-in-jme-3-4/44426/9 and mentioned in #1491, I added getter and setter methods for the field of view for perspective cameras (calling these methods on an orthogonal camera has no effect). These can be easily tested on any example that uses the FlyByCamera (i.e., 90% of them), but I added a new test for it just the same that actually displays the current FOV.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
+++ b/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
@@ -382,7 +382,10 @@ public class FlyByCamera implements AnalogListener, ActionListener {
      * @param value zoom amount
      */
     protected void zoomCamera(float value){
-        cam.setFov(cam.getFov() + value * 0.1F * zoomSpeed);
+        float newFov = cam.getFov() + value * 0.1F * zoomSpeed;
+        if (newFov > 0) {
+            cam.setFov(newFov);
+        }
     }
 
     /**
--- a/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
+++ b/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
@@ -382,7 +382,7 @@ public class FlyByCamera implements AnalogListener, ActionListener {
      * @param value zoom amount
      */
     protected void zoomCamera(float value){
-        cam.setFov(cam.getFov() - value * 0.1F * zoomSpeed);
+        cam.setFov(cam.getFov() + value * 0.1F * zoomSpeed);
     }
 
     /**
--- a/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
+++ b/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
@@ -382,28 +382,7 @@ public class FlyByCamera implements AnalogListener, ActionListener {
      * @param value zoom amount
      */
     protected void zoomCamera(float value){
-        // derive fovY value
-        float h = cam.getFrustumTop();
-        float w = cam.getFrustumRight();
-        float aspect = w / h;
-
-        float near = cam.getFrustumNear();
-
-        float fovY = FastMath.atan(h / near)
-                / (FastMath.DEG_TO_RAD * .5f);
-        float newFovY = fovY + value * 0.1f * zoomSpeed;
-        if (newFovY > 0f) {
-            // Don't let the FOV go zero or negative.
-            fovY = newFovY;
-        }
-
-        h = FastMath.tan( fovY * FastMath.DEG_TO_RAD * .5f) * near;
-        w = h * aspect;
-
-        cam.setFrustumTop(h);
-        cam.setFrustumBottom(-h);
-        cam.setFrustumLeft(-w);
-        cam.setFrustumRight(w);
+        cam.setFov(cam.getFov() - value * 0.1F * zoomSpeed);
     }
 
     /**
--- a/jme3-core/src/main/java/com/jme3/renderer/Camera.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/Camera.java
@@ -586,8 +586,7 @@ public class Camera implements Savable, Cloneable {
      * @return Frame of view angle along the Y in degrees, or 0 if the camera is in orthogonal mode.
      */
     public float getFov() {
-        if (!this.parallelProjection)
-        {
+        if (!this.parallelProjection) {
             float fovY = frustumTop / frustumNear;
             fovY = FastMath.atan(fovY);
             fovY /= 0.5F * FastMath.DEG_TO_RAD;
@@ -603,12 +602,10 @@ public class Camera implements Savable, Cloneable {
      * @param fovY   Frame of view angle along the Y in degrees. This must be greater than 0.
      */
     public void setFov(float fovY) {
-        if (fovY <= 0)
-        {
+        if (fovY <= 0) {
             throw new IllegalArgumentException("Field of view must be greater than 0");
         }
-        if (this.parallelProjection)
-        {
+        if (this.parallelProjection) {
             throw new IllegalArgumentException("Cannot set field of view on orthogonal camera");
         }
         float h = FastMath.tan(fovY * FastMath.DEG_TO_RAD * .5f) * frustumNear;
--- a/jme3-core/src/main/java/com/jme3/renderer/Camera.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/Camera.java
@@ -600,10 +600,14 @@ public class Camera implements Savable, Cloneable {
      * Sets the field of view when the camera is in perspective mode. Note that this method has no
      * effect when the camera is in orthogonal mode.
      *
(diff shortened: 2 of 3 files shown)


jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
```diff
@@ -382,28 +382,7 @@ protected void rotateCamera(float value, Vector3f axis){
...
-        }
-
-        h = FastMath.tan( fovY * FastMath.DEG_TO_RAD * .5f) * near;
-        w = h * aspect;
-
-        cam.setFrustumTop(h);
-        cam.setFrustumBottom(-h);
-        cam.setFrustumLeft(-w);
-        cam.setFrustumRight(w);
+        cam.setFov(cam.getFov() - value * 0.1F * zoomSpeed);
```
stephengold, line 385: Shouldn't that be `cam.getFov() + value * 0.1F * zoomSpeed`?

Markil3: You would think so. However, when I did that, scrolling up would zoom out. That didn't strike me as very intuitive, so I flipped it.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -580,6 +581,43 @@ public void setFrustumTop(float frustumTop) {
         onFrustumChange();
     }
 
+    /**
+     * Obtains field of view when the camera is in perspective mode.
+     *
+     * @return Frame of view angle along the Y in degrees, or -1 if the camera is in orthogonal mode.
+     */
+    public float getFov() {
+        return !this.parallelProjection ? currentFov : -1;
```
stephengold, line 590: I think an orthogonal camera should return zero, not -1.

Markil3: Perhaps. I picked -1 because that is my goto "invalid" value for numbers. However, since 0 isn't a valid FOV either, that could work, too.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -580,6 +581,43 @@ public void setFrustumTop(float frustumTop) {
...
+    }
+
+    /**
+     * Sets the field of view when the camera is in perspective mode. Note that this method has no
+     * effect when the camera is in orthogonal mode.
+     *
+     * @param fovY   Frame of view angle along the Y in degrees.
+     */
+    public void setFov(float fovY) {
+        if (!this.parallelProjection && fovY > 0) {
```
stephengold, line 600: I think passing a non-positive argument should throw an exception.

Markil3: Makes sense. Should we throw an exception for trying to set the FOV on orthogonal cameras as well, or just let them silently fail?

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -152,6 +152,7 @@
      * Distance from camera to bottom frustum plane.
      */
     protected float frustumBottom;
+    private float currentFov;
```
stephengold, line 155: Shouldn't this new field be initialized in the constructor and the `read()` method? Shouldn't it be updated in `onFrameChange()`? I'm thinking it might be safer not to cache the fov value...

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -580,6 +581,43 @@ public void setFrustumTop(float frustumTop) {
...
+        }
+    }
+
+    /**
+     * Obtains the aspect ratio when the camera is in perspective mode.
+     *
+     * @return Width:Height ratio, or -1 if the camera is in orthogonal mode.
+     */
+    public float getAspect() {
+        return !this.parallelProjection ? (width / height) : -1;
```
stephengold, line 618: An orthogonal camera should have a valid aspect ratio, not -1.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
Markil3: Considering that this value is primarily used for calculating frustums and nothing else, that makes sense. I think I found a method for reverse engineering them somewhere that I can use...

Markil3:
I honestly have no idea why the Linux javadoc task failed. Is there a way to rerun it? Perhaps it was just a hiccup, especially since everything else (including Linux on JDK11) worked.

stephengold:
The only way I know to rerun a check/job is to rerun them all. Note the "Re-run jobs" button on the right side:
  https://github.com/jMonkeyEngine/jmonkeyengine/pull/1492/checks

Markil3:
I'm not seeing the button. I'm guessing that I don't have the appropriate permissions to so. Its not a big deal. We can rerun them after we finish the last of the issues for this.

jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
stephengold: Changing the semantics of `FlyByCamera.setZoomSpeed()` would be a different PR.

Markil3: Right, I just double-checked the behavior with the master branch, and scrolling up does in fact zoom out. I thought it was something I changed by mistake, but apparently its "intended" behavior. I'll use the plus sign.

Markil3:
O.K., I guess the failed job was a fluke. It worked fine this time.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
stephengold: I lean toward throwing an exception. Silent failures lead to subtle bugs.

jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
```diff
@@ -382,28 +382,7 @@ protected void rotateCamera(float value, Vector3f axis){
...
-        }
-
-        h = FastMath.tan( fovY * FastMath.DEG_TO_RAD * .5f) * near;
-        w = h * aspect;
-
-        cam.setFrustumTop(h);
-        cam.setFrustumBottom(-h);
-        cam.setFrustumLeft(-w);
-        cam.setFrustumRight(w);
+        cam.setFov(cam.getFov() + value * 0.1F * zoomSpeed);
```
stephengold, line 385: It seems like a test is needed to prevent the FOV from going zero or negative. Currently, this code would throw an exception in that case. The old code never tried to set a zero or negative FOV.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -580,6 +580,57 @@ public void setFrustumTop(float frustumTop) {
...
     }
 
+    /**
+     * Obtains field of view when the camera is in perspective mode.
+     *
+     * @return Frame of view angle along the Y in degrees, or 0 if the camera is in orthogonal mode.
+     */
+    public float getFov() {
+        if (!this.parallelProjection)
+        {
```
stephengold, line 590: The opening brace (`{`) should go on the same line as the `if`. 3 instances in this file.

jme3-examples/src/main/java/jme3test/renderer/TestAspectFov.java
```diff
@@ -0,0 +1,120 @@
...
+        fov.setText("FOV: " + cam.getFov());
+        vars.vect1.subtractLocal(0, fov.getLineHeight(), 0);
+        vars.release();
+    }
+
+    @Override
+    public void onAnalog(String name, float value, float tpf) {
+        switch (name) {
+            case FOV_IN:
+                cam.setFov(cam.getFov() - tpf * 10);
```
stephengold, line 113: This line can cause a crash by resulting in a negative FOV. Please add a test to avoid crashing.

jme3-examples/src/main/java/jme3test/renderer/TestAspectFov.java
```diff
@@ -0,0 +1,120 @@
...
+import com.jme3.scene.Spatial;
+import com.jme3.util.TempVars;
+
+/**
+ * Tests the setting of the FOV and aspect ratios.
+ *
+ * @author Markil 3
+ */
+public class TestAspectFov extends SimpleApplication implements AnalogListener {
+    final String FOV_IN = "fovIn", FOV_OUT = "fovOut";
```
stephengold, line 53: Please make the constants `private static` and declare each variable on a separate line.

stephengold (changes requested):

stephengold (approved):
Thanks for the changes. Looks fine now.

Issue #1056, opened 2019-03-31
Proposal: Adopt a standard code format


riccardobl:
I know this might raise some eyebrows, but hear me out, the code style in jme is pretty much the one suggested by google, with some inconsistency https://google.github.io/styleguide/javaguide.html#s4-formatting .

I think we should at least consider adopting  it officially either in an incremental way or with a single formatting commit, in one of the future major releases.

This would provide a  baseline and a quick way to fix formatting issues in PRs.

pspeed42:
I'm kind of on board with this.  In the long run, it turns out not to matter too much to have a detailed style guide but it can make the bulk of code easier to read.  

I think incremental may be best for at least one reason: using the "blame" option in git is waaaay easier if there aren't giant formatting commits.  GIT doesn't make it easy to jump to the previous event for some line of code (as far as I know).

We generally (sort of) adopted a "when in Rome" style policy where when one is editing surgically, try to keep the style of the file they are in... or at least don't reformat the whole file.  Reformatting sort of == "I'm taking over responsibility for maintaining this file".

I'm not sure what the final solution should be.

I think for new code we definitely should point someone to a style guide.  The google one is nearly perfect except the +2 space indent thing.  (There is only one other small point I disagree with but I'm in the minority on this with the rest of the world.  "if (condition)" versus "if( condition )")

Anyway, we should be careful to reject contributions with only small problems that don't really matter but should definitely have some minimum standards.  Like, I'm not going to reject code because the array initialization is eight-space indented instead of four but I will certainly reject CapitilizedDotNetStyle method and variable names that look like class names.  Or for sure reject empty catch blocks, throwing NullPointException on purpose, etc..  Things that really matter to code quality over time.

stephengold:
Because I use the Netbeans-based IDE, I find it convenient to use the Netbeans formatter with default settings: in new code I indent each block 4 spaces and put a space after the "if" keyword.

Being a traditionalist, I limit my lines to 80 characters.

Like Paul, I try to follow the "when in Rome" policy for surgical edits.

tlf30:
I think this would be great. At work we have a netbeans format config that allows us to right click -> Format, and our company coding format gets applied. That way when people forget and don't format their code right, netbeans will auto-fix it. It would be nice to have for JME.

Ali-RS:
I also use the default code formatting style in Netbeans both in my own codes and when making PR for JME .

Ali-RS:
[...]

PR #1494, opened 2021-03-27
Added JmeSurfaceView.java class , for embedding jme-game in a custom xml UI-design.

Scrappers-glitch:
Embedded jme Game on android view to manipulate jme games on custom xml UI designs , could be used within :

* android.app.Activity;
* androidx.appcompat.app.AppCompatActivity; or ComponentActivity.
* androidx.appcompat.app.AlertDialog;
* android.app.Fragment;
* androidx.appcompat.app.Fragment;

OR any OGLES Compatible Context based container.

### DIFF
--- a/jme3-android/src/main/java/com/jme3/app/JmeSurfaceView.java
+++ b/jme3-android/src/main/java/com/jme3/app/JmeSurfaceView.java
@@ -1,3 +1,34 @@
+/*
+ * Copyright (c) 2009-2020 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
 package com.jme3.app;
 
 import android.app.Activity;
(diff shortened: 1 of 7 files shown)


tlf30:
Can you please ensure formatting is correct. I see several formatting errors.

Scrappers-glitch:
> Embedded jme Game on android view to manipulate jme games on custom xml UI designs , could be used within :
> *  android.app.Activity;
> * androidx.appcompat.app.AppCompatActivity; or ComponentActivity.
> * androidx.appcompat.app.AlertDialog;
> * android.app.Fragment;
> * androidx.appcompat.app.Fragment;

> OR any OGLES Compatible Context based container.

Scrappers-glitch:
> Can you please ensure formatting is correct. I see several formatting errors.

Yep , sure , which formatting errors ? the title or in code ?

Scrappers-glitch:
Okay, assuming the 'formatting errors' are the extra white spaces between the if statement conditions & methods , i have removed them , you can refer to a specific if you find something wrong.

stephengold:
The JMonkeyEngine project plans to move toward the [Google Style Guide](https://google.github.io/styleguide/javaguide.html#s4-formatting) which requires whitespace after `if` and on both sides of binary operators such as "!=", "+", and "=". See the discussion at #1056.

jme3-android/src/main/java/com/jme3/app/JmeSurfaceView.java
```diff
@@ -0,0 +1,467 @@
+package com.jme3.app;
```
stephengold, line 1: Please add the JME copyright notice and license.

stephengold:
If @Scrappers-glitch concurs, I could run the new file through a formatter before integrating this PR. That might be easier (for everyone) than trying to format the code by hand.

Scrappers-glitch:
> If @Scrappers-glitch concurs, I could run the new file through a formatter before integrating this PR. That might be easier (for everyone) than trying to format the code by hand.

I will reformat the code now ,& review it again , no worry.👍

Scrappers-glitch:
I think it's better to change the annotations NonNullable from org.jetbrains to androidx.annotations package , so we are on the safe side , & reformat the code, then the PR would be ready to be merged.

Scrappers-glitch:
So @stephengold , 
this : 
```java
        if ( legacyApplication != null ){
```
or this : 
```java 
        if (legacyApplication != null){
```

jme3-android/src/main/java/com/jme3/app/JmeSurfaceView.java
Scrappers-glitch: > So @stephengold ,
> this :
> 
> ```java
>         if ( legacyApplication != null ) {
> ```
> 
> or this :
> 
> ```java
>         if (legacyApplication != null) {
> ```

Ali-RS:
I think `if (legacyApplication != null) {` style is what mostly used in JME.

Scrappers-glitch:
> I think `if (legacyApplication != null) {` style is what mostly used in JME.

Thank you i ensured that , one another question though about the interfaces , so having them in the same class or different files are better ? , because i like to have them at the same file since they are used only by this file.

pspeed42:
I don't see any interfaces defined in this code.

I do see some scary "synchronized" keywords, though.  At least one seems unnecessary because nothing in the method requires synchronization.

The one on update() worries me, though.  Is there a chance that update() will be called from multiple threads at the same time?

Note: it's very very important that JME's update() is called from the same thread every time.  It could break some applications if it isn't.

Scrappers-glitch:
@pspeed42 thanks for reviewing .

> I don't see any interfaces defined in this code.

At the end of the code , you will find them , anyway i have separated them into files , to be with jme style.



> The one on update() worries me, though. Is there a chance that update() will be called from multiple threads at the same time?

That was a long story of using FutureTasks to delay the add of the GlSurfaceView to the users ' screen without delaying jme-Render & it gets to be successful when using synchronized() because that basically makes jme SystemListener gets notified of the changes in the UI thread , but then i discovered a better way which is using a Handler & postDelay it , i see it's better optimized for android UI , the reason for all this , is that it gives an option for the user to use a SplashScreen while loading the scene asynchronously , so after your comment i tested it again w/o synchronized & it seems to be the same so i removed it as no use of it now & they make confusions as well.

if you donot want to delay the add of glsurfaceView use ```startRenderer(JmeSurfaceView.NO_DELAY);``` , but it's better to at least use dealy of 150 ms , again this doesn't delay the renderer , it's not even related to jme.

The legacyApplication update() method gets updated by the ```SystemListener``` interface , UI isn't involved in the initialize() or update() of jme game.

All the new updates with reformating code would be in the next commit.

pspeed42:
Mmm... nested in with the methods like that, I missed the interfaces.

Inner classes/interfaces should go at the bottom.

In this case, it seems like you only use interfaces to avoid having the application extend this class?  It does feel weird to have them in the class like this... which is usually a sign that something is off.

Scrappers-glitch:
> In this case, it seems like you only use interfaces to avoid having the application extend this class? It does feel weird to have them in the class like this... which is usually a sign that something is off.

i like to enclose all things that are used by each other inside one enclosure , so the answer because they are used by this class only , but anyway i have separated them into files as jme does.

Scrappers-glitch:
> > In this case, it seems like you only use interfaces to avoid having the application extend this class? It does feel weird to have them in the class like this... which is usually a sign that something is off.
> 
> i like to enclose all things that are used by each other inside one enclosure , so the answer because they are used by this class only , but anyway i have separated them into files as jme does.



> Mmm... nested in with the methods like that, I missed the interfaces.

& As you quoted here , someone may literally miss them in the mess of methods.

Ali-RS:
By the way, if you use IntelliJ IDE do a `Code -> Reformat Code`. That will fix the issue with white spaces.

Scrappers-glitch:
> By the way, if you are on IntelliJ IDE do a `Code -> Reformat Code` that will fix the formatting issues.

I know , i have taken the time to fix them myself to be able to retain the way that i would use with jme style , the new commits are done.

Ali-RS (changes requested):

stephengold:
It's too late to make changes to this PR. If there's something you want to change, make a new PR.

PR #1495, opened 2021-03-27
un-deprecate LoopMode, since it's still needed for cinematic events

stephengold:
The `LoopMode` enum is part of the "com.jme3.animation" package: the old animation system, which @Nehon deprecated in 2018 when he introduced the new system. However, `LoopMode` is also used extensively in the "com.jme3.cinematic" package, which is not deprecated.

As long as it has a legitimate use, `LoopMode` should not be deprecated.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/animation/LoopMode.java
+++ b/jme3-core/src/main/java/com/jme3/animation/LoopMode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012 jMonkeyEngine
+ * Copyright (c) 2009-2021 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -35,7 +35,6 @@ package com.jme3.animation;
  * <code>LoopMode</code> determines how animations repeat, or if they
  * do not repeat.
  */
-@Deprecated
 public enum LoopMode {
     /**
      * The animation will play repeatedly, when it reaches the end


tlf30:
If it is only used in the cinema package, perhaps it can be moved there.

stephengold:
> If it is only used in the cinema package, perhaps it can be moved there.

I had the same thought. But I imagine that would break a lot of existing apps.

tlf30:
> > If it is only used in the cinema package, perhaps it can be moved there.
> 
> I had the same thought. But I imagine that would break a lot of existing apps.

That is what version numbers are for 🤣.

It would be an easy fix for people.

stephengold:
What would be the benefit of moving the enum to another class?

tlf30:
We could potentially remove the entire animation package in a future release.

MeFisto94:
Hint: The fact that it was used in Animation and Cinematics hints that it is more a util or general class, so moving it into cinematic may or may not cause the same issue further down the line. E.g. we may want Loop Modes for the current animation as well, you may want that for sound etc.

But then one could argue if we want dedicated enums for enhanced type safety.

stephengold:
> We could potentially remove the entire animation package in a future release.

In that case, wouldn't the future release be a better time to introduce such breaking changes?

tlf30:
> > We could potentially remove the entire animation package in a future release.
> 
> In that case, wouldn't the future release be a better time to introduce such breaking changes?

Yeah, that would probably be a good idea. Perhaps a copy can be created in the cinema package for specific use there, and keeping LoopMode deprecated. I'm not entirely sure what this would look like, but functions would have to be deprecated in the cinema package that use the animation's LoopMode, and new functions created that use the new cinema package one. Then in a future release we just remove the deprecated functions and the entire animation package. 

Just a thought.

stephengold:
As things stand, moving the enum to the cinematics package seems inevitable. But now that v3.4 is into beta, the move will wait for another release. Perhaps v3.5 will trim JME back to a single animation system, and we can move the enum then.

PR #1505, opened 2021-04-04
FieldSerializer: do not ignore final fields.

Ali-RS:
This patch allows serializing final fields for sending over the network.

Note this is possible due to obtaining reflection access to the final field otherwise it will throw IllegalAccessException.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/c3e06bd507cb22de81cf91e095dca4ca2f29b4e1/jme3-networking/src/main/java/com/jme3/network/serializing/serializers/FieldSerializer.java#L104


stephengold (approved):

Ali-RS:
@stephengold thanks for reviewing the PR.

PR #1508, opened 2021-04-05
add a CinematicEvent for the new animation system

stephengold:
This adds a new class `AnimEvent` which approximates the function of `AnimationEvent` in the old animation system. For testing, the `TestCinematic` application was modified to use `AnimEvent` in place of `AnimationEvent`.


stephengold:
Another step toward making the old animation system obsolete!

PR #1516, opened 2021-04-08
privatize package-private loggers

stephengold:
Many JME loggers are declared package-private (with no access modifier). Probably these should all be `public`, `protected`, or `private`. Of these choices, `public` seems the safest.

In a few cases, I added the `final` modifier as well.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/tween/Tweens.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/Tweens.java
@@ -47,7 +47,7 @@ import java.util.logging.Logger;
  */
 public class Tweens {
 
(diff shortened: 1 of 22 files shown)


pspeed42:
final yes.  The rest... hm.

Public is not really right.  One of the ancient-lore style reasons for the package-friendly convention is because when inner classes access the private fields of outer classes then they have to generate some fix-up code (because normally they can't see private members).  So it became a common idiom to just use package-friendly with the idea that nothing is going to cross-talk to those loggers anyway.

Making them public seems weird and it makes them part of the public API for sure.  

If we are going to get spun up about this then it's probably better to make them private.

stephengold:
Thanks for the feedback, @pspeed42 ! I've made them `private` instead.

Issue #1445, opened 2020-12-13
LWJGL3 context restart does not re-enable rendering


tlf30:
When restarting the LWJGL3 context, the window does not resume rendering, but the context does restart. 
See https://hub.jmonkeyengine.org/t/white-screen-on-context-restart/43866/18 for more information and test case.

tlf30:
This was fixed in v3.4.0 with PR #1526. Big thanks to @Markil3!

Issue #844, opened 2018-03-15
Nifty Gui is shown too dark in jMonkeyEngine when app is restarted AND gamma correction is enabled


DomenicDev:
Hello,

I use jMonkeyEngine with Nifty GUI and might discovered a bug. When you have gamma correction enabled and you restart the app with `app.restart()` the gui looks way to dark after the restart. When gamma correction is disabled however, the gui looks fine even after the restart.

Maybe `app.restart()` is doing something which leads to this behavoir?

Here is the related [forum thread](https://hub.jmonkeyengine.org/t/nifty-does-not-properly-display-gui-after-app-restart-when-gamma-correction-is-enabled/40213/
).

Best regards
Domenic

PR #1524, opened 2021-04-14
Calls the renderer's setMainFrameBufferSrgb method after a context restart

Markil3:
As discussed in https://hub.jmonkeyengine.org/t/white-screen-on-context-restart/43866, this should fix a problem with gamma correction not being applied after a context restart.


Markil3:
I was browsing the issues, and this should actually solve #844.

Markil3:
In light of my discoveries in https://hub.jmonkeyengine.org/t/white-screen-on-context-restart-lwjgl-garbage-collection-issue/43866/28?u=markil3, I think we should hold off on this one. My [new code](https://github.com/jMonkeyEngine/jmonkeyengine/compare/master...Markil3:contextRestart) would solve both this one and #1445, but the memory issue does make me hesitate a little to commit it immediately.

Markil3:
Ultimately, #1526 is the more flexible way to pull this off, and I think we should go with that one instead.

PR #1526, opened 2021-04-15
Refactors the context restart code.

Markil3:
https://hub.jmonkeyengine.org/t/white-screen-on-context-restart/43866/28?u=markil3
Now most of the code that goes into initializing the renderer is now repeated every time. This is accomplished by moving most of the logic from the initContextFirstTime() methods into a new initContext(boolean) method. On the first boot, the code ultimately executes the same way. However, every context restart calls initContext(false) now as well. This will repeat all the code from before save for the creation of the GL and Renderer instances as well as input initialization.

At the moment, SimpleApplication now has a new default binding for the TAB key that restarts the context. This is to make it easier to test various examples, and I plan on removing it before we actually merge it.

This should solve issues #844 and #1445. This commit should also replace PR #1524.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
+++ b/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
@@ -69,6 +69,7 @@ public abstract class SimpleApplication extends LegacyApplication {
     public static final String INPUT_MAPPING_CAMERA_POS = DebugKeysAppState.INPUT_MAPPING_CAMERA_POS;
     public static final String INPUT_MAPPING_MEMORY = DebugKeysAppState.INPUT_MAPPING_MEMORY;
     public static final String INPUT_MAPPING_HIDE_STATS = "SIMPLEAPP_HideStats";
+    public static final String INPUT_RESTART_CONTEXT = "SIMPLEAPP_Restart";
 
     protected Node rootNode = new Node("Root Node");
     protected Node guiNode = new Node("Gui Node");
@@ -92,6 +93,8 @@ public abstract class SimpleApplication extends LegacyApplication {
                 if (stateManager.getState(StatsAppState.class) != null) {
                     stateManager.getState(StatsAppState.class).toggleStats();
                 }
+            }else if (name.equals(INPUT_RESTART_CONTEXT)){
+                restart();
             }
         }
     }
@@ -219,6 +222,7 @@ public abstract class SimpleApplication extends LegacyApplication {
 
             if (context.getType() == Type.Display) {
                 inputManager.addMapping(INPUT_MAPPING_EXIT, new KeyTrigger(KeyInput.KEY_ESCAPE));
+                inputManager.addMapping(INPUT_RESTART_CONTEXT, new KeyTrigger(KeyInput.KEY_TAB));
             }
 
             if (stateManager.getState(StatsAppState.class) != null) {
@@ -227,6 +231,7 @@ public abstract class SimpleApplication extends LegacyApplication {
             }
 
             inputManager.addListener(actionListener, INPUT_MAPPING_EXIT);
+            inputManager.addListener(actionListener, INPUT_RESTART_CONTEXT);
         }
 
         if (stateManager.getState(StatsAppState.class) != null) {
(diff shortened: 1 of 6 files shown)


jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
```diff
@@ -92,6 +93,8 @@ public void onAction(String name, boolean value, float tpf) {
                 if (stateManager.getState(StatsAppState.class) != null) {
                     stateManager.getState(StatsAppState.class).toggleStats();
                 }
+            }else if (name.equals(INPUT_RESTART_CONTEXT)){
```
Markil3, line 96: The new input binding is for any testing anyone wants to do. Just let me know when we are ready to merge, and I'll remove it and put it in TestContextRestart.

stephengold:
I confirmed that this PR resolves issue #844, by running the test case from https://hub.jmonkeyengine.org/t/nifty-does-not-properly-display-gui-after-app-restart-when-gamma-correction-is-enabled/40213/19 and repeatedly pressing the 'T' and 'Tab' keys.

jme3-examples/src/main/java/jme3test/renderer/TestContextRestart.java
```diff
@@ -0,0 +1,94 @@
...
+import com.jme3.math.Quaternion;
+import com.jme3.renderer.RenderManager;
+import com.jme3.renderer.ViewPort;
+import com.jme3.scene.Geometry;
+import com.jme3.scene.control.AbstractControl;
+import com.jme3.scene.shape.Box;
+import com.jme3.system.AppSettings;
+
+/**
+ * Tests how gamma correction works after a context restart.
```
stephengold, line 50: Please document the new test. For instance, how does one determine whether it succeeded or failed? Can it be run from the test chooser (which skips `main()`)?

jme3-examples/src/main/java/jme3test/renderer/TestContextRestart.java
```diff
@@ -0,0 +1,94 @@
...
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package jme3test.renderer;
+
+import com.jme3.app.SimpleApplication;
+import com.jme3.input.KeyInput;
```
stephengold, line 35: Please delete the 3 unused imports.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglContext.java
```diff
@@ -174,7 +174,28 @@ protected int getNumSamplesToUse() {
         return samples;
     }
 
+    /**
+     * Reinitializes the relevent details of the context. For internal use only.
```
stephengold, line 178: Please correct the spelling "relevent" -> "relevant".

stephengold:
I ran `TestContextRestart` with both jme3-lwjgl and jme3-lwjgl3, and noticed it behaves differently depending on the LWJGL version.

With jme3-lwjgl (LWJGL 2.9.3) pressing 'Tab' causes the box to disappear. With jme3-lwjgl3 (LWJGL 3.2.3 build 13) pressing 'Tab' causes a momentary flicker but the box does not disappear.

stephengold:
@Markil3 do you intend to act on my comments?

Markil3:
Things have gotten a bit crazy over on my end. I'll try to finish these by the weekend, though. Sorry! Completely forgot!

Markil3:
So, I investigated your issue, and it seems to be linked to the AppSettings.setRenderer setting. I can change this to whatever I want on jme3-lwjgl3, but trying to set this on jme3-lwjgl seems to cause problems. I'm investigating it now, but at least it seems to work on the recommended library.

Markil3:
Well, this is interesting: moving the call to initialize the GL renderer so that it only triggers upon the first run seems to solve the issue. However, trying to do the same on the jme3-lwjgl3 library breaks it if I'm using something other than OpenGL 2.0.

No idea what's going on underneath, but it works without any major problems, so I guess I'll just apply that fix.

jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglContext.java
```diff
@@ -251,42 +251,65 @@ protected int getNumSamplesToUse() {
         return samples;
     }
 
+    /**
+     * Reinitializes the relevent details of the context. For internal use only.
```
stephengold, line 255: Please correct the spelling of "relevant". "Relevent" is not a word in English.

stephengold:
In the interest of keeping things moving, I've decided to integrate this and fix the spelling later.

stephengold:
@Markil3 Thank you for contributing to JMonkeyEngine!

stephengold:
Verified the fix in v3.4.0-beta4.

Issue #1528, opened 2021-04-22
JME3 missing support for MacOSX_ARM64


kode54:
Native library support can be expanded to include arm64 support for Apple Silicon Macs. The architecture name starts with "arm", and there is no 32 bit to support on macOS ARM. I already have tested lwjgl built for fat binaries, for use with Minecraft, it only required either Azul's Zulu builds of Java, or possibly Amazon's builds, but I haven't tested the latter.

I tried an existing built application jar with the arm64 java, but obviously, it failed when it unpacked its i386/x86_64 libopenal.dylib and couldn't load it. Technically, you can even use lipo to bundle the ppc32 and ppc64 builds into it as well? Or does arm64 supporting lipo disable support for even i386? I guess that could mean that those get bundled into yet another binary in the jar.

Ali-RS:
Support for ARM64 (Windows and macOS) should be available when LWJGL3 v 3.3.0 released. It is in Early Access. You can get it from here if you want to give it a try: https://www.lwjgl.org/download

More details: https://github.com/LWJGL/lwjgl3/issues/601

stephengold:
In order for JME to load the correct natives, we'll need to add a new value to `com.jme3.system.Platform` and add cases to `JmeSystemDelegate.is64Bit()` and `JmeSystemDelegate.getPlatform()` to recognize such systems.

@kode54, what are typical values of `System.getProperty("os.name")` and `System.getProperty("os.arch")` on arm64 Apple Silicon Macs? (These may be obtained by entering "java -XshowSettings:all" at a shell prompt.)

Ali-RS:
Oh, Isee, you are right.

kode54:
```
os.arch = aarch64
os.name = Mac OS X
```

tlf30:
It looks like the windows arm64 `os.arch` is set to `aarch64` in the jdk.
https://github.com/openjdk/jdk/blob/bebfae48e3f1643f2456d680d170dcc22f7231bf/src/java.base/windows/native/libjava/java_props_md.c#L574

The `os.name` does not look like it gets modified at all.

stephengold:
I'll extend PR #1530 appropriately.

kode54:
@tlf30 Maybe you could tell me where I could find a native aarch64 build of openjdk for Windows? I could test things there as well, when the time comes.

tlf30:
A pre-release build can be found here:
[...]

PR #1530, opened 2021-04-23
address issue #1528 (distinguish ARM macOS/Windows from other platforms)

stephengold:
We want to be able to distinguish ARM Macs from X86/PPC Macs so that, as soon as LWJGL v3.3 released, JME can also support such platforms.


tlf30:
LWJGL v3.3 also adds support for Windows ARM64. It would be great to get that in here too, unfortunately I am not sure on the platform strings for that as I do not own a ARM64 windows machine.

stephengold:
I agree, but unless we obtain property strings soon, let's make that a separate PR.

stephengold:
Made WIP while I incorporate new information posted at issue #1528.

stephengold:
Ready for review again.

tlf30:
I don't have a way to test, but the changes look good.

PR #1537, opened 2021-05-04
AnimEvent: solve bug where onStop() kills prior event on the same layer

stephengold:
This solves a bug seen in `TestJaime` where anim events in the `Cinematic` were occasionally suppressed. This happened when 2 events were scheduled to play consecutively on the same layer and the later event's `onStart()` was invoked before the earlier event's `onStop()`. In this situation `onStop()` would remove the wrong action from the layer.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -457,6 +457,37 @@ public class AnimComposer extends AbstractControl {
     }
 
     /**
+     * Access the manager of the named layer.
+     *
+     * @param layerName the name of the layer to access
+     * @return the current manager (typically an AnimEvent) or null for none
+     */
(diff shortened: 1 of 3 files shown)


capdevon:
Hi Stephen,
perhaps it is better to use the `animComposer.getAction(actionName)` method instead of the `animComposer.action(actionName)` method which is also used to create actions.

as-is
```        
Action currentAction = composer.getCurrentAction(layerName);
Action eventAction = composer.action(actionName);
if (currentAction == eventAction) {
	composer.removeCurrentAction(layerName);
}
```

to-be
```        
Action currentAction = composer.getCurrentAction(layerName);
Action eventAction = composer.getAction(actionName);
if (currentAction == eventAction) {
	composer.removeCurrentAction(layerName);
}
```

stephengold:
Thanks for the suggestion.

stephengold:
I confirmed that the issue is resolved in v3.4.0-beta3

PR #1539, opened 2021-05-06
Revived broken android examples

Scrappers-glitch:
Revived Broken android examples, through a runnable android module, to run : 
<ul>
<li>Load the module ```jme3-android-tests``` inside android studio</li>
<li>Change/Update/Downgrade dependencies & run on emulator or a physical device</li>
</ul>


stephengold:
Thank you for your contribution to the project. Some questions:

1. Which of the added files are your own work, created from scratch?
2. What issue(s) will this PR solve?
3. How have you tested this PR?
4. If we integrate this, will the Wiki documentation need further updates?
5. Can we avoid duplicating the following files, which already exist in the root project?
  + gradlew, gradlew.bat, gradle-wrapper.properties, gradle-wrapper.jar
  + jbullet.jar, stack-alloc.jar
  + .gitignore files (merge into the one in the root project, perhaps?)

And a request: please add the JMonkeyEngine copyright notice and license to ExampleInstrumentedTest.java

Scrappers-glitch:
> * Which of the added files are your own work, created from scratch?

None of those are mine, it's @iwgeric 's project, but there were missing build files that makes the project non-runnable on android studio, my own work would be submitted in a separate PR.

> * What issue(s) will this PR solve?

Adding Missing build files that cannot be added manually (must be added via the android sdk)

> * If we integrate this, will the Wiki documentation need further updates?

i think the answer no, you don't need to, i haven't added anything new, just refractored @iwgeric 's project into a new 
 separate folder named ```jme3-android-tests```, so that anyone who wants to run this example, he/she would proceed to open that folder only on android studio & everything is fine.
 
 
> * How have you tested this PR?

Ofc, i have tested it multiple times on android studio, before submitting a PR, what it basically does is it lists all the ```jme3-examples``` inside the ```jme3-examples``` module (folder) & the user can choose what to run....



> * gradlew, gradlew.bat, gradle-wrapper.properties, gradle-wrapper.jar

Since android gradle plugin is different from java standard gradle, because android runs on Dalvik-VM (using .dex decoding) not JVM (ie, no main method), so they aren't duplicates, those are the files which were missing that break the application from running.....

Check this if you need more info : https://developer.android.com/studio/releases/gradle-plugin



> * jbullet.jar, stack-alloc.jar

I cannot do ```implementation project(':jme3-jbullet')``` from module (folder) 'jme3-android-tests' inside the module 'jme3-android-examples' which is positioned inside the root module 'jmonkeyengine'.



> * .gitignore files (merge into the one in the root project, perhaps?)

Yeh, we can remove this, it doesn't matter,



> And a request: please add the JMonkeyEngine copyright notice and license to ExampleInstrumentedTest.java

I am going to add jme3 copyright in all the missing files.

So, to sum up, ```jme3-android-examples``` folder would contain : 
### a couple of independant android projects that could be ran directly from android studio by opening them separate
### currently one of those examples is @iwgeric 's project which is this commit
### other examples would be a very simple android examples on how to use ```AndroidHarness``` & ```JmeSurfaceView``` & ```Android UI``` with jme, but i will submit this in a separate commit.

Ali-RS:
So will this not work outside the Android Studio?

Might be wrong but I guess the original idea was to be able to build the APK using the gradle then run the APK in a simulator or from Android Studio.

Scrappers-glitch:
> So will this not work outside the Android Studio?

Only Android Gradle plugin can : 
* Build this app in debugging mode using : ```./gradlew assembleDebug``` 
* Build this app & run on current Active emulator or connected device using : ```./gradlew installDebug```
* Build a release app using keysigning
* run this application whether using ```adb install debug.apk``` on your connected physical device or emulator....

Android Studio uses ```android gradle plugin``` & ```adb platform-tools``` to hide those commands from the user & ease the android environment, but you cannot run directly from intellij without having the Android SDK loaded up, build files created by android gradle plugin not java standard one, important build scripts like gradlew, settings.gradle, etc are inside android module....

You can run this example anywhere assuming that : ```settings.gradle``` is present with a valid android SDK location on your hardware (Android Studio does that for the user when you first open the project using Android Studio on your laptop), gradlew bash script is a must for android apk build, it gets generated directly.

Connecting to my physical device using ```adb TCP/IP Communication feature on a Local Network``` & building android apk using ```./gradlew``` build script w/o android studio : 

image

Scrappers-glitch:
> > So will this not work outside the Android Studio?
> 
> Only Android Gradle plugin can :
> 
> * Build this app in debugging mode using : `./gradlew assembleDebug`
> * Build this app & run on current Active emulator or connected device using : `./gradlew installDebug`
> * Build a release app using keysigning
> * run this application whether using `adb install debug.apk` on your connected physical device or emulator....
> 
> Android Studio uses `android gradle plugin` & `adb platform-tools` to hide those commands from the user & ease the android environment, but you cannot run directly from intellij without having the Android SDK loaded up using `local.properties` from your local disk, build files created by android gradle plugin not java standard one example is `./gradlew` build script, important build scripts like `gradlew`, `settings.gradle`, etc are a must inside an android module....
> 
> You can run this example anywhere assuming that : `local.properties` is present with a valid android SDK location on your hardware (Android Studio does that for the user when you first open the project using Android Studio on your laptop), gradlew bash script is a must for android apk build, it gets generated directly.
> 
> Connecting to my physical device using `adb TCP/IP Communication feature on a Local Network` & building android apk using `./gradlew` build script w/o android studio :
> 
> image

Scrappers-glitch:
Added Jme3 Copyright notice.. :-)

MeFisto94:
Is it really a different gradlew file or just another version? Because typically the android plugin is just a line in the build.gradle file and work automatically. 

I don't see how the binary version of the gradle _wrapper_ (which mostly downloads and launches gradle) needs to be different. 

I am also kinda bothered by jbullet files, but that is mostly due to how badly we support that via maven and stuff (there has been no commit to it since 10? years) and while also making it the primary choice as physics engine, especially on android. 

Anyway: ideally this should work without gradle wrapper and while it's not much cleaner, maybe a dependency using file trees may work [though ideally stack-alloc.jar and jbullet.jar should be part of jme3-jbullet.jar, which is however an upstream issue ever since and also bothering me on the SDK].

Besides that: Hooray! We need proper android support and when leveraging avd we can maybe even have some kind of automated testing further down the line. But also just the ability to download an apk and see how jme works on your mobile device helps us find specific bugs on specific devices etc, so thanks for the PR

Scrappers-glitch:
> I am also kinda bothered by jbullet files, but that is mostly due to how badly we support that via maven and stuff (there has been no commit to it since 10? years) and while also making it the primary choice as physics engine, especially on android

As for ```jbullet```, it's not my choice, the android example author leaves no choice for a physics engine than jbullet, because jme3 examples uses jbullet, & all this example application does is that it lists all jme3-examples & the user selects an example to run.....

In my real apps, i prefer to use ```Minie```, i also prefer dummy android examples which is not the case here....but i want to revive this anyway.

MeFisto94:
> Something very odd happens here, @MeFisto94 , @Ali-RS , @stephengold What do you think, should we wait & investigate more before merging this PR ? ,

Well, in general, we shouldn't merge something as examples which kinda doesn't work without us having a clue what is going on.

> Hmm, yes it seems you are right, i rechecked back the gradlew of both the gradle and the android gradle plugin, the only different tho is a conditional if statement.

So GradleWrapper rarely changes content, but indeed that means a small version mismatch. More important is the gradle version as defined in the gradle/ folder. You can then use `gradlew wrapper` or even `gradlew wrapper --wrapper-version 7.0` or something to update it (do not use 7.0 though).

> gradle-wrapper.properties build file was missing from this android module, afaik this build file specifies the corresponding gradle version used by the compatible android gradle plugin.....that's why i got :

The wrapper should be only needed on the root project, you then do things like `gradlew jme3-android:test`

> besides, this seems to be never true :

Check if this is a property, much like the native builds. Like -PbuildAndroidExamples=true or something.

> i think this PR would fix most of this distraction & confusion by just making the android project independant of jme....but you will need to implement the jme classes from mavenCentral not local disk....

While this is easy, there's a reason it has been done that way (same for the SDK): You can't publish a new jme version just to see if that fixes an example. You need it to work with "live" code.


But now to recap: What is the state, what is the problem?
The exceptions aren't there on your PR, fine, but can you also make it work without the custom gradlew and stuff? Like just bumping some version number somewhere? e.g. android plugin or gradle itself?

Automated Testing will still be a more distant future thing, though a very interesting one. E.g. running avd inside github actions with video acceleration and also somehow validating stuff works (which is/used to be) a research topic on it's own for Apps

stephengold:
> What do you think, should we wait & investigate more before merging this PR?

I'm in no hurry to integrate this PR. I haven't even decided whether or not to include it in the v3.4.0 milestone.

Because I don't have a full understanding of JME's Android support, I'm waiting for @Scrappers-glitch and the various reviewers to reach a consensus among themselves.

Please continue the discussion!

Scrappers-glitch:
> You can't publish a new jme version just to see if that fixes an example. You need it to work with "live" code.

yeh i understand.....So that's why, i thought we may have a one more try.



> : What is the state, what is the problem?
> The exceptions aren't there on your PR, fine, but can you also make it work without the custom gradlew and stuff? Like just bumping some version number somewhere? e.g. android plugin or gradle itself?

After All, i did is : 
* removing the ```never true string conditional if statement```, the ```buildAndroidExamples == "true"``` inside the ```settings.gradle```.....

* specified the local dir of my android sdk inside ```local.properties```....this shouldn't be commited with the engine, because it gets generated automatically by android studio according to the site of sdk on local disk.

* About ```gradle-wrapper.properties``` no problem, it was just a file confusion, it's fine.
& i got the android example working.....before that it didn't work.....but stills the jme3-examples not working.....

now the only problems are :

* From where does ```buildAndroidExamples``` var comes ...? & how i can make it "true", & why it's string not boolean !
* The exception raised by ```AndroidImageLoader``` when i try to load any jme3-example...as if ```jme3-android-natives``` is not accessible at the site

i think it won't hurt if i take more time to understand what causes this exception, if you have a local clone of jme repo on your disk try opening ```jmonkeyengine``` root project on android studio, & try to make it to run through these options : 

* removing the ```never true string conditional if statement```, the ```buildAndroidExamples == "true"``` inside the ```settings.gradle```.....

* specify the local dir of my android sdk manually inside ```local.properties```

Scrappers-glitch:
> I haven't even decided whether or not to include it in the v3.4.0 milestone.

I think, you needn't worry about that anyway, afterall it's just a fix & may be we need a proper documentation on how to deploy android application examples within jmonkeyengine repo, because here the situation seems a little bit complicated than Android Out of the Box.....

I mean jme3.3.2 also needs this fix too...

Scrappers-glitch:
> Check if this is a property, much like the native builds. Like -PbuildAndroidExamples=true or something.
> 
> >

I can see it here  : 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/aa8cceebc81aa126c1a9bcbb167f7dd976504628/gradle.properties#L20

but i cannot find a true assignment...& it should be boolean not String ..?

MeFisto94:
Soo, as you can see, the same is true for the natives. They are false by default.
https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/settings.gradle#L37
is where it is then compared, so it's String on both sides (properties are always strings, I think). 

Also I _think_ groovy isn't so picky about the types, like maybe `"true" == true` works. But don't quote me on that, in this case it should be both strings. So it should also work if you either just change the gradle.properties _or_ use `-PbuildAndroidExamples=true`, I think that's the way to override properties.

If those don't work, we may need some debugging and also looking at bullet-native (which has been removed, need to look into older commits), because there it definitely did work.

> The exception raised by AndroidImageLoader when i try to load any jme3-example...as if jme3-android-natives is not accessible at the site

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-android-examples/build.gradle#L54
That is unexpected, but let's fix the issues one at a time and then see

Scrappers-glitch:
> _or_ use `-PbuildAndroidExamples=true`, I think that's the way to override properties.

Should this statement be elsewhere in build files, enclosed in a statement that detects the SDK location ?!

MeFisto94:
Well, the intention of this statement clearly is, that you need to explicitly need to enable it.
That makes sense for the _native_ builds, that clearly take more time and when the native code hasn't been changed, you can also download our artifacts.

For Android, one could argument, that the presence of an sdk dir can cause examples to be built. But:
A) that will change the behavior, it's up to us to decide if that's good or not.
B) We need to ensure that it's the correct sdk.dir, I guess, because from what I saw, it currently contains some random path by default,
C) There are also things like the CI which explicitly has android SDKs but only wants to compile and not run examples (which however, is due to change in the future hopefully).

But I guess that needs discussion

Scrappers-glitch:
Today, i just changed : 
```java
if(buildAndroidExamples == "true") {

}
```
To

```java
if(buildAndroidExamples){

}
```
& it magically works fine 😂😆, do you have any explanations


About jme3-android-natives, as far as i know to use JNI directly & build natives you need to have 2 plugins installed :
1) Android NDK
2)Android Native/SDK Side by side
I have both installed before now, but they cannot build the current android native module

At least, now the android examples inclusion is fixed, i may need your help to build jme3 natives, any ideas so far ?

stephengold:
Since the natives are build by continuous integration at GitHub Actions, I suggest looking in the workflows:
 + https://github.com/jMonkeyEngine/jmonkeyengine/blob/a1ad505dc5ea540b9edcc161f1261e24afcf3c7d/.github/workflows/main.yml#L72

Try the command used there:
```sh
./gradlew -PuseCommitHashAsVersionName=true --no-daemon -PbuildNativeProjects=true :jme3-android-native:assemble
```

Scrappers-glitch:
> Since the natives are built by continuous integration at GitHub Actions, I suggest looking in the workflows:
> 
> * https://github.com/jMonkeyEngine/jmonkeyengine/blob/a1ad505dc5ea540b9edcc161f1261e24afcf3c7d/.github/workflows/main.yml#L72
> 
> Try the command used there, which is:
> 
> ```shell
> ./gradlew -PuseCommitHashAsVersionName=true --no-daemon -PbuildNativeProjects=true :jme3-android-native:assemble
> ```

Oh, thanks @stephengold for this point, a good point tho, I was a bit to lose my mind trying to find a bug within ```Android.mk``` files really !

Scrappers-glitch:
I have tried to build & install the apk with those flags, but still gets the exception of cannot load native libraries : 
```bash
└──╼ $./gradlew -PuseCommitHashAsVersionName=true -PbuildNativeProjects=true :jme3-android-native:assemble installDebug 
```

stephengold:
> still not working

How not working?

Scrappers-glitch:
I mean still giving me the same exception ```UnSatisfiedLinkError: cannot find decodejme.so```

stephengold:
Have you tried setting the extractNativeLibs flag to true in the "application" element of the "AndroidManifest.xml" manifest?

Scrappers-glitch:
> Have you tried setting the extractNativeLibs flag to true in the "application" element of the "AndroidManifest.xml" manifest?

Hi @stephengold , sorry for late reply, it doesn't matter here, because minSdkVersion = 15 (ICS) specified in gradle which is minSdkVersion < 23(MM), so native libs don't get compressed at the installation time so no need to  decompress them, this is invalid operation, i know it's my fault "i haven't documented the ```extractNativeLibs``` properly", i have corrected that on the forum with android developer link & image, which you can now use as a reference to Minie as well....

And, just to prove that, the PR works without ```extractNativeLibs```, so it must be something else concerned about NDK that i don't know, so i will keep searching....i haven't that much experience with NDK but will try....

Scrappers-glitch:
I will try to download ```jme3-android-natives.jar``` & include it as a jar dependency & see what's the reaction !, if that works, i think we might be building a jar file from ```jme3-android-natives``` during gradlew build to be included inside the apk on installation.....

Scrappers-glitch:
> I will try to download `jme3-android-natives.jar` & include it as a jar dependency

It works, when doing ```implementation files('libs/jme3-android-native.jar')``` after downloading from mavenCentral(like what gradle dependency does(fetching a build)), so i have suspected now, there are problems with gradle that builds the android natives locally(during gradle build), so android java native module isn't build as jar at the installation time.....

This doesn't happen when doing : ```    implementation "org.jmonkeyengine:jme3-android-native:3.4.0-beta4"``` because ```main.yml``` takes the role for providing jars remotely while no one takes this role locally ??!.......

-> using ```implementation files('libs/jme3-android-native.jar')``` would directly provide java module or native libs to be converted to .so for cpp/c or .dex for java.

Scrappers-glitch:
Oh, that's seems to be true, i have just opened ```jmonkeyengine/jme-android-native/build/libs/jme3-android-native-3.5.0-SNAPSHOT.jar``` & only ```META-INF``` manifest file is there, no actual code.......
image

Scrappers-glitch:
Okay, here's my final investigations, the real issue of ```UnSatisfiedLinkError``` is because of android cannot locate the NDK while building (libraries need to be build then copied to a new Build file) which makes jme3-android-natives unBuildable, finally i have found that part that controls the building : 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/035820e3f063e9a520b28e12717735ce8a77aaf9/jme3-android-native/decode.gradle#L64

after an attempt to build i get this permission error, despite specifying ```chmod  +x studio.sh ./studio.sh```: 
```
Execution failed for task ':jme3-android-native:buildNativeLib'.
> A problem occurred starting process 'command '/home/twisted/Android/Sdk/ndk/ndk/22.0.7026061''

Cause: error=13, Permission denied
```

despite of specifying my ndk dir explicitly in here : 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/035820e3f063e9a520b28e12717735ce8a77aaf9/build.gradle#L170
for the ```ndkCommonPath``` string using : 
```
        ndkBuildPath = "/home/twisted/Android/Sdk/ndk/ndk/22.0.7026061"

```
Sorry, i am not good at gradle....but trying to understand, i think the problem was to identify a ```dynamic``` path to the ndk location which is dynamic not manual !

Scrappers-glitch:
> despite of specifying my ndk dir explicitly in here :
> https://github.com/jMonkeyEngine/jmonkeyengine/blob/035820e3f063e9a520b28e12717735ce8a77aaf9/build.gradle#L170
> 
> 
> for the `ndkCommonPath` string using :
> ```
>         ndkBuildPath = "/home/twisted/Android/Sdk/ndk/ndk/22.0.7026061"
> ```
> 
> Sorry, i am not good at gradle....but trying to understand, i think the problem was to identify a `dynamic` path to the ndk location which is dynamic not manual !

Okay, it works now :-) 

### Reading this code again, i seem to come up with the steps that made it work : 

1)  specify the absolute file path for your ndk to build android natives (decodeJme, openALsoft) manually in here : 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/035820e3f063e9a520b28e12717735ce8a77aaf9/gradle.properties#L31

or Dynamically from ```local.properties``` using : 
```
Properties properties = new Properties()
   //override ndkPath with a valid dir if local.properties has a valid path for android ndk
    Properties properties = new Properties()
    FileInputStream fis = new FileInputStream(project.rootProject.file('local.properties'))
    if(fis.available()){
        properties.load(fis)
        if(properties.contains('ndk.dir')) {
            if(new FileInputStream(properties.getProperty('ndk.dir').toString() + File.separator + ndkBuildFile).available()) {
                ndkPath = properties.getProperty('ndk.dir').toString()
            }
        }
    }
    // ndkPath is defined in the root project gradle.properties file
    String ndkBuildPath = ndkPath + File.separator + ndkBuildFile
```
so, i think we should pump this snippet into here : 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/035820e3f063e9a520b28e12717735ce8a77aaf9/build.gradle#L165

2) Multiple opinions : either...

#### do : 
```
if(buildAndroidExamples){
    include 'jme3-android-examples'
}
```
#### instead of :
```
if(buildAndroidExamples == "true"){
    include 'jme3-android-examples'
}
```
#### to load directly android module using android studio. 

#### or use : ```└──╼ $./gradlew -PbuildAndroidExamples=true installDebug``` with ```buildAndroidExamples == "true"``` directly, but android studio cannot access this.....

#### or do a dynamic linking again to the SDK this time (which is the best of all) in here : 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/035820e3f063e9a520b28e12717735ce8a77aaf9/settings.gradle#L37

#### using :

```
if(buildAndroidExamples == "true" || !sdkPath.isEmpty()){
    include 'jme3-android-examples'
}
```
#### after initializing that ```property var``` inside ```gradle.properties``` with empty string.

#### then, assigning it to a valid ```sdk.dir``` inside the local.properties in here : 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/035820e3f063e9a520b28e12717735ce8a77aaf9/build.gradle#L36

using : 
```
    //override sdkPath for a valid dir in local.properties
    Properties properties = new Properties()
    FileInputStream fis = new FileInputStream(project.rootProject.file('local.properties'))
    if(fis.available()){
        properties.load(fis)
        if(properties.contains('sdk.dir')) {
            if (new FileInputStream(properties.getProperty('sdk.dir').toString() + File.separator + "platforms").read() != -1) {
                sdkPath = properties.getProperty('sdk.dir').toString()
            }else {
                sdkPath=""
            }
        }else{
            sdkPath=""
        }
    }else{
        sdkPath=""
    }
```

i prefer the last option in ## 2 because it's more dynamic......

Scrappers-glitch:
The point of this PR was to revive `jme3-android-examples` which is broken from multiple aspects : 
1) It uses local android native object files which is suspected to be built using a local ndk after changing the ndkPath and compileAndroidNatives to true in gradle.properties (you cannot run them if you don't have an ndk).
2) It uses `jbullet` which depends on stack-alloc and a couple of other dependencies (deprecated).
3) It presents old android strategies.

The main goal was : 
- Run the test cases with one click, using precompiled natives (may be downloading from maven automatically using gradle before build), so even tho you don't have the NDK, you can run android examples.

So, i guess i will close this PR until i find a way to resolve those multiple aspects that cannot be done through a single PR/Change.

Issue #1548, opened 2021-05-11
MorphControl doesn't override cloneFields(), jmeClone(), read(), nor write()


stephengold:
As was pointed out at the Forum, the `MorphControl` class supposedly implements `JmeCloneable` and `Savable`, yet none of the relevant methods are overridden.

https://hub.jmonkeyengine.org/t/issue-with-multiple-models-and-morph-animations/44598

Obviously this makes it impossible to load a `MorphControl` from assets.

Also, the OP reported that after cloning a model with a `MorphControl`, the clone and the original morphed in synchrony.

In my own tests, cloning a model with a `MorphControl` resulted in a crash:
```text
RendererException: Cannot recompile shader source
```

stephengold:
You're quite right.

Ali-RS:
After testing on a bunch of different models I confirmed that works fine to move 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/226176b379e723495613e0f59b6ea21e9eb03d7b/jme3-core/src/main/java/com/jme3/anim/MorphControl.java#L79-L82

into setSpatial() method.

```
    @Override
    public void setSpatial(Spatial spatial) {
        super.setSpatial(spatial);

        // gathering geometries in the sub graph.
        // This must be done in the update phase as the gathering might add a matparam override
        targets.clear();
        if (spatial != null) {
            spatial.depthFirstTraversal(targetLocator);
        }
    }
```

Ali-RS:
And this line

https://github.com/jMonkeyEngine/jmonkeyengine/blob/226176b379e723495613e0f59b6ea21e9eb03d7b/jme3-core/src/main/java/com/jme3/anim/MorphControl.java#L90

should be changed like

`for (Geometry geom : targets.getArray())`

stephengold:
A 2nd PR sounds to me like a good plan.

stephengold:
Okay, @Ali-RS. It's your turn now!

Ali-RS:
I'm on it :)

PR #1553, opened 2021-05-13
MorphControl: serialize target list & add minor cleanups

Ali-RS:
See: 
https://hub.jmonkeyengine.org/t/morphcontrol-question/44599?u=ali_rs & https://github.com/jMonkeyEngine/jmonkeyengine/issues/1548#issuecomment-839956156

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/MorphControl.java
+++ b/jme3-core/src/main/java/com/jme3/anim/MorphControl.java
@@ -31,12 +31,10 @@
  */
 package com.jme3.anim;
 
-import com.jme3.export.InputCapsule;
-import com.jme3.export.JmeExporter;
-import com.jme3.export.JmeImporter;
-import com.jme3.export.OutputCapsule;
-import com.jme3.export.Savable;
-import com.jme3.material.*;
+import com.jme3.export.*;
+import com.jme3.material.MatParam;
+import com.jme3.material.MatParamOverride;
+import com.jme3.material.Material;
 import com.jme3.renderer.*;
 import com.jme3.scene.*;
 import com.jme3.scene.control.AbstractControl;
@@ -47,6 +45,7 @@ import com.jme3.util.SafeArrayList;
 import com.jme3.util.clone.Cloner;
 import java.io.IOException;
 import java.nio.FloatBuffer;
+import java.util.ArrayList;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


stephengold:
I believe it's unnecessary to serialize the list of targets, since the list will get regenerated by `setSpatial()` during loading.

Ali-RS:
>I believe it's unnecessary to serialize the list of targets, since the list will get regenerated by setSpatial() during loading.

I was thinking like so but found out setSpatial() is not called when loading. You can confirm that by running TestMorph.

jme3-core/src/main/java/com/jme3/anim/MorphControl.java
```diff
@@ -79,22 +82,31 @@
...
+        super.setSpatial(spatial);
+
+        // Remove matparam override from the old targets
+        for (Geometry target : targets.getArray()) {
+            target.removeMatParamOverride(nullNumberOfBones);
         }
+
         // gathering geometries in the sub graph.
-        // This must be done in the update phase as the gathering might add a matparam override
+        // This must be done in the update phase as the gathering might add a matparam override otherwise it will
```
stephengold, line 94: I don't understand the comment about "must be done in the update phase". Can you clarify?

Ali-RS: My understanding is that the author means this should not be done in `controlRender()` because it can add a MatParamOverride which can throw IllegalStateException if added in the render phase. 

See: https://hub.jmonkeyengine.org/t/morphcontrol-question/44599/8?u=ali_rs

stephengold: That made sense when it was in `controlUpdate()`. But it's in `setSpatial()` now, so it's not being done in the update phase.

Ali-RS: I have tested this on bunch of different models and I did not noticed any issue when done in `setSpatial()`.

so maybe I should remove that comment or change it to something like this:

```
// This must not be done in the render phase as the gathering might add a matparam  
// override which will cause an IllegalStateException if done in the render phase.
```

what do you think?

stephengold: That should be fine.

Ali-RS: Do you mean using the flag?

stephengold: I meant changing the comment. I don't see the need for a flag.

Ali-RS: I see. Will update the comment.

Ali-RS: Fixed the comment.

stephengold:
Thanks for your latest contribution to JMonkeyEngine!

Ali-RS:
I appreciate you for reviewing and merging the PR. :)

Issue #37, opened 2014-03-24
Add warning to material when more than 16 textures are used


ghost:
Reference on googlecode: https://code.google.com/p/jmonkeyengine/issues/detail?id=383

shadowislord:
If more than 16 textures are used then an `ArrayIndexOutOfBoundException` is thrown in `GLRenderer.bindTextureAndUnit`

stephengold:
Forum topic: https://hub.jmonkeyengine.org/t/12-textures-splatting-in-new-terrainlighting-j3md/17561

PR #1561, opened 2021-05-17
Resolved Issue#37 for ArrayIndexOutOfBoundException 

Mezage:
Resolves issue #37

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
@@ -2416,12 +2416,10 @@ public final class GLRenderer implements Renderer {
 
         //ArrayIndexOutOfBoundException thrown when
         //more than 16 units are used per material
-        if (unit >= 17) {
-            if (logger.isLoggable(Level.WARNING)) {
-                logger.log(Level.WARNING,
-                        "Limit of 16 textures have been exceeded for this material.");
-            }
-            return;
+
+        if (unit >= 16) { // base index = 0
+            throw new ArrayIndexOutOfBoundsException("Due to the limitation of GL2, only 16 units are guaranteed to " +
+                    "be supported. Failed to bind the texture");
         }
 
         if (context.boundTextures[unit] == null || context.boundTextures[unit].get() != img.getWeakRef().get()) {
@@ -2431,6 +2429,7 @@ public final class GLRenderer implements Renderer {
         } else {
             statistics.onTextureUse(img, false);
         }
+
     }
     
     /**
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
@@ -2413,7 +2413,18 @@ public final class GLRenderer implements Renderer {
             gl.glActiveTexture(GL.GL_TEXTURE0 + unit);
             context.boundTextureUnit = unit;
         }
-        if (context.boundTextures[unit]==null||context.boundTextures[unit].get() != img.getWeakRef().get()) {
+
+        //ArrayIndexOutOfBoundException thrown when
+        //more than 16 units are used per material
+        if (unit >= 17) {
+            if (logger.isLoggable(Level.WARNING)) {
+                logger.log(Level.WARNING,
+                        "Limit of 16 textures have been exceeded for this material.");
+            }
+            return;
+        }
+
+        if (context.boundTextures[unit] == null || context.boundTextures[unit].get() != img.getWeakRef().get()) {
             gl.glBindTexture(target, img.getId());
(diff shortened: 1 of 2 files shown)


stephengold:
Thank you for your contribution to JMonkeyEngine. Please let us know when you believe this is ready to be merged.

Mezage:
This was my team's first time submitting a pull request, so we had decided to create a draft first. But if you think it looks good we can submit an official one.

stephengold:
I can tell you I'm going to have some feedback on this, but I won't go into specifics before you think it's ready.

stephengold:
The comments say that an `ArrayIndexOutOfBoundException` is thrown, but that's no longer true.

Unit numbers start from 0, so the `if  (unit >= 17)` test appears to be off by one. Have you tested this code?

I don't believe logging a warning adequately addresses the issue. The log message doesn't indicate which material caused the overflow. Since it doesn't halt the application, the unbound texture might lead to subsequent errors that would confuse diagnosis. And if logging were disabled, the early return would merely exacerbate the situation. Wouldn't it be better to throw a custom exception with a full description of the cause?

To demonstrate how this PR would work in practice, I think it should include a test case of a material with 17 textures. Test cases can be added to the `jme3-examples` subproject.

Mezage:
Thanks for the feedback, we tried to incorporate some of your suggestions in our latest push

stephengold:
Thanks for following up. 
With 1d263e7, this PR is much improved. It's still not ready to be integrated, however.

+ The new diagnostic message is confusing because JMonkeyEngine hides details such as GL2 and "supported units" from the app developer. I suggest: "Tried to assign more than 16 texture params in a single material."
+ `TestIssue37` is a unit test. Unit tests should be automated (using JUnit 4) and added in jme3-core/src/test instead of jme3-examples.
+ TestIssue37.java should include the JMonkeyEngine copyright notice and BSD license at the top of the file. (Note: by submitting the file as part of a PR, you are asserting authorship and assigning your rights to project.)
+ Being a unit test, `TestIssue37` doesn't test how the code added to `GLRenderer` works in a realistic situation. For a new diagnostic message, this is more crucial even than unit testing!

I suggest actually defining a `Material` with 17 `Texture2D` params (perhaps by copying "ShowNormals.j3md" to a new file in jme3-testdata and modifying it) then creating a new jme3-example app (perhaps based on `HelloJME3`) that loads the J3MD, adds 17 `Texture2D` parameters to it, and applies it to a box.

When I ran such an app, it crashed in a different part of `GLRenderer`:
```text
java.lang.ArrayIndexOutOfBoundsException: 16
	at com.jme3.renderer.opengl.GLRenderer.bindTextureOnly(GLRenderer.java:2444)
	at com.jme3.renderer.opengl.GLRenderer.setupTextureParams(GLRenderer.java:2345)
	at com.jme3.renderer.opengl.GLRenderer.setTexture(GLRenderer.java:2622)
```
so you probably need to add code to the `bindTextureOnly()` method.

stephengold:
Even better might be to add code to `Material.updateShaderMaterialParameters()`. That's where unit numbers are actually assigned.

MeFisto94:
I am late to this, but isn't the number of textures dynamic in the meantime, like 16 is just the bare minimum that has to be supported by the Hardware.

That way I wonder what the intention of the original issue was.
Not throwing but silently ignoring the issue would be an inacceptable behavior change, because one would wonder why the texture is not rendered, without having a clue.

So maybe someone with more experience around this engine part may have an opinion there.

stephengold:
> I am late to this, but isn't the number of textures dynamic in the meantime, like 16 is just the bare minimum that has to be supported by the Hardware.

Yes, the number of textures bound to units is dynamic. However, each render context has a hardcoded limit of 16 texture units:
```java
    /**
     * Current bound texture IDs for each texture unit.
     *
     * @see Renderer#setTexture(int, com.jme3.texture.Texture)
     */
    public final WeakReference<Image> boundTextures[] = new WeakReference[16];
```
Perhaps this originally had something to do with GL2 minimum hardware requirements, but it's currently intrinsic to JME's GL renderer.

stephengold:
If we think JME might someday support more than 16 bound textures, then the range check(s) should be in `GLRenderer`, of course, not in `Material`.

The advantage of issuing the diagnostic message in `Material` is that then `Material` can include the material definition name, parameters, and overrides in the message, to help pinpoint the cause. So if the limit is renderer-specific, then some new interface (such as an error return value, callback, or custom exception) might be desired so that the renderer can report the situation back to `Material`, which would then issue the message.

Mezage:
Thanks for the specific suggestions, I'm not sure when I'll have time to adjust the PR but your comments will be helpful if I get to working on this!

stephengold:
From reviewing this PR, I learned a lot about issue #37. I think I'll submit my own PR to address it.

stephengold:
The issue in question was closed by PR #1574.

PR #1565, opened 2021-05-22
Test and solution for issue #1564 (setting Material parameter of type IntArray)

### LINKED ISSUE #1564: Material parameter of type IntArray is empty after resetting it
I found a bug related to Material parameters of the type VarType.IntArray.

If you set the parameter for the first time, everything works fine.
However, if you change the parameter later on, it won't be set within the Shader.
The parameter get erased and is empty.

I noticed the bug inside the class Uniform.
When resetting the parameter, the buffer gets resized but not populated with the right array data.
I can provide a pull request to fix the problem.

theMinka:
This pull request contains a fix for #1564.
It also added a minor improvement in the Uniform class and a UnitTest.

### DIFF
--- a/jme3-core/src/test/java/com/jme3/shader/UniformTest.java
+++ b/jme3-core/src/test/java/com/jme3/shader/UniformTest.java
@@ -1,0 +1,342 @@
+/*
+ * Copyright (c) 2009-2021 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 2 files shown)


stephengold:
Thank you for your contribution to the project. This looks very good!

I wish the unit test included some comments describing what properties of `Uniform` are being tested.

jme3-core/src/test/java/com/jme3/shader/UniformTest.java
```diff
@@ -0,0 +1,327 @@
...
+import java.nio.FloatBuffer;
+import java.nio.IntBuffer;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+
+import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+
+public class UniformTest {
```
stephengold, line 47: I wish the unit test included some comments describing what properties of Uniform are being tested.

stephengold (approved):

PR #1570, opened 2021-06-08
Convert a few manual tests to unit tests

MeFisto94:
For the physics tests, I've verified that reverting the related commits indeed flags the tests as failed.


MeFisto94:
I'm currently uncertain about the tests failing, for me jpeg was there when I used `usePlatformConfig = true`, maybe this is a result of the headless environment not supporting AWT/jpeg?

MeFisto94:
My force push should have addressed the issues, but the CI had connection reset intermittent connection errors.

stephengold:
You should be able to re-run the tests. Or would you like me to do it?

MeFisto94:
I already did, but now they failed at another step. The Asset Manager setup is kinda picky, sometimes it does not pickup the Desktop.cfg, this time it seems the way I did it made it ignore the default locators. 
I could swear that all tests ran fine on my headless environment, but I can now see the failing test on my regular system and work towards progress

MeFisto94:
FWIW the change was to add `JmeSystem.newAssetManager(TestLocators.class.getResource("/com/jme3/asset/Desktop.cfg"));` so that the Desktop.cfg is added instead of just using `new DesktopAssetManager(true)`, because `true` will use the platform config, which is probably detected as the Headless System

stephengold:
Is this PR still in progress, or is it ready for review?

MeFisto94:
It's ready for review, the last push only fixed some codacy warnings related to tests not being in camelCase

Issue #1284, opened 2020-02-08
ShaderNodes: ColorToGrey has a wrong implementation 


MeFisto94:
So we are talking about this:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/resources/Common/MatDefs/ShaderNodes/Basic/colorToGrey.frag

It should extract the luminance of a color (or in other words: convert RGB to Greyscale).

I think the way it has been done only works when r == g == b and then dividing by 3 again (because currently grey can be everything between 0 and 3. So it would only work when only one channel is used).

Also, when r == g == b and we'd do `(r + g+ b) / 3`, we could also take `r`, as simple as that.

So: For images already being Greyscale in RGB, we can just take one component.
For Images not being Greyscale, we need a more complicated code like https://github.com/flixel-gdx/flixel-gdx/blob/master/flixel-core/src/org/flixel/data/shaders/blend/luminosity.glsl

Luminance is only Lines 15-21 and we can add a proper RGB->HSL conversion as well.
The license is BSD.

Tagging @neph1 as he contributed the initial ColorToGrey implementation.

stephengold:
Sounds reasonable.

PR #1572, opened 2021-06-10
Gray fix2

Mezage:
Still trying to understand how fragments work, advice would be nice.
Attempt at patching issue #1284


Sailsman63:
Before settling on an implementation, it might be nice to define what kind of results we want from a greyscale transformation, which would be more useful in a shader. See https://en.wikipedia.org/wiki/HSL_and_HSV#Disadvantages for a quick comparison of some of the various ways of transforming RGB into "Lightness/Brightness" of some kind.

Even if the decision is to stick with the basic Arithmetic Mean, as implemented here (technically *I*ntensity from the HSI color space) I question whether the gating check is worth it:

- You're adding two comparisons and a branch instruction *per pixel*, with a fairly high likelihood of having to do the sum and division anyway. I'd expect it to mostly be slower than just doing the math. (Though, someone would have to benchmark on a variety of hardware to be sure.)
- While the speed difference is likely to be minimal either way, you're adding cognitive overhead to the code.

scenemax3d:
Hi @Mezage ,
Are you still interested in this PR?
If so, can you provide the following information:

1. Does it break any current feature / interface
2. How well it’s tested from 0 - 10
3. Does it needs any new documentation
4. Release notes

Thank you

Issue #1349, opened 2020-04-17
FastMath.sphericalToCartesianZ() does not reverse cartesianZToSpherical()


stephengold:
There's a logic error in `FastMath.sphericalToCartesianZ()`, such that the transformation fails to reverse the effect of `cartesianZToSpherical()` when `latitude != longitude`.

Here's a test case:
```java
        final Vector3f in2 = new Vector3f(-1.9f, +5.8f, +8.1f);
        final Vector3f spherical2 = FastMath.cartesianZToSpherical(in2, null);
        final Vector3f out2 = FastMath.sphericalToCartesianZ(spherical2, null);
        assertEquals(in2.x, out2.x, 1e-5f);
        assertEquals(in2.y, out2.y, 1e-5f);
        assertEquals(in2.z, out2.z, 1e-5f);
```

Code inspection discovered that, while `cartesianZToSpherical()` stores the longitude (theta) coordinate in `store.z`, `sphericalToCartesianZ()` expects the longitude to be in `sphereCoords.y`.

PR #1573, opened 2021-06-10
Fix the issue #1349

wei-shiuan-chang:
This pull request is to solve the issue #1349.
According to the definition of each axis and the process to get x, y, and z of Spherical coordinates from the method cartesianZToSpherical, the method sphericalToCartesianZ should be modified.


stephengold:
Thanks for your contribution to JMonkeyEngine!

stephengold (approved):
Looks good!

Issue #352, opened 2015-09-24
Asset paths that begin with "/" don't work on Android


shadowislord:
Due to the changes in jME 3.1 to use Android's `AssetManager` instead of the `ClasspathLocator`, paths that begin with a forward slash no longer work.

zzuegg:
Rather strange, there is already some code in the AndroidLocator to remove a leading /

nordfalk:
OK, I have used several days on this one!

Here is what extensive debugging finally brought to me.
As you can see from https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-android/src/main/resources/com/jme3/asset/Android.cfg the com.jme3.asset.plugins.AndroidLocator gets as / as path:

```
# Android specific locators
LOCATOR / com.jme3.asset.plugins.AndroidLocator
```

This means that the locator gets rootPath = "/" .... and now look at https://github.com/jMonkeyEngine/jmonkeyengine/blob/445f7ed010199d30c484fe75bacef4b87f2eb38e/jme3-android/src/main/java/com/jme3/asset/plugins/AndroidLocator.java#L26 , in the case where the AssetKey starts with a "/" (like "/Textures/...")

String assetPath = rootPath + key.getName(); // will start with TWO slashes, like "//Textures/...

The // Fix path issues will remove ONE of the slashes, but not the other, which will make the system look in the root of the Linux Android file system :-(


Proposed fix 1: Avoid / as path 

```
# Android specific locators
LOCATOR . com.jme3.asset.plugins.AndroidLocator
```
(best would be to set rootPath = "" but that isnt possible as it would not be valid syntax)

Proposed fix 2:  ... or change to a while loop instead of an if statement:

```
  // Fix path issues
        while (assetPath.startsWith("/")) {
            // Remove leading /
            assetPath = assetPath.substring(1);
        }
```

Proposed fix 3:  replace "//" with "/" before removing heading slash

```
        assetPath = assetPath.replace("//", "/");
        if (assetPath.startsWith("/")) {
            // Remove leading /
            assetPath = assetPath.substring(1);
        }

```


This bug affects all people that uses the SDK terrain editor (jMonkeyEngine 3.1-beta1-b002-SNAPSHOT) as it wrongly puths forward slashes in the paths of the textures in the .j3o file.

See also discussions
[...]

PR #1577, opened 2021-06-21
support Android assets paths starting with /

scenemax3d:


Ali-RS:
Looks fine to me.

stephengold:
Please explain the issue addressed by this PR in more detail. Why would an Android asset path contain consecutive slashes?

scenemax3d:
> Please explain the issue addressed by this PR in more detail. Why would an Android asset path contain consecutive slashes?

When DesktopAssetsManager registers its locators, it sets a slash **rootPath** for each locator including AndroidLocator. Combine that with assets path which begins with a slash and the current slash cleaning code in the locator's locate method cannot cope with it because it assumes no more than one slash in the beginning of the path.

tlf30:
This is an interesting issue. I feel like others including myself would have run into before. Can you please provide a simple test case?

scenemax3d:
> This is an interesting issue. I feel like others including myself would have run into before. Can you please provide a simple test case?

Put any asset in your application's assets folder, for example: **assets/Textures/my_icon.png**
The try to load it in your JME application: assetManager.loadTexture("**/**Textures/my_icon.png"); notice the **slash** in the beginning of the path...
**result**: no locator including AndroidLocator will be able to locate this texture

stephengold (approved):

MeFisto94:
> result: no locator including AndroidLocator will be able to locate this texture

So this is also an issue for desktop? if so, that may need a solution at another place, though I wonder if the code was ever meant to support leading `/` paths, because jme examples never have that, afaik.

If that's a goal, it's also important to note that unit tests may be wanted for that, writing one for that is trivial.

stephengold:
It's not an issue for desktops, because the jme-vehicles project uses such asset paths, and it works on desktops.

neph1:
Related it seems: [https://github.com/jMonkeyEngine/jmonkeyengine/issues/352](https://github.com/jMonkeyEngine/jmonkeyengine/issues/352)

stephengold:
Good catch, @neph1 !

PR #1580, opened 2021-07-05
Fix title (and registry entry) for Settingsdialog

Pesegato:
...the developer probably want it to be named as its application

### DIFF
--- a/jme3-desktop/src/main/java/com/jme3/app/SettingsDialog.java
+++ b/jme3-desktop/src/main/java/com/jme3/app/SettingsDialog.java
@@ -169,6 +169,7 @@ public final class SettingsDialog extends JFrame {
             logger.log(Level.WARNING,
                     "Failed to load settings", ex);
         }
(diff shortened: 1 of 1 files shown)


stephengold:
I haven't forgotten about this PR. I just need some focused time to think about it.

stephengold:
I don't understand the motivation for this PR.

When I run a JME application named "Maud", I expect the settings dialog to have the title "Display Settings for Maud" and the application window to have the title "Maud". And that's exactly what I see in practice. Are you seeing something different?

Pesegato:
If you delete the registry entries and start from scratch, then you won't see "Maud".

stephengold:
My JME applications explicitly set the window title in `main()` like so:
```java
        settings.setTitle(windowTitle);
        application.setSettings(settings);

        application.start();
```

You're proposing that JME should override that setting?

Pesegato:
Moving the discussion to the forum

https://hub.jmonkeyengine.org/t/pr1580/45105

PR #1581, opened 2021-07-06
correct errors in documentation

stephengold:
This is mostly for typos in Java comments.


capdevon:
Hi Stephen, great job as always! There would also be the names of these two methods that do not reflect the rules of naming convention of the java language. They are found in the `MotionPath `class and are these:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/cinematic/MotionPath.java#L148
```java
private Geometry CreateLinearPath() { ... }
```

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/cinematic/MotionPath.java#L158
```java
private Geometry CreateCatmullRomPath() { ... }
```

You could correct the initial letter of the method to be lowercase. There should be no problem as these are private methods used internally only by the MotionPath class.

stephengold:
You're right that those 2 methods should be renamed. However, it should be a separate PR.

capdevon:
Okay, I've opened a new issue. [#1587](https://github.com/jMonkeyEngine/jmonkeyengine/issues/1587)

PR #1588, opened 2021-07-09
Upgrade Gradle to 6.9

tonihele:
Testing to upgrade to newer build tools


stephengold:
Thanks for working on this.

tonihele:
There are couple of things. I mostly removed the deprecated Gradle things. These that are left will prevent upgrade to Gradle 7.

Couple of puzzling things. Nifty, LWJGL3 and jBullet.. Those three are needed to be exposed (directly available to be used) to jme3-examples and jme3-vr. I left them exposed from the projects they are declared in. When I say exposed I mean I left the dependency marked as _compile_ rather than changing it to _implementation_. This can be solved by adding those libraries directly as dependency to jme3-examples and jme3-vr, which I think would be correct, since they are directly needed there. I'm unsure did this hiding of the external APIs affect any projects that use JME.

Deprecated:
[log.txt](https://github.com/jMonkeyEngine/jmonkeyengine/files/6794955/log.txt)

tonihele:
> Couple of puzzling things. Nifty, LWJGL3 and jBullet.. Those three are needed to be exposed (directly available to be used) to jme3-examples and jme3-vr. I left them exposed from the projects they are declared in. When I say exposed I mean I left the dependency marked as _compile_ rather than changing it to _implementation_. This can be solved by adding those libraries directly as dependency to jme3-examples and jme3-vr, which I think would be correct, since they are directly needed there. I'm unsure did this hiding of the external APIs affect any projects that use JME.

I think it is correct that they are exposed through JME. So they must remain so. That is my conclusion, I'm not confused about this anymore :)

tonihele:
I think it is ready now. Only these 

> Copying or archiving duplicate paths with the default duplicates strategy has been deprecated. 

are left. I did remove the POM hacks from the common.gradle. Since the Bintray thing doesn't seem to exist anymore and I couldn't find a way to do it in the maven-publish plugin. I assume this is fine. So not quite ready for Gradle 7 but a good step forward.

stephengold:
How about specifying a duplicates strategy?

tonihele:
> How about specifying a duplicates strategy?

That'll get rid of the warning yes. I just didn't know if it is a correct solution or not. I can try to specify that.

stephengold:
Looks pretty good.

I think more of the `compile` dependencies should be changed `api` instead of `implementation`. For instance, "jme3-android" should have an `api` dependency on "jme3-core" because "jme3-android" exposes methods with classes like `ViewPort` in their signatures.

Also, please remove the trailing whitespace that was added to the "build.gradle" in the project root.

tonihele:
> I think more of the `compile` dependencies should be changed `api` instead of `implementation`. For instance, "jme3-android" should have an `api` dependency on "jme3-core" because "jme3-android" exposes methods with classes like `ViewPort` in their signatures.

All of the jme project dependencies should be marked as API to all projects?

stephengold:
Not necessarily. As I understand it, the best practice is to specify an `api` dependency when methods are exposed that use classes from the dependent library in their signatures. Thus, any library that uses `Vector3f` or `ViewPort` or `Spatial` in a public method signature should specify an `api` dependency on "jme3-core". But libraries that only use such classes internally can specify an `implementation` dependency.

Without tool support, it'll be very difficult to get these distinctions perfect on the first attempt. When in doubt, I learn toward specifying `api` dependencies in library builds. And I believe that non-library projects (such as "jme3-examples" and "jme3-testdata") should never specify `api` dependencies.

Does that help?

tonihele:
> Without tool support, it'll be very difficult to get these distinctions perfect on the first attempt. When in doubt, I learn toward specifying `api` dependencies in library builds. And I believe that non-library projects (such as "jme3-examples" and "jme3-testdata") should never specify `api` dependencies.

Yep, lets go with this one then. If somebody edits the code, I doubt he/she then remembers to check whether now a dependency needs to be declared as API. And basically that is how it was before. The deprecated compile keyword is synonymous to api keyword (although not recommended replacement by default since one does not usually want to do this exposing).

stephengold:
Sounds good.

tonihele:
Maybe ok?

I noticed that the jme-example run wasn't working anymore. Somebody maybe already tried to make it a _application_ judging from the commented out code. That was a problematic route. So was the manual run task. I had to rename the variable. Also I took the mainClass stuff out from the other builds. I think this is some old Netbeans stuff, I have it too on my projects. Netbeans 12 doesn't need or do stuff like this anymore.

MeFisto94:
One thing that I had with the original buildscript (haven't tested the migration yet), was that IntelliJ insisted on knowing the Android SDK Location and when rejecting, refused to sync the project, rendering the IDE useless when not having Android Studio / some SDK Path.

Is this still the case and can something be done?

Regarding the examples: They should really be runnable, `gradlew run` should start them I guess. What I did around that was that some dependencies didn't properly build because they have been depending on some other target that run didn't cause, iirc.

tonihele:
> One thing that I had with the original buildscript (haven't tested the migration yet), was that IntelliJ insisted on knowing the Android SDK Location and when rejecting, refused to sync the project, rendering the IDE useless when not having Android Studio / some SDK Path.
> 
> Is this still the case and can something be done?
> 
> Regarding the examples: They should really be runnable, `gradlew run` should start them I guess. What I did around that was that some dependencies didn't properly build because they have been depending on some other target that run didn't cause, iirc.

This is still the case. Intellij doesn't want to sync without Android studio. I haven't found a solution to this.

The examples work now. In IDE if you try to run the project, it runs the examples. As it used to.

tonihele:
> One thing that I had with the original buildscript (haven't tested the migration yet), was that IntelliJ insisted on knowing the Android SDK Location and when rejecting, refused to sync the project, rendering the IDE useless when not having Android Studio / some SDK Path.
> 
> Is this still the case and can something be done?

https://youtrack.jetbrains.com/issue/IDEA-265504

stephengold:
These changes interfere with my ability to directly run a single jme3-test app from the IDE. Before, I could open a class and use Shift-F6 to run it directly. Afterward, pressing Shift-F6 brings up the test chooser.

That's with  jMonkeyEngine SDK v3.2.4-stable-sdk1

tonihele:
Works with Netbeans 12 based SDK 3.3. Also need to always remember to go to console and "gradlew wrapper" to download the appropriate Gradle version. Netbeans doesn't do that. 

The Gradle support in old Netbeans 8 is really poor. It is just... old. And I do remember clearing out some Netbeans specific hacks as they are not needed anymore with newer NBs. And our SDK is also nicely available for NB 12.

stephengold:
I'm gradually transitioning to IDEA, but for now I mostly use JME SDK 3.2.4. Could the Netbeans-specific hacks be added back in without breaking anything?

tonihele:
Maybe, they are these https://github.com/jMonkeyEngine/jmonkeyengine/pull/1588/files#diff-e036a141ae65c0211a306ecb8e1bbc9402063cda41586ae8e07474f399b7c371L5-L16. To all of the builds. Extra stuff will of course increase complexity, so one might want to weight this against do we really want to support 7 year old version of Netbeans.

stephengold:
Perhaps a better question is, "Do we really want to support our release manager?"

If reducing complexity of the build scripts is important, we could simply delete the "run" task from "jme3-examples/build.gradle". When I tested this, it allowed me to run a single jme3-test app from the IDE using Shift-F6.

Renaming the "run" task to "runChooser" or "runExamples" would have the same effect.

tonihele:
> we could simply delete the "run" task from "jme3-examples/build.gradle". When I tested this, it allowed me to run a single jme3-test app from the IDE using Shift-F6.

If the _run_ can be removed, that, in my opinion, is the best option. I just didn't know what the log and assert stuff did there. If these two are not needed then yes, delete _run_.

tonihele:
And out of curiosity, what is wrong with SDK 3.3? :D It seems that many stick to 3.2.4.

stephengold:
> If the _run_ can be removed, that, in my opinion, is the best option. I just didn't know what the log and assert stuff did there. If these two are not needed then yes, delete _run_.

I believe the purpose of the "run" task is to run the test chooser from a command line, or by right-clicking on "jme3-examples" in the Projects window of NetBeans. Useful, I suppose.

Much of the complexity results from configuring assertions based on the "assertions = " line in "jme3-examples/gradle.properties". If I ever wanted to run the test chooser *without* assertions, I would prefer to edit "jme3-examples/build.gradle" directly, instead of going through an extra level of indirection.

Issue #1587, opened 2021-07-09
The name of a couple of methods in the MotionPath class starting with an uppercase letter


capdevon:
The name of a couple of methods in the `MotionPath` class starting with an uppercase letter, violating the coding conventions of the java language.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/cinematic/MotionPath.java#L148
```java
private Geometry CreateLinearPath() { ... }
```

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/cinematic/MotionPath.java#L158
```java
private Geometry CreateCatmullRomPath() { ... }
```

There should be no problem correcting them to lowercase, as the two methods are private.

PR #1589, opened 2021-07-10
renaming Java identifiers

stephengold:
This PR renames local variables, method arguments, private fields, and such in the Java code. In most cases the old names contained typos or spelling errors. A few names violated coding conventions (see issue #1587, for instance) or were abbreviated to the brink of becoming unintelligible.

There's plenty of room for further work along these lines.

### DIFF
--- a/jme3-android/src/main/java/com/jme3/app/state/VideoRecorderAppState.java
+++ b/jme3-android/src/main/java/com/jme3/app/state/VideoRecorderAppState.java
@@ -227,7 +227,7 @@ public class VideoRecorderAppState extends AbstractAppState {
         private int width;
         private int height;
         private RenderManager renderManager;
-        private boolean isInitilized = false;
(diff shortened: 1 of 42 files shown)


pspeed42:
By the way, regarding things like "matrixes" versus "matrices", both are correct despite what spellcheckers will tell you and arguably the -xes version is easier to non-english natives to parse.

Furthermore, as an English speaker/writer, I've learned that for things like matrix/vertex and syllabus/hippopotamus style plural rules, the -es version is always correct while the -ices or -i rule gets inconsistently applied.  So I've personally opted to go the "easy more intelligible way" in all such cases.  I suggest the same for an international open source code base.

stephengold:
If "matrixes" is acceptable, then I can revert at least one of the changes.

pspeed42:
Yeah, the other changes looked good to me.

PR #1592, opened 2021-07-30
Buffered mjpeg file writer

tonihele:
This optimizes the VideoRecorderAppState video file writing.

- The old version writes really small chunks, 4 bytes, even 1 byte chunks at times, PER FRAME!
- The new version uses buffer (8192 bytes) to write bigger chunks
- In reality much of the data written at a time (the individual frames) is a bigger chunk than 8192 -> written at one go
- Implements AutoCloseable, in my mind this is logical. Like writing a ZIP file etc. similar API

Basically this improves performance by eliminating some of the system calls.

New:
image

Old:
image


stephengold:
Thanks for your contribution.

PR #1593, opened 2021-07-31
Restart audio context if current device is "disconnected"

tonihele:
Okay, finally something that it is worthy in my mind; recovering from audio failure! The implementation is based on: http://forum.lwjgl.org/index.php?topic=7081.msg37006#msg37006. Tested on Linux & Windows and seems to work.

To try it out:

- Use i.e. headphones when launching a jME app
- Disconnect headphones

New behavior in this case:

- The disconnection is detected fairly soon, audio renderer is re-created, creates some noise on the log
- Lost bytes in the buffer are not recovered :(
- New audio goes to the new default device

Old behavior in this case:

- No sound, ever, again
- Our app plays background tunes -> CPU spikes permanently as we start feeding audio to void
- Behavior of disconnected device is documented here: https://javadoc.lwjgl.org/org/lwjgl/openal/EXTDisconnect.html
- It kinda pretends to work

What this does **not** do:

- Replugging headphones doesn't change audio back to headphones -> we wont react to changes in default devices, only disconnected/broken
- I did try to also do this, but I couldn't detect this situation. At least with the headphones, the headphones was always the default whether they were plugged in or not. Probably works if the default device is changed by the user for real. But since the automatic headphone thing doesn't work, I don't see this as a must have (or a regular use case).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/audio/openal/ALAudioRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/audio/openal/ALAudioRenderer.java
@@ -33,18 +33,16 @@ package com.jme3.audio.openal;
 
 import com.jme3.audio.*;
 import com.jme3.audio.AudioSource.Status;
+import static com.jme3.audio.openal.AL.*;
 import com.jme3.math.Vector3f;
 import com.jme3.util.BufferUtils;
 import com.jme3.util.NativeObjectManager;
-
 import java.nio.ByteBuffer;
 import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
 import java.util.ArrayList;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
-import static com.jme3.audio.openal.AL.*;
 
 public class ALAudioRenderer implements AudioRenderer, Runnable {
 
@@ -70,6 +68,7 @@ public class ALAudioRenderer implements AudioRenderer, Runnable {
     private boolean audioDisabled = false;
     private boolean supportEfx = false;
     private boolean supportPauseDevice = false;
+    private boolean supportDisconnect = false;
     private int auxSends = 0;
     private int reverbFx = -1;
     private int reverbFxSlot = -1;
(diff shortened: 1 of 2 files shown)


stephengold:
Thanks for the contribution. Looks OK to me, but I'll leave this PR open for a few days in case someone wants to review it.

jme3-core/src/main/java/com/jme3/audio/openal/ALAudioRenderer.java
```diff
@@ -858,6 +862,29 @@ public void update(float tpf) {
...
+    private void checkDevice() {
+
+        // If the device is disconnected, pick a new one
+        if (isZombieDevice()) {
+            logger.log(Level.INFO, "Current audio device disconnected.");
+            restartAudioRenderer();
+        }
+    }
+
+    private boolean isZombieDevice() {
```
Ali-RS, line 874: Can we use a better method name?

tonihele:
Sure, isDisconnected? Not a fan of the zombie term myself. Just pulled that
from their terminology

Ali-RS:
Yes, `isDisconnected` sounds good to me.

tonihele:
I could check after this to make the switch if default audio device is changed although it is a shame that plugging headphones in and out would not trigger this change. But at least there would be a way to fiddle the sounds back to working order without restarting the whole application. With all this we would be ahead of the game compared to most Java games :)

Hard to measure the CPU impact these added features give. Depending on this we could check this default device thing maybe not as often as the disconnected...

MeFisto94 (approved):
I don't see a CPU concern. It's true that we could just as well check every other second (or buffer size) for disconnected devices, the method is probably very JIT friendly (a condition that is almost final and a simple native call) anyway.

Also the functionality is really short/fast so I don't think that will be a problem.

tonihele:
Oh and Java native sound (AudioSystem) does react to changes in default audio device. It doesn't always work perfectly though, but most of the time and in Linux as well. Could be nice to have the OpenAL implementation more closer to that. I don't think it is entirely possible to get on par since as far as I understand the Java native sound implementation is integrated better to the OS than OpenAL.

PR #1599, opened 2021-08-28
Add note to javadoc for Quaternion lookAt method

duncanj:
Adds a small note to the javadoc for Quaternion.lookAt, to explain that the method may produce invalid results under some circumstances.


stephengold:
Looks good to me. Thanks!

Issue #893, opened 2018-09-08
jMonkey Window on Mac Retina uses pixels instead of coordinates, so the viewport does not fill the window


bashburn:
I'm running a MacBook Pro with a Retina display. I have a very basic jMonkey application just trying to make sure everything is working. I set the application resolution as 1024x576 through the `AppSettings`. The window launches and fills the area according to the screen coordinates, but the viewport is in the lower left corner and appears to just be using pixels. On Retina screens, coordinates do not equal pixels. I have added `-Dorg.lwjgl.opengl.Display.enableHighDPI=true -Dsun.java2d.dpiaware=true` to the application start parameters which didn't seem to affect anything.
Here is a screenshot of what the window looks like: <img width="1023" alt="screen shot 2018-09-08 at 9 17 19 am" src="https://user-images.githubusercontent.com/454331/45254709-9bf0c700-b34a-11e8-8089-1eb7a9c89b4a.png">

Here are my `AppSettings`:
`    AppSettings settings = new AppSettings(true);
    settings.setRenderer(AppSettings.LWJGL_OPENGL41);
    settings.putBoolean("RendererDebug", true);
    settings.setWidth(Integer.valueOf(System.getProperty("ORS_SCREEN_WIDTH", "1024")));
    settings.setHeight(Integer.valueOf(System.getProperty("ORS_SCREEN_HEIGHT", "576")));
    settings.setVSync(Boolean.valueOf(System.getProperty("ORS_SCREEN_VSYNC", "false")));
    settings.setFullscreen(Boolean.valueOf(System.getProperty("ORS_SCREEN_FULLSCREEN", "false")));
 `

Here are the logs prior to my application:
`Sep 08, 2018 9:17:12 AM com.jme3.system.JmeDesktopSystem initialize
INFO: Running on jMonkeyEngine 3.2-stable
 * Branch: HEAD
 * Git Hash: f85624a
 * Build Date: 2018-01-21
Sep 08, 2018 9:17:12 AM com.jme3.system.lwjgl.LwjglContext printContextInitInfo
INFO: LWJGL 3.1.2 build 29 context running on thread main
 * Graphics Adapter: GLFW 3.3.0 Cocoa NSGL dynamic
Sep 08, 2018 9:17:12 AM com.jme3.renderer.opengl.GLRenderer loadCapabilitiesCommon
INFO: OpenGL Renderer Information
 * Vendor: Intel Inc.
[...]

PR #1607, opened 2021-09-15
fix incorrect input cursor position on MacOS with retina screen in use

jmecn:
On MacOS, with retina screen in use, the input cursor position is incorrect. As the window size is not equals to camera resolution.

This PR use glfwGetWindowContentScale to get the scaling factor on window resolution, and apply it to GlfwMouseInput#onCursorPos callback. So we have the correct cursor position on MacOS.

For more information, there are the links:

* https://hub.jmonkeyengine.org/t/switching-to-lwjgl-3/41222/11
* https://github.com/glfw/glfw/issues/845
* https://www.glfw.org/docs/latest/window_guide.html#window_scale

This PR is also trying to fix part of issue #893


stephengold:
Is this ready for review or integration? If not, please mark it as a draft.

jmecn:
> Is this ready for review or integration? If not, please mark it as a draft.

It's ready, sir.

jme3-core/src/main/java/com/jme3/system/AppSettings.java
```diff
@@ -289,6 +289,7 @@
         defaults.put("SwapBuffers", true);
         defaults.put("OpenCL", false);
         defaults.put("OpenCLPlatformChooser", DefaultPlatformChooser.class.getName());
+        defaults.put("UseRetinaFrameBuffer", false);// MacOS spec
```
tonihele, line 292: This is just a opinion but I think the default should be true? I would imagine someone would like to use all this native behaviour by default (given it also works perfectly in jME by default). In GLFW this is by default true also.

jme3-lwjgl3/src/main/java/com/jme3/input/lwjgl/GlfwMouseInput.java
```diff
@@ -141,11 +144,14 @@ public GlfwMouseInput(final LwjglWindow context) {
     }
 
     private void onCursorPos(final long window, final double xpos, final double ypos) {
+        float[] xScale = new float[1];
+        float[] yScale = new float[1];
+        glfwGetWindowContentScale(window, xScale, yScale);
```
tonihele, line 149: And was it that the callback doesn't work? That need to call this on every cursor move? Maybe not a heavy call...

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -534,6 +540,18 @@ protected void runLoop() {
...
 
+        // Update the frame buffer size from 2nd frame since the initial value
+        // of frame buffer size from glfw maybe incorrect when HiDPI display is in use
+        if (framesAfterContextStarted < 2) {
+            framesAfterContextStarted++;
+            if (framesAfterContextStarted == 2) {
+                int[] width = new int[1];
+                int[] height = new int[1];
+                glfwGetFramebufferSize(window, width, height);
+                listener.reshape(width[0], height[0]);
```
tonihele, line 551: Well, nobody likes this hack. But you did a very good job testing it and reporting it to upstream. Maybe this, if there is no other choice... should be conditional? Reshape might be heavy and there is no reason to call it every time on frame 2.

For these HiDPI displays I know that the initial size is wrong, so this is a fix for https://github.com/jMonkeyEngine/jmonkeyengine/issues/893?

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -688,4 +706,28 @@ public void destroy(boolean waitFor) {
...
+     * This is especially important for text and any UI elements. If the pixel dimensions of
+     * your UI scaled by this look appropriate on your machine then it should appear at a
+     * reasonable size on other machines regardless of their DPI and scaling settings. This
+     * relies on the system DPI and scaling settings being somewhat correct.
+     *
+     * @param store A vector2f to store the result
+     * @return The window content scale
+     * @see <a href="https://www.glfw.org/docs/latest/window_guide.html#window_scale">Window content scale</a>
+     */
+    public Vector2f getWindowContentScale(Vector2f store) {
```
tonihele, line 723: Would be great if this would be actually part of some interface (JmeContext?). And also required to be implemented by LWJGL 2 (this is fairly easy there, no need to even do a dummy implementation). Of course a bit extra work already.. But would enable jME apps to scale their UIs finally. Also a callback/listener?

stephengold:
@tonihele please continue the conversations you started or else mark them as "resolved".

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
tonihele: So could it be somehow changed that when everything is done do a glfwPollEvents()? But what I'm really after is, if the currentDimension <> newDimensions -> reshape? Not that reshape every application on frame 2.

tonihele:
Yep, sorry it took so long.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
stephengold: @jmecn Do you understand what tonihele is requesting?

jmecn: @stephengold @tonihele 
I add a resolution check before that reshape, would it be better?

stephengold:
Unless there are objections, this will be integrated in about 24 hours.

tonihele (approved):

stephengold:
Changed the default setting for "UseRetinaFrameBuffer" to `false` in an attempt to solve issue #1745.

stephengold:
Further fixes proposed in PR #1746.

Issue #1089, opened 2019-05-12
Issue in TestGltfLoading with "zophrac" model


Ali-RS:
Running TestGltfLoading with zophrac model 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/2124e3e86bf898d6c7a7bc90a191cf639e39bacb/jme3-examples/src/main/java/jme3test/model/TestGltfLoading.java#L119

results in an assertion error 

```
java.lang.AssertionError
	at com.jme3.anim.MorphTrack.setKeyframes(MorphTrack.java:125)
	at com.jme3.anim.MorphTrack.<init>(MorphTrack.java:74)
	at com.jme3.scene.plugins.gltf.GltfLoader.readAnimation(GltfLoader.java:842)
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:134)
	at com.jme3.scene.plugins.gltf.GltfLoader.load(GltfLoader.java:78)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:259)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:373)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:416)
	at jme3test.model.TestGltfLoading.loadModel(TestGltfLoading.java:239)
	at jme3test.model.TestGltfLoading.simpleInitApp(TestGltfLoading.java:119)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:237)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:475)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:574)
	at com.jme3.system.lwjgl.LwjglWindow.create(LwjglWindow.java:423)
	at com.jme3.app.LegacyApplication.start(LegacyApplication.java:463)
	at com.jme3.app.LegacyApplication.start(LegacyApplication.java:424)
	at com.jme3.app.SimpleApplication.start(SimpleApplication.java:125)
	at jme3test.model.TestGltfLoading.main(TestGltfLoading.java:69)

```

Not sure if this an issue with the model itself or a bug in Gltf loader.
Tested with the master branch.

stephengold:
After creating an account at Sketchfab and downloading the "zophrac" model, I was able to reproduce this issue.

[...]

PR #1608, opened 2021-09-25
Improve MorphTrack API

danielperano:
The MorphTrack's current API has some inconsistencies with TransformTrack's API and some limitations as well (no ability to get/set the desired number of morph targets). This patch makes MorphTrack's API consistent with TransformTrack's API and adds a getter/setter for the number of morph targets.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
+++ b/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
@@ -99,14 +99,14 @@ public class MorphTrack implements AnimTrack<float[]> {
      * @throws IllegalArgumentException if weights is an empty array or if
      *         the number of weights violates the frame count constraint
      */
-    public void setKeyframesWeight(float[] weights){
+    public void setKeyframesWeight(float[] weights) {
         if (times == null) {
             throw new IllegalStateException("MorphTrack doesn't have any time for key frames, please call setTimes first");
         }
         if (weights.length == 0) {
             throw new IllegalArgumentException("MorphTrack with no weight keyframes!");
         }
-        if(times.length * nbMorphTargets != weights.length){
+        if (times.length * nbMorphTargets != weights.length) {
             throw new IllegalArgumentException("weights.length must equal nbMorphTargets * times.length");
         }
 
@@ -148,10 +148,10 @@ public class MorphTrack implements AnimTrack<float[]> {
      *                 -- do not modify after passing it to this setter)
      */
     public void setKeyframes(float[] times, float[] weights) {
-        if(times != null){
+        if (times != null) {
             setTimes(times);
         }
-        if(weights != null){
+        if (weights != null) {
             setKeyframesWeight(weights);
         }
     }
@@ -159,7 +159,7 @@ public class MorphTrack implements AnimTrack<float[]> {
     /**
      * @return the number of morph targets
      */
-    public int getNbMorphTargets(){
+    public int getNbMorphTargets() {
         return nbMorphTargets;
     }
 
@@ -173,8 +173,8 @@ public class MorphTrack implements AnimTrack<float[]> {
      * @throws IllegalArgumentException if the number of weights and the new
      *         number of morph targets violate the frame count constraint
      */
-    public void setNbMorphTargets(float[] weights, int nbMorphTargets){
-        if(times.length * nbMorphTargets != weights.length){
+    public void setNbMorphTargets(float[] weights, int nbMorphTargets) {
+        if (times.length * nbMorphTargets != weights.length) {
             throw new IllegalArgumentException("weights.length must equal nbMorphTargets * times.length");
         }
         this.nbMorphTargets = nbMorphTargets;
--- a/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
+++ b/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
@@ -101,10 +101,13 @@ public class MorphTrack implements AnimTrack<float[]> {
         if (weights.length == 0) {
             throw new RuntimeException("MorphTrack with no weight keyframes!");
         }
+        if(times.length * nbMorphTargets != weights.length){
+            throw new RuntimeException("weights.length must equal nbMorphTargets * times.length");
+        }
 
         this.weights = weights;
 
-        assert times != null && times.length == weights.length * nbMorphTargets;
+        assert times.length * nbMorphTargets == weights.length;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stephengold:
Thanks for the contribution.

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
```diff
@@ -113,24 +132,33 @@ public void setTimes(float[] times) {
...
 
     /**
-     * Set the weight for this morph track
+     * Sets the weights for this morph track
      *
-     * @param times    a float array with the time of each frame (alias created
-     *                 -- do not modify after passing it to this setter)
-     * @param weights  the weights of the morphs for each frame (alias created
-     *                 -- do not modify after passing it to this setter)
+     * @param times        a float array with the time of each frame
```
stephengold, line 137: The "alias create" comment applies here, too.

stephengold (commented):
Nice work. I noted one javadoc detail that should be restored.

One undocumented/unchecked assumption is that the number of weights equals the number of frames times the number of morph targets. (This becomes clearer if you look at FrameInterpolator.java.) The pre-existing assertion in `setKeyframesWeight()` is wrong! It seems risky to allow changing the number of morph targets without supplying new weights. Could you change `setNbMorphTargets()` to require new weights?

danielperano:
> One undocumented/unchecked assumption is that the number of weights equals the number of frames times the number of morph targets.

This seems to be a critical condition that will cause either erratic behavior or array-out-of-bounds errors if it does not hold. Rather than check this in an assertion, I think we should check that the invariant holds when setting the weights in `setKeyframesWeight()` - setting weights should happen extremely infrequently, so I don't see a performance objection to checking this in an `if` branch rather than an assertion.

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
danielperano: Noted and applied both here and to the new `weights` parameter of `setNbMorphTargets()`.

stephengold:
Good thinking. I'll review this again once the build error at line 168 is resolved.

danielperano:
Lovely, not sure how I missed that - IDEA decided to stop doing checks, apparently. It's fixed now.

stephengold:
Won't `setNbMorphTargets()` usually cause the assert to fail now?

Also, I thought you were going to change the `assert` to an `if`. Did you change your mind?

danielperano:
I was holding off on the checks changes until hearing back from you since it's a departure from current convention. It's done now.

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
```diff
@@ -101,10 +101,13 @@ public void setKeyframesWeight(float[] weights){
         if (weights.length == 0) {
             throw new RuntimeException("MorphTrack with no weight keyframes!");
         }
+        if(times.length * nbMorphTargets != weights.length){
+            throw new RuntimeException("weights.length must equal nbMorphTargets * times.length");
```
stephengold, line 105: The exception type can be more specific. I advise using `IllegalArgumentException`. I encourage you to also improve line 102 along those lines.

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
```diff
@@ -101,10 +101,13 @@ public void setKeyframesWeight(float[] weights){
...
+        }
 
         this.weights = weights;
 
-        assert times != null && times.length == weights.length * nbMorphTargets;
+        assert times.length * nbMorphTargets == weights.length;
```
stephengold: This assertion is now 100% redundant and should be deleted.

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
```diff
@@ -165,8 +168,11 @@ public int getNbMorphTargets(){
      * @param nbMorphTargets the desired number of morph targets
      */
     public void setNbMorphTargets(float[] weights, int nbMorphTargets){
-        setKeyframesWeight(weights);
+        if(times.length * nbMorphTargets != weights.length){
+            throw new RuntimeException("weights.length must equal nbMorphTargets * times.length");
```
stephengold, line 170: should also be an `IllegalArgumentException`

stephengold (changes requested):
Better, but still not ready for integration in my opinion.

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
danielperano: Agreed - I wrote this out with `IllegalArgumentException` at first but changed to the current `RuntimeException` since that's what's currently used. `TransformTrack` also similarly misuses `RuntimeException`, but since the change will have little to no effect on existing exception handlers I think we should change all similar uses to `IllegalArgumentException`.

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
```diff
@@ -102,35 +127,58 @@ public MorphTrack(Geometry target, float[] times, float[] weights, int nbMorphTa
...
+            setTimes(times);
+        }
+        if(weights != null){
+            setKeyframesWeight(weights);
+        }
+    }
 
-            this.weights = weights;
+    /**
+     * @return the desired number of morph targets
```
stephengold, line 160: Please delete the word "desired".

jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
```diff
@@ -88,6 +88,31 @@ public MorphTrack(Geometry target, float[] times, float[] weights, int nbMorphTa
         return weights;
     }
 
+    /**
+     * Set the weight for this morph track. Note that the number of weights
```
stephengold, line 92: Please change "weight" to "weights".

stephengold (changes requested):

stephengold:
Thanks again, Daniel.
I think this PR is ready to be integrated, but I'll leave it open for 24 hours in case someone else wants to review it.
I believe this will resolve one of our old issues: #1089.

stephengold (approved):

Ali-RS:
It Will be nice to fix the formatting issues before merging. (the missing spaces in `if(` and `){`)

stephengold:
You're right, @Ali-RS.

Don't sweat this one, @danielperano --- I'll standardize the formatting for you.

stephengold:
@Ali-RS please verify that the changed lines are now formatted per our standards

Ali-RS:
Looks good to me. Thanks @stephengold

PR #1613, opened 2021-10-01
Wrap the input stream in BufferedInputStream

tonihele:
Introducing buffered stream for the ImageIO. Otherwise it reads in very small chunks and this is not desirable. Lots of calls. This is a performance improvement.

Old (starts with very small, and then proceeds with 512, maybe strides, dunno, variable read lengths):
image

New:
image

### DIFF
--- a/jme3-desktop/src/main/java/com/jme3/texture/plugins/AWTLoader.java
+++ b/jme3-desktop/src/main/java/com/jme3/texture/plugins/AWTLoader.java
@@ -41,6 +41,7 @@ import com.jme3.util.BufferUtils;
 import java.awt.Transparency;
 import java.awt.color.ColorSpace;
 import java.awt.image.*;
+import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
@@ -176,7 +177,7 @@ public class AWTLoader implements AssetLoader {
 
     public Image load(InputStream in, boolean flipY) throws IOException{
         ImageIO.setUseCache(false);
-        BufferedImage img = ImageIO.read(in);
+        BufferedImage img = ImageIO.read(new BufferedInputStream(in));
         if (img == null){
             return null;
         }
--- a/jme3-desktop/src/main/java/com/jme3/texture/plugins/AWTLoader.java
+++ b/jme3-desktop/src/main/java/com/jme3/texture/plugins/AWTLoader.java
@@ -177,7 +177,7 @@ public class AWTLoader implements AssetLoader {
 
     public Image load(InputStream in, boolean flipY) throws IOException{
         ImageIO.setUseCache(false);
-        BufferedImage img = ImageIO.read(new BufferedInputStream(in));
+        BufferedImage img = ImageIO.read(in);
         if (img == null){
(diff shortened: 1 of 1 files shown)


Ali-RS:
Thanks

tonihele:
I'm actually looking into many of these, and one I suspected the most was the BlockLanguageParser reading one char at the time. All the usages just pass the non-buffered stream there, yet it reads the files completely at one go... Well, I'll keep digging a bit and see if I find some others and make separate PRs then.

pspeed42:
Note: while probably nothing is using this method this way, ImageIO.read() does not close the stream and so this is a change in behavior if anyone was using AWTLoader to read multiple images from a single stream.  (This would break that code.)

It makes me wonder if JME's callers are closing the stream that is passed to this method.  If it is then that is also the place where wrapping in BufferedInputStream should be.  If it's not then we are maybe leaking file handles all over the place.

So while I agree that we should be wrapping our streams for performance, it's sometimes not quite black and white in deep cases like this.

Ali-RS:
Maybe better to move the changes into

https://github.com/jMonkeyEngine/jmonkeyengine/blob/143161e85930f16f3dc4ab25f10f6d33510c7e01/jme3-desktop/src/main/java/com/jme3/texture/plugins/AWTLoader.java#L188

which properly close the stream.

tonihele:
I think the buffered stream itself doesnt hold any resources. Closing it
actually just closes the underlying stream. So this change shouldnt affect
any streams left open etc. The normal assetLoader path closes all streams
therefore. Right?

MeFisto94:
But it reads stuff buffered, affecting `super.position`, so basically imagine a stream of multiple files and typically the loader only reads exactly the content. Buffered will read more, meaning the next invocation on super will be beyond that data, I guess.

tonihele:
> So while I agree that we should be wrapping our streams for performance, it's sometimes not quite black and white in deep cases like this.

Yep, it is very hard to determine the place for these...

tonihele:
> But it reads stuff buffered, affecting `super.position`, so basically imagine a stream of multiple files and typically the loader only reads exactly the content. Buffered will read more, meaning the next invocation on super will be beyond that data, I guess.

Yeah, true, now I get it. Yep, better to change it where @Ali-RS pointed out. There it is always the fresh stream popped out and used. So yeah, I'll move it.

pspeed42:
I brought this up because I've learned the lesson the hard way in some of my own utilities over the years. :)

Rule of thumb I use: you can only safely wrap a stream in a buffered stream where the stream was created.

It's also basically required for writing to streams and so the pattern holds in both directions (which is at least nice).

tonihele:
Alright, here we go. Ready on my part. I converted those also to try-with resources, I guess that is ok.

pspeed42:
Yeah, the in = null bit was a dumb anti-idiom even in the "old way".

Try-with-resource gets rid of a lot of weird habits some Java devs learned somehow.

stephengold:
Please update the copyright years in line 2.

tonihele:
> Please update the copyright years in line 2.

Cool, I was able to do it on the Github directly. Doneded!

PR #1618, opened 2021-10-11
Action & BaseAction classes java docs

pavly-gerges:
-Added Java docs.
-Added jme3 copyright licence.
-Code improvements.


stephengold:
I applaud your willingness to contribute, and I agree that these classes need javadoc. However, this PR is very confusing and needs much improvement. Explaining how to improve it will require much effort and probably multiple review passes.

For instance:
+ "Base implementation interface" is confusing since `Action` is not itself an interface.
+ "specialized to the new animation system" is confusing since there is no comparable class in the old animation system.
+ The sample code in the javadoc probably does more harm than good. The `<BR>` tags make the code hard to read, and it's unclear what the sample code is intended to illustrate.
+ "Inspired by Nehon" is probably misleading. I don't know for certain, but I assume Nehon wrote the code. "Inspired by" implies someone other than Nehon was the author.

Scrappers-glitch:
Hi there, yep i totally agree with you, i would change the "Inspired" to "Created By"

"Interface" : is  a famous misnomer used very often on design patterns to describe abstract things that are open for modification, and it doesn't need to be 100% percent java interface, i mean to use it then you have to extend it.

"Base Implementation" : This keyword is used before in jme tweens by @pspeed42 to describe the `AbstractTweens` which is a jme interface also that must be extended to create a tween, Nehon implied that he has based his whole animation system on the Tweens API, so `Action` & `BlendableAction` classes mimics the `AbstractTween` Class in behavior but they are "specialized" for Nehon's Animation system.

"Specialized" : means that starting from BlendableAction class, Nehon has stopped using raw Tweens inside the descendant classes, it turns to be a concrete implementation starting from this class, plus when you try to use a raw Tween with the new Anim System, the Action class encloses it in a base action, so its true that you cannot use raw tweens(unless you have actions to extract from them).

You will notice every single parameter is termed Tween based on Tween interfacing API, and there's a built in interoperability of tweens & Anim Actions within the `BaseAction` class.

Sample code meant to show the simple way things work, i have inspired this idea from some java util classes.

pspeed42:
"Base implementation" is clear.
"Base implementation interface" could do just as well without the word "interface".  From one reading, it's redundant and from another it's incorrect.  Either way, I'm not sure it adds anything to the statement but confusion.

pspeed42:
Another small nit-pick but it seems unusual to use <h2> to highlight the beginning class javadoc the way you do.

Scrappers-glitch:
Are there any other requested changes ?

PR #1621, opened 2021-10-16
solve issue 1619 (add texture-compression formats)

### LINKED ISSUE #1619: Add support for BC7 texture compression
BC7 hardware compression is part of DirectX11

On the zip file there is a PNG and its BC7 equivalent.

[16.zip](https://github.com/jMonkeyEngine/jmonkeyengine/files/7337119/16.zip)

tonihele:
Resolves #1619 

This PR adds support for BPTC compressed textures. Namely BC6 and BC7 compressed formats. Earlier, BC1-BC5 formats were already somewhat supported, that support is now enhanced a bit by making the DDS loader support them more widely. The file format varies widely, both DX9(?) style and the DX10 style files are supported for both BPTC and RGTC (BC4, BC5) compressed textures.

The BPTC was _mainlined_ in OpenGL 4.2. I put the constants to the EXT instead and check the capability from both the OpenGL version and the extension. There wasn't a file for GL4.2, so I didn't create one as I was little unsure how to integrate that. Therefore the GLExt houses these new constants. The version + extension checking I hope to make sure we utilize the feature even if the graphics driver doesn't give out all capabilities properly. Doesn't cost a thing...

There are some caveats. Support for flipping textures is limited. Not implemented at all for BC6 nor BC7. This is said to be extremely difficult but maybe not impossible. I can create a ticket for these flip issues.

Here is a report of findings of these formats (I used Compressonator to create these, the file formats vary wildly it seems):

- BC1 - Ok
- BC2 - Ok
- BC3 - Ok
- BC4 - Image appears as red
- BC4_S - Image appears as red, no flip support
- BC5 - Image colors weird
- BC5_S - Image colors weird,  no flip support
- BC6 - Ok, no flip support
- BC6_SF - Ok, no flip support
- BC7 - Ok, no flip support


tonihele:
The BC4 and BC5 image colour issue is maybe expected. Since they store the colour info on the red or the red-green channel respectively. But I'm unsure that do we need to use some swizzle or handle this in a shader separately? I don't have enough knowledge on the use cases or whatnots on RGTC textures.

stephengold:
How about providing test app(s) for this feature?

Issue #1361, opened 2020-04-23
TestMultiRenderTarget renders a black display


stephengold:
In the v3.1 branch, `TestMultiRenderTarget` renders a nice picture-in-picture containing 2 images of HoverTank. But in v3.2 and v3.3 branches, the display is all black aside from the stats view.

Here's typical console output:
```text
Apr 23, 2020 5:51:02 AM com.jme3.system.JmeDesktopSystem initialize
INFO: Running on jMonkeyEngine 3.3.0-SNAPSHOT
 * Branch: v3.3
 * Git Hash: 4b62b70
 * Build Date: 2020-04-23
Apr 23, 2020 5:51:02 AM com.jme3.system.lwjgl.LwjglContext printContextInitInfo
INFO: LWJGL 2.9.3 context running on thread jME3 Main
 * Graphics Adapter: null
 * Driver Version: null
 * Scaling Factor: 1
Apr 23, 2020 5:51:02 AM com.jme3.renderer.opengl.GLRenderer loadCapabilitiesCommon
INFO: OpenGL Renderer Information
 * Vendor: NVIDIA Corporation
 * Renderer: GeForce GT 545/PCIe/SSE2
 * OpenGL Version: 4.6.0 NVIDIA 390.116
 * GLSL Version: 4.60 NVIDIA
 * Profile: Compatibility
Apr 23, 2020 5:51:02 AM com.jme3.audio.openal.ALAudioRenderer initOpenAL
INFO: Audio Renderer Information
 * Device: OpenAL Soft
 * Vendor: OpenAL Community
 * Renderer: OpenAL Soft
 * Version: 1.1 ALSOFT 1.15.1
 * Supported channels: 64
 * ALC extensions: ALC_ENUMERATE_ALL_EXT ALC_ENUMERATION_EXT ALC_EXT_CAPTURE ALC_EXT_DEDICATED ALC_EXT_disconnect ALC_EXT_EFX ALC_EXT_thread_local_context ALC_SOFT_loopback
[...]

PR #1625, opened 2021-10-23
Rework Test MultiRenderTarget to actually show how to use multiple

MeFisto94:
Color Targets to render a scene separated by R, G and B values. It leaves out the complexity of the original test that was rendering a scene deferred.

As discussed in #1361 

Edit: Note to myself - I should remove the random color and disable the flyCam, so this can be asserted via [jme3-testing](https://github.com/MeFisto94/jme3-testing)

### DIFF
--- a/jme3-examples/src/main/java/jme3test/post/TestMultiRenderTarget.java
+++ b/jme3-examples/src/main/java/jme3test/post/TestMultiRenderTarget.java
@@ -87,7 +87,7 @@ public class TestMultiRenderTarget extends SimpleApplication implements ScenePro
      *
      * @param args unused
      */
-    public static void main(String[] args){
+    public static void main(String[] args) {
         TestMultiRenderTarget app = new TestMultiRenderTarget();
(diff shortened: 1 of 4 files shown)


stephengold:
Ready for review?

stephengold:
Thanks for creating this PR.

It needs some additional documentation. There should be class-level javadoc describing what it does ("demonstrates blah using 4 pictures, each displaying a different render target blah blah"). The pictures don't have descriptive names, so comments should explain how each one differs from the others. Also, a high-level description of the scene ("five cubes, each with a different blah blah"). Also some explanation of what "ExtractRGB.j3md" does.

Stylistically, there should be a space character between ")" and "{" in line 55 and both sides of each slash in lines 141-154. Also, each `Picture` field should be declared on a separate line.

stephengold:
@MeFisto94 are you still available to work on this?

stephengold:
I think this is ready for integration.

PR #1626, opened 2021-10-23
New animation system javadoc

pavly-gerges:
The docs, since I unfortunately wrongly disrupted my master branch with some forks, no changes are made to the already committed files since the last PR, so you can continue revising from where you have stopped.


jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -33,38 +127,62 @@ public double getLength() {
...
+     * the animation mask holds data(subsets of an object) about which part of the animation would be played.
+     * @return the animation mask instance.
+     */
     public AnimationMask getMask() {
         return mask;
     }
 
+    /**
+     * Sets the animation mask for this action.
+     * the animation mask holds the info about which part(subsets of an object) of the animation to play, like joints in #{@link com.jme3.anim.ArmatureMask#addFromJoint(Armature, String)}
```
stephengold, line 166: Add a space between "data" and "("
Change "like" to "such as",  which is more precise language.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -33,38 +127,62 @@ public double getLength() {
...
+     * @param mask the animation mask instance.
+     */
     public void setMask(AnimationMask mask) {
         this.mask = mask;
     }
 
+    /**
+     * Checks for the forward boolean flag.
```
stephengold, line 174: "Checks for" implies that the flag might not exist. Change "Checks for" to "Tests" or "Tests the value of".
Also, "boolean" is redundant with "flag". Use one word or the other, not both.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -33,38 +127,62 @@ public double getLength() {
...
+     * @return true if the animation action meant to be seek in a forward manner, false otherwise.
```
stephengold, line 175: "meant to" implies design intent. If the forward flag is controlled dynamically by software, use a different verb.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -33,38 +127,62 @@ public double getLength() {
...
+     */
     protected boolean isForward() {
         return forward;
     }
 
+    /**
+     * The forward flag, controls the animation action directionality, forward = true : ensures that animation would seek in a forward manner, however
```
stephengold, line 182: No comma between "flag" and "controls"

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -33,38 +127,62 @@ public double getLength() {
...
+     * forward = false : ensures that animation would seek in a backward manner.
+     * @param forward true if the animation is meant to be seek forward, false otherwise.
```
stephengold, line 184: Again, "meant to" is misleading.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -84,8 +201,7 @@ protected void setForward(boolean forward) {
     @Override
     public Action jmeClone() {
         try {
-            Action clone = (Action) super.clone();
-            return clone;
+            return (Action) super.clone();
```
stephengold, line 204: A documentation PR shouldn't change executable code. Please revert this change.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
```diff
@@ -6,10 +6,35 @@
 
 import java.util.List;
 
+/**
+ * A Concrete fast Implementation for the interface #{@link Action},
```
stephengold, line 10: This implementation is fast compared to what?
Also, "Concrete" and "Implementation" should not be capitalized.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
```diff
@@ -6,10 +6,35 @@
...
+ * used to create a subset of actions gathered from the extracted actions of the assigned tween.
+ *
+ * Example of operation :
+ * <pre class="prettyprint">
+ * //create a base action from a tween.
+ * final BaseAction action = new BaseAction(Tweens.parallel(clipAction0, clipAction1));
+ * //set the action properties - utilized within the #{@link Action} class.
```
stephengold, line 17: What does the example actually do?

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
```diff
@@ -6,10 +6,35 @@
...
+ * animComposer.setCurrentAction("basicAction", ActionState.class.getSimpleName());
+ * </pre>
+ * <b>Created by Nehon.</b>
+ */
 public class BaseAction extends Action {
 
     final private Tween tween;
 
+    /**
+     * Gathers the actions from a tween into a subset collection #{@link Action#actions},
```
stephengold, line 34: What is a "subset collection"?

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
```diff
@@ -6,10 +6,35 @@
...
+     * which joins anim actions hierarchy later on when instantiated by #{@link BlendableAction} implementation entites.
```
stephengold, line 35: What are "implementation entites"?
The abstract language used here obscures the key information, namely "How is this software useful?"

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
```diff
@@ -19,10 +44,18 @@ public BaseAction(Tween tween) {
         subActions.toArray(actions);
     }
 
+    /**
+     * Used to add the tween actions only if they are types of anim #{@link Action} class.
```
stephengold, line 48: "Used to" is unnecessary padding.
Describing in detail what the method does isn't very useful, particularly for a private method. Javadoc for private methods is only visible when browsing the source code. If we can't summarize what the method does at a higher level, there's no point in creating javadoc for it.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
```diff
@@ -19,10 +44,18 @@ public BaseAction(Tween tween) {
...
+     * Otherwise if they are in a raw tween state #{@link Tween}.
+     * then fetch the children actions of this tween & add them to subset of actions collection.
```
stephengold, line 50: In javadoc , the "&" character introduces an HTML entity. To avoid a javadoc error, use "&amp;" or (even better) spell it out as "and".

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
```diff
@@ -32,6 +65,8 @@ private void gatherActions(Tween tween, List<Action> subActions) {
 
     @Override
     public boolean interpolate(double t) {
+        /*interpolate -> calls doInterpolate(t:Double) -> which interpolate between
```
stephengold, line 68: Change "interpolate" to "interpolates".

stephengold (changes requested):
This is not a complete review. Before investing more review time, I want to evaluate the response to the changes requested today.

Scrappers-glitch:
@stephengold thanks for reviewing, give me sometime to read those requested changes and see.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -1,22 +1,112 @@
...
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
 package com.jme3.anim.tween.action;
 
 import com.jme3.anim.AnimationMask;
+import com.jme3.anim.Armature;
```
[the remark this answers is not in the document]
Scrappers-glitch: it is used by a reference link at the setMask() method documentation.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -1,22 +1,112 @@
...
 import com.jme3.anim.tween.Tween;
 import com.jme3.util.clone.Cloner;
 import com.jme3.util.clone.JmeCloneable;
 
+/**
+ * Base implementation (1st level impl/prototype) of the interface #{@link Tween} for the new animation system.
```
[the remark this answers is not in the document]
Scrappers-glitch: Again, Nehon has built his animation api based on Paul's Tweens from lemur.

In `Tweens` API, you will notice this pattern : 
Tween (interface)
|
Sequence, Parallel, ....., AbstractTween (Default implementation of tween, resembles functional code aka interpolates transforms)
|
Default Tweens

WHILE in the new system, Nehon build the system on 2 phases of Abstraction levels or layers : 
Tween
|
Action (Non-functioning tween, its a tween-collective phase)
|
BlendableAction (functioning tween), BaseAction (functioning tween, used to gather tweens and interpolate them)
|
ClipAction (the main functioning class), BlendAction (for blending between 2 actions using a blendSpace).

I can prepare and supply a UML diagram for the animation system to understand its core, we can have another discussion on forums for this anyway.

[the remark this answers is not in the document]
Scrappers-glitch: I think the production level would be the ClipAction, BaseAction, and BlendAction, those are the ones we use in our code tho.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -1,22 +1,112 @@
...
+ * The Action class encloses an array of {@link Tween} and converts them into jme animation actions if they actually aren't.
```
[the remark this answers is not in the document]
Scrappers-glitch: It's determined by the subclass `BaseAction` (notice that in the constructor block of this class), that can hold as many actions as you wish, so the final pattern performance is determined by the `Tweens` utility class.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -1,22 +1,112 @@
...
+ *
+ * NB : this implementation mimics the {@link com.jme3.anim.tween.AbstractTween}, but it delegates the interpolation method #{@link Tween#interpolate(double)}
+ * to the #{@link BlendableAction} class.
+ *
+ * Subclasses need to make a constructor call passing in the Tween Actions to enclose them & override #{@link Action#interpolate(double)}.
+ *
+ * Example of operation(Uses Abstract Factory Pattern Method(interpolate/doInterpolate)) :
```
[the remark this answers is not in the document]
Scrappers-glitch: To show a code that mimics the AbstractTween code, its an example for custom use, to do it in a one phase of abstraction layer, not 2 phases (as Nehon did).

But anyway normal users will not ever use this class in their code, they would rather use `ClipAction`, `BaseAction`, `BlendAction` and the `Tweens` utility class.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -1,22 +1,112 @@
...
+ *      protected abstract void doInterpolate(double t);
+ *  }
+ * </pre>
+ * <b>Created by Nehon.</b>
+ */
 public abstract class Action implements JmeCloneable, Tween {
 
     protected Action[] actions;
     private double length;
+    //the default speed is 1, which results in a normal tpf.
```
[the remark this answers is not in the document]
Scrappers-glitch: At the `AnimComposer.Layer` class, this formula is applied on the time per frames, before interpolating : 

`time += tpf * currentAction.getSpeed() * ac.globalSpeed`

if the speed is 1 then (time = time + tpf), so here tpf shows a normal value and not a scaled value.

I know its somehow frustrating to address everything, if you have a better alias let me know.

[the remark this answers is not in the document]
Scrappers-glitch: A quick look on the Tween interface docs shows : 
```java
 /**
     * Sets the implementation specific interpolation to the
     * specified 'tween' value as a value in the range from 0 to
     * getLength().  If the value is greater or equal to getLength()
     * then it is internally clamped and the method returns false.
     * If 't' is still in the tween's range then this method returns
     * true.
     *
     * @param t animation time (in de-scaled seconds)
     * @return true if t&gt;length(), otherwise false
     */
    boolean interpolate(double t);
```
t is a de-scaled value unless its scaled using the `action.speed` attribute.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -1,22 +1,112 @@
...
     private double speed = 1;
     private AnimationMask mask;
     private boolean forward = true;
 
+    /**
+     * Instantiate an action object which encloses a couple of tween animation actions.
```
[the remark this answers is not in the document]
Scrappers-glitch: I will change `couple` to `one or more`

i think empty args aren't allowed here and null args as well (i don't know if we should really document that, since most users won't use this class).

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -25,6 +115,10 @@ protected Action(Tween... tweens) {
         }
     }
 
+    /**
+     * Gets the length aka the duration of the current whole action that encloses Tweens.
+     * @return the length of the action.
```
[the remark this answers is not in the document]
Scrappers-glitch: I will do it.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -33,38 +127,62 @@ public double getLength() {
...
+     * @return the speed of frames.
+     */
     public double getSpeed() {
         return speed;
     }
 
+    /**
+     * Sets the speed of the frames of this action which gets applied for the Layer enclosing this action.
+     * Gets applied each advance() by this formula : time += tpf * currentAction.getSpeed() * ac.globalSpeed.
+     * @param speed the speed of frames.
```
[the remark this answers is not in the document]
Scrappers-glitch: yes, its 1.0, i will document that one.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
Scrappers-glitch: I will change `Checks for` to `tests the value of`.

In my code (and on other APIs like android/pi4j), sometimes `enum` values are used as flags (ie : to trigger something to behave based on a value), that's why i have used this terminology.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
Scrappers-glitch: The user can choose what to do and the `BlendableAction` class would apply the user choice on Runtime, but the default is `forward = true`.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
Scrappers-glitch: Alright, i will.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: thanks

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
pspeed42: That whole "(impl/prototype)" parenthetical is at best redundant and at worst confusing because "prototype" is an overloaded word that on the whole of language more commonly refers to something different than meant here.  The "Base implementation" and "what it does" are all that's needed.  The extra language doesn't add anything.

(Frankly, if JME wanted to 'adapt' tweens into something with more action-y juice when action should have been an interface with a base implementation.)

Also, you use other software pattern words in an odd way.  "doInterpolate()" is not a factory method by any stretch of the meaning of "factory pattern" that I know of.  A factory creates something.  I suspect you mean a different word here.

To me, Action seems to be an adapter class.  It adapts (one or more) more general Tweens into JME's "Action" which provides more animation-specific settings beyond what a generic tween would provide.  But since it is not implementing a new interface, I guess we could call it a decorator if we wanted to get fancy with our software pattern names.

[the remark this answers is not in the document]
Scrappers-glitch: I agree with the name "adapter", but the system uses the abstract factory pattern starting from `BlendableAction` and `doInterpolate()` is a real factory method, you will override to create your interpolation.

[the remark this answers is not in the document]
pspeed42: re: "//override this factory method to do interpolation"

Is it doing an interpolation or is it creating an object?

I actually really don't like the example provided as abstract AbstractAction extends Action (which is already abstract) feels very strange.  And it is encouraging the very very worst kind of class hierarchy (micro-inheritance over simple composition).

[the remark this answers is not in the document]
Scrappers-glitch: Its doing the interpolation, since no instance return.

Alright, no problem, I will remove the example code from all classes docs.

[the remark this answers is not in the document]
Scrappers-glitch: What about a `Creational pattern` instead of `Abstract factory pattern` ?

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: If that's a debate about its usage, I will remove the example code.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
pspeed42: Factory, creational... they are all the same.  They imply that you are CREATING something.  Which you say you are not.

Factory pattern:
```
public class MyCoolFactory implements SomeFactory {
    // factory method
    public TheThing createTheThing() {
        return new SpecialTheThing();
    }
}
```

If you aren't creating something then it is definitely 100% not a factory pattern.  And "creational" is just another name for "factory" here... and even has "Create" right in the name.

If all you are doing is overriding some behavior then you are overriding some behavior.  There is no special name for that other than "object oriented programming" or "specialization".  You are just extending the API of the class to allow more customization of behavior on some thing partially implemented.  That the specific example is a little upside down doesn't really change what it is.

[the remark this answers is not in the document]
Scrappers-glitch: You are right, but I am not sure if that's the case here as `doInterpolate` returns a boolean, if it its false then no interpolation is created, if its true then the interpolation is created.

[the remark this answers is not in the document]
pspeed42: I think just remove the example.  The engine has other examples that extend this class already, no?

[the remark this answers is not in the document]
Scrappers-glitch: I will remove the abstract factory pattern keyword from the docs.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: 1. If the class were only for internal use, then it wouldn't need example code. Instead, it would need a comment like "internal use only". But I don't believe that's true.

2. Since the class isn't strictly for internal use, example code is a fine idea. But the example should illustrate a more typical use, such as testing the current state of a `Layer`:
```java
        // Determine the current action in the composer's walking layer.
        Action action = animComposer.getCurrentAction("walking layer");
        Action standAction = animComposer.action("stand");
        if (action == standAction) {
            // not traveling --- do something safe
        } else {
            // traveling --- do something exciting
        }
```

3. Or, if you prefer the current example, then please add some introductory explanation. A comment like "The following code sample implements an abstract Tween similar to AbstractTween" might help.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: So, to recap : 
1) Rephrase this class doc including removal of redundant words.
2) Removal of example code.
3) Removal of Abstract factory pattern keyword.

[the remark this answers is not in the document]
Scrappers-glitch: @pspeed42 yes there are examples, do you want to document them by a see tag ?

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: I am not sure about that, but sounds a good idea anyway, I will rethink of a better testcase, but can I do this in a different PR ? So remove it for now and do it as an enhancement latter on ?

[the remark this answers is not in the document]
pspeed42: @stephengold It's not that it's for internal use only but there are plenty of classes in JME that stand as example use-cases without including a (bad one) in the javadoc.  We could just link to them until we have a more succinct and less abstract/wrongy one.  (I get the impression that if I fully understood this example then I could point to a much better way to do what it's trying to do... that doesn't involve extending the Action class.  Hard to say.)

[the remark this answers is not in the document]
pspeed42: @Scrappers-glitch and just to be clear, I'm not trying to put down the work you've been doing.  I think that it's great that you've taken on this task... but examples that are included in javadoc should be succinct and as perfect as possible.  (In fact, I'd argue that they could/should leave out the clone boiler plate, etc. and leave that to actual .java examples.)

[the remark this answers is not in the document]
Scrappers-glitch: Well, clearly I get your point, no problem at all, I realized that better I do an example that would involve inheriting the whole pattern and not just a non-functioning class, may be we could have something later on the wiki for custom usages.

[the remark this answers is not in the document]
Scrappers-glitch: Just in case, 
`
(I get the impression that if I fully understood this example then I could point to a much better way to do what it's trying to do... that doesn't involve extending the Action class. Hard to say.)`

The example I did, merges both the `BlendableAction` and `Action` classes, in a monophasic abstraction level..., but let's just remove this and focus on the real java docs for now.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: I think the javadoc should say that subclasses determine how the tweens are combined.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: Removing the example seems like a good solution.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: So `speed` doesn't affect `tpf` in any way. Instead of "a normal tpf" say something like "animation time = real time" or "animation clips run at their designed speed".

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
stephengold: So it's not a subset of anything. What this constructor does is to an empty action, one without any sub-actions. The proposed javadoc implies the opposite.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: I will change `normal tpf` to `animation clips run at their normal speed`.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: Which subclasses ? There are no subclasses for `BaseAction`, may be you mean its determined by subclasses of `Tween` ?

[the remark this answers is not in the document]
Scrappers-glitch: I will document that if so.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
Scrappers-glitch: I think removal of the `subset` keyword would resolve this.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: No, I mean the subclasses of `Action`. The manner in which an action's tweens are combined is determined by the subclass, not by `Action` itself.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
stephengold: That's better.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: Yes, its determined by the BaseAction, and its the only job of BaseAction.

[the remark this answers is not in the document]
Scrappers-glitch: I will document that too.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
Scrappers-glitch: @stephengold this example is used in production code, would i remove it too ? 

This is the github version : 
https://github.com/Scrappers-glitch/DBTraining/blob/06e968a5fc94d16f577629b869193c392083e9e0/app/src/main/java/com/scrappers/dbtraining/mainScreens/prefaceScreen/renderer/animationWrapper/EmitterTween.java#L310

Scrappers-glitch:
I have made the requested changes for the Action class, and rephrased some parts, hope its better now, please review them.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -38,68 +38,42 @@
...
     private AnimationMask mask;
     private boolean forward = true;
 
     /**
-     * Instantiate an action object which encloses a couple of tween animation actions.
+     * Instantiate an action object which holds one or more tweens.
+     * If the tweens are {@link Action}, they would be added directly to {@link Action#actions}.
+     * If the tweens aren't {@link Action}, the Action class would delegate that to {@link BaseAction}, and {@link BaseAction} would try to extract the
+     * tween actions from the tweens arg and interpolate them.
+     * Notes :
```
Scrappers-glitch, line 72: Those notes are to illustrate the behavior of the subclasses towards this constructor, and that's why not all subclasses of Action make a `super(tweens)` call.

Scrappers-glitch, line 72: Those notes are to illustrate the behavior of the subclasses towards this constructor, and that's why not all subclasses of Action make a `super(tweens)` call.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -38,68 +38,42 @@
...
+     * - If intentions are to make a holder class, then subclasses have to call this constructor, examples : {@link BlendableAction}, {@link BaseAction}.
```
Scrappers-glitch, line 73: Those notes are to illustrate the behavior of the subclasses towards this constructor, and that's why not all subclasses of Action make a `super(tweens)` call.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -38,68 +38,42 @@
...
+     * - If intentions are to make an implementation of {@link Action} that shouldn't hold tweens of actions, then subclasses shouldn't call this
+     * constructor, examples : {@link ClipAction}, {@link BlendAction}.
+     *
```
Scrappers-glitch, line 76: Those notes are to illustrate the behavior of the subclasses towards this constructor, and that's why not all subclasses of Action make a `super(tweens)` call.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -70,9 +70,9 @@
      * If the tweens aren't {@link Action}, the Action class would delegate that to {@link BaseAction}, and {@link BaseAction} would try to extract the
      * tween actions from the tweens arg and interpolate them.
      * Notes :
-     * - If intentions are to make a holder class, then subclasses have to call this constructor, examples : {@link BlendableAction}, {@link BaseAction}.
+     * - If intentions are to make a holder class, then subclasses have to call this constructor, examples : {@link BlendableAction}, {@link BaseAction}, {@link BlendAction}.
      * - If intentions are to make an implementation of {@link Action} that shouldn't hold tweens of actions, then subclasses shouldn't call this
-     * constructor, examples : {@link ClipAction}, {@link BlendAction}.
+     * constructor, examples : {@link ClipAction}.
      *
```
Scrappers-glitch, line 76: Those notes are to illustrate the behavior of the subclasses towards this constructor, and that's why not all subclasses of Action make a `super(tweens)` call.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -25,46 +89,76 @@ protected Action(Tween... tweens) {
...
+     * the animation mask holds the info about which part (subsets of an object) of the animation to play, such as joints in {@link com.jme3.anim.ArmatureMask#addFromJoint(Armature, String)}.
+     * @param mask the animation mask instance.
+     */
     public void setMask(AnimationMask mask) {
         this.mask = mask;
     }
 
+    /**
+     * Tests the value of the forward flag.
+     * @return true if the animation action is in a forward manner state.
```
stephengold, line 151: It's not clear what a "forward manner state" is, nor what `false` indicates. I believe this method tests whether the action is running forward (true) or backward (false).

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -84,7 +177,7 @@ protected void setForward(boolean forward) {
     @Override
     public Action jmeClone() {
         try {
-            Action clone = (Action) super.clone();
+            final Action clone = (Action) super.clone();
```
stephengold, line 180: Adding `final` shouldn't be part of this PR.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
stephengold: Better. I would say, "A simple implementation of the Action interface."

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
stephengold: Please provide a more specific comment here. Something like "The following example runs two clips in sequence, at double speed, stopping after 5 seconds."

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
stephengold: Then please remove the word "subset".

stephengold (changes requested):
You still need to apply the changes you agreed to make.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
Scrappers-glitch: That's true indeed.

jme3-core/src/main/java/com/jme3/anim/tween/action/BaseAction.java
Scrappers-glitch: If you think that the new change is good and clarify things, please resolve this conservation and continue reviewing.

Thanks.

Scrappers-glitch:
I have cleaned up `BlendableAction` before being reviewed to minimize time reviewing it.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
-
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * A concrete implementation used to blend between 2 actions #{@link BlendAction#firstActiveIndex}, #{@link BlendAction#secondActiveIndex} & apply them on their targets,
+ * blending between 2 actions is done by interpolating the set of transforms serially starting from the firstAction transforms, moving on to the second one transforms,
+ * using the blendWeight as a delta value, then interpolating the resulted value with the current target local transforms,
+ * and finalizing by applying the final collected result of interpolation to the target as a local transform matrix.
```
stephengold, line 44: I find this description cryptic and confusing for many, many reasons. Instead of describing the algorithm step by step, summarize how the class is used. Something like "An implementation of BlendableAction that blends a list of actions using a specific BlendSpace."

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
+ *
+ * If the blendWeight is :
```
stephengold, line 46: Since `blendWeight` is a private field, it shouldn't be mentioned in the class javadoc.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
+ * <b>Created by Nehon.</b>
+ */
 public class BlendAction extends BlendableAction {
 
     private int firstActiveIndex;
     private int secondActiveIndex;
     final private BlendSpace blendSpace;
-    private float blendWeight;
-    final private double[] timeFactor;
+    private double[] timeFactor;
```
stephengold, line 58: Why did you delete `blendWeight` here? Why did remove the `final` qualifier from `timeFactor`?

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
     final private Map<HasLocalTransform, Transform> targetMap = new HashMap<>();
 
+    /**
+     * Creates an action that utilizes a blend space to blend between 2 selected actions from the collection of actions provided.
```
stephengold, line 62: Only 2 actions are used? That doesn't seem right.

And if it *is* correct, then what determines which 2 actions are used?

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
+     * @param blendSpace the blendSpace, used for calculating the blendWeight
```
stephengold, line 63: Don't refer to private fields in javadoc.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
         blendSpace.setBlendAction(this);
 
-        for (BlendableAction action : actions) {
+        //prepare for interpolating actions by gathering targets & timeFactors based on the actionLength & the global length.
+        gatherTargetsFromActions(actions);
+        gatherTimeFactorPerAction(actions);
+    }
+
+    /**
+     * Used for gathering the timeFactor or the timeDelta value for the inscribed actions.
```
stephengold, line 77: What are "inscribed actions"?

stephengold: This description doesn't clearly explain how the methoddoes at a high level. As the method is private, it might be better to delete the javadoc and let the source code "speak for itself".

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
+                this.timeFactor[i] = actionLength / getLength();
+            }else{
+                //assign a default value if durations are negative
+                this.timeFactor[i] = 1;
+            }
+        }
+    }
+
+    /**
+     * Used for gathering the targets & prepare them from the parsed anim actions args.
```
stephengold, line 98: Parsing refers to creating a hierarchy from a string of text or tokens. This method doesn't parse anything.

This javadoc obscures what the method actually does and doesn't say much that isn't conveyed by the menthod name. It might be better to delete the javadoc.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
+     * @param actions the actions used for gathering targets.
+     */
+    private void gatherTargetsFromActions(final BlendableAction... actions){
+        for (final BlendableAction action : actions) {
+            //stretch the blend action length(duration) to the max possible length to accommodate the actions.
```
stephengold, line 103: Not the "max possible length" but the "length of the longest action".

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
-                double actionLength = this.actions[i].getLength();
-                if (actionLength > 0 && getLength() > 0) {
-                    this.timeFactor[i] = this.actions[i].getLength() / getLength();
-                }
-            }
-        }
     }
 
+    /**
+     * The factory method, overridden to manage the interpolation.
```
stephengold, line 124: How is `doInterpolate()` a factory method?

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
+        interpolateAction(secondActiveAction, Math.min(Math.max(blendWeight, 0f), 1f), t * timeFactor[secondActiveIndex]);
+        //release the resources.
+        resetTransformationDelegation(firstActiveAction, secondActiveAction);
+    }
 
-        //Second action should be interpolated
-        secondActiveAction.setWeight(blendWeight);
-        secondActiveAction.interpolate(t * timeFactor[secondActiveIndex]);
+    /**
+     * Resets the delegation & release the resources.
```
stephengold, line 145: Ampersands in javadoc should be "&amp;" or "and".

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
+     * @param firstAction the firstActiveAction.
+     * @param secondAction the secondActiveAction.
+     */
+    private void activateTransformationDelegation(final BlendableAction firstAction, final BlendableAction secondAction){
+        firstAction.setCollectTransformDelegate(this);
+        secondAction.setCollectTransformDelegate(this);
+    }
 
+    /**
+     * Interpolates the actions according to a blendingWeight(delta value).
```
stephengold, line 166: The Greek letter "delta" usually stands for change. The blend weight is neither a delta value nor a delta factor.

stephengold (changes requested):
I got a little further on this review pass, but didn't get to review all the changes. I'd like to review this PR yet again after all the current changes have been addressed.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -1,22 +1,86 @@
...
 import com.jme3.anim.AnimationMask;
+import com.jme3.anim.Armature;
 import com.jme3.anim.tween.Tween;
 import com.jme3.util.clone.Cloner;
 import com.jme3.util.clone.JmeCloneable;
 
+/**
+ * Base implementation of the interface {@link Tween} for the new animation system.
+ *
+ * The Action class gathers the animation actions from a {@link Tween} array argument into {@link Action#actions}, and it extracts the plain {@link Tween} instances (non-action tween)
```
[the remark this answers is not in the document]
Scrappers-glitch: it implies the constructor argument, i would remove the word argument since it adds nothing.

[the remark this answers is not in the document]
Scrappers-glitch: > "gathers" refers to combining separate items into one. Perhaps the verb you want here is "converts"?

Well, not really, if they are instanes of Action then we don't convert and add them directly to action[].

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -25,46 +89,75 @@ protected Action(Tween... tweens) {
         }
     }
 
+    /**
+     * Gets the length aka the duration of the current whole action that encloses Tweens.
```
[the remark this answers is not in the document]
Scrappers-glitch: The Action class creates an `action[]` collection that holds the tweens, so it have to have tweens.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: I think keeping javadoc for a private method won't hurt after all javadoc is for us is for reading only, can i just rephrase it and keep it ?

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: I haven't applied the changes requested by Paul on BlendAction yet.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: The blending weight is a changeable value through the `BlendSpace` either Linear or non-Linear equations can change it.

stephengold: Just because a value changes over time does not make it a delta.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: But its an adjustable field through the blendSpace.
`blendWeight = blendSpace.getWeight()`

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: The global field `blendWeight` here is used only at one place so no need to be global then.

The `timeFactor` field was final as it was used inside the constructor directly, but when moving what's inside the constructor to relevant methods, the `final` needs to be removed.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: Successive actions, blending is done between 2 actions only, you can choose those via your BlendSpace, i will re-document that to clarify it.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: The problem is this isn't really a private field, its just a reference holder for the `public BlendSpace#getWeight()`.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: Right, i will change it.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -25,46 +89,75 @@ protected Action(Tween... tweens) {
...
+    /**
+     * Tests the value of the forward flag.
+     * @return true if the animation action is running forward, false otherwise.
+     */
     protected boolean isForward() {
         return forward;
     }
 
+    /**
+     * The forward flag controls the animation action directionality.
```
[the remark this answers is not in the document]
Scrappers-glitch: Yes, but what about if someone inherits it, should he be able to know what it does ?

Scrappers-glitch:
@stephengold Could you please review the `Action` class again ? 

I am really sorry for these stupid mistakes, i will try to omit these mistakes as possible on the other un-revised classes.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: Alright, i would remove the keyword 'delta'.

Scrappers-glitch:
I'd like to finish the review of the `Action` and `BaseAction` first then move on to other classes.

stephengold:
> I am really sorry for these stupid mistakes

None of the mistakes I've seen so far are "stupid".

Be gentle on yourself. You've assigned yourself a daunting task. Documenting someone else's code is always a challenge. The new animation system is particularly obscure. And English is (I think) a language especially unsuited to writing technical documentation.

The good news is that your skills are improving. You're already better at writing documentation than you were a few weeks ago. I hope you see that.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: I now think "gathers" is OK.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: so delete "that encloses Tweens"

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: Yes, it needs javadoc. I request additional javadoc that will advise app developers to use the `setSpeed()` method instead.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
stephengold: Then mention that instead, something like "using the weight from the BlendSpace"

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
stephengold: Code changes shouldn't be included in a documentation PR. Please revert those changes. You can submit them in the future as part of a different PR.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: > I now think "gathers" is OK.

I have replaced gathers with collects, is that okay ?

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
stephengold: "the BlendSpace that calculates the weight applied to each Action"

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: Yep, i have removed nonsense keywords and rephrased some, please check the new today's commits.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
stephengold: Yes, if you can rephrase it to be accurate and not obscure.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
stephengold: Sure. You might also consider "organizes".

stephengold:
After you've made the changes you promised, I'll make another review pass.

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
```diff
@@ -25,59 +90,99 @@ protected Action(Tween... tweens) {
...
+     * @param speed the speed of frames.
+     */
     public void setSpeed(double speed) {
         this.speed = speed;
-        if( speed < 0){
-            setForward(false);
-        } else {
-            setForward(true);
-        }
+        setForward(!(speed < 0));
```
stephengold, line 134: This change makes the logic less clear. Also, a documentation PR shouldn't change the code.

Scrappers-glitch: Well, sometimes in order to do some docs, code needs to be simplified.

The original code was : 
```java 
if(speed < 0) {
setForward(false);
}else{
setForward(true);
}
```
So if its simplified then : setForward(!(speed < 0));

But also can be : setForward(speed > 0);

And I think setForward(speed > 0) is the most clear form, in which the animation runs forward if the speed is positive !

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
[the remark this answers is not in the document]
Scrappers-glitch: Great ! i will provide that !

jme3-core/src/main/java/com/jme3/anim/tween/action/Action.java
pspeed42: Not sure why the code had to change for the javadoc to explain what is happening.  Seems like the javadoc describes what the method does whether the old code is used or not.

...javadoc changes should cause 0 regressions.  The only way to guaranteed 0 regressions is to not change the code.

(Strictly speaking, setForward(!(speed < 0)); and setForward(speed > 0); are not equivalent.  setForward(speed >= 0) would be equivalent... and while it _probably_ won't cause a regression, "no code changes" is 100% guaranteed to not cause a regression. ;))

Else we end up with a situation where we can't integrate **javadoc** to some version just because it might change behavior for someone.  Way more maintenance burden than should be required for documentation.

stephengold: Please put the code back the way it was. If you believe the code should be improved, open and issue or a separate PR for that change. A PR should focus on one thing, and in this case that's clearly javadoc.

Scrappers-glitch: I will revert all the changes in code for now.

stephengold:
@Scrappers-glitch Are you still available to work on this PR?

Scrappers-glitch:
> @Scrappers-glitch Are you still available to work on this PR?

Yeah ofc, I have not leaved the project, but I am currently have ongoing exams that will extend up to two weeks.

stephengold:
Thanks for the update ... and best of luck with your exams!

Scrappers-glitch:
This is still on my todo, give me a few months, i will free somethings out from my stack and then return for those requested changes, i have a better suggestion for this, may be i will close this PR and separate the work into multiple organized PRs based on the packages, we will work on each package separately and i will make sure our conversation here is not wasted out by picking up the important requested changes !

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
 import com.jme3.math.Transform;
-
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * A concrete implementation used to blend between 2 actions #{@link BlendAction#firstActiveIndex}, #{@link BlendAction#secondActiveIndex} & apply them on their targets,
+ * blending between 2 actions is done by interpolating the set of transforms serially starting from the firstAction transforms, moving on to the second one transforms,
+ * using the blendWeight as a delta value, then interpolating the resulted value with the current target local transforms,
```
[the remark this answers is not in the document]
Scrappers-glitch: @stephengold I would like to discuss some of these changes here, before shipping the documentation to a newly separated PR.

I looked up the original diagram created by Nehon on forums, the most righteous expression that could be used to describe the "blend weight" is "blending driver value", What do you think?

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: This won't be shipped on the new PR, no worries.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: Thanks, I will take this into consideration.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
```diff
@@ -1,129 +1,265 @@
...
 import com.jme3.anim.util.HasLocalTransform;
 import com.jme3.math.Transform;
-
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * A concrete implementation used to blend between 2 actions #{@link BlendAction#firstActiveIndex}, #{@link BlendAction#secondActiveIndex} & apply them on their targets,
+ * blending between 2 actions is done by interpolating the set of transforms serially starting from the firstAction transforms, moving on to the second one transforms,
```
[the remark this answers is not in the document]
Scrappers-glitch: > Interpolation is not performed serially. Please delete the word "serially".

I will see how I should describe this more clearly, it is kinda an incremental pattern, interpolation is made on the previously interpolated value, I guess that "serially" is not the right word to describe this.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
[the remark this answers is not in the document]
Scrappers-glitch: I think I have now remembered why I chose to use "delta" to describe the blendweight, the `ClipAction` class calls the `Transform#interpolateTransforms(Transform t1, Transform t2, float delta)` which passes the weight as the delta value of interpolation!!

I will let you decide on this, but in my opinion, it's mathematically a delta value.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: "Blends" is a meaningless keyword. Me as an animation noob doesn't have any idea of what Blending actually is, the word is vague and needs a good description, most of the time, the best way is to describe things mathematically.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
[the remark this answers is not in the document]
stephengold: I still think "delta" is confusing, both here and in "Transform.java". It's the parameter of the interpolation, which controls the relative weight given to each input, right? To me "delta" describes the difference between 2 inputs.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: Yes, it should be 'Wrapped Actions', almost all implementations of "Action" wrap actions or other tween actions except "ClipAction" which wraps the actual animation data.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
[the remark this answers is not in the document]
Scrappers-glitch: > I looked up the original diagram created by Nehon on forums, the most righteous expression that could be used to describe the "blend weight" is "blending driver value", What do you think?

It's good to see the diagram and understand what a blending driver is, but in reality, people won't look-up this diagram, it must be crystal clear "What's the blending?" and "What's the weight of blending?" which could summarised in this sentence "a delta value (or scale factor) for animation transformation data interpolation".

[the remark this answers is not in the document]
Scrappers-glitch: > I still think "delta" is confusing, both here and in "Transform.java". It's the parameter of the interpolation, which controls the relative weight given to each input, right? To me "delta" describes the difference between 2 inputs.

Okay, so, does the "scale factor of interpolation" satisfies a good description for the blending weight?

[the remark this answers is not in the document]
Scrappers-glitch: Btw, the parameter of linear interpolation is calculated mathematically by dividing the value of "x(in) - x0" by the value of "x1 - x0", so that is why it is called delta, IMHO, it describes the change of x(in) with respect to the origin point (x0) in the way to x1.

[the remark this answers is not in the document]
pspeed42: It's just the "value for interpolation".  I agree delta makes no sense.  0 means the first value, 0.5 means halfway between the two values, 1 means the second value.

re: "it describes the change of x(in) with respect to the origin point (x0) in the way to x1." ...which would be fine if it wasn't rescaled to be between x0 and x1 again.  It's not a delta.

JME tends to call it "scale" but that's not really correct either.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
pspeed42: It's more important for this to describe what it's doing and not how it's doing it.  I find your description no more illuminating than reading the code... and the code is cryptic enough, too... (and I suspect some methods and parameters could have better names).  Without seeing examples, I would have had no idea what this code was doing, and as detailed as it is, I'd have been no better off with the javadoc above.

I don't know what the solution is but it feels like we are not getting closer to it.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendableAction.java
```diff
@@ -7,6 +38,20 @@
 import com.jme3.util.clone.Cloner;
 import java.util.Collection;
 
+/**
+ * An implementation of the Action interface, used to manage the interpolation code
```
pspeed42, line 42: Quibble: Action is not an interface.  This is a "specialization" of the Action "base class".

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
Scrappers-glitch: I will try to summarize the action of this class as much as I can.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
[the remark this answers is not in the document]
Scrappers-glitch: Ok, I will use "value for interpolation" instead of "delta" to avoid confusion. Thanks.

pspeed42:
As a general comment, every time I pick up this animation system again I feel forced to relearn all of its weird concepts from scratch again... usually be reading the code and finding working examples.  While I appreciate the effort that has gone into the documentation so far, I'm not sure it's really alleviated that problem for me.

It's unfortunate that nehon never gave us many bread crumbs because I'm sure the author would have had insights into what the reasoning was that we have to reverse engineer.

That said, I'm not sure what the solution really is, either.  The javadoc should help tell us how to use these classes with some high level explanations but we somehow miss this mark.  It always takes me a lot of time to re-wrap my head around it and so I can't really provide much help on how to push it closer to that goal.

...so in the end, my comments are not helpful, I guess.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
pspeed42: But note: the code rewritten in English is not really a summary unfortunately.  But neither do I know how to fix it.

pspeed42:
Put another way, the javadoc for these classes should be like stereo instructions.  They should tell me how to plug things together, what the impedance ranges are so I don't blow something up, what the knobs do, etc..  I don't really need circuit diagrams and chip documentation.

First and foremost, the javadoc should be answering "why" and "when" do I use a particular class.  "How" do I set it up to do that?  What are the limitations I can expect?  The details of internal workings are for internal code comments for someone trying to change the class later.  Javadoc is for the users.  Dumb it down a lot.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendableAction.java
Scrappers-glitch: Great, I am going to remove all the architectural keywords, since the original author is not here, I will try to stick more to describing the behavior of the system.

Scrappers-glitch:
> and if someone ever comes up with a really good explanation about the difference between actions and anim clips and th

The only thing I know currently about this from the code, and some words from the original author on forums is that `AnimClip` is where the original animation data are wrapped, while `Actions` is where the porting of Tween to the animation system came from, Nehon has built an abstraction layer over Tweens and didn't use the Tweens API directly, I think that is a flaw in the design and maybe it's a residual part from the old State-Machine pattern, but maybe I am wrong, maybe it supplies out some good code tricks that wouldn't have been if he used the Tween API directly, I have to really look back in forums to understand more as I weren't there during these days.

Ali-RS:
My two cents:

I think the current flow in design is due to having a backward playback (negative speed) and the other thing is that the animation mask is passed by the animation layer to the action. Because of those the Action class needs to "visit" the passed tweens (that can be chained) and extract the actions (ClipAction and BlendActions).

To solve those, for backward playback, we can use an Invert tween and instead of passing mask by layer, we can require the user to set the mask on the action when creating the action.

Now we should be able to remove BaseAction and Action classes and just keep the BlendableAction, ClipAction, and BlendAction. (we can rename them if required.)

And the next step could be to remove the AnimLayer and instead use TweenAnimation and AnimationState (imported from Lemur) to run animations. For this, AnimComposer needs to be refactored to use the "anim channels", very similar to the EffectControl in Lemur.

>difference between actions and anim clips and the weird wiring between them

AnimClip (IMO could be renamed to AnimData or ClipData), as it contains the animation track data which is savable to j3o whereas the ClipAction (can not think of a better name ATM) is the Tween ("runtime" thing) that uses the anim data and can not be saved itself.

Scrappers-glitch:
> I think the current flow in the design is due to having a backward playback (negative speed) and the other thing is that the animation mask is integrated into the animation layer (the mask is passed by the anim layer to the action). Because of those the Action class needs to "visit" the passed tweens (that can be chained) and extract the actions (ClipAction and BlendActions).
> 
> To solve those, for backward playback, we can use an Invert tween and instead of passing mask by layer, we can require the user to set the mask on the action when creating the action.
> 
> Now we should be able to remove BaseAction and Action classes and just keep the BlendableAction, ClipAction, and BlendAction. (we can rename them if required.)
> 
> And the next step could be to remove the AnimLayer and instead use TweenAnimation and AnimationState (imported from Lemur) to run animations. For this, we should refactor AnimComposer to use "channels" like the Lemur EfectControl I think.

It seems a good idea, though it smells of re-designing the whole system or parts of it and re-designing a system without understanding some of its parts might break things out.

Out of innovation, we may try this on a fork or a new API before thinking to change the current API, and I am ready to participate in this, it will be a great learning experience!

I hope there won't be a de novo new Animation System after this :-).

Ali-RS:
>Out of innovation, we may try this on a fork or a new API before thinking to change the current API, and I am ready to participate in this, it will be a great learning experience!

That sounds a good idea.

Also, I think "speed" should **not** be an attribute on the tween/action but it should be an attribute on the thing that "runs" the tween. (i.e. AnimLayer or TweenAnimation)

Scrappers-glitch:
> > Out of innovation, we may try this on a fork or a new API before thinking to change the current API, and I am ready to participate in this, it will be a great learning experience!
> 
> That sounds a good idea.
> 
> Also, I think "speed" should **not** be an attribute on the tween/action but it should be an attribute on the thing that "runs" the tween. (i.e. AnimLayer or TweenAnimation)

After documenting the base API, I will open a forum thread to discuss this as a community project so we can reach a better solution.

Issue #1622, opened 2021-10-17
Android image loaders should use BufferedInputStream


Ali-RS:
Similar to AWTLoader (#1613), the AndroidBufferImageLoader and AndroidNativeImageLoader should be updated to use BufferedInputStream.

stephengold:
Thanks for documenting this issue.

PR #1627, opened 2021-10-23
BufferedInputStream enhancement android

Scrappers-glitch:
Targeting issue #1622

### DIFF
--- a/jme3-android/src/main/java/com/jme3/texture/plugins/AndroidBufferImageLoader.java
+++ b/jme3-android/src/main/java/com/jme3/texture/plugins/AndroidBufferImageLoader.java
@@ -84,7 +84,7 @@ public class AndroidBufferImageLoader implements AssetLoader {
         options.inPurgeable = true;
         options.inSampleSize = 1;
         
-        try(final BufferedInputStream bin = new BufferedInputStream(assetInfo.openStream())) {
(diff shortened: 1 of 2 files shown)


Scrappers-glitch:
@Ali-RS Hello, 

if you have already tested the `AndroidBufferImageLoader` and its working fine, then no need to change it, and this PR would be ready to be merged.

Ali-RS:
Thanks. 
Going to merge this in next 24H.

jme3-android/src/main/java/com/jme3/texture/plugins/AndroidBufferImageLoader.java
```diff
@@ -158,8 +153,7 @@ public Object load(AssetInfo assetInfo) throws IOException {
         data.flip();
         
         bitmap.recycle();
-        
-        Image image = new Image(format, width, height, data, ColorSpace.sRGB);
-        return image;
+       
+        return new Image(format, width, height, data, ColorSpace.sRGB);
```
stephengold, line 157: I prefer the old code, which assigned the new object to a variable and returned the variable. That made it easy to examine the new object in the debugger.

Scrappers-glitch: Changes Applied

stephengold:
My thanks to you both!

PR #1629, opened 2021-10-24
Support texture flipping for signed versions of RGTC 1 and 2

### LINKED ISSUE #1623: Add support for texture flipping of BC4 and BC5 signed formats
Add support for texture flipping of BC4 and BC5 signed formats. This is maybe just a trivial task of setting the block size right.

You can use DDS as the texture container and the code handling the flipping is in https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-core/src/plugins/java/com/jme3/texture/plugins/DXTFlipper.java

tonihele:
Resolves #1623 

Okay, it was easy...

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/texture/plugins/DXTFlipper.java
+++ b/jme3-core/src/plugins/java/com/jme3/texture/plugins/DXTFlipper.java
@@ -220,9 +220,11 @@ public class DXTFlipper {
                 type = 3;
                 break;
             case RGTC2:
+            case SIGNED_RGTC2:
                 type = 4;
                 break;                
             case RGTC1:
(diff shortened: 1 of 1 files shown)


stephengold:
Have you tested the new features?

tonihele:
> Have you tested the new features?

I think so yes. With this https://github.com/jMonkeyEngine/jmonkeyengine/blob/5e2350cc5abd3e9d7551a0b417b4eeb6eedb25e2/jme3-examples/src/main/java/jme3test/texture/dds/TestLoadDds.java.

stephengold:
Thanks for your latest contribution!

PR #1635, opened 2021-11-02
ignore return values from AnimMigrationUtils.migrate()

stephengold:
Our best guidance for how to use `AnimMigrationUtils` is found in 5 example apps in jme3-test. However, 2 of those apps ( `TestJaime` and `TestAttachmentsNode`) assign return values from `migrate()` to local variables, giving the impression that the model might be converted by creating a new `Spatial`. 

In fact, `migrate()` converts models in place. Its return value is simply the argument. The assignments are redundant and also misleading, so this PR removes them from `TestJaime` and `TestAttachmentsNode`.


stephengold:
Unless there's substantive feedback, I plan to integrate this in 24 hours.

PR #1638, opened 2021-11-05
solve issue #1696 (missing check for GLFW_KEY_UNKNOWN)

stephengold:
This addresses the issue by returning `null` if the key is unknown.


stephengold:

Issue #1612, opened 2021-09-30
NPE while loading glTF model "Hover bike - The Rocket"


stephengold:
To reproduce the issue;

1. download the Autoconverted format glTF from https://sketchfab.com/3d-models/hover-bike-the-rocket-8b2e5bfca78e41c791b4e5b5e8c04512
2. unzip the files into "jme3-examples/src/main/resources"
3. add `assetManager.loadModel("scene.gltf");` to the example app of your choice

Typical stack trace:
```text
com.jme3.asset.AssetLoadException: An error occurred loading scene.gltf
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:181)
	at com.jme3.scene.plugins.gltf.GltfLoader.load(GltfLoader.java:106)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:439)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:444)
...
Caused by: java.lang.NullPointerException
	at com.jme3.anim.SkinningControl.getAttachmentsNode(SkinningControl.java:403)
	at com.jme3.scene.plugins.gltf.GltfLoader.setupControls(GltfLoader.java:1151)
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:166)
	... 10 more
```

No matter how corrupt the model is, the loader shouldn't throw an NPE!

Forum discussion at https://hub.jmonkeyengine.org/t/how-to-load-gltf-model-correctly/44946

For reasons I don't yet understand, the loader isn't finding anywhere to add the `SkinningControl`. The following workaround allows the model to load, but probably isn't the best solution:
```patch
# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -1148,7 +1148,11 @@
                 continue;
             }
[...]

PR #1639, opened 2021-11-05
solve issue #1612 (NPE while importing from erroneous glTF)

stephengold:
This addresses issue #1612, a `NullPointerException` that occurs while importing a model from an invalid glTF file.

The proposed fix detects that the model's `SkinningControl` isn't added to any `Spatial`. In that case, it logs a warning and adds the control to the model's root node. This allows the import to proceed. However, the resulting model may be incorrect.


stephengold:
Unless there's substantive review, I plan to integrate this in about 24 hours.

PR #1641, opened 2021-11-05
throw better exceptions in TransformTrack

stephengold:
`jme3.anim.TransformTrack` throws generic runtime exceptions in 7 places. This PR changes those exceptions to more specific types, as recommended by Codacy and other static analysis tools.

This PR also checks for null arguments that would otherwise cause NPEs.

It also tries to make the exception message clearer.


Scrappers-glitch:
Yeh, i think this is better, and in case anyone needs to reference the details about the difference between these exceptions : 

image

image

source : Java the complete 11th e, oracle's press.

stephengold:
Please note: the 7 generic runtime exceptions were introduced by @nehon, who is no longer active with JMonkeyEngine: https://github.com/jMonkeyEngine/jmonkeyengine/commit/c3cb4ef97f6c8a8815da57bb9c993cc8f7edfcb6

PR #1642, opened 2021-11-05
improve the TransformTrack javadoc

stephengold:
The javadoc for the `TransformTrack` class is confusing because it implies that the track is a "joint track" and also that the target is a "bone". But the track's target might actually be a `Spatial`, and (in the new animation system) bones are called "joints".

The javadoc is also padded with low-value information such as "a float array" and "of this track".

The javadoc also fails to document some important details:
+ which arguments can be null,
+ which getter returns an internal array, and
+ which setter creates an alias, 

This PR attempts to remedy these defects. Aside from rearranging fields, only comments are affected.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/TransformTrack.java
+++ b/jme3-core/src/main/java/com/jme3/anim/TransformTrack.java
@@ -54,7 +54,7 @@ public class TransformTrack implements AnimTrack<Transform> {
     private HasLocalTransform target;
 
     /**
-     * transforms and times for keyframes
+     * Transforms and times for keyframes.
      */
     private CompactVector3Array translations;
     private CompactQuaternionArray rotations;
@@ -68,7 +68,7 @@ public class TransformTrack implements AnimTrack<Transform> {
     }
 
     /**
-     * Creates a transform track for the given target
+     * Creates a transform track for the given target.
      *
      * @param target       the target Joint or Spatial of the new track
      * @param times        the time for each keyframe, or null for none
@@ -85,7 +85,7 @@ public class TransformTrack implements AnimTrack<Transform> {
     }
 
     /**
-     * Copies the rotations
+     * Copies the rotations.
      *
      * @return a new array, or null if no rotations
      */
@@ -94,7 +94,7 @@ public class TransformTrack implements AnimTrack<Transform> {
     }
 
     /**
-     * Copies the scales
+     * Copies the scales.
      *
      * @return a new array, or null if no scales
      */
@@ -103,7 +103,7 @@ public class TransformTrack implements AnimTrack<Transform> {
     }
 
     /**
-     * Gives access to the keyframe times
+     * Gives access to the keyframe times.
      *
      * @return the pre-existing array
      */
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/anim/TransformTrack.java
```diff
@@ -189,12 +197,16 @@ public void setKeyframesRotation(Quaternion[] rotations) {
 
 
     /**
-     * Set the translations, rotations and scales for this bone track
+     * Set the translations, rotations, and/or scales
```
Scrappers-glitch, line 200: Should we document that : 
1) the arrays translations, rotations, scales, times have to be of the same size. (strict note should be provided for this)
2) the net result is a `Transformation Matrix of (L x 3)` ; where L is the length of the times for clarifying how the api organize and arrange the inputs before processing the animation.

jme3-core/src/main/java/com/jme3/anim/TransformTrack.java
```diff
@@ -290,7 +301,7 @@ public HasLocalTransform getTarget() {
     }
 
     /**
-     * Replace the target of this track, which might be a Joint or a Spatial.
+     * Replace the target, which might be a Joint or a Spatial.
```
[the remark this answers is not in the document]
stephengold: Nodes, geometries, etcetera are subclasses of `Spatial`, so mentioning them in addition to `Spatial` would complicate the explanation without adding any real value.

jme3-core/src/main/java/com/jme3/anim/TransformTrack.java
```diff
@@ -131,9 +136,10 @@ public void setTimes(float[] times) {
     }
 
     /**
-     * Set the translations for this joint track
+     * Set the translations
```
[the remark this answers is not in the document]
stephengold: I'll do that.

jme3-core/src/main/java/com/jme3/anim/TransformTrack.java
stephengold: I agree with 1 but not 2. The result isn't a "matrix" because the rotations are quaternions, not vectors.

Scrappers-glitch: Well, that makes sense, could we correct that to a multidimensional matrix or just say that each row in this L x 3 is a frame that uses a corresponding time...?

Scrappers-glitch: The problem is the first translation, the first rotation, the first scale and the first time should not be discrete, they should be enclosed in an entity called Frame or be documented that they represent a Frame.

stephengold: > could we correct that to a multidimensional array or just say that each row in this L x 3 is a frame
No, that would be misleading, because "array" and "matrix" both imply that all elements are of the same type. They aren't of the same type because the translations are vectors and the rotations are quaternions.

The class javadoc describes the track as "a list of transforms and times for keyframes" and the method javadoc describes each argument as being "for each frame". Short of creating a "Keyframe" class and rewriting the entire module, I don't see how to make this any clearer.

stephengold:
I've updated the javadoc to conform to the latest style proposal at issue 1098.

stephengold:
If there's no further discussion, I plan to integrate this in 24 hours.

PR #1643, opened 2021-11-06
solve issue 1555 (add API to determine max line width)

stephengold:
This PR adds a `getMaxLineWidth()` method to the `Renderer` interface. It also adds a test to jme3-examples.


stephengold:

PR #1650, opened 2021-11-12
SurfaceView Maintenance

Scrappers-glitch:
A maintenance PR for the `JmeSurfaceView`.

## New features introduced : 

1) Optionally  binding the jmeSurfaceView to the android activity life cycle, it helps in `pausing`, `resuming` game and openGL context, and also helps in releasing memory when the game is destroyed.

2) Added new Listeners events : `OnLayoutDrawn()`, `OnRendererStarted()`.

4) Added a `DestructionPolicy` flag, which uses a static game state `GameState` to determine whether to destroy the game with the activity on exiting the activity or to keep it for a second use.
https://youtu.be/NBWyBzr_YXU

5) Optimizations to the `ErrorDialog` includes : showing a full error stack and `Copy crash log` button.

6) Added other flags such as : `showErrorDialog`, `exitOnEscPressed`, `showEscExitPrompt`.

7) Added a getter field for gl_es version.

8) Documentations and code optimizations following best practices.

`NB : changes here are tested more than once on pre-existing jme3 android apps since jme3.4.0, and they won't likely affect user code.`

### Screenshots : 

Copy crash log feature

I will try to make a video to preview these changes on a live android test.

### DIFF
--- a/jme3-android/src/main/java/com/jme3/app/jmeSurfaceView/JmeSurfaceView.java
+++ b/jme3-android/src/main/java/com/jme3/app/jmeSurfaceView/JmeSurfaceView.java
@@ -243,12 +243,14 @@ public class JmeSurfaceView extends RelativeLayout implements SystemListener, Di
 
     /**
      * Instantiates a surface view holder with XML attributes and a default style attribute.
+     * On instantiating this surface view, the holder is bound directly to the
+     * parent context life cycle.
      *
      * @param context      the parent context.
      * @param attrs        a collection of attributes describes the tags in an XML document.
      * @param defStyleAttr an attribute in the current theme that contains a
      *                     reference to a style resource that supplies
-     *                     defaults values.
+     *                     defaults values. Can be 0 to not look for defaults.
      * @see android.content.res.Resources.Theme#obtainStyledAttributes(AttributeSet, int[], int, int)
      */
     public JmeSurfaceView(@NonNull Context context, @Nullable AttributeSet attrs, int defStyleAttr) {
@@ -259,12 +261,15 @@ public class JmeSurfaceView extends RelativeLayout implements SystemListener, Di
 
     /**
      * Instantiates a surface view holder with XML attributes, default style attribute and a default style resource.
+     * On instantiating this surface view, the holder is bound directly to the
+     * parent context life cycle.
      *
      * @param context      the parent context.
      * @param attrs        a collection of attributes describes the tags in an XML document.
-     * @param defStyleAttr an attribute in the current theme that contains defaults.
+     * @param defStyleAttr an attribute in the current theme that contains defaults. Can be 0 to not look for defaults.
      * @param defStyleRes  a resource identifier of a style resource that
      *                     supplies default values, used only if defStyleAttr is 0 or can not be found in the theme.
+     *                     Can be 0 to not look for defaults.
      * @see android.content.res.Resources.Theme#obtainStyledAttributes(AttributeSet, int[], int, int)
      */
     public JmeSurfaceView(Context context, AttributeSet attrs, int defStyleAttr, int defStyleRes) {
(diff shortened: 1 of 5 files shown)


stephengold:
Thanks for your contribution to the Engine!

jme3-android/src/main/java/com/jme3/app/jmeSurfaceView/JmeSurfaceView.java
```diff
@@ -80,209 +91,411 @@
...
+     * Adds the glSurfaceView to the screen immediately, saving the current app instance.
+     */
+    protected void addGlSurfaceView(){
+        /*jme Renderer joins the UIThread at that point*/
+        JmeSurfaceView.this.addView(glSurfaceView);
+        /*dispatch the layout drawn event*/
+        if(onLayoutDrawn != null){
+            onLayoutDrawn.onLayoutDrawn(legacyApplication, this);
+        }
+        /*set the static memory to hold the game state, only if the game life cycle is not bounded to the activity life cycle*/
```
Scrappers-glitch, line 314: Comment should be changed to `Only if the destruction policy uses KEEP_WHEN_FINISHED policy`
In either polices, the user can choose if he/she wants to keep the surface view updated with a activity life cycle or not !

I will correct this.

Scrappers-glitch: This comment is wrongly added before the think of adding a destruction policy to control this.

Scrappers-glitch:
I will review this PR again today to finally assure the changes.

stephengold:
I'd feel more comfortable with this if it were reviewed by someone other than the author. I can't give good feedback because I'm not familiar with Android.

Unless there's substantial review, I intend to integrate this in 24 hours.

Scrappers-glitch:
> I'd feel more comfortable with this if it were reviewed by someone other than the author. I can't give good feedback because I'm not familiar with Android.
> 
> Unless there's substantial review, I intend to integrate this in 24 hours.

You can give it a quick look, if you find something that is odd, point it out and i will try to modify it to be as friendly and good as i can.

stephengold:
Like I said, I can't give this a proper review.

One thing I'm curious about is the addition of `enum GameState` with no values. It looks to me a utility class, but I've never seen a utility class defined as an `enum`. Why did you do it that way?

stephengold:
Also, if this isn't ready to be integrated, please mark it as a "draft" so it won't get integrated prematurely.

Scrappers-glitch:
> Like I said, I can't give this a proper review.
> 
> One thing I'm curious about is the addition of `enum GameState` with no values. It looks to me a utility class, but I've never seen a utility class defined as an `enum`. Why did you do it that way?

A shortcut to the utility class instead of creating a class and an object, if you don't like it or feeling its off, I would convert it to a static subclass.

stephengold:
> A shortcut to the utility class

I've no idea whether this is an acceptable idiom or not.

Scrappers-glitch:
> > A shortcut to the utility class
> 
> I've no idea whether this is an acceptable idiom or not.

I see, but the enum here acts as just a scope, so instead of saying `setLegacyApplication()` directly, we say `GameState.setLegacyApplication()` which keeps our game instance as a static state.

pspeed42:
I've never seen anyone use an enum to hack the singleton pattern.  It's weird and I don't recommend it.

What advantage does this have over a static method on a regular class?  Other than being confusing?

Scrappers-glitch:
> I've never seen anyone use an enum to hack the singleton pattern. It's weird and I don't recommend it.
> 
> What advantage does this have over a static method on a regular class? Other than being confusing?

Alright, I will use a static subclass instead.

Scrappers-glitch:
It's now ready to be merged, i have converted the `enum` to `static subclass` and add the missing constructor and some more clarifications to the constructors, those constructors are default styling constructors.

stephengold:
You said this PR was done, and I integrated it. Why are you still pushing commits to it?

Scrappers-glitch:
> You said this PR was done, and I integrated it. Why are you still pushing commits to it?

I have not pushed anything else ?

stephengold:
You're right about the content of the merged commit. Thanks for pointing that out.

Sorry for the confusion.

PR #1652, opened 2021-11-13
Pie chart blend space

Scrappers-glitch:
An `Enhancement` to the new animation system : 

- Involves controlling the `blendWeight`, using a pie chart area and a pie chart sector.
- Area of pie chart and sector area can be altered during runtime by altering both the radius and the sector angle respectively.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/PieChartSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/PieChartSpace.java
@@ -53,8 +53,10 @@ public class PieChartSpace implements BlendSpace{
     protected float area;
     //pie-chart sector area
     protected float sectorArea;
-    protected int firstActionIndex;
     protected boolean sectorAreaManualAdjustment;
+    //exposing the action indices
+    private int firstActionIndex = 0;
+    private int secondActionIndex = 0;
 
     /**
      * Instantiates a default pie chart space implementation.
@@ -62,7 +64,7 @@ public class PieChartSpace implements BlendSpace{
      * Default angle = 45 degrees.
      * Default area = 0.125 of unit circle area (0.125 * PI).
      */
-    public PieChartSpace(){
+    public PieChartSpace() {
         this(1f, 45f);
     }
 
@@ -73,7 +75,7 @@ public class PieChartSpace implements BlendSpace{
      * @param radius circle radius.
      * @param angle sector angle in degrees.
      */
-    public PieChartSpace(final float radius, final float angle){
+    public PieChartSpace(final float radius, final float angle) {
         //implicit suppression to extrapolation
         //clamp values in the range :  r = [0, 1] and angle = [0, 360].
         this.radius = radius % 1.1f;
@@ -83,17 +85,16 @@ public class PieChartSpace implements BlendSpace{
     @Override
     public void setBlendAction(BlendAction action) {
         this.action = action;
-        //calculate the area of the pieChart
-        area = FastMath.PI * FastMath.pow(radius, 2f);
-        //calculate the sector area
-        sectorArea = (angle / 360f) * area;
+        //use default actions indices
+        action.setFirstActiveIndex(action.getActions().length - 2);
+        action.setSecondActiveIndex(action.getActions().length - 1);
     }
 
     @Override
     public float getWeight() {
-        //keep the values updated with the loop (coherent update) as long as the manual adjustment is false.
+        calculatePieChartTotalArea();
         if(!sectorAreaManualAdjustment) {
-            setBlendAction(action);
+            calculateSectorArea();
         }
         //calculate the unit circle area.
         final float areaOfUnitCircle = FastMath.PI;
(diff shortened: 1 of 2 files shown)


stephengold:
This will need a license and an example app.

Scrappers-glitch:
> This will need a license and an example app.

Example app should be in a different PR or can be on the same PR ?

stephengold:
Same PR would be best, I think.

Scrappers-glitch:
Added a simple testcase for the new piechart blend space.

Scrappers-glitch:
Idk why this happens : 
`
1s
Run gradle/wrapper-validation-action@v1
  with:
    min-wrapper-count: 1
    allow-snapshots: false
  env:
    JAVA_HOME_11.0.13_x64: /Users/runner/hostedtoolcache/jdk/11.0.13/x64
    JAVA_HOME: /Users/runner/hostedtoolcache/jdk/11.0.13/x64
    JAVA_HOME_11_0_13_X64: /Users/runner/hostedtoolcache/jdk/11.0.13/x64
Error: getaddrinfo ENOTFOUND downloads.gradle-dn.com
`

Sailsman63:
The build failure seems to be due to a transient dns glitch in the wrapper-validation-action used by the CI workflow. See :
https://github.com/gradle/wrapper-validation-action/issues/33

The build script is locked to v1, should probably be bumped to 1.0.4 to improve stability against this.

stephengold:
I'll re-run the GitHub jobs in the hope of more successful checks.

I've upgraded wrapper-validation-action in the "master" branch at 58203937828dca4dacfb6560af97d54557eda56c .

Ali-RS:
Can you explain what is the use case for a pie chart blend space?

Scrappers-glitch:
> Can you please explain what is the use case for a pie chart blend space? How it is different from the default blend space in JME? And how many actions we are supposed to blend with this blend space?
> 
> Thank you

Hello @Ali-RS , 

1) BlendSpaces in general : they are mathematical spaces used to adjust the value of the `BlendWeight`, the `blend weight` is the delta value of interpolation, used to serially interpolate transforms (blend). 

2) `PieChartSpace` introduces a different way of adjusting the `blendWeight` based on the `Area of a circle sector`, so multiple  cases to be considered : 
  - Max Blending value => can be achieved when (theta = 360) and radius = 1 (aka unit circle).
  - Blending value between 0 and 1 => can be achieved when (theta = ]0, 360[ ) or (radius = ]0, 1[ ).
  - Min Blending value (approaching zero or zero) => can be achieved by either using (theta = 0) or (radius = 0) or both. 
  - An infinite number of blending values can be achieved when playing with a radius and an angle :-).
  
 ==> Parameters you can change to tweak the effect : 
 1) Radius of the circle (adjusts the final scale factor).
 2) Angle of the sector (adjusts the final unscaled sector area to be used as a blending step).
 3) The Sector area directly.
 
 3) How many actions ? 
 
well, i thought of this question yesterday too, in general `blending` is done between 2 actions only, so passing an arg with more than 2 actions, needs a subsequent check for a progressive time if it reaches the end of the (length1 + length2) and if so then advance the `firstActiveIndex` and `secondActiveIndex`.

Currently, this blend space doesn't handle this, since t is not passed as an argument from `getWeight()`, so its a little bit chaotic about which would run (if more than 2 blendable actions are inserted).

this blend space should work of as many actions as you want, and it would actually blend between each 2 successive actions, 
the current testcase (in this PR) uses only 2 actions (it works fine), but if you try to put more than 2 actions and the system would be confused about which is the firstActiveIndex and which is the secondActiveIndex.

I have looked now at `LinearBlendSpace`, it adjusts the firstActiveIndex and the secondActiveIndex based on the `value`, so they aren't automatically adjusted, they use a user value, i think i can do something similar, but i am trying to make it more like robust emerging from the system itself (to be as "you don't have to check for the time passed and compare it with the length of the actions on every update pass").

 So, its currently 2 only and no handling for 3 or more.

Scrappers-glitch:
I think i will have to make another testcase `TestPieChartSpaceMultipleActions` so we can gain the max efficiency of blend spaces.

Ali-RS:
Thanks @Scrappers-glitch

Scrappers-glitch:
Thanks for your reply, it actually points us to an important part, which is liberating the `indices` of the blendable actions through the blend space, I will do this change, this change would open the way to manually alters the indices in case you want to blend between 3 or more actions  🙂😉.

Scrappers-glitch:
> Maybe adding a test with Jame (the monkey model) walk/run blending action both with liner space and pie chart space playing side by side would be a nice demo.
> 
> You can copy/paste this example and modify it if you like.

I plan to do some special use cases on another PR for the new Animation system including how to build your own blend space, using base actions to fire events on some moment of time in seconds and I will include this interesting testcase of using both linear and piechart  blend spaces on Jaime.

Scrappers-glitch:
@Ali-RS Alright, you could check it now, added blending between more than 2 actions, but the user has to alter the slider on which 2 actions to play (use `setFirstActionIndex(int i)` and `setSecondActionIndex(int j)`).

Scrappers-glitch:
I have nothing else to add on this, please continue reviewing it.

Scrappers-glitch:
As requested, a very representation for this function : 

Piechart

### Parameters that alter the function : 
- Circle radius (will alter circle area).
- Sector angle (will alter blend step).

Both of which will calculate the blendStep which is the user's sector area with respect to the unit circle area.

### For users : 
- One can make the radius constant in his/her app and alter the angle towards `360` which will the alter the blend step.
- Or make the angle constant and alter the radius towards 1 which will also alter the blend step.
- Or change both the angle and radius according to your preferences.

Sailsman63:
Note: as you've written the diagram, you essentially have `πr⁴*(angle)`, due to multiplying user area by sector area.

Scrappers-glitch:
> Note: as you've written the diagram, you essentially have `πr^4*(angle)`, due to multiplying user area by sector area.
> 
> EDIT: for visual clarity, the superscript4 was way smaller than I anticipated

the `pi` is cancelled, the final value is the `blend step` which would be used in linear interpolation.

Sailsman63:
But that still begs the question "why?"

What problem are you trying to solve that has an intuitive model in terms of `r^4 * (theta/360)`, and why implement it in this series of steps with interim results when you can just calculate `r^4 * (theta/360)` in a single line of code with no local/instance variables?

Scrappers-glitch:
> But that still begs the question "why?"
> 
> What problem are you trying to solve that has an intuitive model in terms of `r^4 * (theta/360)`, and why implement it in this series of steps with interim results when you can just calculate `r^4 * (theta/360)` in a single line of code with no local/instance variables?

Do you want it in a single step ? I think being in multiple steps makes you understand better.

It's just a scaled sector area...notice that r is a fraction between zero and one so the `r^4` would scale down the sector area because of applying the power law on a fraction.

PR #1653, opened 2021-11-13
apply preferred style to com.jme3.renderer

stephengold:
This PR illustrates minimal changes to make the `com.jme3.renderer` package comply with the coding style proposed at issue #1098.

Marked as "draft" because the PR includes 2 API changes that should be integrated separately, as discussed at #1098.


jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -1480,13 +1555,17 @@ public Vector3f getScreenCoordinates(Vector3f worldPosition, Vector3f store) {
     }
 
     /**
+     * Determines the display width.
```
Scrappers-glitch, line 1558: Should we change this to `Gets` or `Retrieves`.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -1480,13 +1555,17 @@ public Vector3f getScreenCoordinates(Vector3f worldPosition, Vector3f store) {
...
      * @return the width/resolution of the display.
      */
     public int getWidth() {
         return width;
     }
 
     /**
+     * Determines the display height.
+     *
      * @return the height/resolution of the display.
```
Scrappers-glitch, line 1569: And this too.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
stephengold: The choice between "Gets" and "Determines" isn't addressed in the style guide.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
stephengold: The choice between "Gets" and "Determines" isn't addressed in the style guide.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
Scrappers-glitch: Yes, but determines is a little bit confusing for a getter method, unless this getter is internally changing code.

stephengold: How about "Returns"?

Scrappers-glitch: I think its better than determines.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -485,7 +483,7 @@ public void resize(int width, int height, boolean fixAspect) {
     }
 
     /**
-     * <code>getFrustumBottom</code> returns the value of the bottom frustum
+     * Returns the value of the bottom frustum
```
Sailsman63, line 521: Completely aside from the format/styling concerns, I question the value of this, and other "Trivial" Javadoc comments in this file. It adds no details that cannot be gleaned from the method signature, and therefore does not add value above what the javadoc tool would create on it's own. Why does it exist?

Sailsman63, line 521: Completely aside from the format/styling concerns, I question the value of this, and other "Trivial" Javadoc comments in this file. It adds no details that cannot be gleaned from the method signature, and therefore does not add value above what the javadoc tool would create on it's own. Why does it exist?

stephengold: The style guide says that javadoc is optional for "simple, obvious" methods like getFoo, in cases where there really and truly is nothing else worthwhile to say but "Returns the foo".

stephengold: This PR show minimal changes. Deleting the entire javadoc seemed like a bigger change than deleting one word.

I agree that the current javadoc isn't valuable. Honestly, I haven't studied the frustum math to figure out what "the value of the bottom frustum" really is. Is it the y/z slope, the y-coordinate at the near plane, or the y-coordinate at the far plane? So I believe `getFrustumBottom()` needs *some* javadoc, but I don't yet know enough to write it myself. Writing javadoc for someone else's code can be very frustrating!

Sailsman63: Good point, though in this case the code literally passes through the value of a field.

I hear you on documenting other people's code. In the case of the frustrum stuff here, I'd think that the javadoc should go on the field declarations, as they are `protected` and could, in theory, be overridden by extension code.

stephengold: In theory, yes.

stephengold: Ok, there's a proposed change (to the style guide) that would remove the option of omitting javadoc for "simple, obvious" methods. Discussion of this proposed change is at #1098.

stephengold:
I believe this PRs usefulness as a conversation piece is over. I'll revert the 2 API changes and make another pass over the files.

stephengold:
I plan to integrate this in 24 hours.

Issue #1566, opened 2021-05-25
extraction of the inner class Layer -contained in the AnimComposer- in a separate class, with getter/setter methods for AnimationMask


capdevon:
I open this issue to keep track of the discussion on the forum: 
https://hub.jmonkeyengine.org/t/question-about-new-animation-system-part-4-layers/44643

In short, we talk about transforming the inner class `Layer` - contained in the `AnimComposer` class - into a separate class with getter/setter methods for the `AnimationMask` plus other optimizations. This makes it possible to change the mask at runtime in a graphic editor or in other contexts as needed.

```
/**
 * You could refine the class name in AnimLayer as it is only used in the AnimComposer class
 */
public class Layer implements JmeCloneable {

    private AnimComposer ac;
    private Action currentAction;
    private AnimationMask mask;
    private double time;
    private Object manager;

    public Layer(AnimComposer ac) {
        this.ac = ac;
    }

    /**
     * This function is called only in the controlUpdate method of the AnimComposer class, 
     * so it may become protected ???
     * 
     * @param tpf
     */
    protected void advance(float tpf) {
        time += tpf * currentAction.getSpeed() * ac.getGlobalSpeed();
        // make sure negative time is in [0, length] range
        if (time < 0) {
            double length = currentAction.getLength();
            time = (time % length + length) % length;
        }

    }

    @Override
    public Object jmeClone() {
        try {
            Layer clone = (Layer) super.clone();
            return clone;
        } catch (CloneNotSupportedException ex) {
            throw new AssertionError();
        }
    }

    @Override
    public void cloneFields(Cloner cloner, Object original) {
        ac = cloner.clone(ac);
        currentAction = null;
    }

    public AnimationMask getAnimationMask() {
        return mask;
    }

    public void setAnimationMask(AnimationMask mask) {
        this.mask = mask;
    }

    public Object getManager() {
        return manager;
    }

    public void setManager(Object manager) {
        this.manager = manager;
    }

}
```

Now we need to add a couple of methods to the `AnimComposer` class to access the Layers
```
public Layer getLayer(String name) {
    return layers.get(name);
}

public Collection<Layer> getLayers() {
    return layers.values();
}
```

These 2 methods in the `AnimComposer` class would become superfluous: 
```
    /**
     * Access the manager of the named layer.
     *
     * @param layerName the name of the layer to access
     * @return the current manager (typically an AnimEvent) or null for none
     */
[...]

PR #1656, opened 2021-11-15
convert inner class Layer to a top-level class

stephengold:
This PR consists mainly of refactoring to address issue #1566:
+ All the functionality of the `Layer` class was moved to a new top-level class named `AnimLayer`.
+ Functionality from `AnimComposer.controlUpdate()` was combined with `Layer.advance()` to create a new `AnimLayer.update()` method, resulting in a cleaner interface between the 2 classes.
+ The `AnimLayer` constructor takes 2 arguments instead of one, avoiding the need for an `AnimLayer.setMask()` method.

In addition to `AnimComposer.getLayer()`, some new capabilities were added:
+ `AnimComposer.getLayerName()` to find the name of a specified layer,
+ `AnimComposer.getLayerNames()` to enumerate all layer names, and
+ `AnimLayer.getMask()` to access a layer's mask.

`AnimComposer` is shorter and simpler now, but still more complex than it has to be. At some point we may want to deprecate (and possibly delete) all or some of the redundant methods:
+ `AnimComposer.getCurrentAction()`,
+ `AnimComposer.getLayerManager()`,
+ `AnimComposer.getTime()`,
+ `AnimComposer.setCurrentAction()`,
+ `AnimComposer.setLayerManager()`, and
+ `AnimComposer.setTime()`.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -63,13 +63,13 @@ public class AnimComposer extends AbstractControl {
 
     private Map<String, Action> actions = new HashMap<>();
     private float globalSpeed = 1f;
-    private Map<String, Layer> layers = new LinkedHashMap<>();
+    private Map<String, AnimLayer> layers = new LinkedHashMap<>(4);
 
     /**
      * Instantiate a composer with a single layer, no actions, and no clips.
      */
     public AnimComposer() {
-        layers.put(DEFAULT_LAYER, new Layer(this));
+        layers.put(DEFAULT_LAYER, new AnimLayer(this, null));
     }
 
     /**
@@ -135,14 +135,14 @@ public class AnimComposer extends AbstractControl {
      * @return The action corresponding to the given name.
      */
     public Action setCurrentAction(String actionName, String layerName) {
-        Layer l = layers.get(layerName);
+        AnimLayer l = layers.get(layerName);
         if (l == null) {
             throw new IllegalArgumentException("Unknown layer " + layerName);
         }
         
         Action currentAction = action(actionName);
-        l.time = 0;
-        l.currentAction = currentAction;
+        l.setCurrentAction(currentAction);
+
         return currentAction;
     }
     
@@ -162,12 +162,13 @@ public class AnimComposer extends AbstractControl {
      * @return The action corresponding to the given name.
      */
     public Action getCurrentAction(String layerName) {
-        Layer l = layers.get(layerName);
+        AnimLayer l = layers.get(layerName);
         if (l == null) {
             throw new IllegalArgumentException("Unknown layer " + layerName);
         }
-        
-        return l.currentAction;
+
+        Action result = l.getCurrentAction();
+        return result;
     }
     
     /**
@@ -183,13 +184,12 @@ public class AnimComposer extends AbstractControl {
      * @param layerName The name of the layer we want to remove its action.
      */
     public void removeCurrentAction(String layerName) {
-        Layer l = layers.get(layerName);
+        AnimLayer l = layers.get(layerName);
         if (l == null) {
             throw new IllegalArgumentException("Unknown layer " + layerName);
         }
-        
-        l.time = 0;
-        l.currentAction = null;
+
+        l.setCurrentAction(null);
     }
     
     /**
@@ -208,11 +208,13 @@ public class AnimComposer extends AbstractControl {
      * @return the time (in seconds)
      */
     public double getTime(String layerName) {
-        Layer l = layers.get(layerName);
+        AnimLayer l = layers.get(layerName);
         if (l == null) {
             throw new IllegalArgumentException("Unknown layer " + layerName);
         }
-        return l.time;
+
+        double result = l.getTime();
+        return result;
     }
     
     /**
@@ -231,19 +233,15 @@ public class AnimComposer extends AbstractControl {
      * @param time the desired time (in seconds)
      */
     public void setTime(String layerName, double time) {
-        Layer l = layers.get(layerName);
+        AnimLayer l = layers.get(layerName);
         if (l == null) {
             throw new IllegalArgumentException("Unknown layer " + layerName);
         }
-        if (l.currentAction == null) {
+        if (l.getCurrentAction() == null) {
             throw new RuntimeException("There is no action running in layer " + layerName);
         }
-        double length = l.currentAction.getLength();
-        if (time >= 0) {
-            l.time = time % length;
-        } else {
-            l.time = time % length + length;
-        }
+
+        l.setTime(time);
     }
 
     /**
@@ -324,8 +322,7 @@ public class AnimComposer extends AbstractControl {
      * @param mask the desired mask for the new layer (alias created)
      */
     public void makeLayer(String name, AnimationMask mask) {
-        Layer l = new Layer(this);
-        l.mask = mask;
+        AnimLayer l = new AnimLayer(this, mask);
         layers.put(name, l);
     }
 
@@ -376,9 +373,8 @@ public class AnimComposer extends AbstractControl {
      * Reset all layers to t=0 with no current action.
      */
     public void reset() {
-        for (Layer layer : layers.values()) {
-            layer.currentAction = null;
-            layer.time = 0;
+        for (AnimLayer layer : layers.values()) {
+            layer.setCurrentAction(null);
         }
     }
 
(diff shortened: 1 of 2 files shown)


capdevon:
Hi Steven, well done. Can I suggest some small changes?

stephengold:
go ahead ...

Ali-RS:
Thanks @stephengold

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -0,0 +1,225 @@
...
+
+    /**
+     * Assigns the specified manager. This cancels any manager previously
+     * assigned.
+     *
+     * @param desiredManager the desired manager (such as an AnimEvent, alias
+     *     created) or null for unknown manager
+     */
+    public void setManager(Object desiredManager) {
+        this.manager = desiredManager;
```
capdevon, line 149: Simplify the variable name 'desiredManager' to 'manager'. The javadoc already clarifies the meaning of the parameters very well.

```java
    public void setManager(Object manager) {
        this.manager = manager;
    }
```

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * Changes the animation time, wrapping the specified time to fit the
+     * current Action. An Action be running.
+     *
+     * @param desiredAnimationTime the desired time (in scaled seconds)
+     */
+    public void setTime(double desiredAnimationTime) {
```
capdevon, line 158: Simplify the variable name 'desiredAnimationTime' to 'animTime'. The javadoc already clarifies the meaning of the parameters very well. eg: 
```java
@param animTime the desired time (in scaled seconds)
```

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * Updates the animation time and the current Action during a
+     * controlUpdate().
+     *
+     * @param realSeconds the amount to advance the current Action, in real
+     *     seconds
+     */
+    void update(float realSeconds) {
```
capdevon, line 174: it would be more intuitive to keep the variable name tpf that is passed by the controlUpdate() method.

change parameter name realSeconds -> tpf

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -456,19 +421,59 @@ public void setGlobalSpeed(float globalSpeed) {
...
+        return result;
+    }
+
+    /**
+     * Find the name of the specified layer.
+     *
+     * @param layer the layer to find
+     * @return the layer's name, or null if not found
+     */
+    public String getLayerName(AnimLayer layer) {
```
capdevon, line 457: the behavior of this method is a bit unusual, why don't we add the 'name' attribute to the AnimLayer class instead of managing it with a map?
We could do an optimization: add the 'name' attribute to the AnimLayer class + public getter method, and replace the map with a list in the AnimComposer class. These are all internal changes that do not impact the external use of the classes.

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
stephengold: ok

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
stephengold: ok

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
stephengold: "tpf" assumes the reader knows what those initials stand for, doesn't communicate the units, and assumes that the method will be called once per frame. Doesn't seem more intuitive to me.

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
stephengold: I considered adding the name to `AnimLayer`. I'm willing to do that change. However, I don't think a `List` is a good idea because it's inefficient in the common case, which is looking up the layer for a given name. There's no limit on the number of layers in an `AnimControl`.

capdevon: ok for the map, but we add the name attribute to the AnimLayer class so this method is no longer needed

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
capdevon: I always thought that the controlUpdate() method was invoked once per frame, didn't it? The javadoc already describes all this information very well. We could add: `@param tpf (time per frame) the amount to advance the current Action, in real seconds`

stephengold: It doesn't seem strange to me. The names of formal arguments should make sense in the context where they're seen. In the context of a scene-graph control, "frame" usually means a render cycle. In the context of an animation action, "frame" usually means keyframe, so any mention of "frame" here could cause confusion.

I try not to rely much on javadoc to convey what methods and args do. Many dev tools provide formal argument names without javadoc, so devs often make assumptions about what methods and arguments do based on their names, leading to mistakes like [this one](https://hub.jmonkeyengine.org/t/solved-charactercontrol-update-not-called-automatically/45101).

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
stephengold: yes

stephengold:
If there's nothing more to discuss, I'd like to integrate this in about 24 hours.

capdevon:
Sorry Stephen, I appreciate the work you do and the time you put into it, I didn't mean to annoy you. Ignore my latest proposal and integrate the PR.

stephengold:
I'm not annoyed with you---quite the opposite. I requested review, and you gave it to me. I appreciate the feedback you gave me.

There was one point of disagreement, which is fine. When I saw no reply in 22 hours, I assumed our conversation had gone as far as it could. To keep things moving, I announced my intent to integrate the change.

If you want to resume the conversation, please do so, and I will hold off integration.

capdevon:
ok then I misunderstood. Even for me there are no problems if we have different ideas about something. I enjoy collaborating with you. In the last review I was wondering what you thought of renaming the variable `realSeconds` to `deltaTime`. Just because I'm used to thinking of seconds as integer and not fractional values. If you prefer realSeconds, that's no problem. Close the review and integrate the PR. I'm curious to try it in the next alpha release :)

stephengold:
In `realSeconds`, "real" is meant to indicate that the time interval hasn't yet been multiplied by the animation speedup factor---not to emphasize that it's a floating-point value. Can you suggest a better way to indicate that?

capdevon:
Ok for me, let's go integrate the PR.

pspeed42:
One issue with "seconds" is that it's not really the important part of what it is... just the units.  For example, it's not the "seconds since the animation started" which is what I might have assumed if I didn't already know what JME updates tended to do.

It would be like calling a velocity parameter just "meters" because that's the scale of the units.

I'm really not a fan of tpf (or indeed any of JME's time handling) but at least it is used consistently.  I like having seconds in the name (and seconds never means 'integer' to me in a game engine) but it would be clearer if delta or interval appeared in the parameter name.

To me: deltaSeconds would be clear enough.  And anyone confusing "seconds" for "integer" need only look at the parameter type and then quickly correct there way of thinking that integer seconds would ever be useful in a game engine. ;)  The "realness" was already right there in the parameter type, after all.  We are arguing over units, not precision.

stephengold:
Thanks for the suggestions, Paul. Like you, I'm not 100% satisfied with the current arg name.

The way I see it, `AnimLayer` deals with 2 kinds of time. There's "real" time, which is passed in by `AnimComposer`, and there's "animation" time, which is tracked by the layer's `time` field. Animation time resets periodically, and speedup factors in the `Action` and `AnimComposer` convert between the two. So in my mind, there are "real seconds" and "animation seconds".

It occurs to me that AnimCompser's `tpf` might not reflect real wall-clock time, since it could've been distorted by the `speed` field in `SimpleApplication.update()`. Perhaps I should call it "app time" instead.

What do you think of `appDeltaTimeInSec`?

pspeed42:
Heheh... I thought "real" as opposed to "integer".  So, yeah, "real" is a bad word because also one's perspective on what real means depends on what direction they are coming from.  Is the real animation time or the real world time or the real app time... etc..

I'm ok with lengthier descriptive parameter names.  They are a side-effect of the broken frame time design in jme that uses a single float to represent an amorphous value.  I've always regretted not finding a way to inject a "frame info" object way back in 3.0 alpha development but I was still really new back then.  (With that we could have even supported different kinds of time... app time versus game time, etc.)

So, yeah, more words is probably better.

PR #1658, opened 2021-11-16
jme3-core/src/test:  upgrade MockitoJUnitRunner from deprecated version

stephengold:
While reviewing some deprecation warnings, I discovered that 3 tests in jme3-core were using a deprecated version of `MockitoJUnitRunner`.

See https://stackoverflow.com/questions/41909538/mockitojunitrunner-is-deprecated

This PR corrects the issue.


stephengold:
Unless there's substantive review, I plan to integrate this in about 24 hours.

PR #1659, opened 2021-11-16
Upgraded to LWJGL 3.3.0

tlf30:
@stephengold I have resolved the compilation issue with OpenVR inside jme-vr module, and updated the lwjgl version to 3.3.0.

I apologize, IDEA is trying to trim trailing whitespace, and no matter what I hit, I cannot get it to leave it alone. All other changes are whitespace changes, the only change is on line 233 (and the imports, which again IDEA does on its own...)

I guess this is a step in the right direction for the new style guide 🤣

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -25,7 +25,7 @@ tasks.withType(JavaCompile) { // compile-time options:
 }
 
 ext {
-    lwjgl3Version = '3.2.3' // used in both the jme3-lwjgl3 and jme3-vr build scripts
+    lwjgl3Version = '3.3.0' // used in both the jme3-lwjgl3 and jme3-vr build scripts
 }
 
 repositories {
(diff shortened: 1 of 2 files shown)


stephengold:
Quite right that we don't want to reformat existing files in a non-style PR. Unfortunately, I'm still a raw beginner in IDEA. you could try deleting the ".vscode" folder in the project's root directory. If that doesn't work, search Google or ask someone more experienced, such as Riccardo.

Luckily, GitHub provides an option to ignore blanks in diffs, so it's fairly easy for me to see which changes are substantive.

tlf30:
@stephengold I got it, I had to make the changes by hand in notepad++

stephengold:
Thanks. I'll leave this open for a few days. It's not that I'm expecting any review, but so I can run a few tests before integration.

stephengold:

PR #1660, opened 2021-11-17
Un-deprecate AndroidBufferImageLoader

### LINKED ISSUE #1634: Un-deprecate AndroidBufferImageLoader
Hi

`AndroidBufferImageLoader` which internally uses the Android Bitmap class to load images, supports loading a variety of texture formats (such as WebP images) that are not supported by the `AndroidNativeImageLoader`.

I am not sure why it is deprecated, should we un-deprecate it for v3.5?

Ali-RS:
Fix #1634


stephengold:
Unless there's substantial review, I plan to integrate this in 24 hours.

Scrappers-glitch:
Nice work, a question : Does the system directly picks up AndroidBufferImageLoader when we try to load a webp image.?

Ali-RS:
>Does the system directly pick up AndroidBufferImageLoader when we try to load a webp image.?

It should, with the new entry I added for webp inside Android.cfg.

stephengold:
Thank you for your contributions!

Issue #1661, opened 2021-11-18
JME 3.5.0-alpha1 incompatible with Java 7


stephengold:
The distributed file "jme3-core-3.5.0-alpha1.module" at MavenCentral contains the following:
```json
"attributes": {
        "org.gradle.category": "library",
        "org.gradle.dependency.bundling": "external",
        "org.gradle.jvm.version": 8,
```

Since it specifies JVM version 8, newer build tools won't allow it to be used in projects that specify compatibility with Java 7.

This issue arose from 2 changes to "common.gradle":
+ Jdk 11 compatibility (ad2ba95e993d3ad16f95c189e25c1b2c54763826) which bumped `sourceCompatibility` from '1.7' to '1.8' and
+ Gradle 6.9 (PR #1588) which caused Gradle to publish a .module file to MavenCentral

Returning to 1.7 compatibility will require some changes to our source code, to avoid using Java 8 features such as lambdas.

Ali-RS:
Why should we return to java 1.7 compatibility?

Note even LWJGL 3 requires Java 8 or later to build and run.

stephengold:
The .module file breaks existing Gradle projects that worked with JME 3.4.0, such as Heart.

Ali-RS:
Can we prevent .module file from getting published to maven?

tlf30:
I would say that it is time to drop support for java 7. At this point I would say it is also time to start looking at dropping 8 and moving to 11 as the main lts supported version since 17 is the main lts now. Most projects are only supporting the moat recent two LTS versions. If we continue to support java 7 we will never be able to use the new features in java.

stephengold:
Both ideas (moving to 8 and moving to 11) are worth considering.

I remember hearing we needed Java 7 compatibility to run on Android devices, but that was many years ago. Back in May of 2020, Adi Barda still had a requirement for Java 8 compatibility, but maybe that's passed now.

I don't feel comfortable dropping support for 7 without first raising the idea at the Hub/Forum.

stephengold:
I see you took care of that for me. Thanks!

https://hub.jmonkeyengine.org/t/jme3-minimum-java-version/45111

Ali-RS:
>I remember hearing we needed Java 7 compatibility to run on Android devices

[...]

PR #1662, opened 2021-11-18
address issue #1661 (incompatible with Java 7)

stephengold:
This addresses issue #1661.

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -12,12 +12,6 @@ version = jmeFullVersion
 
 sourceCompatibility = '1.7'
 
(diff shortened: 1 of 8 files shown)


Ali-RS:
Probably `jme3-plugins` should be excluded as well. GltLoader uses java.util.Base64 which is added in Java 8.

stephengold:
This is on hold, waiting for the [Forum/Hub discussion](https://hub.jmonkeyengine.org/t/jme3-minimum-java-version/45111) to settle down.

stephengold:
#1661 has been labelled as "wontfix", so I'm closing this PR.

PR #1664, opened 2021-11-19
jme3-examples:  simplify 6 unnecessary uses of unary minus

stephengold:
Static analysis by IntelliJ IDEA flagged this weird code. This PR makes the examples a tiny bit more readable.


stephengold:
Unless there's substantial review, I plan to integrate this in about 24 hours.

Issue #809, opened 2018-01-16
No gamma correction field in Android JmeHarness


jayfella:
There is no protected field available in com.jme3.app.AndroidHarness to enable/disable gamma correction. This can be fixed by adding a protected field and adding it to the AppSettings initialization code.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-android/src/main/java/com/jme3/app/AndroidHarness.java#L223

Nehon:
Gamma correction won't work on android unless it's using opengl es 3. The setting is not exposed on purpose.
Maybe we should allow it only if opengl version is es 3 but some test have to be done to ensure that it's working exactly the same than on desktop

For information we need those 2 things: 
this extension
https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_framebuffer_sRGB.txt
and srgb image formats
https://www.khronos.org/opengl/wiki/Image_Format#sRGB_colorspace

Scrappers-glitch:
I will investigate this problem & see if i can do something here.....

Markil3:
Have you made any leeway on this, @Scrappers-glitch?

Scrappers-glitch:
> Have you made any leeway on this, @Scrappers-glitch?

i have tried to enable the gamma correction via this snippet : 
```
  ActivityManager am = (ActivityManager) context.getSystemService(Context.ACTIVITY_SERVICE);
        ConfigurationInfo info = am.getDeviceConfigurationInfo();
   // above OGLES3.0, enable sRGB
            if (info.reqGlEsVersion >= 0x30000) {
                 renderer.setMainFrameBufferSrgb(settings.getBoolean("GammaCorrection"));
            }
```
 for OGLES3.0 & up from ```OGLESContext.java``` class, directly using the ```AppSettings#setGammaCorrectionEnabled(true)``` before initializing the renderer here : 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/33a75dddddb8258a35732d8c5a79230e1d3bc0b6/jme3-android/src/main/java/com/jme3/system/android/OGLESContext.java#L214


>Generally saying, to enable gamma correction, you have to enable the sRGB frameBuffer from the renderer class using: 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/33a75dddddb8258a35732d8c5a79230e1d3bc0b6/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java#L3224

[...]

Issue #781, opened 2017-12-02
GammaCorrectionFilter is deprecated in 3.2


stephengold:
I'm unhappy that GammaCorrectionFilter is deprecated in JME 3.2.0-alpha1 . SkyControl uses it for an off-screen render, where AppSettings isn't a viable solution. I could write my own filter, but that seems like a waste, given that the Engine already provides that functionality.

empirephoenix:
What is the reason that it was deprecated? Eg. what does the commit say?

stephengold:
https://github.com/jMonkeyEngine/jmonkeyengine/commit/f7181c38dc1720d3df491b7ccac048ffc59817d6#diff-6d6b25ca99b0a3b73d214a0a675b890f

```
Removed all xxx15.frag shaders and used GLSLCompat to have the same shader whatever version it's used for
```

Nehon:
The reason is that it's just wrong to handle gamma correction like this. There is a lot more to it than just correcting the last image of the filter stack.
Also it should not be needed anymore, you can set gamma correction on in the app settings, and it will handle it properly even with post process.
Steve has a corner case where he has a custom offscreen render, we need to discuss this as there are a couple of things to know to do it properly.

stephengold:
@pspeed42  pointed out that what I want is actually a contrast filter. From digital photography, I'm accustomed to adjusting contrast using a gamma function. For reference:

https://en.wikipedia.org/wiki/Image_editing#Gamma_correction

stephengold:
In addition to the deprecation warning at compile time, I get an AssetNotFoundException at runtime (in JME 3.2.0-alpha1):
```
Dec 03, 2017 5:30:55 PM jme3utilities.sky.test.TestGlobeRenderer simpleInitApp
INFO: jme3-core version is "jMonkeyEngine 3.2-alpha1"
Dec 03, 2017 5:30:55 PM jme3utilities.sky.test.TestGlobeRenderer simpleInitApp
INFO: SkyControl version is "master 0.9.6+1"
Dec 03, 2017 5:30:55 PM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
com.jme3.asset.AssetNotFoundException: Common/MatDefs/Post/GammaCorrection.j3md
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:370)
[...]

PR #1665, opened 2021-11-19
ColorAdjustmentFilter

Scrappers-glitch:
A color filter used to adjust the color contrast and brightness based on a simple tone map operator function and adjust scales on the final result at the final pass before rendering.

Supports and tested on both OGL and OGLES (android).

Discussion : https://hub.jmonkeyengine.org/t/gammacorrectionfilter/45076/33


Scrappers-glitch:
Targeting both : #781 #809

stephengold:
Thanks for working to resolve these issues.

stephengold:
Let us know when this is ready for review.

Scrappers-glitch:
> Let us know when this is ready for review.

I have changed the default exp value to 2.2, and documenting some default values.

It's now ready to be reviewed.

jme3-effects/src/main/resources/Common/MatDefs/Post/ColorContrast.frag
```diff
@@ -30,61 +30,56 @@
...
-    color.r = (color.r - minBrightness) / (maxBrightness - minBrightness);
-    color.g = (color.g - minBrightness) / (maxBrightness - minBrightness);
-    color.b = (color.b - minBrightness) / (maxBrightness - minBrightness);
-
-
-    color.r = max(color.r, 0.0);
-    color.g = max(color.g, 0.0);
-    color.b = max(color.b, 0.0);
-
+    color.rgb = (color.rgb - vec3(m_minBrightness)) / (vec3(m_maxBrightness) - vec3(m_minBrightness));
```
Scrappers-glitch, line 69: The brightness value can exceed 1.0 by setting the minBrightness to 0 and maxBrightness to less than 1.

jme3-effects/src/main/resources/Common/MatDefs/Post/ColorContrast.frag
```diff
@@ -30,61 +30,56 @@
...
 
     //2) apply transfer functions on different channels.
-    color.r = pow(color.r, exp_r);
-    color.g = pow(color.g, exp_g);
-    color.b = pow(color.b, exp_b);
+    color.r = pow(color.r, m_exp_r);
+    color.g = pow(color.g, m_exp_g);
+    color.b = pow(color.b, m_exp_b);
 
+    //3) apply a final scale factor, between 0.0 and 1.0.
```
Scrappers-glitch, line 78: The color channels scales control 'how much of our changes would be applied'.

Ali-RS:
I did a quick look through and noticed fields `scale_r`,`exp_r`,... and getters `getScale_r`, `getExp_r`,... I think to follow JME (and generally Java conventions) it isbetter to name them like rScale, rExp,... and `getRScale`, `getRExp`,.... that's how rest of the JME code is following.

Scrappers-glitch:
Alright, I will change them, please continue reviewing.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+ * of the scene textures based on a simple transfer function.
+ *
+ * The transfer function involves setting the brightness of colors based on a maxValue and a minValue,
+ * then setting the contrast of each color channel independently using a power law function before the textures' colors are processed by the
+ * rasterizer and the final result can be scaled again on runtime using scale factors from 0.0 to 1.0.
+ *
+ * @author pavl_g.
+ */
+public class ContrastAdjustmentFilter extends Filter {
+    //the different channels exponents
```
stephengold, line 58: "channels" here is a possessive plural, so it needs an apostrophe:  channels'

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+    protected float exp_r = 2.2f;
+    protected float exp_g = 2.2f;
+    protected float exp_b = 2.2f;
+    //the minimum value and the maximum value of the brightness parameter
```
stephengold, line 62: There's no brightness parameter. The min is a parameter and the max is a parameter.

Also, the comments should clarify that these control the min/max *input* values, not the min/max *output* values.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+    //the final pass scale factor
+    protected float scale_r = 1f;
+    protected float scale_b = 1f;
+    protected float scale_g = 1f;
+    protected Material material;
+
+    /**
+     * Instantiates a default color contrast filter, default brightness and default scale.
+     * Default values :
+     * - Exponents = 2.2 on all channels.
```
stephengold, line 74: 2.2 is a strange choice for a default. I think the default exponents should all be 1.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+import com.jme3.math.Vector2f;
+import com.jme3.math.Vector3f;
+import com.jme3.post.Filter;
+import com.jme3.renderer.RenderManager;
+import com.jme3.renderer.ViewPort;
+import java.io.IOException;
+
+/**
+ * A color filter used to change the contrast for each color channel independently and the brightness of colors
+ * of the scene textures based on a simple transfer function.
```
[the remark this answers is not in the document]
stephengold: You might also note that the alpha channel passes through the filter unchanged.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+    protected float scale_r = 1f;
+    protected float scale_b = 1f;
+    protected float scale_g = 1f;
+    protected Material material;
+
+    /**
+     * Instantiates a default color contrast filter, default brightness and default scale.
+     * Default values :
+     * - Exponents = 2.2 on all channels.
+     * - Brightness = 1.0f (normal).
```
stephengold, line 75: You should document the default values for both the min and max brightness.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+     * - Scale = 1.0f.
+     */
+    public ContrastAdjustmentFilter() {
+        super(ContrastAdjustmentFilter.class.getName());
+    }
+
+    /**
+     * Instantiates a color contrast filter with a default brightness and default scale.
+     * Default values :
+     * - Brightness = 1.0f (normal).
```
stephengold, line 85: Document both max brightness and min brightness.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+ *
+ * The transfer function involves setting the brightness of colors based on a maxValue and a minValue,
+ * then setting the contrast of each color channel independently using a power law function before the textures' colors are processed by the
+ * rasterizer and the final result can be scaled again on runtime using scale factors from 0.0 to 1.0.
+ *
+ * @author pavl_g.
+ */
+public class ContrastAdjustmentFilter extends Filter {
+    //the different channels exponents
+    protected float exp_r = 2.2f;
```
stephengold, line 59: Please change all Java field names to camelCase without underscores.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+    protected float exp_g = 2.2f;
+    protected float exp_b = 2.2f;
+    //the minimum value and the maximum value of the brightness parameter
+    protected float minBrightness = 0f;
+    protected float maxBrightness = 1f;
+    //the final pass scale factor
+    protected float scale_r = 1f;
```
stephengold, line 66: Please change all java field names to camelCase without underscores.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+     * Instantiates a color contrast filter with a default brightness and default scale.
+     * Default values :
+     * - Brightness = 1.0f (normal).
+     * - Scale = 1.0f.
+     *
+     * @param exp_r the red color exponent.
+     * @param exp_b the blue color exponent.
+     * @param exp_g the green color exponent.
+     */
+    public ContrastAdjustmentFilter(float exp_r, float exp_b, float exp_g) {
```
stephengold, line 92: Please change argument names to camelCase without underscores.

stephengold: Also, I think typical use cases would use the same exponents for each color channel. Perhaps we should provide a single-argument constructor for that case.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+        super(ContrastAdjustmentFilter.class.getName());
+        this.exp_r = exp_r;
+        this.exp_g = exp_g;
+        this.exp_b = exp_b;
+    }
+
+    /**
+     * Instantiates a color contrast filter by adjusting different parameters.
```
stephengold, line 100: "by adjusting different parameters" is so vague that it adds nothing. Delete it or else provide more details.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+     *
+     * @param exponents  the exponents applied to the colors in order, where x = r, y = g, z = b.
+     * @param brightness the brightness applied to the textures in order, where x = minBrightness, y = maxBrightness.
```
stephengold, line 103: `brightness` is not applied to textures, but to a rendered scene.

"in order" adds no useful information, since max and min do not have a natural ordering.

Also, should clarify that `brightness` applies to the input levels, not output levels.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+        super(ContrastAdjustmentFilter.class.getName());
+        this.exp_r = exp_r;
+        this.exp_g = exp_g;
+        this.exp_b = exp_b;
+    }
+
+    /**
+     * Instantiates a color contrast filter by adjusting different parameters.
+     *
+     * @param exponents  the exponents applied to the colors in order, where x = r, y = g, z = b.
```
stephengold, line 102: I think `exponents` and `scales` should be ColorRGBA, not Vector3f. That way you wouldn't have to document which component corresponds to each color channel.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,309 @@
...
+     * Retrieves the red channel exponent.
+     * Default value = 2.2.
+     * @return the red channel exponent.
+     */
+    public float getExp_r() {
+        return exp_r;
+    }
+
+    /**
+     * Sets the color channels brightness using a maxValue and a minValue based on this equation :
```
stephengold, line 168: This is misleading because it doesn't set the color channels' brightness. It sets the range of input levels.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
stephengold: "Brightness" refers to the color as a whole. When discussing a single color channel, the term "level" is commonly used.

stephengold (changes requested):
This is a partial review. Many of my comments are about issues that arise throughout ContrastAdjustmentFilter.java: the choice of defaults, the use of underscores in variable names, the terminology used to discuss colors, etcetera. Please apply my requested changes throughout the PR. I'd like to review this PR again after my requested changes are made.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
Scrappers-glitch: That would leave the `default` constructor with no use.

If i am going to make it 1.0, i will remove the default constructor (the first constructor on this class)...so what do you think ?

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
Scrappers-glitch: That would make me add an exponent and a scale to the alpha channel...if there are no issues from adjusting the alpha channel then i will add it.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
Scrappers-glitch: > You might also note that the alpha channel passes through the filter unchanged.

Using `ColorRGBA` instead of `Vector3f` for both the contrast and scale would force me to add an alpha channel to the play.

I don't mind adding an alpha channel. Do you mind adding it ?

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
Scrappers-glitch: If there are issues, then we better use the alpha value as a scale value to the color channel before being passed to the fragment shader ie: scale the channels on java on a scale from 0 to 1.

Scrappers-glitch:
@stephengold @Ali-RS Hello there, i have made the requested changes, i hope everything is clear now, please review it again.
Thanks.

jme3-effects/src/main/resources/Common/MatDefs/Post/ColorContrast15.frag
```diff
@@ -0,0 +1,89 @@
...
+    // limit the chromaticity space into the +ve quadrant
+    color.rgb = abs(color.rgb);
+
+    //2) apply transfer functions on different channels.
+    color.r = pow(color.r, m_redChannelExponent);
+    color.g = pow(color.g, m_greenChannelExponent);
+    color.b = pow(color.b, m_blueChannelExponent);
+
+    //3) apply a final scale factor, between 0.0 and 1.0.
+    color.r = color.r * min(max(m_redChannelScale, 0.0), 1.0);
```
stephengold, line 83: Invoking min() and max() in the shader is inefficient. It would be better for Java to ensure the scales passed to the shader are all in the range (0, 1).

jme3-examples/src/main/java/jme3test/post/TestContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,101 @@
...
+    public void simpleUpdate(float tpf) {
+        counter += tpf;
+        if (counter >= 4f) {
+            counter = 0f;
+        }
+        if (contrastAdjustmentFilter != null) {
+            //adjust the transfer function during runtime
+            contrastAdjustmentFilter.setScales(counter * 0.5f, counter, counter * 0.5f)
+                                    .setExponents(counter, counter * 2f, counter)
+                                    .setInputRange(0, 1 / counter);
```
stephengold, line 98: What happens here when `counter` resets to 0? Looks to me like the input upper limit goes to `NaN`!

jme3-examples/src/main/java/jme3test/post/TestContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,101 @@
...
+
+        //add light
+        final AmbientLight ambientLight = new AmbientLight(ColorRGBA.White);
+        rootNode.addLight(ambientLight);
+
+        //setup the filter
+        final FilterPostProcessor postProcessor = new FilterPostProcessor(assetManager);
+        contrastAdjustmentFilter = new ContrastAdjustmentFilter();
+        //adjusting some parameters
+        contrastAdjustmentFilter.setExponents(1.8f, 1.8f, 2.1f)
```
stephengold, line 80: Why bother invoking the setters in `simpleInit()` when they're going to be immediately overridden by `simpleUpdate()`?

stephengold (changes requested):

jme3-examples/src/main/java/jme3test/post/TestContrastAdjustmentFilter.java
Scrappers-glitch: Good point, I will remove this for the sake of simplicity.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+ * A color filter used to adjust the contrast for color channels of the rendered scene by using a
+ * simple transfer function which involves adjusting the input range of the color channels based on
+ * an upper and a lower limit, adjusting the exponent of different color channels and scaling the output values
+ * before being processed by the fragment shader.
+ *
+ * @author pavl_g.
+ */
+public class ContrastAdjustmentFilter extends Filter {
+
+    protected float redChannelExponent;
```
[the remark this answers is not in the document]
Scrappers-glitch: Yes I thought of that too, but telling others that this is a redChannel exponent adds nothing in fact.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+    public float getBlueChannelExponent() {
+        return blueChannelExponent;
+    }
+
+    /**
+     * Sets the color channels input range using a lowerLimit and an upperLimit based on this equation :
+     * color.rgb = (color.rgb - lowerLimit) / (upperLimit - lowerLimit)
+     * where; increasing the lowerLimit value and increasing the difference between upperLimit and lowerLimit would decrease the input range,
+     * while decreasing the lowerLimit value and increasing the difference between upperLimit and lowerLimit would increase the input range.
+     * The final input range value is always above 0.0.
```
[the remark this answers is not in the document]
Scrappers-glitch: The function output is an absolute value.

jme3-examples/src/main/java/jme3test/post/TestContrastAdjustmentFilter.java
Scrappers-glitch: Yep, I will limit that to 1.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+        setInputRange(lowerLimit, upperLimit);
+
+        //final pass scales
+        setScales(redChannelScale, greenChannelScale, blueChannelScale);
+    }
+
+    @Override
+    protected Material getMaterial() {
+        if (material == null) {
+            throw new IllegalStateException("Cannot create a color filter from a null reference !");
```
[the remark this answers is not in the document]
Scrappers-glitch: Well I don't know but I want to be safe.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+     * @param lowerLimit the lower value of the color channels input range, default is 0f.
+     * @param upperLimit the higher value of the color channels input range, default is 1f.
+     * @return this filter instance for a chain call.
+     */
+    public ContrastAdjustmentFilter setInputRange(float lowerLimit, float upperLimit) {
+        this.lowerLimit = lowerLimit;
+        this.upperLimit = upperLimit;
+
+        if (material == null) {
+            return this;
```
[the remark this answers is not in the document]
Scrappers-glitch: Although I feel the code is very clear, like 'if the material is not initialized' then skip setting the material data until we initialize it via the FilterProcessor, but I will try to simplify it more.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
Scrappers-glitch: Things like calling this factory method from outside this class would make the code fail (ofc almost no one may do this).

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+     * Default value = 1.0f
+     *
+     * @return the blue channel exponent.
+     */
+    public float getBlueChannelExponent() {
+        return blueChannelExponent;
+    }
+
+    /**
+     * Sets the color channels input range using a lowerLimit and an upperLimit based on this equation :
```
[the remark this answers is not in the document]
Scrappers-glitch: Sorry, I don't understand here, what should I do about this ?

jme3-effects/src/main/resources/Common/MatDefs/Post/ColorContrast15.frag
Scrappers-glitch: Although i see no big difference, but I will move the code to the java side for clean purposes.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+import com.jme3.post.Filter;
+import com.jme3.renderer.RenderManager;
+import com.jme3.renderer.ViewPort;
+import java.io.IOException;
+
+/**
+ * A color filter used to adjust the contrast for color channels of the rendered scene by using a
+ * simple transfer function which involves adjusting the input range of the color channels based on
+ * an upper and a lower limit, adjusting the exponent of different color channels and scaling the output values
+ * before being processed by the fragment shader.
```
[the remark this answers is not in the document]
Scrappers-glitch: I wonder what's the exact life cycle of jme filters, because every time I misunderstand something about this, I searched the wiki but there weren't satisfying answers.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+     */
+    public float getBlueChannelScale() {
+        return blueChannelScale;
+    }
+
+    @Override
+    protected void initFilter(AssetManager manager, RenderManager renderManager, ViewPort vp, int w, int h) {
+        //validate app
+        if (manager == null || renderManager == null || vp == null || w == 0 || h == 0) {
+            return;
```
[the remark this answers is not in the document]
capdevon: These checks are not necessary. Looking at the `FilterPostProcessor` class, the `assetManager` and `renderManager` variables are never null and width and length are always greater than 0.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
capdevon: in this method the '`material`' should always be different from null because it should be initialized in the `initFilter() `method as suggested by the javadoc: 
```java
Initialization of sub classes filtersThis method is called once when the filter is added to the FilterPostProcessorIt should contain Material initializations and extra passes initialization
Overrides: initFilter(...) in Filter
Parameters:
- manager the assetManager
- renderManager the renderManager
- vp the viewPort where this filter is rendered
- w the width of the filter
- h the height of the filter
```

None of the current filters in the com.jme3.post.filters package raise exceptions in this method.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
stephengold: I think you're asking about the rendering pipeline, not the life cycle. I'm no expert on rendering, so I'm not the best person to instruct you.

Here's something I do know: when writing documentation, it's dangerous to use words you don't understand. Misleading documentation is worse than incomplete documentation.

The documentation for `ContrastAdjustmentFilter` should focus on what makes it different from other JME filters. Since all JME filters are used to post-process rendered scenes, that fact doesn't need to be documented here.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
stephengold: The null check should be removed.

jme3-effects/src/main/resources/Common/MatDefs/Post/ColorContrast15.frag
stephengold: The shader gets executed on every frame; `setScales()` typically gets executed *much* less often. That's the difference.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
Scrappers-glitch: I will remove all the null checks.
@capdevon @stephengold Thanks for the review, keep reviewing other parts.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Instantiates a default color contrast filter, default input range and default scale.
+     * Default values :
+     * - Exponents = 1.0f on all channels.
+     * - Input Range Lower Limit = 0f.
+     * - Input Range Upper Limit = 1f.
+     * - Scale = 1.0f on all channels.
+     */
+    public ContrastAdjustmentFilter() {
+        this(1f);
```
capdevon, line 76: The default constructor should set the filter `name`. Use the filters that already exist as a reference. 
Can you change it like this?

```java
public ContrastAdjustmentFilter() {
    super("ContrastAdjustmentFilter");
}
```

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+     * Instantiates a color contrast filter with a specific exponent, default scale and default input range.
+     *
+     * @param exponent an exponent to apply on all channels.
+     */
+    public ContrastAdjustmentFilter(float exponent) {
+        setExponents(exponent, exponent, exponent);
```
capdevon, line 85: You shouldn't call public methods in the constructor because methods can be overridden.
Use the filters that already exist as a reference.
Can you change it like this?

```java
public ContrastAdjustmentFilter(float exponent) {
    this();
    this.redChannelExponent = exponent;
    this.greenChannelExponent = exponent;
    this.blueChannelExponent = exponent;
}
```

`this();` is used to invoke the default constructor that sets the name of the filter.

capdevon:
All current filters have getter / setter methods for all variables. Could you add the setters methods for the 8 class variables to make it easier to build a test gui (with both Lemure and NetBeans)?


See this example using the Lemur library to read and modify filter values:

```java
ContrastAdjustmentFilter filter = new ContrastAdjustmentFilter();
fpp.addFilter(filter);
viewPort.addProcessor(fpp);

Container container = new Container();
container.addChild(new Label("ContrastAdjustmentFilter", new ElementId("title")));

PropertyPanel panel = container.addChild(new PropertyPanel("glass"));
panel.addFloatProperty("redExp", filter, "redChannelExponent", 0f, 255f, 1f); 		//use getRedChannelExponent/setRedChannelExponent
panel.addFloatProperty("greenExp", filter, "greenChannelExponent", 0f, 255f, 1f); 	//use getGreenChannelExponent/setGreenChannelExponent
panel.addFloatProperty("blueExp", filter, "blueChannelExponent", 0f, 255f, 1f); 	//use getBlueChannelExponent/setBlueChannelExponent

panel.addFloatProperty("lowerLimit", filter, "inputRangeLowerLimit", 0f, 1f, 0.1f); //use getInputRangeLowerLimit/setInputRangeLowerLimit
panel.addFloatProperty("upperLimit", filter, "inputRangeUpperLimit", 0f, 1f, 0.1f); //use getInputRangeUpperLimit/setInputRangeUpperLimit

panel.addFloatProperty("redScale", filter, "redChannelScale", 0f, 1f, 0.1f); 		//use getRedChannelScale/setRedChannelScale
panel.addFloatProperty("greenScale", filter, "greenChannelScale", 0f, 1f, 0.1f); 	//use getGreenChannelScale/setGreenChannelScale
panel.addFloatProperty("blueScale", filter, "blueChannelScale", 0f, 1f, 0.1f); 		//use getBlueChannelScale/setBlueChannelScale

panel.addBooleanProperty("enabled", filter, "enabled");

guiNode.attachChild(container);
```

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
Scrappers-glitch: Good points, thanks.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
Scrappers-glitch: I completely agree, calling an overridable method in a constructor is not right, I will apply these changes.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
stephengold: Here's a suggestion:
```java
/**
     * Sets the input range for all color channels.
     * 
     * <p>An input channel at the lower limit yields an output level of 0.
     * 
     * <p>An input channel at the upper limit yields an output level of channel scale.
     *
     * @param lowerLimit the desired upper limit level, default is 0.
     * @param upperLimit the desired upper limit level, default is 1.
     * @return this filter instance for chaining.
     */
```

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
stephengold: Absolute value refers to the function returning `(x < 0) ? -x : x`.
I don't see an absolute value here.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
Scrappers-glitch: Thank you !

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
Scrappers-glitch: The abs() is done on the shader side, I do it to the java side with the max and min().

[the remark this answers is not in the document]
stephengold: The `abs()` in the shader affects the input levels, not the input range value. Please correct the javadoc.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,296 @@
...
+    protected float greenChannelExponent;
+    protected float blueChannelExponent;
+    //the lower value and the upper value of the inputRange
+    protected float lowerLimit;
+    protected float upperLimit;
+    //the final pass scale factor
+    protected float redChannelScale;
+    protected float greenChannelScale;
+    protected float blueChannelScale;
+    protected Material material;
```
capdevon, line 65: The variable `material` already exists in the `Filter` super class and is 'protected' therefore visible in the subclasses. Please remove this duplicate.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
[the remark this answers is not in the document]
Scrappers-glitch: I will move it to java instead, then I will see if the docs will need to be corrected based on the new code.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
Scrappers-glitch: I will make it private.

Scrappers-glitch: I will, thank you !

stephengold:
@Scrappers-glitch When you find the time, please commit the changes you agreed to make.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
stephengold: Recent commits address these issues. Are they resolved?

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
stephengold: I've delete the field. Conversation resolved?

capdevon: yes ;)

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
capdevon: yes

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,402 @@
...
+        return this;
+    }
+
+    /**
+     * Sets the power-law exponent for the red channel.
+     *
+     * @param exponent the desired exponent (default=1)
+     * @return this filter instance, for chaining
+     */
+    public ContrastAdjustmentFilter setRedExponent(float exponent) {
```
capdevon, line 119: Could you please standardize the getter and setter methods of the variables? for example `getRedChannelExponent` and `setRedExponent` -> could you change to get/set RedExponent, get/set GreenExponent, get/set BlueExponent and so on for the other class attributes?

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,402 @@
...
+        return this;
+    }
+
+    /**
+     * Retrieves the value of the red channel scale that's applied on the final pass.
+     * Default value = 1.0.
+     *
+     * @return the scale of the red channel.
+     */
+    public float getRedChannelScale() {
```
capdevon, line 313: Could you please standardize the getter and setter methods of the variables? for example `getRedChannelScale` and `setRedScale` -> could you change to get/set RedScale, get/set GreenScale, get/set BlueScale ?

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,402 @@
...
+        return this;
+    }
+
+    /**
+     * Returns the lower limit of the input range.
+     * Default value = 0.0.
+     *
+     * @return the lower limit
+     */
+    public float getInputRangeLowerLimit() {
```
capdevon, line 233: Could you please standardize the getter and setter methods of the variables? could you change to getLowerLimit ?

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
```diff
@@ -0,0 +1,402 @@
...
+        return lowerLimit;
+    }
+
+    /**
+     * Returns the upper limit of the input range.
+     * Default value = 1.0.
+     *
+     * @return the upper limit
+     */
+    public float getInputRangeUpperLimit() {
```
capdevon, line 243: Could you please standardize the getter and setter methods of the variables? could you change to getUpperLimit ?

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
stephengold: Done.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
stephengold: Done.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
stephengold: Done.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
stephengold: This conversation should be resolved now.

jme3-effects/src/main/java/com/jme3/post/filters/ContrastAdjustmentFilter.java
capdevon: Well done. I can't resolve the conversation because I didn't open the PR and I don't have write access to the repository on which the PR was opened.

stephengold (approved):

stephengold:

stephengold:
Additional work will be required to resolve #809.

Scrappers-glitch:
> Additional work will be required to resolve #809.

Like what ? Predefining the filter in the harness activity if the srgbFrame is true ?

stephengold:
Something like that, yes. #809 specifically refers to the harness.

Issue #1597, opened 2021-08-21
Missing item in the BinaryExporter documentation


fabrice-ducos:
The [documentation](https://javadoc.jmonkeyengine.org/v3.3.0-beta1/com/jme3/export/binary/BinaryExporter.html) of the format descriptor refers to a 19th item.

However the format descriptor only contains 18 items.

stephengold:
Thanks for informing us of the issue. I see what you mean.

Someone needs to read the code carefully to determine what it actually does.

PR #1666, opened 2021-11-21
clarify and update the J3O format description

stephengold:
While investigating issue #1597, I uncovered inaccuracies in the definition of JMonkeyEngine's native binary format (J3O) presented in the class javadoc for `BinaryExporter`. Perhaps the format was modified at some point without updating the javadoc.

This PR corrects the inaccuracies. It also reformats and simplifies the class javadoc. It also adds comments to the `save()` method to match format items in the documentation with `write()` invocations.


stephengold:
Unless there's substantial feedback, I plan to integrate this PR in about 24 hours.

PR #1668, opened 2021-11-22
set initial window position

bob0bob:
Added the ability under ljwgl3  to be able to position the window instead of forcing it to be center on the screen all the time.   You can set through AppSettings 3 types of values.   CenterWindow, WindowXPosition, WindowYPosition.  By defaul the CenterWindow is set to 'true', so it runs just like it use to do and doesn't affect any previous builds.

  This way the positioning will can be saved and used in the future.


stephengold:
The "v3.4" branch is not open for new features. New features should be directed at the "master" branch.

bob0bob:
The pull request handles this.

Issue #1672, opened 2021-11-23
Logging statement should be guarded


vxel:
In the Node class, two logging statements are not guarded and show a performance impact due to the `this.toString()`.

Line 434 :
`logger.log(Level.FINE, "{0}: Child removed.", this.toString());`

Line 464 : 
`logger.log(Level.FINE, "{0}: All children removed.", this.toString());`

Those statements should be guarded with : 
`if (logger.isLoggable(Level.FINE)) { ... }`

As shown in the profile summary below, this logging is responsible for almost 100% of the CPU used by the method :
![image](https://user-images.githubusercontent.com/28866693/143059036-0845698f-6f02-4f0e-b63f-36d91000d1d9.png)

pspeed42:
:facepalm: The code goes through the trouble of calling the format version but then calls toString() anyway.

Because:
```
logger.log(Level.FINE, "{0}: Child removed.", this);
```
...would have been fine, too.

tonihele:
Oh dear :D Good catch @vxel, planning to make a PR for this? I think it still can make 3.5 release, although that is not for me to decide.

vxel:
Yeah I can prepare a PR for this. The fix is trivial ;-)

PR #1673, opened 2021-11-23
fix issue 1672 (inefficient logging in Node.java)

vxel:
This fixes issue #1672

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Node.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Node.java
@@ -431,7 +431,9 @@ public class Node extends Spatial {
         Spatial child = children.remove(index);
         if (child != null) {
             child.setParent(null);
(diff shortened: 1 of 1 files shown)


stephengold:
Looks good to me! Unless there's substantial review, I will integrated this in about 24 hours.

pspeed42:
I kind of wish we'd just changed the "this.toString()" into just "this" instead.  That's a garbage anti-idiom that would have prevented the whole issue if done properly in the first place.

stephengold:
I like pspeed's solution better than adding guard code. @vxel would you be willing to implement that solution instead?

vxel:
Agree with that too. PR updated.

PS: sorry for the force-push I merged wrong commits...

stephengold:
Force pushes are okay in this case because "fix/1672" is your own private branch. But unnecessary because I'm going to squash the history when I integrate this PR.

stephengold:
Thanks for your contribution, @vxel !

PR #1675, opened 2021-11-23
document the correct build procedure

stephengold:
The instructions in “CONTRIBUTING.md” say to “run ‘gradle build’”, which only works if you’ve installed particular versions of Gradle. Those instructions date back to 2014.

No instructions are ever foolproof. However, this PR tries to improve the odds of success for new users.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,10 +33,36 @@ Developers in the Contributors team can push directly to Main instead of submitt
 - In general, library changes that plausibly might break existing apps appear only in major releases, not minor ones.
 
 
-## Building the engine
+## How to build the Engine from source
 
-1. Install [Gradle](http://www.gradle.org/)
-2. Navigate to the project directory and run 'gradle build' from command line to build the engine.
+### Prerequisites
+
+These steps need only be done once...
+
+1. Install a Java Development Kit (JDK), if you don't already have one.
+2. Set the `JAVA_HOME` environment variable:
+  + using Bash: `export JAVA_HOME="` *path to your JDK* `"`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -33,10 +33,36 @@ Developers in the Contributors team can push directly to Main instead of submitt
...
 
-## Building the engine
+## How to build the Engine from source
+
+### Prerequisites
+
+These steps need only be done once...
+
+1. Install a Java Development Kit (JDK), if you don't already have one.
+2. Set the `JAVA_HOME` environment variable:
```
Sailsman63, line 43: Is this required? I've never set it on linux, and it is not set by any of the java management tooling for my distro.

Basically, setting this allows the user to override the version of the JDK that is on the system path, ie, the version you get if you run `java` or `javac` directly from the command line.

stephengold: In some cases, installing a JDK doesn't add it to the system path. For instance, [the official installation instructions for OpenJDK](https://openjdk.java.net/install/) say to extract the archive, but nothing about adding to the system path.

Sailsman63: Ahh... I see your point. I'd expect this to be rare, though. Most linux users would probably use their package manager, and other JDK packages, such as Oracle and the adoptium project give you an actual installer that automatically sets this stuff up.

stephengold: Rare on Linux. More likely on Windows.

stephengold: I've added a note explaining that JAVA_HOME isn't always needed.

stephengold:
Unless there is further discussion, I plan to integrate this PR in about 24 hours.

Issue #1680, opened 2021-11-24
inefficient use of loggers


stephengold:
Issue #1672 drew my attention to inefficient use of loggers in JMonkeyEngine. A quick scan found dozens of instances that can probably be improved. A common anti-pattern is code that constructs a log message using string concatenation, as in `GltfUtilities`:
```java
                logger.log(Level.WARNING, "Unsupported Vertex Buffer type " + attribute);
```

Most of these should be refactored.

pspeed42:
WARNING is a bad example because it is almost always enabled and so the check is redundant.

Whatever crappy-JUL's version of INFO is... that's sort of on the fence.  Finer than that should either be gated or use the pattern version.  (I prefer gating in general if a conversion is being done because that pattern idiom has an argument limit before it's no better than concatenating a string.)

vxel:
Indeed, I spotted the same issue in a couple of classes in jme3-core. I did not check in other libs.
Not all FINE loggers are problematic though, e.g. those used only during initialization.

Additionnally, I wonder if it is worth the effort to gate logger statements that use multiple parameters, such as : 
`logger.log(Level.FINE, "{0} {1}", new Object[]{myobject1, myobject2})`
Since this statement implies the creation of an object array, I also added a gate, but not systematically.

tonihele:
Isnt there the method that takes in the supplier? That is executed only
when the line is actually logged. Could be used for complex cases that have
a performance impact on creating the message. Im not sure how heavy this
method is though...

vxel:
No, this statement does not use the supplier. I'm pretty sure the performance penalty of the array creation is present, even if the logger is not enabled, because the parameter is evaluated before calling the log method.

It is true that we can use a supplier instead of a gate for log statements like this : 
`logger.log(Level.FINE, "My log : {0}", () -> myMethod())`
This does not seem to be a common practice in jme3, so I use a gate for now.

stephengold:
> I'm preparing a PR fot this. Before submit it, you may take a look and tell me what you think.

I'm glad to know you're working on this, and I look forward to seeing your changes.

However, I'll do that *after* you submit the PR, not before.

pspeed42:
re: logger.log(Level.FINE, "{0} {1}", new Object[]{myobject1, myobject2})

[...]

PR #1682, opened 2021-11-25
Fix issue 1680 (inefficient use of loggers)

vxel:
Fixes issue #1680 

- Only FINE, FINER, FINEST, INFO levels were gated
- A gate is added only if a paremeter needs to be evaluated
- No use of lambda to preserve current practices
- Assumption : methods used in logging parameters did not produce side-effects


stephengold:
Thanks for your help, @vxel!

jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/JInputJoyInput.java
```diff
@@ -262,7 +263,7 @@ protected void addButton( Component comp ) {
                 original = String.valueOf(buttonIndex.size());
             }
             String logicalId = JoystickCompatibilityMappings.remapButton( controller.getName(), original );
-            if( logicalId != original ) {
+            if (logger.isLoggable(Level.FINE) && !Objects.equals(logicalId, original)) {
```
stephengold, line 266: This may fix a logic error in the existing code.

jme3-desktop/src/main/java/com/jme3/system/JmeDesktopSystem.java
```diff
@@ -337,7 +337,9 @@ public void initialize(AppSettings settings) {
             return;
         }
         initialized = true;
-        logger.log(Level.INFO, getBuildInfo());
+        if (logger.isLoggable(Level.INFO)) {
+            logger.log(Level.INFO, getBuildInfo());
```
stephengold, line 341: I think this is okay because `getBuildInfo()` has no side effects.

stephengold: However, since `JmeDesktopSystem` only gets initialized once, this is probably not a worthwhile change.

jme3-examples/src/main/java/jme3test/TestChooser.java
```diff
@@ -148,7 +148,10 @@ private void find(String packageName, boolean recursive,
 
         try {
             Path directory = Paths.get(uri);
-            logger.log(Level.FINE, "Searching for Demo classes in \"{0}\".", directory.getFileName().toString());
+            if (logger.isLoggable(Level.FINE)) {
+                logger.log(Level.FINE, "Searching for Demo classes in \"{0}\".",
```
stephengold, line 152: Optimizing the performance of `TestChooser` is probably not worthwhile.

jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/JInputJoyInput.java
```diff
@@ -284,7 +286,7 @@ protected void addAxis( Component comp ) {
             String name = comp.getName();
             String original = id.getName();
             String logicalId = JoystickCompatibilityMappings.remapAxis( controller.getName(), original );
-            if( logicalId != original ) {
+            if(logger.isLoggable(Level.FINE) && !Objects.equals(logicalId, original)) {
```
stephengold, line 289: This may fix a logic error in the existing code.

jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java
```diff
@@ -217,7 +217,9 @@ public void run(){
         }
 
         loadNatives();
-        logger.log(Level.FINE, "Using LWJGL {0}", Sys.getVersion());
+        if (logger.isLoggable(Level.FINE)) {
+            logger.log(Level.FINE, "Using LWJGL {0}", Sys.getVersion());
```
stephengold, line 221: This is initialization code that's probably not worth optimizing.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -635,7 +635,9 @@ public void run() {
                     + "Must set with JmeContext.setSystemListener().");
         }
 
-        LOGGER.log(Level.FINE, "Using LWJGL {0}", Version.getVersion());
+        if (LOGGER.isLoggable(Level.FINE)) {
+            LOGGER.log(Level.FINE, "Using LWJGL {0}", Version.getVersion());
```
stephengold, line 639: This is initialization code, probably not worth optimizing.

jme3-vr/src/main/java/com/jme3/app/VRApplication.java
```diff
@@ -817,8 +817,10 @@ public void start(JmeContext.Type contextType, boolean waitFor){
             settings = new AppSettings(true);
         }
 
-        logger.log(Level.FINE, "Starting application: {0}", getClass().getName());
-        
+        if (logger.isLoggable(Level.FINE)) {
+            logger.log(Level.FINE, "Starting application: {0}", getClass().getName());
```
stephengold, line 821: This is startup code, probably not worth optimizing.

jme3-vr/src/main/java/com/jme3/app/VRApplication.java
```diff
@@ -1513,7 +1515,9 @@ public void stop(){
      */
     @Override
     public void stop(boolean waitFor){
-        logger.log(Level.FINE, "Closing application: {0}", getClass().getName());
+        if (logger.isLoggable(Level.FINE)) {
+            logger.log(Level.FINE, "Closing application: {0}", getClass().getName());
```
stephengold, line 1519: Code that's executed once, probably not worth optimizing.

jme3-vr/src/main/java/com/jme3/system/lwjgl/LwjglWindowVR.java
```diff
@@ -455,8 +455,9 @@ public void run() {
                     + "Must set with JmeContext.setSystemListener().");
         }
 
- 
-        LOGGER.log(Level.FINE, "Using LWJGL {0}", Version.getVersion());
+        if (LOGGER.isLoggable(Level.FINE)) {
+            LOGGER.log(Level.FINE, "Using LWJGL {0}", Version.getVersion());
```
stephengold, line 459: This is initialization code. Probably not worth optimizing.

stephengold (changes requested):
I reviewed all the changes and didn't find any major problems. However, there were a number of changes I consider unnecessary. Also some refactoring to be scrutinized.

jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/JInputJoyInput.java
vxel: Regarding the original code, it is difficult to determine it if works or not - which is already a problem.
So I would recommend to replace it with code that is obviously correct.
Since this change is related only to the display of the log message, I chose to include it in this PR.

Please tell me explicitely if you want me to preserve the original logic.

jme3-desktop/src/main/java/com/jme3/system/JmeDesktopSystem.java
tonihele: About these 

> not a worthwhile change

They are already done, so there shouldn't be any harm? One millisecond faster app start up time in production with less logging :) Sounds wasteful to me if this comment means that the change should be reverted. Just kinda asking for clarification what exactly is requested by this comment.

jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/JInputJoyInput.java
stephengold: Normally I fuss about changes outside the scope of the PR, but in this case, I don't mind. I commented in order to bring it to the attention of other reviewers.

I don't want you to preserve the original logic.

jme3-desktop/src/main/java/com/jme3/system/JmeDesktopSystem.java
stephengold: I'm requesting that the "not worthwhile" changes be reverted.

vxel: Ok those changes have been reverted.

jme3-lwjgl/src/main/java/com/jme3/input/lwjgl/JInputJoyInput.java
pspeed42: Note: the change is fine but the reason it "worked" before is because the strings are regurgitated by the library and so should always be the same instance from call to call.  Meaning: code never constructs one of these strings.  We are passed them and then in turn pass them around.

...it still looks weird and I'm ok with the change.  Though consider that it is now not catching the same condition that it did before.  Whether that matters or not, I don't know.

My explanation above on how the logical IDs are handled could in theory mean that other code is counting on this instance-sameness and this check was confirming that (and now doesn't anymore).  If that's the case, I would certainly have expected a comment.

stephengold:

stephengold (approved):

PR #1688, opened 2021-11-28
use short-circuit OR in Spatial.requiresUpdates()

stephengold:
This corrects a typo in the jme3-core source code, making it slightly more efficient.


stephengold:
Unless there's substantial discussion, I plan to integrate this in the next day or two.

PR #1689, opened 2021-11-30
new javadoc for Vector3f

stephengold:
`Vector3f` may be the most crucial class for newcomers to learn, but unfortunately its javadoc is often vague, inconsistent, or even misleading.

This PR tries to raise the javadoc to a higher standard:
+ explicitly indicate which methods modify the current instance and which arguments are modified
+ explicitly indicate which methods return a new instance and which can return null
+ document which arguments cannot be null and what effect null arguments have
+ consistently refer to X, Y, and Z as "components" instead of "values" or "fields"
+ in cross-product methods, distinguish the right factor from the left factor
+ don't refer to calculations that modify the current instance as "internal"
+ don't refer to a fraction as a "percentage"
+ don't constantly tout the benefits of chaining


stephengold:

Issue #1193, opened 2019-09-20
jme3test.app.TestApplication hangs with LWJGL3


stephengold:
Current `master` branch running on Windows:
 + With LWJGL2, `TestApplication` completes in < 20 seconds without user input, after displaying a black window and a black fullscreen.
 + With LWJGL3, `TestApplication` displays the black window, then waits for the user to close the window. If the user closes the window, it pauses, displays the black fullscreen, and hangs. ALT-Tab is required to regain control.

stephengold:
Observed similar behaviour on Linux.

louhy:
I get the same result here on Linux. Curiosity struck so here's an analysis. 

TLDR: It seems to me the test is incompatible with the way the LWJGL3 context class works (LwjglWindow.java). LWJGL2 creates a separate thread for rendering loop, 3 uses the _current_ 
 (test!) thread for the rendering loop - so obviously, it blocks.

That's probably all that's needed, but here's my original verbose version: 

This test expects the current thread to be allowed to continue after calling start() (which eventually calls context.create()). That's fine for LWJGL2, which looks like this (see: LwjglDisplay.java): 

```
    public void create(boolean waitFor){
        if (created.get()){
            logger.warning("create() called when display is already created!");
            return;
        }
        new Thread(this, THREAD_NAME).start();
        if (waitFor)
            waitFor(true);
    }
```

But with LWJGL3 this is what it does (see: LwjglWindow.java): 
```
    public void create(boolean waitFor) {
        if (created.get()) {
            LOGGER.warning("create() called when display is already created!");
            return;
        }
        // NOTE: this is required for Mac OS X!
        mainThread = Thread.currentThread();
        run();
    }
```

That run() never lets the current thread continue on (rendering loop), so the test gets stuck. Closing the window I guess kills the context or something, and apparently breaks it out of the loop in run(). Fullscreen there's no window to close, so it gets "stuck".

So I think due to the way the context classes work, I would say this test is just fundamentally incompatible with LWJGL3. 

If the test can be made specific to an LWJGL version, I would say this should be a test for 2 only.

louhy:
Just to see what happens I tried having the LWJGL3 class create a new thread like 2 does, it passes the test and seems to work fine (Linux). 

I see the "required for Mac" comment... wouldn't be hard to do an OS-specific if so that it behaves the same with other OSes. But just to make the test pass, I'm going to assume that's not worth doing. 

[...]

Issue #1558, opened 2021-05-15
TestAWTPanels crashes with LWJGL v3 on Linux


stephengold:
With "jme3-lwjgl3" in JME v3.4.0-beta4, `TestAWTPanels` hangs. No output is generated, and no window appears.

Debugger suggests a deadlock, with the "jME Main" thread parked in line 103:
```java
            panelsAreReady.await();
```

stephengold:
This issue dates back at least to JME v3.2.4-stable.

stephengold:
When I tested #1690 I accidentally used LWJGL v2 instead of LWJGL v3.

I just re-tested TestAWTPanels on Linux with LWJGL v3.3.0 and latest "master". I got the following crash:
```text
> Task :jme3-examples:run
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f59932effc4, pid=9041, tid=0x00007f5965790700
#
# JRE version: OpenJDK Runtime Environment (8.0_212-b04) (build 1.8.0_212-b04)
# Java VM: OpenJDK 64-Bit Server VM (25.212-b04 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libpthread.so.0+0xbfc4]  pthread_mutex_lock+0x4
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/sgold/Git/jmonkeyengine/jme3-examples/hs_err_pid9041.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

> Task :jme3-examples:run FAILED
```

Here's the stack trace from hs_err_pid9041.log:
```text
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libpthread.so.0+0xbfc4]  pthread_mutex_lock+0x4
C  0x0000000000000000

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
[...]

Issue #1514, opened 2021-04-06
jme-lwjgl3 - bug for starting app


supertick:
What is expected :  When starting a SimpleApplication JMonkeyEngine would spawn a thread for internal processing with an option to "waitFor" block until started.

What is observered: The calling thread is endlessly blocked by calling the run() method directly - with a note ...
// NOTE: this is required for Mac OS X!
https://github.com/jMonkeyEngine/jmonkeyengine/blob/v3.2.3-stable/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java#L420-L423

PR submitted - https://github.com/jMonkeyEngine/jmonkeyengine/pull/1513

tonihele:
Hmm, is this the correct issue... But check out what LibGX tries to do with their LWJGL 3 MacOS starting problem:
https://github.com/libgdx/libgdx/pull/6673

Ali-RS:
Fixed in https://github.com/jMonkeyEngine/jmonkeyengine/pull/1690

PR #1690, opened 2021-12-01
fix issues #1193 and #1558 for Linux/Windows (LWJGL v3 hangs)

bob0bob:
TestAWTPanels hangs with LWJGL v3 #1558 /  jme3test.app.TestApplication hangs with LWJGL3 #1193

Made LWJGL3-JME to launch in the same way the LWJGL2-JME launches.  This always Windows and Linux to run, still needs testing on Max OS X


bob0bob:
This is also related to issue  jme-lwjgl3 - bug for starting app #1514

Jeddic:
From what I can tell you just reverted the changes? I can already tell you this breaks osx which is why it was changed in the first place.

bob0bob:
I made more changes, based on the history of the changes and other issues about this issue.  It does end up checking to see if it is a Mac and do something else.   This is solution I'm using on my engine,  it needs to be tested but I worry there is not anyone to test for it.  I will cross my fingers for one to step up to test.

pspeed42:
We should try to come up with code that doesn't use "magic guess the OS" string checks.  Even if that means changing the semantics of what start() means.

stephengold:
Sorry if this is a stupid question, but why use `mainThread.run()` instead of `mainThread.start()`?

bob0bob:
I found a way to check from inside LWJGL3,  they have a call to return a ENUM for the 3 types of hardware that is support.  I will put that in but last night my eclipse build started to mess up and I couldn't get it to run in debug to verify it.

The enum returns,  WINDOWS, LINUX or MAC.

They even have a class to verify it is running on the main thread and return false is not, so you can throw an exception.

Jeddic:
I just confirmed that this crashes on OSX.

bob0bob:
I made changes to remove the string comparison.  This does fix the Windows and Linux.   Jeddic, if you have a MAC,  Please look into a solution for Mac.  Looking into the JME-LWJGL2 library the create() method ends up doing the same for the Mac.

This is moving forward to getting LWJGL3 functioning.   I'm not sure what "CRASH" means,  because it does the run() call on the current thread.   If anything this is not a solution for Mac, but for other 2 OS.

stephengold:
@Jeddic next time, please copy/paste the stack trace

Jeddic:
```
java.lang.ExceptionInInitializerError
	at org.lwjgl.glfw.GLFW.glfwInit(GLFW.java:1046)
	at com.jme3.system.lwjgl.LwjglWindow.createContext(LwjglWindow.java:197)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:515)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:650)
Caused by: java.lang.IllegalStateException: GLFW may only be used on the main thread. This check may be disabled with Configuration.GLFW_CHECK_THREAD0.
	at org.lwjgl.glfw.EventLoop.<clinit>(EventLoop.java:30)
	... 5 more
```
There you go.

Jeddic:
So I tested the current version and it is close, but calling mainthread.run() won't work. There isn't a runnable assigned so it immediately returns and closes the application. Just calling run() directly works for me.

bob0bob:
Jeddic,  thanks.    I but the run() call back for MacOS.   This will not solve the issue for Mac and AWT but will for Linux and Windows.

Jeddic:
Yah, fixing it for osx will probably take changing the canvas to use the lwjgl-awt project.  Let me know if there is anything else that needs to be tested.

stephengold:
If this PR doesn't solve the issue for macOS, then why not remove the `if (Platform.get() == Platform.MACOSX) {` clause?

Also, our preferred Java style specifies that braces are used with `if` statements, even when the body is empty or contains only a single statement, so the `if (waitfor)` clause(s) need braces.

Jeddic:
Well, this pull request should fix the issue for windows / linux. The clause is there specifically so it doesn't break macOS in the process.

bob0bob:
I added the braces for the correct coding.

We have to leave in the check for MAC,  because the Mac needs the run() call.  That call is blocking and non-returning.   But with the thread().start(),  that is a non-blocking call that will allow the return so the rest of the code gets executed after the call to app.start() in main.

stephengold:
Rather than slog through another round of review, I fixed a couple minor issues and integrated the PR.

Thank you for your contribution to the Engine!

stephengold:
Now TestAWTPanels is broken in a new and different way. See issue #1558 for details.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -477,10 +479,21 @@ public void create(boolean waitFor) {
...
+        if (Platform.get() == Platform.MACOSX) {
+            // NOTE: this is required for Mac OS X!
+            mainThread = Thread.currentThread();
+            mainThread.setName("jME3 Main");
+            if (waitFor) {
+                LOGGER.warning("create(true) is not supported for macOS!");
+            }
+            run();
+        } else {
+            new Thread(this, "jME3 Main").start();
```
Ali-RS, line 491: A bit late to the party!

@bob0bob `mainThread` should be initialized to newly create thread inside `else` statement.

bob0bob:
Okay, I checked in the changes and tested it against the example from the forum where it is hanging.

bob0bob:
Do I start a new merge request to get the changes over since it has already been merged.

Ali-RS:
No need for that, it is already solved in #1722.

PR #1691, opened 2021-12-01
Use static final logger

tonihele:
We just took this to use and what we use this for is to generate tangents for thousands of geometries. This said batch also gives out 300 lines of log... This particular line. So this is a small optimization. Follows the standard jME logging style.


stephengold (approved):
Looks good. Thanks!

PR #1692, opened 2021-12-03
Chase cam can rotate fix on re-enable

tlf30:
See discussion: https://hub.jmonkeyengine.org/t/chasecamera-canrotate-not-set-to-true-when-re-enabled-after-being-disabled/45138/14

This fixes the chase camera not correctly setting the canRotate flag upon being re-enabled when drag to rotate is false.


jme3-core/src/main/java/com/jme3/input/ChaseCamera.java
```diff
@@ -547,6 +547,8 @@ public void setEnabled(boolean enabled) {
         this.enabled = enabled;
         if (!enabled) {
```
stephengold, line 548: This code would be more readable if you inverted the test.

stephengold:
Please supply a test case for the issue.

jme3-core/src/main/java/com/jme3/input/ChaseCamera.java
tlf30: I agree, I did not want to modify too much of the existing code. I can easily make that change though.

tlf30:
I'll work on getting a test case built

jme3-core/src/main/java/com/jme3/input/ChaseCamera.java
stephengold: In general, keeping PRs simple is a good thing. However, the resulting source code will be read 100x times more than the PR will, so modifying 4 lines (instead of 2) isn't "too much".

stephengold:
@tlf30 are you still available to work on this PR?

tlf30:
@stephengold yes, sorry I forgot about this one. I'll see about putting something together tonight or tomorrow.

tlf30:
@stephengold let me know if this test case is acceptable.

stephengold:
The test is good. I appreciate the fact that it displays instructions when run. Manydon't, so thanks for that!

With your permission, I'd like to make some minor changes to the test before integrating this PR -- changes such as whitespace, renaming a local variable, avoiding a deprecated method, avoiding code duplication, deleting an unused import, and privatizing a public method. Are you okay with that plan?

Also, I think the inline documentation of the test is misleading. It talks in 2 places about keeping "the correct flags set so that we can still rotate without dragging." But the bug doesn't merely disable rotation without dragging; it totally disables mouse rotation. Or at least that's how it seems to me. Am I missing something?

tlf30:
@stephengold feel free to modify the PR.  Yes you are correct, the flag in question does for the bug disable all movement of the camera, but the flag's actual purpose is to be used to only enable camera movement unless dragging. I can see where my comment is confusing though to what is being addressed in the fix for the bug. 

I copied the code for the test from another test class. Perhaps it would be good to build a template and _standard_ for how tests should be built, I think this would cover most cases (I imagine that some tests may require large custom code though) and provide a good reference for developing tests to keep consistency. But if nothing else at least standard for test programs (and a good time to develop one since we are adding standards elsewhere).  Any thoughts?

stephengold:
> Perhaps it would be good to build a template and standard for how tests should be built, I think this would cover most cases

Test cases are so diverse, I imagine it would be difficult to write a template that would help with even half of them. Currently there are dozens (hundreds?) of apps in jme3-examples, any one of which could serve as the starting point for a new test case.

As I see it is, the real issues is that few people are familiar all the apps in jme3-examples. So it's difficult to choose the best one to use as the starting point for a particular test case.

But if you want to draft a template, go right ahead!

stephengold:
Thanks for your contribution, Trevor!

tlf30:
Thank you Stephen for the cleanup work on this PR! As always you are doing great work for JME.

Issue #1687, opened 2021-11-28
NullPointerException in TestIssue1120 when run from TestChooser


stephengold:
When run from the test chooser, `TestIssue1120` in jme3-examples crashes:
```text
Nov 28, 2021 1:10:12 AM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,6,main]
java.lang.NullPointerException
	at jme3test.bullet.TestIssue1120.simpleInitApp(TestIssue1120.java:142)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:240)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:139)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:221)
	at java.lang.Thread.run(Thread.java:748)
```

The application's `test` field is `null` because it wasn't initialized. The initialization was placed in `main()`, which the test chooser never executes.

bob0bob:
I submitted a pull request for this fix.

PR #1694, opened 2021-12-03
fix issue #1687 (NPE in TestIssue1120)

stephengold:
The fix was to access the `settings` field directly (in `simpleInitApp()`) instead of relying on `main()` to initialize the `test` field.


stephengold:
I decided instead to integrate PR #1697, which makes the exact same change!

Issue #1686, opened 2021-11-28
jmeClone() methods returning a supertype


stephengold:
Similar to #1676, the `jmeClone()` method returns an object of the same type and should be declared accordingly.

I found several dozen examples, mostly in jme3-core. Currently, for example, `AnimClip.jmeClone()` returns `Object`.

stephengold:
I've decided this isn't worth fixing.

PR #1695, opened 2021-12-03
fix issue #1686 (jmeClone() return types)

stephengold:
This is riskier than #1683. For one thing, it affects many commonly extended classes including `BetterCharacterControl` and `AbstractControl`.  Also, it modifies the `JmeCloneable` interface itself.


pspeed42:
Also, I think applications should not be calling jmeClone() directly.  It should only be called by JmeCloner.  Perhaps leaving these as Object will help discourage users from messing up?

...and JmeCloner does not care what the method returns and so Object is fine.

stephengold:
After thinking about this some more, I'm convinced this isn't a worthwhile change.

PR #1697, opened 2021-12-03
Fix issue #1687 (NPE in TestIssue1120)

bob0bob:
When run from the test chooser, TestIssue1120 in jme3-examples crashes:

The application's test field is null because it wasn't initialized. The initialization was placed in main(), which the test chooser never executes.


stephengold:
I self-assigned the issue before I saw this PR, and now we have 2 fixes for #1687!

jme3-examples/src/main/java/jme3test/bullet/TestIssue1120.java
```diff
@@ -65,7 +65,7 @@
  */
 public class TestIssue1120 extends SimpleApplication {
 
-    private static TestIssue1120 test;
+//    private static TestIssue1120 test;
```
stephengold, line 68: Please delete the commented-out code. Then this will be ready for integration.

bob0bob: done

stephengold:
Thanks for your help.
By the way ... best practice is to create a new Git branch name for each pull request.

bob0bob:
This is what I did.  I created a branch for each pull request,  so I could continue working.

stephengold:
Then I don't understand why GitHub was showing 19 commits in the branch. Ah well...

PR #1663, opened 2021-11-18
java comments and javadoc

stephengold:
This PR corrects typos, grammar, and punctuation issues in Java source.

PR #1698, opened 2021-12-03
correct typos in Java comments and javadoc

stephengold:
Very similar to #1663, this PR solves missing words/hyphens, unintentionally repeated words, non-standard punctuation, and similar issues.


stephengold:
If there's nothing to discuss, I plan to integrate this PR in about 24 hours.

PR #1701, opened 2021-12-08
New RectangleMesh shape class

fba-rio:
This RectangleMesh class was created based on the idea proposed at https://hub.jmonkeyengine.org/t/next-release-of-the-engine/44988/170

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Rectangle.java
+++ b/jme3-core/src/main/java/com/jme3/math/Rectangle.java
@@ -152,7 +152,6 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
      * 
      * @param normal the vector to store the result in
      * @return the normal vector of this <code>Rectangle</code>
-     * @see https://web.ma.utexas.edu/users/m408m/Display12-5-4.shtml
      */
     public Vector3f calculateNormal(Vector3f normal) {
         if (normal == null) {
--- a/jme3-core/src/main/java/com/jme3/math/Rectangle.java
+++ b/jme3-core/src/main/java/com/jme3/math/Rectangle.java
@@ -143,7 +143,7 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
      * Returns the coordinates of the 4th corner, calculated by the formula
      * D = (B + C) - A .
      *
-     * @return a new Vector3f
+     * @return the corner location (a new Vector3f)
      */
     public Vector3f getD() {
         return b.add(c).subtractLocal(a);
@@ -152,13 +152,13 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
     /**
      * Returns a normal vector, calculated by the formula
      * <pre>
-     *      (C - A) x (B - A) 
+     *      (C - A) x (B - A)
      * N = -------------------
      *     |(C - A) x (B - A)|
      * </pre>
      *
      * @param normal storage for the normal, or null for a new Vector3f
-     * @return either {@code normal} or a new Vector3f
+     * @return the normal direction (either {@code normal} or a new Vector3f)
      */
     public Vector3f calculateNormal(Vector3f normal) {
         if (normal == null) {
--- a/jme3-core/src/main/java/com/jme3/math/Rectangle.java
+++ b/jme3-core/src/main/java/com/jme3/math/Rectangle.java
@@ -149,7 +149,7 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
 
         Vector3f v1 = c.subtract(b);
         Vector3f v2 = a.subtract(b);
-        normal = v1.cross(v2).normalizeLocal();
+        normal.set(v1.crossLocal(v2).normalizeLocal());
 
         return normal;
     }
--- a/jme3-core/src/main/java/com/jme3/math/Rectangle.java
+++ b/jme3-core/src/main/java/com/jme3/math/Rectangle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2020 jMonkeyEngine
+ * Copyright (c) 2009-2021 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -37,7 +37,7 @@ import java.io.IOException;
 /**
  * <code>Rectangle</code> defines a finite plane within three dimensional space
  * that is specified via three points (A, B, C). These three points define a
- * triangle with the fourth point defining the rectangle ((B + C) - A.
+ * triangle with the fourth point defining the rectangle (C - B) + A.
  *
  * @author Mark Powell
  * @author Joshua Slack
@@ -126,8 +126,37 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
     }
 
     /**
+     * <code>getD</code> returns the fourth point of the rectangle.
+     * The fourth point is given by calculating (C - B) + A.
+     *
+     * @return the fourth point of this rectangle.
+     */
+    public Vector3f getD() {
+        return c.subtract(b).addLocal(a);
+    }
+
+    /**
+     * Computes the normal vector of this <code>Rectangle</code>.
+     * 
+     * @param normal the vector to store the result in
+     * @return the normal vector of this <code>Rectangle</code>
+     * @see https://web.ma.utexas.edu/users/m408m/Display12-5-4.shtml
+     */
+    public Vector3f calculateNormal(Vector3f normal) {
+        if (normal == null) {
+            normal = new Vector3f();
+        }
+
+        Vector3f v1 = c.subtract(b);
+        Vector3f v2 = a.subtract(b);
+        normal = v1.cross(v2).normalizeLocal();
+
+        return normal;
+    }
+
+    /**
      * <code>random</code> returns a random point within the plane defined by:
-     * A, B, C, and (B + C) - A.
+     * A, B, C, and (C - B) + A.
      *
      * @return a random point within the rectangle.
      */
@@ -137,7 +166,7 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
 
     /**
      * <code>random</code> returns a random point within the plane defined by:
-     * A, B, C, and (B + C) - A.
+     * A, B, C, and (C - B) + A.
      *
      * @param result
      *            Vector to store result in
--- a/jme3-core/src/main/java/com/jme3/math/Rectangle.java
+++ b/jme3-core/src/main/java/com/jme3/math/Rectangle.java
@@ -151,7 +151,11 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
 
     /**
      * Returns a normal vector, calculated by the formula
-     * N = |(C - A) x (B - A)| .
+     * <pre>
+     *      (C - A) x (B - A) 
+     * N = -------------------
+     *     |(C - A) x (B - A)|
+     * </pre>
      *
      * @param normal storage for the normal, or null for a new Vector3f
      * @return either {@code normal} or a new Vector3f
--- a/jme3-core/src/main/java/com/jme3/math/Rectangle.java
+++ b/jme3-core/src/main/java/com/jme3/math/Rectangle.java
@@ -37,7 +37,19 @@ import java.io.IOException;
 /**
  * <code>Rectangle</code> defines a finite plane within three dimensional space
  * that is specified via three points (A, B, C). These three points define a
- * triangle with the fourth point defining the rectangle (C - B) + A.
+ * triangle with the fourth point defining the rectangle (B + C) - A.<p>
+ * 
+ * The points of a rectangle are arranged as follows: 
+ * 
+ * <pre>
+ *     C +-----+ D
+ *       |\    |
+ *       | \   |
+ *       |  \  |
+ *       |   \ |
+ *       |    \|
+ *     A +-----+ B
+ * </pre>
  *
  * @author Mark Powell
  * @author Joshua Slack
@@ -127,12 +139,12 @@ public final class Rectangle implements Savable, Cloneable, java.io.Serializable
 
     /**
      * <code>getD</code> returns the fourth point of the rectangle.
-     * The fourth point is given by calculating (C - B) + A.
+     * The fourth point is given by calculating (B + C) - A.
      *
      * @return the fourth point of this rectangle.
      */
     public Vector3f getD() {
-        return c.subtract(b).addLocal(a);
+        return b.add(c).subtractLocal(a);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
```diff
@@ -0,0 +1,163 @@
...
+import com.jme3.export.OutputCapsule;
+import com.jme3.math.Rectangle;
+import com.jme3.math.Vector3f;
+import com.jme3.scene.Mesh;
+import com.jme3.scene.VertexBuffer.Type;
+
+/**
+ * {@code RectangleMesh} is a rectangular plane in space similar to
+ * {@link com.jme3.scene.shape.Quad}, but it uses a
+ * {@link com.jme3.math.Rectangle} as its coordinates system.
```
stephengold, line 48: It doesn't use the Rectangle as its coordinate system, it uses the Rectangle to position its vertices.

stephengold: Also, it's not a rectangular plane. It's a rectangle.

fba-rio: I updated the description according to your suggestions.

jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
```diff
@@ -0,0 +1,163 @@
...
+ * {@link com.jme3.scene.shape.Quad}. It uses a
+ * {@link com.jme3.math.Rectangle} to position its vertices.
+ * 
+ * @author Francivan Bezerra
+ */
+public class RectangleMesh extends Mesh {
+
+    private Rectangle rectangle;
+
+    public RectangleMesh() {
```
stephengold, line 56: This no-arg constructor appears to produce a mesh with all vertices at (0,0,0) which isn't very useful. I suggest making it initialize the `Rectangle` with `Rectangle(new Vector3f(), new Vector3f(1,0,0), new Vector3f(0,1,0))` or something similar.

jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * {@code RectangleMesh} is a rectangle in space similar to
+ * {@link com.jme3.scene.shape.Quad}. It uses a
+ * {@link com.jme3.math.Rectangle} to position its vertices.
+ * 
+ * @author Francivan Bezerra
+ */
+public class RectangleMesh extends Mesh {
+
+    private Rectangle rectangle;
```
stephengold, line 54: The `rectangle` field should be cloned when the mesh is cloned. Please override the `jmeClone()` method.

jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        setBuffer(Type.Normal, 3, new float[] {
+                0, 0,
+                1, 0,
+                0, 1,
+                0, 0,
+                1, 0,
+                0, 1
+        });
```
stephengold, line 139: These normals are only correct for a rectangle in the X-Y plane. Calculate the normal vector using the `com.jme3.math.Triangle` class or similar logic.

jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
fba-rio: Done. But I've overridden cloneFields() instead.

jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
fba-rio: Makes sense. I changed it. Now this constructor creates a square of side=1 as you proposed.

jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
fba-rio: I used the static method `Triangle.computeTriangleNormal()` to fix it.

jme3-core/src/main/java/com/jme3/scene/shape/RectangleMesh.java
stephengold: You did the right thing. My bad!

stephengold:
This is better. There might be advantages in adding methods to the `com.jme3.math.Rectangle` class to calculate `D` and the normal direction vector. What do you think?

stephengold:
Also ... thinking ahead to how we might test `RectangleMesh`. One place it might be useful is in example apps (like `TestCameraNode`, `TestTransparentShadow`, and `TestChaseCamera`) that currently use `Quad` to create a floor for test scenes.

I encourage you to modify these apps (and others) to use `RectangleMesh`.

fba-rio:
Yes, a `getD()` method in `Rectangle` would be helpful. I'm not sure if the normal direction vector calculation should be there though. `Rectangle` in the `math` package essentially is not a mesh. I know that `Triangle` is not a mesh either, but it makes sense to have a method to calculate the normal vector there because meshes are made of triangles.

Anyway, I am going to open a new PR to add `getD()` to `Rectangle`. I can add a method to calculate the normal vector there too if you think it's the right place to have it.

Also, I'll take a look at the examples and modify them to use `RectangleMesh`.

stephengold:
Normals (of rectangles) are used to calculate distances and for other purposes that have nothing to do with meshes. So a method to calculate the normal would have broad application, in my opinion.

One tricky thing about normals is documenting how the sign of the normal is determined...

stephengold:
P.S.: Note that the `Rectangle` class tends to produce parallelograms if the vertices are not chosen carefully!

fba-rio:
I moved the `calculateNormal()` method from `RectangleMesh` to `Rectangle` and created a `getD()` in `Rectangle`.

I used this PR to make the changes in `Rectangle`. I hope it is okay.

stephengold:
It's okay to make both changes in the same PR, since they're so closely related.

However, I think it's unwise to change the semantics of `Rectangle` as you have done. Outher developers may have relied on the old definition of D = (B + C) - A.

fba-rio:
The thing is if you do the math you won't find the fourth point of the rectangle with that formula. Instead, you will find a point outside the rectangle area.

I don't know where the original developers of `Rectangle` got that formula from, but clearly there is a mistake there. Anyway, the formula is never used in the source code, it only appears in the javadoc comments. Changing these comments is like fixing a bug, except that I'm not changing the behavior of any function. No source code using `Rectangle` will be affected. Also, leaving the comments as they are now and calculating `getD()` differently can be confusing.

stephengold:
The old formula assumes that the order of the vertices (as you follow the rectangle's perimeter) is A-B-D-C. This isn't the usual labeling used in math textbooks (which favor A-B-C-D). It was careless of Mark and Joshua to use a non-standard vertex ordering without documenting it, but I wouldn't call it a bug.

I'm very reluctant to change the vertex order. After 12+ years of telling developers that D = ((B + C) - A, it's very likely someone used that formula, even if it wasn't coded into jme3-core. The best we can do is document the gotchas, including this one.

fba-rio:
I Understand. Back to the original formula. I documented the order of the vertices in the first javadoc entry of `Rectangle.java`. Note that the texture coordinates in `RectangleMesh` still have to follow the A-B-C-D order.

fba-rio:
> Also ... thinking ahead to how we might test `RectangleMesh`. One place it might be useful is in example apps (like `TestCameraNode`, `TestTransparentShadow`, and `TestChaseCamera`) that currently use `Quad` to create a floor for test scenes.
> 
> I encourage you to modify these apps (and others) to use `RectangleMesh`.

As suggested, I've modified and tested some example apps, replacing `Quad` with `RectangleMesh`.

jme3-core/src/main/java/com/jme3/math/Rectangle.java
```diff
@@ -125,6 +137,35 @@ public void setC(Vector3f c) {
...
+    public Vector3f getD() {
+        return b.add(c).subtractLocal(a);
+    }
+
+    /**
+     * Computes the normal vector of this <code>Rectangle</code>.
+     * 
+     * @param normal the vector to store the result in
+     * @return the normal vector of this <code>Rectangle</code>
+     * @see https://web.ma.utexas.edu/users/m408m/Display12-5-4.shtml
```
capdevon, line 155: The correct syntax for inserting an external link in the javadoc should be this: @see <a href="URL#value"> label </a>. Despite this, we are sure that it is useful to insert a link to a domain outside the perimeter of the jMonkey engine that could be removed at any time or not accessible for all countries? If the page were suddenly removed we would have a dead link in the javadoc.

capdevon: I would suggest creating a specific test case for this class such as `TestRecatangleMesh` and not modifying the current test cases, where perhaps it would be better to replace the `Quad` with the `CenterQuad`, keeping the creation of floors simple and intuitive for new users. Just my opinion.

fba-rio: Link removed.

stephengold: > perhaps it would be better to replace the `Quad` with the `CenterQuad`, keeping the creation of floors simple and intuitive for new users. Just my opinion.

For floors, `RectangleMesh` is more intuitive because it doesn't require rotation of the `Geometry`. `CenterQuad` is in the X-Y plane, so it requires rotation to create an X-Z floor.

pspeed42:
Though I still wonder how many of these 'new' shape classes could be replaced by a couple of static methods that operate on any mesh.

stephengold:
> a couple of static methods that operate on any mesh.

Operating on *any* mesh is a tall order, since JME meshes are extremely flexible: animated/morphing/non-animated, indexed/non-indexed, 9 mesh modes, 9 buffer formats, and all with/without binormals/normals/tangents/UVs/colors, All this flexibility is great for writing importers and generating custom meshes, not so good if you want to manipulate meshes in a general way.

I've been writing mesh-edit utilities for years, but I still don't grok all the options.

So you write a simple utility to translate the vertex positions, but you forget to update the bounds. Or you forget to translate the bind positions of an animated mesh. Then someone finds a mesh with a double-precision position buffer. Or throws interleaved data at it.

pspeed42:
You know what I meant.  Certainly, the dozens of custom shape objects folks are about to make just to offset this or twist that are all going to use the position buffer, normal buffer, etc..

Two or three static methods and a Quad would have been fine.  Then you get the added benefit that those static methods would work for box, sphere, <insert any mesh with a Position, Normal, Tangent, etc. buffer here>.

"Write a new shape class every time we want to center a thing" doesn't seem like the right answer.

stephengold:
While I certainly agree we should be cautious about adding `Mesh` subclasses to the Engine, this one has an unusually high benefit-to-cost ratio.

While not perfect, I believe this PR is ready to be integrated.

@fba-rio  I'd like to document a few caveats and other details. Would it be OK for me to make last-minute changes to the javadoc before integration?

fba-rio:
> @fba-rio I'd like to document a few caveats and other details. Would it be OK for me to make last-minute changes to the javadoc before integration?

Sure, go ahead!

fba-rio:
>     1. If `RectangleMesh.setNormal()` is used to reverse the normal vectors, the winding order of the triangles will be backward, which might cause unexpected behavior with some materials.
Shouldn't the responsibility to provide a valid vector lie with the developer who calls this method? If not, what do you think should be done in this case?

>     2. The `Rectangle.getD()` method is inefficient, creating 2 garbage instances of `Vector3f` each time it's invoked.
Good point. I made the necessary change.

>     3. `RectangleMesh.updateMesh()` should be `protected` or `private`, not `public`.
It is protected now.

>     4. `Rectangle.getD()` has a misleading name. Perhaps it should be renamed to `calculateD()`.
You're right. Renamed.

I think you've made a mistake in the javadoc of `Rectangle.calculateNormal`. You swapped the (A - B) part of the formula, as seen in the method itself.

stephengold:
> > 1. If `RectangleMesh.setNormal()` is used to reverse the normal vectors, the winding order of the triangles will be backward, which might cause unexpected behavior with some materials.
> 
> Shouldn't the responsibility to provide a valid vector lie with the developer who calls this method? If not, what do you think should be done in this case?

It's not that simple. For any particular parallelogram, there are 2 plausible normal vectors. For instance, a square in the X-Y plane could have normals of either +Z or -Z. I think the most likely use for `setNormal()` is to flip the sign of the normals. And flipping the sign is exactly what should reverse the winding order.

I see many possible solutions. One would be to update the winding order in `updateMesh()` based on the dot product between `rectangle.calculateNormal(null)` and `this.normal`, when the latter is non-null. But I think that's overkill. Since there are really only 2 plausible normals for a non-degenerate parallelogram, we just need to provide a mechanism to switch back and forth between them. So I suggest replacing `setNormal()` with a method named `flip()` or `reverse()` or something similar, a method that both negates the normal vector and reverses the winding order.

An even simpler possibility would be to just get rid of `setNormal()`. In that world, the corners determine the normal. If you want the normal to point in the other direction, you have to shuffle the corners, for instance swapping B and C.

fba-rio:
I took the `flip()` approach and removed `setNormal()`.

stephengold:
What does `flip()` do to the texture coordinates? Is that what a naive user would expect?

pspeed42:
This is the down side to all-in-one classes versus utility methods.  Always one more edge case instead of letting the user just pick the behavior they want.

fba-rio:
> What does `flip()` do to the texture coordinates? Is that what a naive user would expect?

My bad. I used a seamless texture in my tests and ended up neglecting the texture on `flip()`.

stephengold:
How would you like to resolve this issue?

stephengold:
Sorry, I missed 1f20015 for some reason, so ignore my last comment.

I'm still concerned about `flip()` and its documentation. It appears to re-write the position buffer, which isn't at all what I would expect for an indexed mesh.

Please describe what `flip()` does, at a high level, and add that info to the javadoc.

fba-rio:
I've rewritten the `flip()` method. Now it doesn't mess with the position buffer anymore, it's restricted to reverse the normal vector direction and rearrange the index buffer (in the `updateMesh()` method). Also, by doing so it becomes unnecessary to change the texture coordinates.

stephengold (approved):
That should do it! Thanks for your patience.

stephengold:
Thanks for your contribution to the project, Francivan! I enjoyed collaborating with you.

fba-rio:
My pleasure!
And I learned a little bit more in the process.

PR #1702, opened 2021-12-10
deprecate Quaternion.negate()

stephengold:
The `negate()` function in `Quaternion` doesn't follow the JME naming conventions. The other math functions that modify the current instance have "Local" in their names, but `negate()` doesn't.

Since this inconsistency could lead to buggy code, the old method is being deprecated, and a new method `negateLocal()` is being added. Unlike the old method, the new one returns `this` to allow function chaining.


stephengold:
This PR is ready for review.

stephengold:

PR #1703, opened 2021-12-11
add a version of CenterQuad to jme3-core

stephengold:
The `CenterQuad` class in jme3-vr seems to have some general utility, for instance for UI objects that will be rotated and/or scaled. This PR cleans it up and adds it to the jme3-core. Long-term, we may want to deprecate the version in jme3-vr.

Notable changes relative to jme3-vr:
+ no static instances
+ privatized the `updateGeometry()` methods
+ added a no-arg constructor and read/write methods to allow serialization
+ solved a winding-order bug when `width` is negative
+ avoided invocation of overridable methods in constructors
+ improved javadoc


stephengold:
This PR is ready for review.

stephengold:
Oops! Withdrawn in favor of PR #1704.

PR #1704, opened 2021-12-11
jme3-core:  add a CenterQuad class based on the one in jme3-vr

stephengold:
The `CenterQuad` class in jme3-vr seems to have some general utility, for instance for UI objects that will be rotated and/or scaled. This PR cleans it up and adds it to the jme3-core. Long-term, we may want to deprecate the version in jme3-vr.

Notable changes relative to jme3-vr:
+ no static instances
+ privatized the `updateGeometry()` methods
+ added a no-arg constructor and read/write methods to allow serialization
+ solved a winding-order bug when `width` is negative
+ avoided invocation of overridable methods in constructors
+ improved javadoc

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/shape/CenterQuad.java
+++ b/jme3-core/src/main/java/com/jme3/scene/shape/CenterQuad.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright (c) 2009-2021 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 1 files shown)


stephengold:
This PR is ready for review.

stephengold:
Regarding the lack of a no-arg constructor in the jme3-vr version, see issue #1705.

jme3-core/src/main/java/com/jme3/scene/shape/CenterQuad.java
```diff
@@ -0,0 +1,190 @@
...
+     * Instantiate a quad in the X-Y plane with the specified width and height.
+     *
+     * @param width the desired X extent or width
+     * @param height the desired Y extent or height
+     * @param flipCoords true to flip the texture coordinates (v=0 when
+     *     y=height/2) or false to leave them unflipped (v=1 when y=height/2)
+     */
+    public CenterQuad(float width, float height, boolean flipCoords) {
+        updateGeometry(width, height, flipCoords);
+        super.setStatic();
```
capdevon, line 86: This instruction can be removed. The `setsStatic()` method is already invoked in the `updateGeometry()` method.

stephengold: Thanks for the tip. I'll do that.

stephengold:
Unless there's more to discuss, I'd like to integrate this PR in about 24 hours.

PR #1685, opened 2021-11-28
solve issue #1158 (enabling Right-to-Left fonts)

Aufricer:
This pull request is solving the problem with displaying Right to Left texts in JME. fixes #1158 and would also make the additional file for testcase close #1159 unnecessary.

More info could be found in the [forum  ](https://hub.jmonkeyengine.org/t/issue-with-line-wrapping-of-right-to-left-text-in-bitmaptext/42148/5) or in the discussion for issue #1158.

What changed: 
Beside repairing the behaviour of RtL texts, a new parameter in BitMapfont allowing to set the textdirection in the font and in Bitmaptext. Before that was not possible in the font itself. 
An interface "GlyphParser" now allowing the user  of Persian, Arabic or other RtL oriented fonts to use individual parsers for the correct position of letters in a text (that may differ depending on letter before or after that) something that is not needed for  Left to Rights texts.
Also a new discovered bug that brings different results for linewidth from BitmapText.getLineWidth() and BitmapFont.getLineWidth() for textinput with multilines had been fixed.

We have tested it and feel the PR is ready for review.

Edit: Of course it is Right to Left texts (at least I wrote it wrong on each occasion in the text)

PR #1709, opened 2021-12-17
avoid invoking the newly-deprecated BitmapText 2-arg constructor

stephengold:
PR #1685 deprecated the 2-arg constructor for `BitmapText`. That constructor is used extensively in the Engine itself, especially in "jme3-examples".

Since the Engine (and especially "jme3-examples") ought to set a good example, this PR replaces all uses of the 2-arg constructor with the 1-arg constructor.


Aufricer (approved):
Nothing wrong here

stephengold:
Unless there's more to discuss, I plan to integrate this PR in about 20 hours.

PR #1714, opened 2021-12-19
downgrade or delete unnecessary library dependencies

stephengold:
These unnecessary dependencies were discovered with the aid of Intellij IDEA, which is an imperfect tool. So the changes will need careful review and testing.


Joao839:
E

stephengold:
> E

E? Please elaborate.

stephengold:
I don't plan to include this in v3.5, because the valuable changes here (reducing app size at runtime) could also potentially break existing builds. The changes to "test" dependencies are very safe (because they get tested thoroughly by CI) but have correspondingly less benefit (since they don't affect apps).

stephengold:

PR #1715, opened 2021-12-20
rename local variables for clarity

stephengold:
Choosing the wrong name for a local variable can make the code less clear than it ought to be.

Sometimes they're misspelled or abbreviated too far. Sometimes words are run together without using camelCase.


stephengold:
I did a merge to resolve some conflicts, and now the "files changed" tab is empty. Pretty sure that's a bug in GitHub.

Anyway, it makes the PR difficult to review, so I may as well integrate it now.

stephengold:
Integration was a no-op. I'll try to recover my work.

PR #1716, opened 2021-12-21
rename method arguments for clarity

stephengold:
Sometime a formal method argument has a name that obfuscates its purpose, such as a `JmeImporter` named `e` or when words are run together without using camelCase.


stephengold:
Unless there's substantial discussion, I plan to integrate this PR in about 29 hours.

PR #1720, opened 2021-12-24
LwjglCanvas: initialize OpenCL for LWJGL2

AmiroooTheWalnut:
Fixing OpenCL context null issue with LwjglCanvas. It is important that for Swing applications and with this fix, OpenCL won't be available in "simpleInitApp" and everything needs to be initialized inside the "simpleUpdate" with a boolean variable.
The discussion is in the following post:
https://hub.jmonkeyengine.org/t/opencl-context-for-lwjglcanvas-is-always-null/45213/3

### DIFF
--- a/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
+++ b/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
@@ -477,7 +477,10 @@ public class LwjglCanvas extends LwjglAbstractDisplay implements JmeCanvasContex
                 }else{
                     Display.create(acquirePixelFormat(false));
                 }
-
+				if (settings.isOpenCLSupport()) {
+					initOpenCL();
+				}
+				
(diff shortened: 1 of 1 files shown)


stephengold:
The caveat about Swing applications needs to be documented outside the Forum/Hub and this PR.
I'd suggest adding it to the Wiki, but that's such a nightmare these days...

Could you at least document the caveat in the javadoc?

AmiroooTheWalnut:
Sure. I do anything I can to help though I'm not much experienced to work with such big projects. I cancel this pull request and make a new one with javadoc description. Also I try to add things to wiki.

stephengold:
You can amend a PR without cancelling it, simply by pushing more commits to the PR branch. I'll re-open so you can try that if you want...

AmiroooTheWalnut:
Sorry for my late reply. I added a few sentences to the javadoc.

stephengold:
Thank you for your contribution to JMonkeyEngine.

PR #1725, opened 2021-12-27
new javadoc for Quaternion

stephengold:
Despite previous efforts, the `Quaternion` javadoc leaves much to be desired. I plan to clarify what the components represent, what the methods do, which methods modify the object, which arguments can be null, and so on, similar to what I did in PRs #1689 and #1706.


stephengold:
Unless there's substantial discussion, I plan to self-integrate this in about 24 hours.

PR #1726, opened 2021-12-27
fix and test for issue 1723 (more serialization bugs in RenderState)

stephengold:
The solves issue 1723 and adds a JUnit test to check for similar bugs in `RenderState` serialization and/or cloning.


stephengold:

PR #1727, opened 2021-12-31
new javadoc for Matrix3f

stephengold:
The `Matrix3f` javadoc used to be good for a few laughs:
+ "Matrix data is maintained internally"
+ referring to "the provided axis" when 3 axes are provided
+ referring to the current instance as "this matrix4f" and "this instance of Matrix4f"
+ a method claiming to return a null matrix in situations where it actually throws an NPE
+ failing to document the order of the factors for matrix multiplications
+ the `transpose()` method that modifies the current instance
+ the `multLocal()` method that doesn't modify the current instance

I've tried to make the javadoc less ridiculous and more informative.
Future work will include renaming some method arguments.
The good news is that `Matrix3f` isn't used much...


stephengold:

PR #1729, opened 2022-01-04
new javadoc for Vector2f

stephengold:
+ Explain what the class is used for.
+ Consistently refer to its "values" as "components".
+ Conform the javadoc to the project's preferred Java style.
+ Clarify which methods modify the current instance and which don't.
+ Clarify which arguments can be null and which can't.
+ Clarify the order of the factors in cross products.
+ Add javadoc to methods that lacked it.
+ Don't refer to fractions as percentages, which implies they go from 0 to 100!


stephengold:

PR #1731, opened 2022-01-06
Vector2f: save 2 trig calculations in rotateAroundOrigin()

stephengold:
This greatly improves the efficiency of a method ... which is rarely used.


pspeed42:
Hurray for ancient oddly named methods.  Hehe.

stephengold:
Unless there's more to discuss, I plan to integrate this PR in about 24 hours.

PR #1732, opened 2022-01-06
avoid self-closing HTML tags in javadoc

stephengold:
Starting with JDK 8, javadoc regards self-closing HTML tags as errors. This PR corrects 17 such errors.


stephengold:

PR #1733, opened 2022-01-06
deal with unresolved references in javadoc

stephengold:
In javadoc, an unresolved cross reference (`@see` or `@link`) is an error that may suppress javadoc output.

This PR attempts to clear up the remaining unresolved refs. In most cases, I could guess what the author intended. In one case, I deleted the offending reference.


stephengold:

PR #1736, opened 2022-01-09
rename private variables

stephengold:
Improves the readability of Java sourcecode by renaming some private variables, including names where words were over-abbreviated or runtogether without camelCase.


stephengold:
Unless there's something to discuss, I plan to self-integrate in 24 hours.

PR #1737, opened 2022-01-10
correct typos in diagnostic messages

stephengold:
This PR improves the readability of some logger and exception messages, mainly by separating words that were runtogether.


stephengold:
Unless there's something to discuss, I plan to self-integrate this in about 24 hours.

PR #1738, opened 2022-01-11
TerrainQuad:  publicize fixNormals()

stephengold:
Per discussion at the Hub/Forum:  https://hub.jmonkeyengine.org/t/normalrecalccontrol-not-working-outside-of-sdk/44140/6

This makes the `fixNormals()` method public.

### DIFF
--- a/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainQuad.java
+++ b/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainQuad.java
@@ -1601,7 +1601,7 @@ public class TerrainQuad extends Node implements Terrain {
     /**
      * fix the normals on the edge of the terrain patches.
      */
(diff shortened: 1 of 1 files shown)


yaRnMcDonuts:
I believe we also want to publicize the fixNormalEdges() method, I overlooked that when I made my post late last night, but that method is also necessary for generating the normals on the edges of each terrain patch, while fixNormals handles generating the noramls within each individual TerrainPatch.

I also think I found an alternate solution that fixes NormalRecalcControl that I'll submit in another PR after testing a bit more, but I still think it is beneficial to publicize these methods so they can be called directly in situations where you don't need a control constantly checking for updates.

stephengold:
I'm putting this PR on hold until I see your PR.

stephengold:
This PR now publicizes both methods. Unless there's substantial discussion, I plan to integrate this in about 24 hours.

PR #1742, opened 2022-01-12
rename more method arguments for clarity

stephengold:
Similar to #1716.


stephengold:
Unless there's something to discuss, I plan to self-integrate in 24 hours.

PR #1713, opened 2021-12-19
clarify javadoc

stephengold:
Fixes some javadoc errors.
Also corrects capitalization, spelling, noun-verb agreement, and punctuation in javadoc.
Also addresses some awkward phrasing.

PR #1712, opened 2021-12-19
clarify some non-javadoc comments

stephengold:
Corrects punctuation/capitalization/agreement, rewords some awkward phrases, and adds missing space after `//`.

PR #1743, opened 2022-01-12
clarify comments and javadoc

stephengold:
Similar to #1712 and #1713, this addresses comments and javadoc that might be confusing due to issues such as words run together or missing punctuation.


stephengold:
Unless there's something to discuss, I plan to integrate this in about 24 hours.

Issue #1745, opened 2022-01-16
Mouse Coords in LWJGL3 are being multipled by scaling factor on all machines causing coords to be off.


bob0bob:
if you create a window of 100x100 and have a display that reports DPI scaling.  GlfwMouseInput.     private void onCursorPos(final long window, final double xpos, final double ypos) 

this method is multiplying the mouse coord by the monitor scaling even when not being used.  It causes anything based on mouse coords to fail.   Specially GUI, mouse clicks on buttons and such are off by 2.5 times.

stephengold:
Thanks for opening this issue. Please re-test with v3.5.0-beta7.

stephengold:
Default setting for "UseRetinaFrameBuffer" changed from `true` to `false` in "v3.5" branch at c44d74b4 and in "master" branch at a1ae3fd.

stephengold:
I'm working on a PR coming with further changes.

bob0bob:
Retina setting doesn't affect the scaling.  It was added as a "MUST" and not a "IF".

stephengold:
> Retina setting doesn't affect the scaling. It was added as a "MUST" and not a "IF".

I addressed that in PR #1746.

riccardobl:
In mac os HiDPI makes the framebuffer "physically" bigger, making it 2x or 3x the size in pixels of the window, that's the reason for the scaling, but it seems in other systems the hidpi fb is handled transparently and it is made to look as if it had the same number of pixels to the gl api(?).

I think the solution to this issue is to calculate the scaling manually with  glfwGetFramebufferSize/glfwGetWindowSize

PR #1746, opened 2022-01-16
GlfwMouseInput:  scale mouse coords only if Retina AppSetting is true

stephengold:
Proposed fix for #1745.


stephengold:
The fix is somewhat tricky to test. Due to persistent settings and the recent change of default, it's possible to have the Retina flag set to `true` without knowing it. Also, the default LWJGL (in jme3-examples) is still LWJGL v2.

richardTingle (approved):
Looks good to me

stephengold:
Thanks for your help @richardTingle and @bob0bob!

bob0bob:
This fix works for me after getting rid of the "saved" settings,  because JME already set it to "true" by default and re-running brought them back, but when I erased the settings, it worked great.

stephengold:
Thanks, @bob0bob. It looks like this PR will be superseded by #1753 though.

Also, it's integrated into both "master" and "v3.5", so it ought to be closed by now. I wonder why it isn't...

PR #1747, opened 2022-01-16
Configure circular deps as warning on eclipse

riccardobl:
This solves an issue with eclipse based IDEs (eg. vscode with redhat's java plugin) that might mark circular dependencies as errors and stop the project from being imported. 
See image below:
image


stephengold (approved):
Looks good to me!

PR #1749, opened 2022-01-17
Fix error 1280 in MacOS when querying GL_FRAMEBUFFER_SRGB_CAPABLE_EXT

riccardobl:
The line `getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT)` in GLRenderer throw an opengl error 1280 (invalid enum) when the GL debugger is enabled. 
I am not sure why this is the case, however i believe it is safe to assume that every modern macos device is sRGB capable.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
@@ -48,6 +48,8 @@ import com.jme3.scene.VertexBuffer.Usage;
 import com.jme3.shader.*;
 import com.jme3.shader.Shader.ShaderSource;
 import com.jme3.shader.Shader.ShaderType;
+import com.jme3.system.JmeSystem;
+import com.jme3.system.Platform;
 import com.jme3.texture.FrameBuffer;
 import com.jme3.texture.FrameBuffer.RenderBuffer;
 import com.jme3.texture.Image;
@@ -3268,9 +3270,12 @@ public final class GLRenderer implements Renderer {
         setFrameBuffer(null);
 
         if (enableSrgb) {
-            if (!getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT)) {
-                logger.warning("Driver claims that default framebuffer "
-                        + "is not sRGB capable. Enabling anyway.");
+            if (
+                // Workaround: getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT) causes error 1280 (invalid enum) on macos
+                JmeSystem.getPlatform().getOs() != Platform.Os.MacOS
+                && !getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT)
+            ) {
+                logger.warning("Driver claims that default framebuffer " + "is not sRGB capable. Enabling anyway.");
             }
 
             gl.glEnable(GLExt.GL_FRAMEBUFFER_SRGB_EXT);
--- a/jme3-core/src/main/java/com/jme3/system/Platform.java
+++ b/jme3-core/src/main/java/com/jme3/system/Platform.java
@@ -39,109 +39,122 @@ public enum Platform {
     /**
      * Microsoft Windows 32-bit AMD/Intel
      */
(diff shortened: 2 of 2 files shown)


jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -3268,9 +3270,8 @@ public void setMainFrameBufferSrgb(boolean enableSrgb) {
         setFrameBuffer(null);
 
         if (enableSrgb) {
-            if (!getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT)) {
-                logger.warning("Driver claims that default framebuffer "
-                        + "is not sRGB capable. Enabling anyway.");
+            if (JmeSystem.getPlatform() != Platform.MacOSX64 && JmeSystem.getPlatform() != Platform.MacOSX_ARM64 && !getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT)) {
```
stephengold, line 3273: 1. There ought to be a comment here explaining why `MacOSX64` and `MacOSX_ARM64` are exempt from the `getBoolean()` call.
2. Have you seen the 1280 error on both platforms, or just on `MacOSX_ARM64`?
3. If we need to exempt both platforms, then the code would be more readable if you saved the `getPlatform()` result to a local variable and tested that variable, rather than invoking `getPlaform()` twice.

riccardobl: They are the only two architectures that are still supported by apple. And MacOS running on Apple M1 (that is the arm64 processor) can also run x86 64bit apps. 
I exempted only those two because the others are deprecated and i don't know if all the old systems running there support sRGB.
I don't agree with point 3

stephengold: 1. I understand the reason for exempting ARM-based Macs (1280 errors) because I'm reading the code in the context of this PR. But a future developer reading the code without that context won't know about the 1280 errors. So I still think a comment should be added.
2. If you've only seen the 1280 error with `MacOSX_ARM64`, then why do you believe it's necessary to test for `MacOSX64`?

riccardobl: 1. Ok i'll add a comment
2. Because I believe this issue is with the OS and they run the same OS, in any case this only removes a warning, it won't impact with the engine functionality

stephengold: OK.

riccardobl:
I reconsidered this PR, it just removes a warning, so it is better to exclude all the mac os platforms, worst case scenario: the devs won't receive a warning when using sRGB on unsupported deprecated platforms, not a big deal.

There isn't a way to get the generic os name (ie without the arch) from Platform in jme, so i've added that aswell, it will be useful to apply os specific workarounds across the engine.

stephengold:
Another way to determine the OS is to invoke `org.lwjgl.system.Platform`. I'm unsure which is better in the long run.

riccardobl:
We can't use org.lwjgl.system.Platform outside of the jme3-lwjgl3 modules

stephengold:
Regarding "Platform.java", I have some style requests:
+ Please provide javadoc for the new public class and the new public method.
+ Please provide a space character after each new comma, before each new curly brace, and on both sides of each assignment operator.

riccardobl:
done

jme3-core/src/main/java/com/jme3/system/Platform.java
```diff
@@ -151,12 +164,21 @@
     public boolean is64Bit() {
         return is64bit;
     }
+
+    /**
+     * Get the opterating system of this platform
+     * @return the generic name of the operating system of this platform
+     */
+    public Os getOs(){
```
stephengold, line 172: Still needs a space before the `{`.

riccardobl: fixed

stephengold:
Ready for integration, I think.

PR #1750, opened 2022-01-17
Separate concept of window size from default framebuffer size

riccardobl:
This pr makes AppSettings.getWidth/Height return the size of the default framebuffer that is usually what is expected by user code. New methods to set and get window size are added.
The change is backward compatible.

The PR removes also a workaround that caused the engine to be aware of high DPI resolution after the second frame, in my tests waiting two frames has proven to be unnecessary.

EDIT: added rescale listener that propagates the high dpi scaling. Can be used for example to multiply the size of GUI elements.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
+++ b/jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
@@ -66,8 +66,8 @@ public interface SceneProcessor {
      * Called when the scale of the viewport has been changed.
      *
      * @param vp the affected ViewPort
-     * @param x the new scale (in pixels)
-     * @param y the new scale (in pixels)
+     * @param x the new horizontal scale 
+     * @param y the new vertical scale 
      */
     public default void rescale(ViewPort vp, float x, float y){
 
--- a/jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
+++ b/jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
@@ -69,7 +69,7 @@ public interface SceneProcessor {
      * @param x the new horizontal scale 
      * @param y the new vertical scale 
      */
-    public default void rescale(ViewPort vp, float x, float y){
+    public default void rescale(ViewPort vp, float x, float y) {
 
     }
 
--- a/jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
+++ b/jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
@@ -63,6 +63,17 @@ public interface SceneProcessor {
     public void reshape(ViewPort vp, int w, int h);
 
     /**
+     * Called when the scale of the viewport has been changed.
+     *
+     * @param vp the affected ViewPort
+     * @param x the new scale (in pixels)
+     * @param y the new scale (in pixels)
+     */
+    public default void rescale(ViewPort vp, float x, float y){
+
+    }
+
+    /**
      * @return True if initialize() has been called on this SceneProcessor,
      * false if otherwise.
      */
--- a/jme3-core/src/main/java/com/jme3/system/AppSettings.java
+++ b/jme3-core/src/main/java/com/jme3/system/AppSettings.java
@@ -767,8 +767,8 @@ public final class AppSettings extends HashMap<String, Object> {
     /**
      * Set the size of the window
      * 
-     * @param width The width in pixels (default = resolution width)
-     * @param height The height in pixels (default = resolution height)
+     * @param width The width in pixels (default = width of the default framebuffer)
+     * @param height The height in pixels (default = height of the default framebuffer)
      */
     public void setWindowSize(int width, int height) {
         putInteger("WindowWidth", width);
--- a/jme3-core/src/main/java/com/jme3/system/AppSettings.java
+++ b/jme3-core/src/main/java/com/jme3/system/AppSettings.java
@@ -269,6 +269,8 @@ public final class AppSettings extends HashMap<String, Object> {
         defaults.put("CenterWindow", true);
         defaults.put("Width", 640);
         defaults.put("Height", 480);
+        defaults.put("WindowWidth", Integer.MIN_VALUE);
+        defaults.put("WindowHeight", Integer.MIN_VALUE);
         defaults.put("BitsPerPixel", 24);
         defaults.put("Frequency", 60);
         defaults.put("DepthBits", 24);
@@ -735,7 +737,7 @@ public final class AppSettings extends HashMap<String, Object> {
     }
 
     /**
-     * @param value the width for the rendering display.
+     * @param value the width for the default framebuffer.
      * (Default: 640)
      */
     public void setWidth(int value) {
@@ -743,7 +745,7 @@ public final class AppSettings extends HashMap<String, Object> {
     }
 
     /**
-     * @param value the height for the rendering display.
+     * @param value the height for the default framebuffer.
      * (Default: 480)
      */
     public void setHeight(int value) {
@@ -751,7 +753,8 @@ public final class AppSettings extends HashMap<String, Object> {
     }
 
     /**
-     * Set the resolution for the rendering display
+     * Set the resolution for the default framebuffer
+     * Use {@link #setWindowSize(int, int)} instead, for HiDPI display support.
      * @param width The width
      * @param height The height
      * (Default: 640x480)
@@ -761,6 +764,18 @@ public final class AppSettings extends HashMap<String, Object> {
         setHeight(height);
     }
 
+    /**
+     * Set the size of the window
+     * 
+     * @param width
+     *            The width
+     * @param height
+     *            The height (Default: 640x480)
+     */
+    public void setWindowSize(int width, int height) {
+        putInteger("WindowWidth", width);
+        putInteger("WindowHeight", height);
+    }
 
     /**
      * @param value the minimum width the settings window will allow for the rendering display.
@@ -991,7 +1006,7 @@ public final class AppSettings extends HashMap<String, Object> {
     /**
      * Get the width
      *
-     * @return the width of the rendering display (in pixels)
+     * @return the width of the default framebuffer (in pixels)
      * @see #setWidth(int)
      */
     public int getWidth() {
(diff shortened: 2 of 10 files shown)


jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -584,18 +600,18 @@ protected void runLoop() {
...
-
-                if (settings.getWidth() != width[0] || settings.getHeight() != height[0]) {
-                    listener.reshape(width[0], height[0]);
-                }
-            }
-        }
+        // if (framesAfterContextStarted < 2) {
+        //     framesAfterContextStarted++;
+        //     if (framesAfterContextStarted == 2) {
+        //         int[] width = new int[1];
```
stephengold, line 612: Unless there's a plausible reason for this code to be re-enabled in the future, I'd rather you deleted it. Commented-out code without any clear purpose adds to the burden of software maintenance.

riccardobl: Yeah, the comment there says
```
        // Update the frame buffer size from 2nd frame since the initial value
        // of frame buffer size from glfw maybe incorrect when HiDPI display is in use
```
I didn't find this bug to be present, but we might run into it at some point, so i think it's better to leave the supposed fix commented out there.

riccardobl:
Note: I'll need to rebase this PR manually if we decide to merge #1753 since they conflict.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
stephengold: > I didn't find this bug to be present,

The code in question was added by @jmecn. Please talk to them and try to understand why they saw this bug and you don't.

Here's the relevant Forum/Hub topic: https://hub.jmonkeyengine.org/t/switching-to-lwjgl-3/41222

stephengold:
> Note: I'll need to rebase this PR manually if we decide to merge #1753 since they conflict.

We'd better integrate 1753 first then, since I'm hoping to cherry-pick 1753 into the "v3.5" branch.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
riccardobl: I've read the forum post, now it is clear to me that this is a workaround for https://github.com/jMonkeyEngine/jmonkeyengine/issues/1751 ... however it means also that in the first two frames the framebuffer size will be wrong... it might cause unexpected results in user apps.
For now i've updated the PR to add it back, i'll see if i can come up with a different workaround

riccardobl:
rebased

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
riccardobl: I applied a different workaround that doesn't meddle with the initialization flow

stephengold:
I need more time to study this.

stephengold:
1. This PR adds a `rescale()` method to the public `SystemListener` interface. Since there's no default implementation, this is potentially a breaking change. Could we implement this PR with less risk of breaking existing code?
2. I like that you documented `LegacyApplication.rescale()` as "Internal use only". Perhaps do the same for `AndroidHarness`, `AndroidHarnessFragment`, and `JmeSurfaceView`.
3. The javadoc descriptions of the `x` and `y` arguments in `SceneProcessor.rescale()` are identical. Please mention that one is horizontal and the other is vertical. Also, "pixels" isn't the correct unit. Each arg is actually a ratio: frame-buffer pixels per window pixel, right? Also, I have the same concerns aboout `SystemListener.rescale()`.
4. `RenderManager.notifyRescale(float x, float y)` needs javadoc.
5. In the javadoc for `AppSettings.setWindowSize()` please specify units and also document the defaults separately for each arg, i.e. width 640 pixels, height 480 pixels.
6. Why does `AwtPanelsContext.rescale()` throw an `IllegalStateException`? It looks like an unimplemented/unsupported feature, in which case `UnsupportedOperationException` would be more appropriate. Also, the exception should include a message describing what's unsupported (or illegal).
7. I noticed some stylistic errors, which I will enumerate in my next review.

riccardobl:
> This PR adds a rescale() method to the public SystemListener interface. Since there's no default implementation, this is potentially a breaking change. Could we implement this PR with less risk of breaking existing code?

Yes it can be implemented as a default method, they should be supported by android at this point.

> Why does AwtPanelsContext.rescale() throw an IllegalStateException? It looks like an unimplemented/unsupported feature, in which case UnsupportedOperationException would be more appropriate. Also, the exception should include a message describing what's unsupported (or illegal).

I copied the reshape implementation

riccardobl:
Fixed. 
Thanks for catching the problems with the documentation as usual, I never pay enough attention to it.

stephengold:
> I copied the reshape implementation

`AwtPanelsContext.reshape()` is wrong and should be improved (but not in this PR). Please correct and improve `AwtPanelsContext.rescale()`.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -333,6 +339,10 @@ public void invoke(final long window, final boolean focus) {
         setWindowIcon(settings);
         showWindow();
 
+        // HACK: the framebuffer seems to be initialized with the wrong size
+        // on some HiDPI platforms untill glfwPollEvents is called 2 or 3 times
+        for (int i = 0; i < 4; i++) glfwPollEvents();
```
stephengold, line 344: Preferred style would use braces around the body of the loop.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -367,7 +377,31 @@ public void invoke(final long window, final int width, final int height) {
...
+    }
+
+    
+    private void updateDefaultFramebufferSize(boolean updateListener) {
+        // If default framebuffer size is different than window size (eg. HiDPI)
+        int[] width = new int[1];
+        int[] height = new int[1];
+        glfwGetFramebufferSize(window, width, height);
+
+        Vector2f scale=getWindowContentScale(null);
```
stephengold, line 392: Preferred style would have a space on each side of the assignment operator.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -281,11 +282,16 @@ public void invoke(int error, long description) {
 
         final GLFWVidMode videoMode = glfwGetVideoMode(glfwGetPrimaryMonitor());
 
-        if (settings.getWidth() <= 0 || settings.getHeight() <= 0) {
-            settings.setResolution(videoMode.width(), videoMode.height());
+        if (settings.getWindowWidth() <= 0 || settings.getWindowHeight() <= 0) {
+            settings.setWindowSize(videoMode.width(), videoMode.height());
```
stephengold, line 286: Please check whether `videoMode` is null from an error occurring during `glfwGetVideoMode()`.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -333,6 +339,10 @@ public void invoke(final long window, final boolean focus) {
         setWindowIcon(settings);
         showWindow();
 
+        // HACK: the framebuffer seems to be initialized with the wrong size
+        // on some HiDPI platforms untill glfwPollEvents is called 2 or 3 times
```
stephengold, line 343: untill -> until

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -367,7 +377,31 @@ public void invoke(final long window, final int width, final int height) {
...
 
-        framesAfterContextStarted = 0;
+
+        updateDefaultFramebufferSize(false);
+
+    }
+
+    
+    private void updateDefaultFramebufferSize(boolean updateListener) {
+        // If default framebuffer size is different than window size (eg. HiDPI)
```
stephengold, line 387: eg. -> e.g.

jme3-core/src/main/java/com/jme3/system/AppSettings.java
```diff
@@ -761,6 +764,16 @@ public void setResolution(int width, int height) {
         setHeight(height);
     }
 
+    /**
+     * Set the size of the window
+     * 
+     * @param width The width in pixels (default = resolution width)
+     * @param height The height in pixels (default = resolution height)
```
stephengold, line 771: It's not clear what "resolution width" and "resolution height" are. Perhaps you mean the frame-buffer width and height?

jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
```diff
@@ -62,6 +62,17 @@
...
     public void reshape(ViewPort vp, int w, int h);
 
+    /**
+     * Called when the scale of the viewport has been changed.
+     *
+     * @param vp the affected ViewPort
+     * @param x the new horizontal scale 
+     * @param y the new vertical scale 
+     */
+    public default void rescale(ViewPort vp, float x, float y){
```
stephengold, line 72: Preferred style would have a space before the curly brace.

stephengold:
@riccardobl are you still available to work on this PR?

riccardobl:
Sure. I'll fix the issues asap (tomorrow?)

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
riccardobl: This issue is not related to this PR and might need some error handling. It is better to fix it with a commit later

jme3-core/src/main/java/com/jme3/post/SceneProcessor.java
```diff
@@ -62,6 +62,17 @@
      */
     public void reshape(ViewPort vp, int w, int h);
 
+    /**
+     * Called when the scale of the viewport has been changed.
+     *
+     * @param vp the affected ViewPort
+     * @param x the new horizontal scale 
+     * @param y the new vertical scale
```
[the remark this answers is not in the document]
riccardobl: It is just a scale factor, i don't know how to better define it without tying it to the internal logic that might change in the future (or in other renderers).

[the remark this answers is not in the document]
stephengold: What could it be other than "frame-buffer pixels per displayed pixel"?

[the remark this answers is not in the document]
riccardobl: It might be a magnification factor. I am not sure. I wouldn't tie it to the framebuffer.
In the lwjgl3 it is fb/window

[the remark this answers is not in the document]
stephengold: Thanks for that explanation.

PR #1753, opened 2022-01-17
Cross-platform fix for issue 1745 (mouse input scaling)

riccardobl:
Fix issue https://github.com/jMonkeyEngine/jmonkeyengine/issues/1745


bob0bob:
This fix works for me after getting rid of the "saved" settings, because JME already set it to "true" by default and re-running brought them back, but when I erased the settings, it worked great.

stephengold:
I tested this on Windows 11 with display scaling set to 150%, and it worked.

Question: do the `GlfwMouseInputVR` and `LwjglWindowVR` classes in jme-vr need similar modifications?

riccardobl:
It's better to address that in a future PR, because the VR implementation doesn't seem to have any code regarding scaling and high DPI.

stephengold:
In that case, I believe this PR is ready to be integrated into "master".

stephengold:
Thank you, Riccardo!

PR #1754, opened 2022-01-18
Fix issue 1705 (serialize 6 mesh subclasses)

stephengold:
This adds no-arg constructors to 6 `Mesh` subclasses and implements `read()` and `write()` methods as needed.

A few `final` fields had to be made non-final in order for reads to work.

In addition, the automated tests for skeleton-debug mesh serialization (in `TestCloneMesh`) were enabled.


stephengold:
Unless there's something to discuss, I plan to self-integrate this in about 24 hours.

PR #1756, opened 2022-01-19
Add simple math methods to Vector2f

yaRnMcDonuts:


stephengold:
Also looks good. I think this is ready for integration.

stephengold:
Thanks for the contributions!

PR #1757, opened 2022-01-19
Add set methods and constructors to ColorRGBA that take in a Vector param

yaRnMcDonuts:
I've noticed that ColorRGBA has a very useful .toVector4f() method (and also toVector3() method), which are great for instances where you need to do math with a ColorRGBA value returned from an ImageRaster, but there is no equally convenient method to convert that Vector back into a color afterwards, so I thought these 2 methods could be useful (specifically for Vector4f, but I added one for Vector3f too since there was also a toVector3f() method)

I initially thought of adding a new toColorRGBA() method to the Vector classes, but it seemed more appropriate to put this functionality into the ColorRGBA class here with set methods the way I did.

Any thoughts on this addition?

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
+++ b/jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
@@ -203,6 +203,50 @@ public final class ColorRGBA implements Savable, Cloneable, java.io.Serializable
         }
         return this;
     }
+    
+    /**
+     * <code>set</code> sets the values of this <code>ColorRGBA</code> to those
+     * set by a parameter Vector4f.
+     *
+     * @param vec4 The 4 component vector to set this <code>ColorRGBA</code> to.
+     * @return this
+     */
+    public ColorRGBA set(Vector4f vec4) {
+        if (vec4 == null) {
+            r = 0;
+            g = 0;
+            b = 0;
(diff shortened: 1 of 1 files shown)


pspeed42:
I agree with your reasoning on putting these on ColorRGBA.  I often miss these methods myself.

It would be nice if there were constructors, too... but "new ColorRGBA().set(myVec4)" seems like an ok compromise also.

stephengold:
Perhaps the javadoc should specify the correspondences: `r` &larr; `x`, `g` &larr; `y`, `b` &larr; `z`, and `a` &larr; `w`.

stephengold:
@yaRnMcDonuts are you still available to work on this PR?

yaRnMcDonuts:
Yes, my bad for not responding to your last reply, I somehow overlooked it and actually thought this PR got merged with the other similar vector ones I just did. I'm only just realizing that I've been unintentionally deleting some of my github notifications when I press "done" after first opening them and I must have done that with this topic real late when I was tired and forgot.

I changed the wording to mention specifically which component of the vector is copied to which component of the color like you suggested, let me know if it sounds good.

stephengold:
@yaRnMcDonuts Thank you for your latest contribution to JMonkeyEngine!

PR #1758, opened 2022-01-20
Add FOR macro to GLSL preprocessor and J3MD

riccardobl:
This can be used in shaders that need to repeat some code where loops cannot be used (eg. uniforms declaration).
See for example 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-terrain/src/main/resources/Common/MatDefs/Terrain/TerrainLighting.frag#L20-L56

This code can be rewritten like so

```java
#for i=0..12 (#ifdef DIFFUSEMAP_$i $0 #endif)
  uniform sampler2D m_DiffuseMap_$1;
#endfor
```
*Note: the first diffuse map (m_DiffuseMap) will be renamed in m_DiffuseMap0*

As you can see, it makes code much more readable and maintainable.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
@@ -50,6 +50,11 @@ import com.jme3.util.PlaceholderAssets;
 import com.jme3.util.blockparser.BlockLanguageParser;
 import com.jme3.util.blockparser.Statement;
 import com.jme3.util.clone.Cloner;
+
+import jme3tools.shader.Preprocessor;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.*;
@@ -798,6 +803,9 @@ public class J3MLoader implements AssetLoader {
             } else if (key.getExtension().equals("j3md") && key instanceof MaterialKey) {
                 throw new IOException("Material definitions must be loaded via AssetKey");
             }
+
+            in = Preprocessor.apply(in);
+
             loadFromRoot(BlockLanguageParser.parse(in));
         } finally {
             if (in != null){
--- a/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
@@ -50,9 +50,7 @@ import com.jme3.util.PlaceholderAssets;
 import com.jme3.util.blockparser.BlockLanguageParser;
 import com.jme3.util.blockparser.Statement;
 import com.jme3.util.clone.Cloner;
-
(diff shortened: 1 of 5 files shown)


pspeed42:
Does this come up often?

My only hesitation is that this is not compatible with "real" GLSL code.  Similar to our "#import" hack... but that has unarguably been a huge boon to the engine.

For reference, even the C/C++ preprocessor does not have anything like this which is telling in that in 40 years or whatever of C development, they decided cutting and pasting code like this was the better alternative.

It's cool code, though.  I also wonder if shader nodes already had an answer for something like this (I doubt it).

riccardobl:
It comes up often especially if you are trying to build generic shaders.
It is true that it's weird and non standard, but a lot of engines have their own shading language, i don/t think it is a bad idea to add non standard preprocessing directive if they can make things easier.

Another example, here, applying a filter to up to 6 textures at the same time + depth

```glsl
#for i=0..6 ( #ifdef SCENE_$i $0 #endif )
    uniform sampler2D Input$i;
    #ifdef MRT 
        layout(location=$i) 
    #endif 
        out vec4 outScene$i;
#endfor 

#ifdef DEPTH
    uniform sampler2D InputDepth;
#endif

void main(){
    #for i=0..6 ( #ifdef SCENE_$i $0 #endif )
        outScene$i = ...;
    #endfor 

    #ifdef DEPTH
        gl_FragDepth = ...;
    #endif
}
```

yaRnMcDonuts:
This would've definitely been a helpful feature when I was making pbr terrain shaders, I recall trying to figure out if I could define textures in a for loop like this before realizing it was not possible at the time.

If it doesn't break any existent shaders and if users can still code their shaders the normal way, then I'd say it looks like a useful feature to have. And since this PR isn't actually changing any existent shaders to use the feature yet, it seems like the risk of this breaking things is pretty low.

I guess the only question is whether or not there's any significant downside to this like Paul mentioned. It doesn't seem like there would be, but I don't have enough experience to speculate that much, so I suppose the best way to know would be to write 2 versions of the same shader, one with the for loop macro and one the standard way, then compare to make sure there's no significant performance hits or other unexpected issues.

grizeldi:
Sounds like an useful feature to have, it would've definitely come handy in my shaders a couple times by now.

tlf30:
That is a very cool idea. While I have never had the need for it, it looks like it could be useful. 

My only hesitation is that I have never seen anything like it before in any macro based C/C++ precompiler. 

On the other hand, most glsl precompilers have custom macros. For example the engine I am currently working in has
`#injdef` which tells the preprocessor where the engine should inject predefined defines. It also has `#alias` to allow for a variable (such as a uniform, sampler, or push constant) name to be referenced by an aliased name from the engine. So I would say it would not be unexpected for JME to have custom macros as it is common in other engines.

riccardobl:
I think this is mergeable

stephengold:
Thanks for your contribution, Riccardo!

Issue #1761, opened 2022-01-23
Exception in thread "jME3 Audio Decoder" java.lang.AssertionError


wizzardo:
```
Exception in thread "jME3 Audio Decoder" java.lang.AssertionError
	at com.jme3.audio.openal.ALAudioRenderer.updateInDecoderThread(ALAudioRenderer.java:1004)
	at com.jme3.audio.openal.ALAudioRenderer.run(ALAudioRenderer.java:257)
	at java.lang.Thread.run(Thread.java:748)
```

jme: 3.5.0-beta9, os: ubuntu/android
![image](https://user-images.githubusercontent.com/5871626/150684237-5d188460-86ea-47e0-9234-571085b738a4.png)
fails because both statuses are 'Paused'

I got this error multiple times in android emulator, then decided  to reproduce it the desktop version.. had to click play-pause ~50 times before it happened, possible a concurrency issue

stephengold:
Thanks for reporting this issue and providing the stack trace. Do you have any further advice on how to reproduce it?

It's unfortunate that the exception doesn't indicate the values of `oalStatus` and `jmeStatus` ...

wizzardo:
The issue happens (sometimes) when I execute `audioNode.pause()` on a streaming audio, at the point of AssertionError both oalStatus and jmeStatus are `Paused`.

I copied all ALAudioRenderer code and added logging instead of AssertionError.. and so far it looks good, I can see that message in the output, but nothing crashes anymore, so maybe it could be safely removed

stephengold:
Okay, thanks for that info. Your suggested fix sounds plausible to me. It's unfortunate this issue missed the cut-off for 3.5.0-stable, but I expect we'll fix it in the next release.

stephengold:
I managed to reproduce this by modifying `TestMusicStreaming` to repeatedly pause and play the `AudioNode`.

stephengold:
This might be related to #1728, but I don't see how.

stephengold:
I reproduced this issue in the "v3.3" branch, so it's not a regression and definitely not related to #1728.

PR #1762, opened 2022-01-25
solve issue #1761 (AssertionError in ALAudioRenderer)

stephengold:
Delete the offending assertion, which was added in 2014 because someone thought it would be "appropriate".
I tested this fix.


stephengold:

PR #1764, opened 2022-02-01
jme3-nifty should account for gamma correction

stephengold:
This solves an issue with jme3-nifty not taking gamma correction into account. Nifty UIs specify colors by their red, green, and blue components. In order for jme3-nifty to display these colors property, it needs to know whether gamma correction is enabled or not.

If gamma correction is enabled, the application should specify `ColorSpace.sRGB`, in which case `ColorRGBA.setAsSrgb()` will be used to convert Nifty colors to JME colors.

If gamma correction is not enabled, the application should use the old constructor or specify `ColorsSpace.Linear`, in which case `ColorRGBA.set()` will be used to convert Nifty color to JME colors.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/animation/TestCinematic.java
+++ b/jme3-examples/src/main/java/jme3test/animation/TestCinematic.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2022 jMonkeyEngine
  * All rights reserved.
  *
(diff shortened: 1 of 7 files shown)


stephengold:
Here's a screenshot of how `TestIssue99` looked with gamma correction enabled, prior to this change:
linear-gamma-crop

...and here's how it was intended to look (how it looked with gamma correction disabled):
linear-noGamma2-crop

Note how gamma correction made the left side of each scroll panel too bright.

tonihele (approved):
Looks good!

PR #1765, opened 2022-02-01
JoyButtonTrigger:  clarify that the constructor arg is not an axis ID

stephengold:
The sourcecode for the `JoyButtonTrigger` class in jme3-core is confusing because it refers to the 2nd argument of the constructor as an axis ID. It is actually a button ID.

This PR corrects the javadoc and renames the method argument.


stephengold:

PR #1766, opened 2022-02-01
TestPostWater:  avoid using the deprecated BitmapText constructor

stephengold:
Somehow `TestPostWater` was missed when updating jme3-examples to use the new `BitmapText` constructor.


stephengold:
missed in PR #1709, to be specific

stephengold:

PR #1769, opened 2022-02-10
Refactoring of the initialization code of the 'particleMesh' variable.

capdevon:
```java
switch (meshType) {
    case Point:
        particleMesh = new ParticlePointMesh();
        this.setMesh(particleMesh);
        break;
    case Triangle:
        particleMesh = new ParticleTriMesh();
        this.setMesh(particleMesh);
        break;
    default:
        throw new IllegalStateException("Unrecognized particle type: " + meshType);
}
```

This piece of code is repeated 3 times in `ParticleEmitter` class, violating the DRY programming rule. So I encapsulated it in a separate method.

Fixed typo compatability -> compatibility

### DIFF
--- a/jme3-core/src/main/java/com/jme3/effect/ParticleEmitter.java
+++ b/jme3-core/src/main/java/com/jme3/effect/ParticleEmitter.java
@@ -284,6 +284,20 @@ public class ParticleEmitter extends Geometry {
         control = new ParticleEmitterControl(this);
         controls.add(control);
 
+        this.initParticleMesh();
+        this.setNumParticles(numParticles);
+//        particleMesh.initParticleData(this, particles.length);
+    }
+
+    /**
+     * For serialization only. Do not use.
+     */
+    protected ParticleEmitter() {
+        super();
+        setBatchHint(BatchHint.Never);
+    }
+    
+    private void initParticleMesh() {
         switch (meshType) {
             case Point:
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/effect/ParticleEmitter.java
```diff
@@ -284,6 +284,20 @@ public ParticleEmitter(String name, Type type, int numParticles) {
...
+//        particleMesh.initParticleData(this, particles.length);
+    }
+
+    /**
+     * For serialization only. Do not use.
+     */
+    protected ParticleEmitter() {
+        super();
+        setBatchHint(BatchHint.Never);
+    }
```
stephengold, line 298: Why did you move the protected constructor?

capdevon: I inserted the method after the default constructor. I don't know why github points out the differences so weird for this point. If you see the whole file you can check that everything is ok.

stephengold: GitHub's diff algorithm is unhelpful sometimes, and I guess this is one of those times. Sorry for my confusion!

stephengold:
Thanks for your contribution to our project, @capdevon!

PR #1771, opened 2022-02-12
SimpleApplication:  correct 2 typos in the javadoc

stephengold:
In both cases, "applications" is in the possessive case, so an apostrophe is required for clarity.


stephengold:

PR #1772, opened 2022-02-16
InstancedGeometry: added a field to keep track of all instance data b…

Ali-RS:
…uffers (transformInstanceData and globalInstanceData) so we do not need to recreate it every frame which was not good performance-wise.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2022 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
@@ -67,7 +67,7 @@ public class InstancedGeometry extends Geometry {
     private VertexBuffer transformInstanceData;
     private Geometry[] geometries = new Geometry[1];
     // Keep track of both transformInstanceData and globalInstanceData
-    // going to be used by renderer.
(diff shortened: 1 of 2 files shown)


Ali-RS:

jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
```diff
@@ -383,14 +389,18 @@ protected void updateWorldBound() {
 
     @SuppressWarnings("unchecked")
```
stephengold, line 390: I think you can re-enable this warning.

jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
```diff
@@ -383,14 +389,18 @@ protected void updateWorldBound() {
...
     public VertexBuffer[] getAllInstanceData() {
-        ArrayList<VertexBuffer> allData = new ArrayList();
+        return allInstanceData;
+    }
+
+    private void updateAllInstanceData() {
```
stephengold, line 395: I think the `read()` method needs to invoke `updateAllInstanceData()` in order to ensure that `allInstanceData` gets initialized.

jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
```diff
@@ -66,6 +66,9 @@
     private VertexBuffer[] globalInstanceData;
     private VertexBuffer transformInstanceData;
     private Geometry[] geometries = new Geometry[1];
+    // Keep track of both transformInstanceData and globalInstanceData
+    // going to be used by renderer.
```
stephengold, line 70: Why "going to be"?

stephengold (changes requested):

jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
Ali-RS: Done.

Ali-RS:
@stephengold thanks for reviewing.

PR #1770, opened 2022-02-12
AppSettings:  solve 2 javadoc diagnostics

stephengold:
AppSettings.java contains 2 javadoc tags referencing methods that don't exist. This PR corrects the references.

PR #1774, opened 2022-02-18
avoid use of deprecated methods (mostly Class.newInstance())

tonihele:
Removes deprecated code and some warnings.

Compiled with JDK 11 (-Xlint:deprecation)...

- Class.newInstance() was deprecated in Java 9, I replaced it with Class.getDeclaredConstructor().newInstance(); (get empty constructor and call it)
- new Float() is deprecated, unnecessary boxing
- new JsonParser() is deprecated, use the static methods instead

### DIFF
--- a/jme3-android/src/main/java/com/jme3/app/AndroidHarness.java
+++ b/jme3-android/src/main/java/com/jme3/app/AndroidHarness.java
@@ -190,6 +190,7 @@ public class AndroidHarness extends Activity implements TouchListener, DialogInt
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public void onCreate(Bundle savedInstanceState) {
         initializeLogHandler();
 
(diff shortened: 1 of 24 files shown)


stephengold:
"Remove deprecated code" is a bit misleading. Most of the changes seem to be replacing `Class.newInstance()` invocations with alternative code that isn't deprecated. Please edit the title of this PR.

The corrections to `AppSettings` javadoc are unrelated and will be handled by PR #1770. To avoid potential merge conflict, please revert them from this PR.

tonihele:
Doesn't seem to show those as merge conflicts, so auto-merge will probably go just fine. However if this still bothers, I can try tomorrow to revert the file, seems best to use some GIT magic since I can't revert a whole commit (too big commits), just the file in the commit.

I adjusted the PR title and also added a better description of the changes. I also cleaned up the Android project which I was a bit hesitant to do yesterday, since the Java versions on Android.... But these methods are present all in Java 8 at least so I guess everything is fine in that regard.

The code base is surprisingly fine in the deprecation regard for a project of this size. Seems to be mostly now internal deprecations about the animation system and the framebuffer change.

stephengold:
@tonihele do you believe this PR is ready to be integrated?

tonihele:
> @tonihele do you believe this PR is ready to be integrated?

I think so. I didn't have the chance to revert the file you asked. But it looks to me that it will go just fine, without any conflict. So there should be no need to revert the file.

stephengold:
Thanks for your contribution, @tonihele!

PR #1777, opened 2022-02-23
Remove print statements from FBX plugin

wyskoj:
This changes a print statement in `FbxAnimCurveNode.java` to a logger call so it doesn't contaminate output.

### DIFF
--- a/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/anim/FbxAnimCurveNode.java
+++ b/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/anim/FbxAnimCurveNode.java
@@ -103,9 +103,9 @@ public class FbxAnimCurveNode extends FbxObject {
      */
     public Quaternion getQuaternionValue(long time) {
         Vector3f eulerAngles = getVector3Value(time);
-        logger.fine("\tT: " + time + ". Rotation: " +
-                                eulerAngles.x + ", " + 
-                                eulerAngles.y + ", " + eulerAngles.z);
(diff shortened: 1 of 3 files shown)


stephengold:
Thank you for proposing this change.

The FBX loader contains 13 print statements directed to `System.out`. Why does this particular print statement cause an issue?

Does the issue relate to `FbxDump.dumpFile()` or is there some other impact?

wyskoj:
I spent some time resolving warnings logged by JME in my project, and noticed I kept getting this print statement. So this was the only one that affected me, but if you prefer, I could do the same to the rest of the occurrences.

wyskoj:
I know it's not a warning. In the process of resolving warnings, I spotted the output of this print statement in my console. And it doesn't help me to have it there (and frankly I'd like to have it gone). And by making it a `fine` call, it can still be seen—if wanted—by setting the logger level.

So, yes, that is the the only impact—just me being picky on the cleanliness of my console :)

stephengold:
Thank you for your patience. Now that I understand the issue/motivation, I'm OK with this PR as is.

However, I notice that the straightforward conversion of `println()` to `Logger.fine()` results in some inefficient code: converting floats to strings, concatenating them, and then (in the usual case) not using the result. If you're agreeable, I suggest simply commenting out the original `println()`, similar to what was done in `FbxLoader.constructAnimations()`.

wyskoj:
Yes, I do agree. An alternative would be to define the string lazily. But I don't suspect these statements are getting enough attention to warrant implementing that. I'll comment them out.

stephengold (approved):
Looks good. Thanks for your cooperation and thoroughness!

PR #1779, opened 2022-03-04
jme-lwjgl3: config GLFW using AppSettings and org.lwjgl.system.Platform

stephengold:
This a step toward allowing JME applications developed for Linux/Windows to run on macOS without any changes. Since LWJGL v2, doesn't support the newer "Apple Silicon" hardware, the focus is on LWJGL v3. For portability, we want to use the "glfw_async" library (added in LWJGL v3.3.1) when running on macOS. To reduce the risk of regressions, we'll use the old library when running on other operating systems.

For extra flexibility, a new `AppSetting` named "GlfwLibraryName" specifies which library to use: "glfw" or "glfw_async". If this setting is unset or null, `org.lwjgl.system.Platform` is used to determine which library to use.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -64,6 +64,8 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.lwjgl.system.Configuration;
+import org.lwjgl.system.Platform;
 
 import static org.lwjgl.glfw.GLFW.*;
 import static org.lwjgl.opengl.GL11.GL_FALSE;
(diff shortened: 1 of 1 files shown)


riccardobl:
I think it would be better to set Configuration.GLFW_LIBRARY_NAME only if the libraryName is != null , so that the default becomes whatever LWJGL3 think the default is.

stephengold:
I'm converting this to a draft PR both because of the suggestion from @riccardobl and also because I'm unsure the best place to set "glfw_async".

stephengold:
This PR is still on hold because I've been investigating issues with "glfw_async". I'm convinced it's unreliable in LWJGL v3.3.1, causing certain apps to crash (in a highly reproducible fashion) with "BUG IN CLIENT OF LIBDISPATCH: dispatch_sync called on queue already owned by current thread". The same apps work OK with ordinary "glfw" and -XstartOnFirstThread. When I have time, I hope to create a compact test case to send to the LWJGL and/or GLFW folks.

riccardobl:
We replaced glfw with SDL3, so i think this can be closed.

Issue #1780, opened 2022-03-08
RendererException in TerrainLighting.frag


stephengold:
5th generation Mac Mini (Apple Silicon)
OpenJDK 64-Bit Server VM Temurin-17.0.2+8 (build 17.0.2+8, mixed mode)
JME v3.5.0-stable
`setRenderer(AppSettings.LWJGL_OPENGL32)`

```console
Mar 07, 2022 10:10:34 PM com.jme3.renderer.opengl.GLRenderer updateShaderSourceData
WARNING: Bad compile of:
1       #version 110
2       #define SRGB 1
3       #define FRAGMENT_SHADER 1
4       #define WARDISO 1
5       #define DIFFUSEMAP 1
6       #define DIFFUSEMAP_1 1
7       #define DIFFUSEMAP_2 1
8       #define NORMALMAP 1
9       #define NORMALMAP_1 1
10      #define NORMALMAP_2 1
11      #define ALPHAMAP 1
12      #define DIFFUSEMAP_0_SCALE 64.0
13      #define DIFFUSEMAP_1_SCALE 16.0
14      #define DIFFUSEMAP_2_SCALE 128.0
15      // -- begin import Common/ShaderLib/BlinnPhongLighting.glsllib --
16      /*Blinn Phong lighting*/
17      
18      /*
19      * Computes diffuse factor (Lambert)
20      */
21      float lightComputeDiffuse(in vec3 norm, in vec3 lightdir){
22          return max(0.0, dot(norm, lightdir));
23      }
24      
25      /*
26      * Computes specular factor   (blinn phong) 
27      */
28      float lightComputeSpecular(in vec3 norm, in vec3 viewdir, in vec3 lightdir, in float shiny){
29          vec3 H = normalize(viewdir + lightdir);
30          float HdotN = max(0.0, dot(H, norm));
31          return pow(HdotN, shiny);
32      }
33      
34      /*
35      *Computes standard phong specular lighting
36      */
37      float lightComputeSpecularPhong(in vec3 norm, in vec3 viewdir, in vec3 lightdir, in float shiny){
38          vec3 R = reflect(-lightdir, norm);
39          return pow(max(dot(R, viewdir), 0.0), shiny);
40      }
41      
42      
43      /*
44      * Computes diffuse and specular factors and pack them in a vec2 (x=diffuse, y=specular)
45      */
[...]

PR #1781, opened 2022-03-09
solve issue #1780 (built-in matdefs/shaders lack core-profile support)

stephengold:
Added "GLSL150" to matdefs and `#import "Common/ShaderLib/GLSLCompat.glsllib"` to shaders. Not certain I got them all.


stephengold:
I verified that this PR solves issue 1780 on macOS Monterey.
In my opinion this would be worth including in JME 3.5.1 if we decide to go that route.

stephengold:
Optimistically interpreting the thumbs-up emojis as positive code reviews, I plan to self-integrate this PR in about 24 hours ... unless there's something to discuss.

PR #1790, opened 2022-03-29
Add GL debug capabilities

riccardobl:
This PR implements debug capabilities added by the KHR_debug extension that permit to label objects and group calls in OpenGL.
This data is picked up by graphics debuggers such as RenderDOC and used to improve their output, see screenshots below

Screenshot from 2022-03-29 11 14 21
Screenshot from 2022-03-29 11 14 28

This PR is only for the LWJGL3 renderer, the new methods are implemented as default methods of Renderer interface and will do nothing on other renderers.

The feature is disabled by default and can be enabled by setting GraphicsDebug to true in AppSettings.


jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -2995,6 +3034,9 @@ public void setVertexAttrib(VertexBuffer vb, VertexBuffer idb) {
                 attribs[slot] = vb.getWeakRef();
             }
         }
+        if (debug && caps.contains(Caps.GLDebug)) {
+            if (vb.getName() != null) glext.glObjectLabel(GLExt.GL_BUFFER, vb.getId(), vb.getName());
```
Trass3r, line 3038: It should suffice to do this once right after creating the GL object (glGenBuffers). Then you also have less chance of missing something, e.g. index buffers.

PR #1792, opened 2022-03-29
Add Color param to SkyBox

yaRnMcDonuts:
Currently, JME scenes with a skybox can't achieve a realistic day/night cycle because the AmbientLight and DirectionalLight do not affect the skyBox's brightness or color.

This small addition would allow basic skyboxes to work with day/night cycles, as well as any other special effects that could benefit from changing the distant sky color/brightness.

If this addition sounds like a good idea to core devs and is approved, I can add the rest of the code for the new color param to the .jj3md files and make any other adjustments so its mergable.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Misc/Sky.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Misc/Sky.frag
@@ -3,10 +3,12 @@
 
 uniform ENVMAP m_Texture;
 
+uniform vec4 m_Color;
+
 varying vec3 direction;
 
 void main() {
     vec3 dir = normalize(direction);
-    gl_FragColor = Optics_GetEnvColor(m_Texture, dir);
+    gl_FragColor = Optics_GetEnvColor(m_Texture, dir).mult(m_Color);
 }
 


riccardobl:
I think you typoed `.mult` instead of `*`.

I am fine with this addition, but will this work with common skyboxes? I am not entirely convinced since they are usually pretty complex. Can you show how your skybox looks and how you are using this to simulate day/night cycle?

yaRnMcDonuts:
> I think you typoed `.mult` instead of `*`.

Yes my mistake, I just typed up a short draft to display the general idea and I confused it with the java vector multiplication code.

I think for some nature-based SkyBoxes with mostly trees, clouds, and mountains this will work well, but you're right it won't work for complex skyboxes, or ones with that appear to be lit from sources that should still be lit when its night.  

And while using a sky box like this definitely isn't as good for day/night as some other sky options like SkyControl or the other dynamic cloudy sky libraries, I figure that having something in the basic SkyBox for day/night is better than nothing. I like those other options better for clouds, but I also like how custom SkyBoxes can be uniquely made to match the zone, so I'm still switching between different sky options for different scenarios.

An alternative idea that would have better visual results for complex skyboxes would be to have a second cube map and interpolate between the two based on a value sent into the shader, so you could edit the skybox textures to make an alternate set of textures adjusted for night time (add stars, adjust things like sky-scraper windows to look like the lights are on, etc).

yaRnMcDonuts:
The results with just scaling based on a color value are much worse than I expected, even with a generic set of cloudy sky textures that I thought would do well. So you are right that this isn't a great way to do realistic day/night.

I also tried the second idea of making a second cubeMap that has versions of the sky textures adjusted for night. This was going well and theoretically works, but I stopped after I finished making the adjusted set of sky textures and previewed them in JME, because it created noticeable seams between the textures of the sky box.  Even though I applied the same edits to all 6 textures, it seems as though some blurring happened on the edges with some of the color editing effects I used did in Gimp to make the night textures. 

So I think this may not be the best idea to try and use a basic skybox for anything dynamic like day or night, its not necessarily easy to edit existing skybox textures, and the results are still a sky that doesn't move and just changes colors. So maybe its better to just use a more advanced sky library and forgo trying to make skybox more advanced.

yaRnMcDonuts:
Here were the results with interpolating between 2 cubemaps in the shader. It's not bad for a basic skybox, but it's not great for the effort it took to modify all 6 sky textures, especially considering this was a really simple skybox and I still struggled to edit the colors to be darker without creating even more visual artifacts.

https://www.youtube.com/watch?v=5siVvlPOkaU&ab_channel=RyanMcDonough

riccardobl:
Maybe the initial idea can be improved by adjusting saturation and brightness(/value) instead of multiplying by a color?
image


If you want to give it a try, for convenience, I post here the code that you can use to convert from rgb to hsv and back 
```glsl
vec3 rgb2hsv(vec3 c){
    vec4 K = vec4(0.0, -1.0 / 3.0, 2.0 / 3.0, -1.0);
    vec4 p = mix(vec4(c.bg, K.wz), vec4(c.gb, K.xy), step(c.b, c.g));
    vec4 q = mix(vec4(p.xyw, c.r), vec4(c.r, p.yzx), step(p.x, c.r));

    float d = q.x - min(q.w, q.y);
    float e = 1.0e-10;
    return vec3(abs(q.z + (q.w - q.y) / (6.0 * d + e)), d / (q.x + e), q.x);
}

vec3 hsv2rgb(vec3 c){
    vec4 K = vec4(1.0, 2.0 / 3.0, 1.0 / 3.0, 3.0);
    vec3 p = abs(fract(c.xxx + K.xyz) * 6.0 - K.www);
    return c.z * mix(K.xxx, clamp(p - K.xxx, 0.0, 1.0), c.y);
}```

yaRnMcDonuts:
I will give that a try as well whenever I have a chance, and see how the results look. That sounds more promising than scaling just the brightness like I initially did. 

Scaling by a single value would also be a lot easier than my second idea, since it wouldn't require editing and preparing a second set of clean sky textures. 

I think I would like the results of interpolating between 2 cubeMaps if it weren't the fact that I cannot find any image editor that can easily edit 6 textures in a cubemap all at the same time without creating artifacts on the edges.

pspeed42:
For a game that doesn't want to implement a real sky, it might be best to handle this with a hard transition.  Popup a message on the screen "Night has fallen" while fading the whole non-UI screen to black, switch the skybox, then fade in again.

Else, a dynamic sky is probably the only good way.

stephengold:
@yaRnMcDonuts It was a clever idea. Thanks for implementing it and testing it. I'm sorry it didn't pan out.

Issue #1793, opened 2022-03-31
missing reshape() callbacks in LWJGL v3


stephengold:
I expanded `TestContextRestart` in the "master"-branch jme3-examples sub-project to study some issues I'm seeing with context restarts in LWJGL v3. The issues are platform-dependent and reproducible.

Running on current "master"-branch with LWJGL v3 on my WIndows 11 laptop, `reshape()` is invoked just once:
```console
Mar 30, 2022 5:59:43 PM com.jme3.system.JmeDesktopSystem initialize
INFO: Running on jMonkeyEngine 3.6.0-SNAPSHOT
 * Branch: master
 * Git Hash: 30774ef
 * Build Date: 2022-03-30
Mar 30, 2022 5:59:45 PM com.jme3.system.lwjgl.LwjglContext printContextInitInfo
INFO: LWJGL 3.3.1 build 7 context running on thread jME3 Main
 * Graphics Adapter: GLFW 3.4.0 Win32 WGL Null EGL OSMesa VisualC DLL
Mar 30, 2022 5:59:45 PM com.jme3.renderer.opengl.GLRenderer loadCapabilitiesCommon
INFO: OpenGL Renderer Information
 * Vendor: NVIDIA Corporation
 * Renderer: NVIDIA GeForce RTX 2070/PCIe/SSE2
 * OpenGL Version: 3.2.0 NVIDIA 497.33
 * GLSL Version: 1.50 NVIDIA via Cg compiler
 * Profile: Core
Mar 30, 2022 5:59:47 PM com.jme3.audio.openal.ALAudioRenderer initOpenAL
INFO: Audio Renderer Information
 * Device: OpenAL Soft
 * Vendor: OpenAL Community
 * Renderer: OpenAL Soft
 * Version: 1.1 ALSOFT 1.21.1
 * Supported channels: 64
 * ALC extensions: ALC_ENUMERATE_ALL_EXT ALC_ENUMERATION_EXT ALC_EXT_CAPTURE ALC_EXT_DEDICATED ALC_EXT_disconnect ALC_EXT_EFX ALC_EXT_thread_local_context ALC_SOFT_device_clock ALC_SOFT_HRTF ALC_SOFT_loopback ALC_SOFT_loopback_bformat ALC_SOFT_output_limiter ALC_SOFT_pause_device
[...]

PR #1794, opened 2022-03-31
work around missing reshape() callbacks from LWJGL v3 (issue #1793)

stephengold:
Based on the observations documented in #1793, the key issue is that transitions between windowed mode and fullscreen mode in LWJGL v3 don't always generate `reshape()` callbacks. This spoils GUI layouts that need updating each time the dimensions of the graphics context change. The user sees a layout generated for a different-sized framebuffer.

(There's also an issue with the initial `reshape()` callback, which occurs on Windows, but not Linux, when the initial context is created. However, that's an easy issue for apps to work around.)

The key issue manifests differently on Linux than on Windows:
+ On Linux LWJGL v3, windowed-to-fullscreen causes a callback, but not fullscreen-to-windowed.
+ On Windows LWJGL v3, neither transition causes a callback.

This PR proposes that jme3-lwjgl3's  `LwjglWindow` should explicitly invoke `reshape()` on both windowed-to-fullscreen and fullscreen-to-windowed transitions. This results in some redundant callbacks (multiple callbacks for a single `createContext()`). However, I expect most `reshape()` callbacks perform idempotent operations. Also, filtering out redundant callbacks is much easier that compensating for missing callbacks.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2022 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -141,6 +141,7 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
 
     private Thread mainThread;
 
+    private boolean isFullscreen = false;
     private long window = NULL;
     private int frameRateLimit = -1;
 
@@ -357,7 +358,7 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
                 for (WindowSizeListener listener : windowSizeListeners.getArray()) {
                     listener.onWindowSizeChanged(width, height);
                 }
-                updateDefaultFramebufferSize(true);
+                updateDefaultFramebufferSize(true, true);
             }
         });
 
@@ -366,7 +367,7 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
 
             @Override
             public void invoke(final long window, final int width, final int height) {
-                updateDefaultFramebufferSize(true);
+                updateDefaultFramebufferSize(false, true);
             }
         });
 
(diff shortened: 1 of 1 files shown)


stephengold:
Here's a summary:
 + deal with all size/scale changes in a single method, now named `updateSizes()`
 + keep a private record of the framebuffer size, so `updateSizes()` won't be confused by unrelated changes to `AppSettings`

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -349,16 +349,7 @@ public void invoke(final long window, final boolean focus) {
...
             public void invoke(final long window, final int width, final int height) {
-
-                // This is the window size, never to passed to any pixel based stuff!
-                // https://www.glfw.org/docs/latest/window_guide.html#window_size
-                onWindowSizeChanged(width, height);
-
-                // Notify listeners
-                for (WindowSizeListener listener : windowSizeListeners.getArray()) {
-                    listener.onWindowSizeChanged(width, height);
-                }
```
Ali-RS, line 359:
```
// Notify listeners
for (WindowSizeListener listener : windowSizeListeners.getArray()) {
    listener.onWindowSizeChanged(width, height);
}
```

Is this removed by mistake?

stephengold: Good catch! I was supposed to move that loop into `updateSizes()` but somehow it got dropped. I'll fix.

stephengold:
Unless there's more to discuss, I propose to self-integrate this PR in about 24 hours.

stephengold:
In order to backport this into "v3.5" branch, we'd also need to backport #1750. The combination of the 2 PRs seems risky relative to the benefits. Better wait for 3.6 or 4.0

Ali-RS:
Looks like this broke TestAWTPanels. See #1945

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -281,18 +285,13 @@ public void invoke(int error, long description) {
...
 
         final GLFWVidMode videoMode = glfwGetVideoMode(glfwGetPrimaryMonitor());
-
-        if (settings.getWindowWidth() <= 0 || settings.getWindowHeight() <= 0) {
-            settings.setWindowSize(videoMode.width(), videoMode.height());
-        } else {
-            settings.setWindowSize(settings.getWindowWidth(), settings.getWindowHeight());
+        int requestWidth = settings.getWindowWidth();
+        int requestHeight = settings.getWindowHeight();
+        if (requestWidth <= 0 || requestWidth <= 0) {
```
Ali-RS, line 290: `if (requestWidth <= 0 || requestWidth <= 0)`

I think should be 

 `if (requestWidth <= 0 || requestHeight <= 0)`

Ali-RS: Fixed in #1950

stephengold: Good catch!

PR #1796, opened 2022-04-04
refactor to avoid invoking AppSettings.getBoolean() directly

stephengold:
Invoking `AppSettings.getBoolean()` directly is error-prone. Ideally, its use would be limited to custom settings.

This PR refactors some invocations to use special-purpose methods already defined in the `AppSettings` class.


stephengold:
Unless there's more to discuss, I propose to self-integrate this PR in about 24 hours.

PR #1797, opened 2022-04-06
Add java types to VarType and type checks to MatParam

riccardobl:
This PR adds the matching java types to VarTypes and uses them to perform runtime type checks on MatParam.setValue().
Should be merged after https://github.com/jMonkeyEngine/jmonkeyengine/pull/1759


stephengold:
I think this is a good idea. I'm curious why the checks are enabled dynamically.

riccardobl:
While not ideal, i suspect there might be some community code that sets wrong types or changes VarType at runtime (being it protected not final),  since this (even if wrong) would technically work in some circumstances,  i wanted to provide a way to disable the type check without requiring a recompilation of the engine in case this causes existing code to break.

It might be superfluous ... i am not sure

Issue #1809, opened 2022-05-05
A multithreading issue with GltfLoader


foxcc2021:
I use multithreading to load gltf models, when loaded, all model materials are messed up.

After I debug GltfLoader, I found out that there may be  threading issue in the GltfLoader. The variable "defaultMaterialAdapters" cannot be "static". This variable will be shared when all threads load materials from the model.

A simple way to resolve this issue is to remove the "static".
Hope this helps.
I use jme version: 3.5.2-stable
![20220505173134](https://user-images.githubusercontent.com/93635496/166900456-62948438-1335-4563-99c0-0cdada11880f.jpg)

![resolve](https://user-images.githubusercontent.com/93635496/166900714-56e25c54-caf5-4ccd-a1ed-106435b68acc.jpg)
I use this way to resolve my problem.

![err0](https://user-images.githubusercontent.com/93635496/166901053-864bf782-4d8d-4a99-bc30-da51fa42adfa.jpg)
![err1](https://user-images.githubusercontent.com/93635496/166901072-d6d99275-fb64-488c-a350-b8b85237ff03.jpg)
The wrong materials before

![correct](https://user-images.githubusercontent.com/93635496/166901114-9cb068e3-c7a4-42c4-a25b-31d50eeae9db.jpg)
The correct materials after

stephengold:
Thanks for the bug report and suggested fix.

stephengold:
By the way ... creating threads to load models is good trick for reducing latency in a game. Another way to reduce latency is to convert the models to JMonkeyEngine's preferred format (j3o) at build time and then load the j3o instead of the glTF. And of course one can combine both tricks.

The expectation is that the glTF loader won't be used at runtime in a production game. Perhaps that's why the original author of `GltfLoader` didn't worry about thread safety.

Ali-RS:
Apparently, CustomContentManager which handles Gltf extension lading is also not thread-safe:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/CustomContentManager.java#L54-L60

Ali-RS:
A fix is submitted: PR #1886

Ali-RS:
Merged in commit https://github.com/jMonkeyEngine/jmonkeyengine/pull/1886/commits/92f99ef17e614507e473f157e321d4e06d324751

PR #1810, opened 2022-05-06
solve issue #1809 (GltfLoader is not thread-safe)

stephengold:
In order to ensure initialization of the `defaultMaterialAdapters` map, I created a no-arg constructor for the loader. Otherwise this is a straightforward fix.


stephengold:

PR #1811, opened 2022-05-07
Allow registering default material adapters in GltfLoader

Minikloon:
I want to use Gltf but I don't want to use PBR, so I made this adapter: https://gist.github.com/Minikloon/93728e7ad494373b70753c072e36a77a

Now I don't want to register it on every model key!

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -89,15 +89,20 @@ public class GltfLoader implements AssetLoader {
     private Vector3fArrayPopulator vector3fArrayPopulator = new Vector3fArrayPopulator();
     private QuaternionArrayPopulator quaternionArrayPopulator = new QuaternionArrayPopulator();
     private Matrix4fArrayPopulator matrix4fArrayPopulator = new Matrix4fArrayPopulator();
-    private static Map<String, MaterialAdapter> defaultMaterialAdapters = new HashMap<>();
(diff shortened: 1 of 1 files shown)


Minikloon:
rip conflict with https://github.com/jMonkeyEngine/jmonkeyengine/pull/1810

pspeed42:
Do you load GLTF directly into your game or do you use the recommended and much smarter approach of converting it to j3o first?  Because it's quite common to have to fix a dozen things during the gltf->j3o conversion process and this could in theory just be another standard thing to fix during conversion.

For example, JmeConvert allows you to specify scripts to post-process the model during conversion and it would be a pretty standard script to convert all materials to something other than PBR.

Minikloon:
I saw from JmeConvert that you use the adapters map of the GltfModelKey. I could've used that but then I'd want a ModelKeyFactory.

If I wanted to setup an art pipeline rn, I'd override com.simsilica.jmec.Convert::runProcessors and run java code there. But even then it's post-processing. Hooking into the loading dodges having to fully understand the jme layout.

pspeed42:
re: "I'd override com.simsilica.jmec.Convert::runProcessors and run java code there."

If you are calling it from code then you can already run Java code there and if you are calling it from the command line then you already can use groovy which is like Java++.

I'm not against fleshing out the GltfLoader API but in the long run you will be happier with an asset pipeline that gets you to a j3o file that is ready to load, properly scaled, cleaned up, generated j3m files, etc... rather than constantly hacking in "one more thing" at load time.

stephengold:
@minikloon Why did you close this PR?

Minikloon:
I'll re-open if I work more with jMonkey.

stephengold:
But why not just leave it open until then?

Minikloon:
I don't like PRs in limbo, they make me worry that I will have to make changes _someday_.

stephengold:
good to know

Issue #1798, opened 2022-04-13
filtered scenes are squeezed by resizable windows


stephengold:
I've been converting apps to use resizable windows. Mostly this just involves invoking `AppSettings.setResizable()` and adjusting the GUI layout each time the window gets resized.

I've encountered an issue involving filters in resizable windows, and I believe it's a bug in the Engine. When I resize a window to a new aspect ratio, the rendered scene gets squashed to "fit" the new aspect ratio. This only happens when there are filters. In an unfiltered scene, no squashing occurs.

Here's a simple testcase based on `TestCartoonEdge`:

```java
package jme3test.post;

import com.jme3.app.SimpleApplication;
import com.jme3.light.DirectionalLight;
import com.jme3.material.Material;
import com.jme3.math.ColorRGBA;
import com.jme3.math.FastMath;
import com.jme3.math.Quaternion;
import com.jme3.math.Vector3f;
import com.jme3.post.FilterPostProcessor;
import com.jme3.post.filters.CartoonEdgeFilter;
import com.jme3.scene.Geometry;
import com.jme3.scene.Node;
import com.jme3.scene.Spatial;
import com.jme3.scene.Spatial.CullHint;
import com.jme3.system.AppSettings;
import com.jme3.texture.Texture;

public class TestIssueXXXX extends SimpleApplication {

    private FilterPostProcessor fpp;

    public static void main(String[] args) {
        AppSettings s = new AppSettings(true);
        s.setResizable(true);
        TestIssueXXXX app = new TestIssueXXXX();
        app.setSettings(s);
        app.start();
    }

    private void setupFilters() {
        fpp = new FilterPostProcessor(assetManager);
        int numSamples = getContext().getSettings().getSamples();
        if (numSamples > 0) {
            fpp.setNumSamples(numSamples);
        }
        CartoonEdgeFilter toon = new CartoonEdgeFilter();
        toon.setEdgeColor(ColorRGBA.Yellow);
        fpp.addFilter(toon);
        viewPort.addProcessor(fpp);
    }

    private void makeToonish(Spatial spatial) {
        if (spatial instanceof Node) {
            Node n = (Node) spatial;
            for (Spatial child : n.getChildren()) {
                makeToonish(child);
            }
        } else if (spatial instanceof Geometry) {
            Geometry g = (Geometry) spatial;
            Material m = g.getMaterial();
            if (m.getMaterialDef().getMaterialParam("UseMaterialColors") != null) {
[...]

PR #1812, opened 2022-05-08
solve issue #1798 (filtered scenes are squeezed)

stephengold:
I've verified this fix on Linux using the `TestIssue1798` app in jme3-examples. I also tested with no filters in the FPP and without the FPP to ensure I did not break those situations.


stephengold:
Now I've also tested all 3 situations on macOS.

stephengold:
I've now tested all 3 cases on Windows 11 as well.

Unless there's discussion, I intend to integrate this PR in about 26 hours.

PR #1814, opened 2022-05-08
Add Quaternion::rotateTowards

Minikloon:
It's basically this: https://docs.unity3d.com/ScriptReference/Quaternion.RotateTowards.html
I honestly don't know much about how Quaternions work, I took the code from: https://gist.github.com/aeroson/043001ca12fe29ee911e

Two things I don't know:
* The linked implementation uses slerpUnclamped. There's no such method in jMonkey. My test code worked with just clamp.
* I don't know if the angle() method requires the quaternions to be normalized.

I copied the naming convention that slerp uses, but I think they should be rotateTowards/rotateTowardsLocal (and slerp/slerpLocal) to match the other methods in the maths package.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Quaternion.java
+++ b/jme3-core/src/main/java/com/jme3/math/Quaternion.java
@@ -1429,7 +1429,7 @@ public final class Quaternion implements Savable, Cloneable, java.io.Serializabl
      * See {@link #rotateTowardsLocal(Quaternion, float)} for details
      *
      * @param initial the initial rotation
-     * @param target the desired rotation
+     * @param target the desired rotation (may be modified)
      * @param maxRadDelta the maximum angular step taken during interpolation
      * @return the (modified) current instance (for chaining)
      */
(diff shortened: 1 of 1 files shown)


stephengold:
I'm skeptical whether someone who doesn't understand quaternions should be contributing code to the `Quaternion` class. Setting that aside for now, I have a few questions:

1. What are the intended use cases for the new methods?
2. What tests have you run to convince you that the methods are reliable?
3. What happens if a large negative value (such as -6) is passed for `maxRadDelta'?

Other notes:
+ The 2-argument `rotateTowards()` should be renamed `rotateTowardsLocal()` because it modifies the current instance. (The `slerp()` method is, unfortunately, a bad example.)
+ The `angle()` method should be renamed `angleBetween()` to match the analogous `Vector3f` method.
+ JME's quaternion slerp method sometimes modifies the target. Hence, so does `rotateTowards()`, and this oddity should be documented.
+ The angles are in radians, and this detail should be documented.
+ As written, the `angle()` method is accurate only for unit (normalized) quaternions. Hence, the other new methods are also limited to unit quaternions, and this limitation should be documented.

Minikloon:
My use case is that I have a character and a third-person chase camera.
I'm adding a feature to auto-adjust the camera to go behind the character and I want the camera to move at constant speed.

I will have to get back to you on tests.

> * The 2-argument `rotateTowards()` should be renamed `rotateTowardsLocal()` because it modifies the current instance. (The `slerp()` method is, unfortunately, a bad example.)

Actually renamed both, cause they both modify the current instance.

Issue #1813, opened 2022-05-08
GltfLoader loses animation data


stephengold:
Forum thread: https://hub.jmonkeyengine.org/t/incomplete-animation/45558

I've tracked this down to the `TrackData.update()` method. For the "character.gltf" model's "stand" animation, the track for node 4 ("all") has 3 translation keyframes prior to `TrackData.update()` but only 1 keyframe after the invocation.

`timeArrays.get(0)` is rotation data, for which there is a single keyframe at t=0.
`timeArrays.get(1)` is translation data, for which there are 3 keyframes: at t=0, t=2.25, and t=4.25.

Those last 2 keyframes never get merged into the `keyFrames` list, so translation data is lost.

stephengold:
"TrackData.java" implements a merge sort using spaghetti code with poor encapsulation. But in case nobody feels like rewriting it from scratch, here is a straightforward patch that I believe will solve this issue:

```patch
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TrackData.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TrackData.java
@@ -90,7 +90,14 @@
                         KeyFrame kf = keyFrames.get(j);
                         if (Float.floatToIntBits(kf.time) != Float.floatToIntBits(time)) {
                             if (time > kf.time) {
-                                continue;
+                                if (j < keyFrames.size() - 1) {
+                                    // Keep searching for the insertion point.
+                                    continue;
+                                }
+                                kf = new KeyFrame();
+                                kf.time = time;
+                                // Add kf after the last keyframe in the list.
+                                keyFrames.add(kf);
                             } else {
                                 kf = new KeyFrame();
                                 kf.time = time;
```

PR #1815, opened 2022-05-08
solve issue #1813 (glTF loader loses keyframes)

stephengold:
This PR implements the "straightforward patch" solution to issue #1813.


Minikloon:
I tested this change and it does fix my issue, very cool!

stephengold:
> I tested this change and it does fix my issue, very cool!

I'm sure you weren't the first person to be affected by #1813, but you were the first to report it with enough details for it to be reproduced. Thank you!

It's a nasty bug, an old bug, and it might otherwise have persisted for years. I'm very grateful for your diligence.

stephengold:
Unless there's substantial discussion, I intend to integrate this PR in about 24 hours.

PR #1817, opened 2022-05-10
re-organize jme3-jbullet sourcecode

stephengold:
In the wake of PR #1474, the source code for the jme3-jbullet library remains split between 2 directories:  jme3-jbullet/src/main/java/com/jme3/bullet and jme3-bullet/src/common/java/com/jme3/bullet

The jme3-jbullet library shows no signs of dying, and the meanwhile the split has proven confusing and inconvenient. I therefore propose to merge the "common" source files into the jme3-jbullet directory.


MeFisto94:
On a similar note, we may want to integrate jbullet directly, as a submodule? into the repository? or at least in some kind of pipeline, since we're going to have to fix jbullet bugs as well and normen found our version

stephengold:
> we may want to integrate jbullet directly

The difficulty is that I haven't managed to build jbullet without using Ant. I know there are ways to integrate Ant tasks into Gradle builds, but I haven't made them work for this particular case. In order to publish a jbullet library, I build the source jar, javadoc jar, pom, and module using Gradle, but build the class jar using Ant.

stephengold:
Barring further discussion, I plan to self-integrate this PR in about 24 hours.

Issue #1816, opened 2022-05-09
Typo in method name 'TrackData.checkTimesConsistantcy'


capdevon:
Should be 'Consistency'.

I haven't had a chance to verify the source code, but if it's only used in the `TrackData` class, the method visibility could change to `private`

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TrackData.java#L173

stephengold:
It's not used anywhere else in the Engine. However, it's the sort of utility method that an application might want to use, so we should probably create a new method and deprecate the old one.

I'll take care of it.

PR #1818, opened 2022-05-10
jme3-plugins:  rename the TrackData.checkTimesConsistantcy() method

stephengold:
to address issue #1816

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TrackData.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TrackData.java
@@ -187,6 +187,7 @@ public class TrackData {
 
     @Deprecated
     public void checkTimesConsistantcy() {
+        checkTimesConsistency();
     }
 
     private void populateTransform(Type type, int index, List<KeyFrame> keyFrames, KeyFrame currentKeyFrame, TransformIndices transformIndices) {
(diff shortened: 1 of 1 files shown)


jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TrackData.java
```diff
@@ -165,19 +165,30 @@ public void update() {
...
+     * {@link #times}, {@link #translations}, {@link #rotations}, and
+     * {@link #scales} vectors all have the same length, if present.
+     *
+     * @throws AssetLoadException if the lengths differ
+     */
+    public void checkTimesConsistency() {
         if ((translations != null && times.length != translations.length)
                 || (rotations != null && times.length != rotations.length)
                 || (scales != null && times.length != scales.length)) {
             throw new AssetLoadException("Inconsistent animation sampling ");
```
capdevon, line 184: space at the end of the error message is needed?

stephengold: can be removed

stephengold: However, I think removal shouldn't be part of this PR.

stephengold:
Unless there's more to discuss, I plan to integrate this PR in about 24 hours.

PR #1820, opened 2022-05-16
Update the Gradle wrapper version to 6.9.2

tonihele:
I think you already upgraded Gradle to 6.9.2 but this part was left out.


stephengold (commented):
Yes.

Issue #1678, opened 2021-11-23
Google Play Store Console pre-launch report shows errors due to usage of non-SDK API classes


digitprop:
When submitting an Android JME app in the Google Play Store, for target API level 30 and higher, the Play Store Console reports multiple errors. These are due to AndroidBufferAllocator accessing fields of java.nio.ByteBufferAsFloat and similar (ByteBufferAsChar, ByteBufferAsInt, etc). These classes are not part of the SDK-approved Android API, and as such their use is not allowed for API levels 30 and higher.

This is the stack trace:

```
StrictMode policy violation: android.os.strictmode.NonSdkApiUsedViolation: Ljava/nio/ByteBufferAsCharBuffer;->bb:Ljava/nio/ByteBuffer;
	at android.os.StrictMode.lambda$static$1(StrictMode.java:407)
	at android.os.-$$Lambda$StrictMode$lu9ekkHJ2HMz0jd3F8K8MnhenxQ.accept(Unknown Source:2)
	at java.lang.Class.getDeclaredField(Native Method)
	at com.jme3.util.AndroidBufferAllocator.<clinit>(AndroidBufferAllocator.java:70)
	at java.lang.Class.classForName(Native Method)
	at java.lang.Class.forName(Class.java:454)
	at java.lang.Class.forName(Class.java:379)
	at com.jme3.util.BufferAllocatorFactory.create(BufferAllocatorFactory.java:25)
	at com.jme3.util.BufferUtils.<clinit>(BufferUtils.java:66)
	at com.jme3.util.BufferUtils.createIntBuffer(BufferUtils.java:926)
	at com.jme3.renderer.android.AndroidGL.<init>(AndroidGL.java:47)
	at com.jme3.system.android.OGLESContext.initInThread(OGLESContext.java:205)
	at com.jme3.system.android.OGLESContext.onSurfaceCreated(OGLESContext.java:184)
	at android.opengl.GLSurfaceView$GLThread.guardedRun(GLSurfaceView.java:1560)
	at android.opengl.GLSurfaceView$GLThread.run(GLSurfaceView.java:1285)
```

This is the static initializer of AndroidBufferAllocator where the error occurs:

```
    static {
        for (String className : wrapperClassNames) {
            try {
                Class clazz = Class.forName(className);

                // loop for all possible field names in android
                for (String fieldName : possibleBufferFieldNames) {
                    try {
[...]

Issue #1674, opened 2021-11-23
ReflectionAllocator is broken on JDK 16


Ali-RS:
`SEVERE: Buffer cannot be destroyed: java.nio.DirectFloatBufferU`

Even by adding this JVM arg (which used to solve that on jdk 11+)

"--add-opens=java.base/jdk.internal.ref=ALL-UNNAMED"

you can test it with

https://github.com/jMonkeyEngine/jmonkeyengine/blob/d76d4dc8c67bc99c5fd197d41a102639194a4e3b/jme3-examples/src/main/java/jme3test/app/TestReleaseDirectMemory.java#L43

Forum post:
https://hub.jmonkeyengine.org/t/jme-on-jdk-16/44411/6?u=ali_rs

Ali-RS:
We might consider deprecating ReflectionAllocator for the 3.5 release.

stephengold:
Deprecating `ReflectionAllocator` is another idea that should be discussed at the Forum/Hub ASAP.

Ali-RS:
Sorry @stephengold, I thought I have deleted that comment. I guess it's to late to delete it now but I will modify it to sense a bit nice;)

You are right deperecation is another topic that should be discoussed on forum.

Regards

stephengold:
Here's the diagnostic message I got when I ran `TestReleaseDirectMemory` on 64-bit Linux using OpenJDK 16.0.1: 
```text
Dec 08, 2021 12:55:58 PM com.jme3.system.JmeDesktopSystem initialize
INFO: Running on jMonkeyEngine 3.5.0-SNAPSHOT
 * Branch: master
 * Git Hash: 645b1bb
 * Build Date: 2021-12-08
Inconsistency detected by ld.so: dl-lookup.c: 111: check_match: Assertion `version->filename == NULL || ! _dl_name_match_p (version->filename, map)' failed!
```

stephengold:
Using OpenJDK 14.0.2, I get additional diagnostic messages:
```text
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.jme3.util.ReflectionAllocator (file:/home/sgold/Git/jmonkeyengine/jme3-core/build/libs/jme3-core-3.5.0-SNAPSHOT.jar) to method sun.nio.ch.DirectBuffer.cleaner()
WARNING: Please consider reporting this to the maintainers of com.jme3.util.ReflectionAllocator
[...]

PR #1821, opened 2022-05-20
Android: Implemented AndroidNativeBufferAllocator - Deprecated AndroidBufferAllocator

Scrappers-glitch:
This PR targets this issue #1678 and adds these : 
1) Deprecation of `AndroidBufferAllocator`.
2)  Update jme3-copyright.


Ali-RS:
Thanks @Scrappers-glitch

Ali-RS:
@Scrappers-glitch, unfortunately, I am not qualified to review this PR but I have a few questions to ask.

Have you tested this PR on an android device?
Did you precompile natives or they will be generated by GitHub CI?
Which ARM architectures are supported? 

Thanks

jme3-android-native/src/native/jme_bufferallocator/com_jme3_util_AndroidNativeBufferAllocator.c
```diff
@@ -0,0 +1,85 @@
...
+    LOGI("Buffer released (mem_address, size) -> (%p, %lu)", buffer, sizeof(buffer));
+    // avoid accessing this memory space by resetting the memory address
+    buffer = NULL;
+    LOGI("Buffer mem_address formatted (mem_address, size) -> (%p, %lu)", buffer, sizeof(buffer));
+}
+
+JNIEXPORT jobject JNICALL Java_com_jme3_util_AndroidNativeBufferAllocator_createDirectByteBuffer
+(JNIEnv * env, jobject object, jlong size)
+{
+    void* buffer = malloc(size);
```
riccardobl, line 75: The jvm specification requires buffers to be zeroed.
We should either have a memset that sets the memory to 0s or use calloc instead of malloc here.

riccardobl (changes requested):

riccardobl:
Thank you @Scrappers-glitch . 

We need to merge this before continuing to work on https://github.com/jMonkeyEngine/jmonkeyengine/issues/1674 , https://hub.jmonkeyengine.org/t/buffer-allocator-implementation-set-by-lwjglcontext-gets-ignored-sometimes/45594/4

Scrappers-glitch:
> Have you tested this PR on an android device?

@Ali-RS yep on my physical device (Xiaomi Redmi Note 6 pro running on LineageOS-18), the test on the WIP screenshot, i tested the `TestInstancedNode` testcase and debugged using android logs.

> Did you precompile natives or they will be generated by GitHub CI?

I have NDK within the SDK locally on my machine, so i just changed the `ndkPath` on the gradle.properties file to my ndk path and the `compileAndroidNatives` to true that is locally on my laptop so i can test it, but for jme3-android-native yep it will be generated by the CI.

> Which ARM architectures are supported?

All those ABIs (Application Binary Interfaces) are supported  : 
1) [armeabi-v7a](https://developer.android.com/ndk/guides/abis#v7a)
2) [arm64-v8a](https://developer.android.com/ndk/guides/abis#arm64-v8a)
3) [x86](https://developer.android.com/ndk/guides/abis#x86) for desktop android_x86 (and emulators).
4) [x86_64](https://developer.android.com/ndk/guides/abis#86-64) for desktop android_x86 (and emulators).

By doing this at the `Application.mk` : 
`APP_ABI := all`

I have left respective links for the resources i used for this PR, navigate them if there is something you want to understand.

jme3-android/src/main/java/com/jme3/util/AndroidBufferAllocator.java
```diff
@@ -40,6 +40,7 @@
 
 /**
  * @author Jesus Oliver
+ * @deprecated implemented {@link AndroidNativeBufferAllocator} instead.
  */
 public class AndroidBufferAllocator implements BufferAllocator {
```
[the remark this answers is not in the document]
Scrappers-glitch: Ops, i reverted this by wrong while removing my test changes from `AndroidBufferAllocator` (as i was adding some logging to this class to see its behavior), thanks for pointing this, appreciate that.

jme3-android-native/src/native/jme_bufferallocator/com_jme3_util_AndroidNativeBufferAllocator.c
Scrappers-glitch: I am not sure if i am fully understanding you, so we will do these steps after the basic `malloc(size_t size)`: 

- if (`buffer != 0`) then ` memset (buffer, 0, size);`.
- OR just replace the whole thing with a single line : `void* buffer = calloc (1, size)`.

Which one do you think is better ?

Scrappers-glitch: I suggest using calloc from stdlib, much cleaner and reliable.

pspeed42:
Yes, calloc.

In my hardcode C programming days, malloc was always looked on with suspicion... too often where malloc goes so does memory bugs.

jme3-android-native/src/native/jme_bufferallocator/com_jme3_util_AndroidNativeBufferAllocator.c
pspeed42: Yes, calloc.

In my hardcode C programming days, malloc was always looked on with suspicion... too often where malloc goes so does memory bugs.

jme3-android/src/main/java/com/jme3/util/AndroidNativeBufferAllocator.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+
+/**
+ * Allocates and destroys direct byte buffers using native code.
+ *
+ * @author pavl_g.
+ */
+public final class AndroidNativeBufferAllocator implements BufferAllocator {
```
Scrappers-glitch, line 42: I am wondering, following our discussion on forum which states unifying a class name for all buffer allocators across the different modules and call them in a static initializer, so do i need to change the class name to `NativeBufferAllocator` so it will be `com.jme3.util.NativeBufferAllocator` ?

jme3-android-native/src/native/jme_bufferallocator/com_jme3_util_AndroidNativeBufferAllocator.c
Scrappers-glitch: Alright, i will go with `calloc(1, size)`, thanks.

jme3-android/src/main/java/com/jme3/util/AndroidNativeBufferAllocator.java
riccardobl: Nah, it is better to keep it this way in this pr.
We will do the renaming or add wrapper classes for this and the other allocators in a pr that deals specifically with the other issue.

jme3-android-native/src/native/jme_bufferallocator/com_jme3_util_AndroidNativeBufferAllocator.c
Scrappers-glitch: Oh i forgot this part, i think we should implement our own `calloc()`, so we can check if the malloc is zero and this means the memory is exhausted and the program should throw a fatal error (exiting with ENOMEM error code) describing that the device is out of memory :

```c
#include <errno.h>

void* xcalloc (size_t size) {
  void* buffer = malloc (size);
  // sanity check the buffer
  if (buffer == 0) {
     LOGE("Error device is out of memory, cannot allocate (%p, %ul), buffer, size);
     LOGE("Exiting with error code %i ", errno);
     exit(errno);
   } else {
     memset (buffer, 0, size);
     LOGI("Buffer created successfully (%p, %ul)", buffer, size);
   }
  return value;
}
```
Are we okay with this change ? 

Generally, combining those : 
- https://www.gnu.org/software/libc/manual/html_node/Allocating-Cleared-Space.html
- https://www.gnu.org/software/libc/manual/html_node/Malloc-Examples.html

pspeed42: Pretty classic problem, though... how are you going to log anything if you can't allocate?

...why switch back to malloc, though?

Scrappers-glitch: > Pretty classic problem, though... how are you going to log anything if you can't allocate?
> 
> ...why switch back to malloc, though?

1) This was an example from GNU C docs...but it was using fatal(..) which is printf or perror, so could use perror instead but i am not sure if it will show the error for jni.

2) I haven't switched back to malloc, instead i implemented calloc locally as stated by GNU docs : 
```c
You could define calloc as follows:

void *
calloc (size_t count, size_t eltsize)
{
  void *value = reallocarray (0, count, eltsize);
  if (value != 0)
    memset (value, 0, count * eltsize);
  return value;
}
```

riccardobl: The code allocates memory only here, so I think there is no reason to create the xcalloc function since checks can be done here in the only place where they are needed.
Also i think calloc should be used anyway, since the code you posted is only one possible implementation, platforms could implement it differently in their stdlib

pspeed42: Yes, optimal to use calloc instead of rolling your own probably-degraded version.

Scrappers-glitch: I just want to handle memory errors in some logs, suggest me a better solution if you know, something like, may be `perror()` is sufficient : 

[Code] : 
```c
#include <stdio.h>
#include <errno.h>
#include <stdlib.h>
#include <string.h>

void* xcalloc (size_t size) {
  void* buffer = malloc (size);
  // sanity check the buffer
  if (buffer == 0) {
     perror("Exiting with error code");
     exit(errno);
   } else {
     memset (buffer, 0, size);
     printf("Buffer created successfully (%p, %lu)", buffer, size);
   }
  return buffer;
}

int main() {
    void* buffer = xcalloc(50000000000000000*50000000000000000);
    return 0;
}

```
[Debug logs] : 
```c
Exiting with error code: Cannot allocate memory
```

Scrappers-glitch: > The code allocates memory only here

I haven't tested what will happen if the device is out of memory (eg. allocating multiple large buffers at the same time)....the crash should be obvious that is the application is halted because we are out of memory...but at least from my opinion we should handle memory errors properly for future debugging purposes, of note, emulators are small memory virtual devices so expect to have these kinds of problems especially on emulators.

riccardobl: I think the logging is fine, this will be probably  triggered when the app tries to allocate a very large memory area, so there will probably be enought memory available for the logging.

Scrappers-glitch: Alright, i think this may be better if you insist to use calloc, what do you think ? : 
```c
#include <errno.h>
#include <stdlib.h>
#include <stdbool.h>

bool isDeviceOutOfMem(void* buffer) {
   return buffer != NULL && errno == ENOMEM;
}

void* buffer = calloc(1, size);
if (isDeviceOutOfMem(buffer)) { 
   perror("Device is out of memory");
   exit(errno);
} else { 
  LOGI("Buffer created successfully %p %ul ", buffer, size);
}
```

jme3-android/src/main/java/com/jme3/util/AndroidBufferAllocator.java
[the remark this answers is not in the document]
Scrappers-glitch: Done

Scrappers-glitch:
Ops, `@deprecated` mentioned someone by wrong :-).

jme3-android-native/src/native/jme_bufferallocator/com_jme3_util_AndroidNativeBufferAllocator.c
Scrappers-glitch: Tested this and the perror shows nothing on android, `ANDROID_LOG_FATAL` should be used instead, i will implement this approach if there is no objection.

Scrappers-glitch: This what the test shows, on regular run : 
```java
....
I/AndroidNativeBufferAllocator: Buffer released (mem_address, size) -> (0x78c08ab320, 8)
    Buffer mem_address formatted (mem_address, size) -> (0x0, 8)
    Buffer created successfully (mem_address, size) -> (0x78d0948130 65536)
    Buffer released (mem_address, size) -> (0x78c08c3950, 8)
    Buffer mem_address formatted (mem_address, size) -> (0x0, 8)
    Buffer created successfully (mem_address, size) -> (0x78d0938120 65536)
I/AndroidNativeBufferAllocator: Buffer released (mem_address, size) -> (0x78c08b3530, 8)
....
```

When setting buffer = NULL and errno = ENOMEM explicitly to mimic no memory : 
```java
I/JmeSurfaceView: Starting a new Game State
I/JmeSystem: Running on jMonkeyEngine 3.4.0-SNAPSHOT
     * Branch: android-native-bytebuffer
     * Git Hash: f46d469
     * Build Date: 2022-05-21
I/OGLESContext: Android Build Version: {0}
I/JmeSurfaceView: Game returns from the idle mode
I/JmeSurfaceView: Game goes idle
I/JmeSurfaceView: Context stops, but game is still running
I/JmeSurfaceView: Starting a new Game State
I/OGLESContext: Android Build Version: {0}
I/JmeSurfaceView: Game returns from the idle mode
I/AdrenoGLES: QUALCOMM build                   : f2ab992, I401605978b
    Build Date                       : 09/28/19
    OpenGL ES Shader Compiler Version: EV031.27.05.01
    Local Branch                     : 
    Remote Branch                    : 
    Remote Branch                    : 
    Reconstruct Branch               : 
    Build Config                     : S L 8.0.11 AArch64
I/AdrenoGLES: PFP: 0x005ff11a, ME: 0x005ff066
I/Gralloc4: mapper 4.x is not supported
W/Gralloc3: mapper 3.x is not supported
I/JmeSurfaceView: JmeSurfaceView's joined the UI thread
I/JmeSurfaceView: JmeSurfaceView's joined the UI thread
A/AndroidNativeBufferAllocator: Device is out of memory
```

riccardobl (approved):

riccardobl:
Perfect, unless there are objections I will merge this PR tomorrow.
Thank you.

Scrappers-glitch:
Btw, this PR should close the issue #1678, Could you please link them, thank you.

Issue #1825, opened 2022-05-29
AreaUtils is in a wrong package


Scrappers-glitch:
I noticed `AreaUtils` class is located inside `com.jme3.scene.control`, proposed fix I guess is to deprecate this class and create a new one under `com.jme3.util`.

PR #1826, opened 2022-05-29
AreaUtils: Migrated package to `com.jme3.util`

Scrappers-glitch:
This PR includes: 
- Migrating the `AreaUtils` class from `com.jme3.scene.control` to `com.jme3.util`.
- Deprecating `com.jme3.scene.control.AreaUtils`.
- Jme3 copyright update.

EDIT: 
- Targeting this issue #1825

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/control/AreaUtils.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/AreaUtils.java
@@ -61,6 +61,6 @@ public class AreaUtils {
     * @return The area in pixels on the screen of the bounding volume.
     */
     public static float calcScreenArea(BoundingVolume bound, float distance, float screenWidth) {
-        return com.jme3.utils.AreaUtils.calcScreenArea(bound, distance, screenWidth);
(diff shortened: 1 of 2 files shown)


riccardobl:
To avoid code duplication, could you make the methods in the old deprecated class to call the methods in the new one?

Scrappers-glitch:
> To avoid code duplication, could you refactor the old deprecated class to keep only the public method and make it call the one in the new class?

Hello, yeah i will consider that.

Scrappers-glitch:
Done, @riccardobl you could review it now.

Ali-RS:
I think the new class should not call the deprecated class, it should be vice-versa.

Scrappers-glitch:
> I think the new class should not call the deprecated class, it should be vice-versa. I think this is also what @riccardobl meant.

Oh, i must have misunderstood this, okay i am going to refactor this.

Ali-RS:
>There is a small question though, should i move the private methods from the deprecated class to the new one ?

Yes.

Scrappers-glitch:
Here we go, you can review this now.

Ali-RS:
Looks good to me, thanks

PR #1829, opened 2022-06-27
fix Transform.toString() method and formatting the TranslucentBucketFilter class code

capdevon:


stephengold:
Changing the `toString()` format might break existing code.

capdevon:
How is such a thing possible? Could you give me an example where the code could break?

tonihele:
Probably not break any code internally, or if that is the case, I would be glad it happens and that can be fixed :) `toString()` should not be used for anything more that i.e. logging purposes. But there is always someone using it on logic level. Illegal and should go straight to jail but what can you do.... Changes are probably very low here but of course they exist.

stephengold:
> How is such a thing possible? Could you give me an example where the code could break?

Yes, I could. But first I'd like to understand the benefit of changing the `toString()` format. What's the motivation there?

pspeed42:
Personally, I don't see a problem with changing toString() output to fix things.  Anyone relying on specific toString() formatting has to know they are on shaky ground.  It's not generally part of any contract and in this case the javadoc is already effectively lying about the output anyway.

stephengold:
@pspeed42 I don't see where the javadoc is lying. Please clarify.
@capdevon Calling it a "clear mistake" isn't sufficient to persuade me. You removed 4 blanks from the output, making it more compact. I don't see any other benefit, and you haven't presented any. In my mind, that slight compaction doesn't justify the risk of breaking apps. Please revert that portion of the PR.

stephengold:
> maybe it's time to make the repository read-only.

@pspeed42 It's not that time. What's happening here is I'm trying to weigh benefits against the risks, both minuscule in this case.
I respect your wisdom and experience. If you ever get the urge to approve/integrate JME PRs, I'd welcome your assistance.

stephengold:
> The javadoc heavily implies a decimal formatting (4 decimal places, not whole numbers for rotation, etc.) but in fact it will be whatever random format that the float toString

If the javadoc is misleading, that's unintentional. How would you improve the javadoc?

capdevon:
I honestly didn't think my change request was so risky and absurd. We are all programmers with years of experience; it is okay to be cautious and evaluate everything carefully, but we all know that anyone would correct that method by removing the spaces. 

I like the engine and all the people behind it. I don't want to fight or disrespect anyone. I'm just trying to help. 

The discussion got stuck on a hypothesis not yet proven: could you please show a concrete example of an application that could break?  Could it be useful to do a survey on the forum to quantify the percentage of risk?

stephengold:
@capdevon I don't see any reason for a Forum survey. The opinions that matter are those of people with write access to the repo. If pspeed or Ali-RS is willing to integrate this PR without my approval, I won't complain.

You say you don't want to fight. Neither do I. So let's collaborate!

If I provide "proof" that this PR is potentially a breaking change, will you make the changes I requested on 29 June, without further argument? Or will you use my labor as a springboard for your next argument?

capdevon:
okay, let's cooperate in a friendly and peaceful way as we always have.

stephengold:
Please consider this code: https://github.com/jMonkeyEngine/jmonkeyengine/blob/37788c50b437487b474d2e554fe6467dd13955be/jme3-core/src/test/java/com/jme3/math/TestTransform.java#L49-L69

capdevon:
At the moment I don't have a fast connection and I can't download the branch (more than 1 Gb) to reset the changes to the `Transform.toString()` method. I synced the branch to the latest master changes via github, and now obviously the changes don't pass the software checks (probably due to the `TestTransform` class?). 

Do you prefer that I close the PR or leave it open in case someone else reports and tries to correct the same problem?

If so, I'll post a separate new PR for formatting the `TranslucentBucketFilter` class code as soon as I get the chance.

stephengold:
Let's close this PR. If you want to track the `toString()` format issue, create a GitHub issue for it.

PR #1830, opened 2022-06-27
Better implementation of the AnimationMask interface

capdevon:
- Homogeneous signature of methods.
- Possibility of chaining methods.
- The 'Armature' is passed only once in the constructor.
- 'Joints' are passed to methods as strings instead of objects.


capdevon:
Any feedback on this proposal? Can it be useful or should I close it?

stephengold:
I think it's worth considering, but my attention has been elsewhere, so I haven't reviewed this in depth. If someone else could review it, that would be great.

Ali-RS:
I am in doubt if we should add this into the engine or not. And if it will be confusing to have two identical implementations that one keeps the Armature instance and one does not.

I prefer core members decide about this and if they are positive about this addition then I can continue reviewing the code.

capdevon:
I agree with you Ali, two identical implementations in core package, would just mess up. The `AnimationMask` interface gives programmers the freedom to write their own functionality and that's perfectly fine. If there is nothing else to discuss, I expect to close this PR within 24 hours.

PR #1838, opened 2022-07-25
Update README.md to include Exotic Matter

entrusc:
Would be great to have our game Exotic Matter also mentioned as its engine is based on jME. Thanks!


Ali-RS:
Sure, thanks for the PR. Glad to see another game made with JME. Wish you the best with your project.

PR #1842, opened 2022-08-08
Fix #1843 (java.util.zip.ZipException in HttpZipLocator)

### LINKED ISSUE #1843: java.util.zip.ZipException in HttpZipLocator
```
assetManager.registerLocator(
                    "https://www.icyboxs.com/icyboxs/Lina.zip",
                    HttpZipLocator.class);

 Node  model =  (Node)assetManager.loadModel("Stand.gltf");
```

Fails with this error:

```
com.jme3.asset.AssetLoadException: An error occurred loading Stand.gltf
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:181)
	at com.jme3.scene.plugins.gltf.GltfLoader.load(GltfLoader.java:106)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at net.jmecn.HelloArmature.model(HelloArmature.java:321)
	at net.jmecn.HelloArmature.initScene(HelloArmature.java:257)
	at net.jmecn.HelloArmature.simpleInitApp(HelloArmature.java:114)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:139)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:221)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: com.google.gson.JsonIOException: java.util.zip.ZipException: invalid distance too far back
	at com.google.gson.internal.Streams.parse(Streams.java:62)
	at com.google.gson.JsonParser.parse(JsonParser.java:84)
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:123)
	... 10 more
Caused by: java.util.zip.ZipException: invalid distance too far back
	at java.base/java.util.zip.InflaterInputStream.read(InflaterInputStream.java:164)
	at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:270)
	at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:313)
	at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:188)
	at java.base/java.io.InputStreamReader.read(InputStreamReader.java:177)
	at com.google.gson.stream.JsonReader.fillBuffer(JsonReader.java:1295)
	at com.google.gson.stream.JsonReader.nextNonWhitespace(JsonReader.java:1333)
	at com.google.gson.stream.JsonReader.consumeNonExecutePrefix(JsonReader.java:1576)
	at com.google.gson.stream.JsonReader.doPeek(JsonReader.java:534)
	at com.google.gson.stream.JsonReader.peek(JsonReader.java:425)
	at com.google.gson.internal.Streams.parse(Streams.java:46)
	... 12 more
```

Link to the file: https://www.icyboxs.com/icyboxs/lina.zip

Ali-RS:
Fix #1843

Reported on the forum: 
https://hub.jmonkeyengine.org/t/httpziplocator-questions/45859

```

com.jme3.asset.AssetLoadException: An error occurred loading Walk.gltf
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:181)
	at com.jme3.scene.plugins.gltf.GltfLoader.load(GltfLoader.java:106)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:439)
	at asset.TestHttpLocator.simpleInitApp(TestHttpLocator.java:33)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:548)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:662)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: com.google.gson.JsonIOException: java.util.zip.ZipException: invalid code lengths set
	at com.google.gson.internal.Streams.parse(Streams.java:62)
	at com.google.gson.JsonParser.parse(JsonParser.java:84)
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:123)
	... 9 more
Caused by: java.util.zip.ZipException: invalid code lengths set
	at java.base/java.util.zip.InflaterInputStream.read(InflaterInputStream.java:164)
	at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:270)
	at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:313)
	at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:188)
	at java.base/java.io.InputStreamReader.read(InputStreamReader.java:177)
Caused by: com.google.gson.JsonIOException: java.util.zip.ZipException: invalid code lengths set

	at com.google.gson.stream.JsonReader.fillBuffer(JsonReader.java:1295)
	at com.google.gson.stream.JsonReader.nextNonWhitespace(JsonReader.java:1333)
Caused by: java.util.zip.ZipException: invalid code lengths set

	at com.google.gson.stream.JsonReader.consumeNonExecutePrefix(JsonReader.java:1576)
	at com.google.gson.stream.JsonReader.doPeek(JsonReader.java:534)
	at com.google.gson.stream.JsonReader.peek(JsonReader.java:425)
	at com.google.gson.internal.Streams.parse(Streams.java:46)
	... 11 more


```

```
com.jme3.asset.AssetLoadException: An error occurred loading Stand.gltf
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:181)
	at com.jme3.scene.plugins.gltf.GltfLoader.load(GltfLoader.java:106)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at net.jmecn.HelloArmature.model(HelloArmature.java:321)
	at net.jmecn.HelloArmature.initScene(HelloArmature.java:257)
	at net.jmecn.HelloArmature.simpleInitApp(HelloArmature.java:114)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:139)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:221)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: com.google.gson.JsonIOException: java.util.zip.ZipException: invalid distance too far back
	at com.google.gson.internal.Streams.parse(Streams.java:62)
	at com.google.gson.JsonParser.parse(JsonParser.java:84)
	at com.jme3.scene.plugins.gltf.GltfLoader.loadFromStream(GltfLoader.java:123)
	... 10 more
Caused by: java.util.zip.ZipException: invalid distance too far back
	at java.base/java.util.zip.InflaterInputStream.read(InflaterInputStream.java:164)
	at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:270)
	at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:313)
	at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:188)
	at java.base/java.io.InputStreamReader.read(InputStreamReader.java:177)
	at com.google.gson.stream.JsonReader.fillBuffer(JsonReader.java:1295)
	at com.google.gson.stream.JsonReader.nextNonWhitespace(JsonReader.java:1333)
	at com.google.gson.stream.JsonReader.consumeNonExecutePrefix(JsonReader.java:1576)
	at com.google.gson.stream.JsonReader.doPeek(JsonReader.java:534)
	at com.google.gson.stream.JsonReader.peek(JsonReader.java:425)
	at com.google.gson.internal.Streams.parse(Streams.java:46)
	... 12 more
```

After comparing the two zips (one generated by [www.ezyzip.com](http://www.ezyzip.com/) and one by linux archive manager both from the same contents and with the same compression level), I noticed that their entry offsets are different in HttpZipLocator. 

Following it, I found that the "extraLen" (Central directory (CEN) header extra field length field offset) used here has different values on these two zip files:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/5ae543cc6f35cd93b53eb511cb912f91e210749f/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java#L239

in the zip file created by www.ezyzip.com (and also in the JME's "wildhouse.zip" that is on google storage), that value is 0 (which is why HttpZipLocator does not show an error) but in the zip file that @icyboxs uploaded and also I created in the Linux archive manager that value is non-zero (36 bytes).

Removing the "extraLen" fixed the issue.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
+++ b/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
@@ -96,6 +96,10 @@ public class HttpZipLocator implements AssetLocator {
         int compSize;
         long crc;
         boolean deflate;
+        // These fields will be fetched later from local file header,
+        // once asset requested.
+        Integer nameLength;
+        Integer extraLength;
 
         @Override
         public String toString(){
@@ -234,10 +238,6 @@ public class HttpZipLocator implements AssetLocator {
         entry.compSize = get32(table, offset + ZipEntry.CENSIZ);
         entry.offset   = get32(table, offset + ZipEntry.CENOFF);
 
-        // We want the offset in the file data:
-        // move the offset forward to skip the LOC header.
-        entry.offset += ZipEntry.LOCHDR + nameLen;
-
         entries.put(entry.name, entry);
         
         return newOffset;
@@ -323,9 +323,23 @@ public class HttpZipLocator implements AssetLocator {
         readCentralDirectory();
     }
 
-    private InputStream openStream(ZipEntry2 entry) throws IOException{
-        InputStream in = readData(entry.offset, entry.compSize);
-        if (entry.deflate){
+    private InputStream openStream(ZipEntry2 entry) throws IOException {
+        if (entry.nameLength == null && entry.extraLength == null) {
+            // Need to fetch local file header to obtain file name length
+            // and extra field length.
+            InputStream in = readData(entry.offset, ZipEntry.LOCHDR);
+            byte[] localHeader = new byte[ZipEntry.LOCHDR];
+            in.read(localHeader);
+            entry.nameLength = get16(localHeader, ZipEntry.LOCNAM);
+            entry.extraLength = get16(localHeader, ZipEntry.LOCEXT);
+            in.close();
+        }
+
+        // We want the offset in the file data:
+        // move the offset forward to skip the LOC header.
+        int fileDataOffset = entry.offset + ZipEntry.LOCHDR + entry.nameLength + entry.extraLength;
+        InputStream in = readData(fileDataOffset, entry.compSize);
+        if (entry.deflate) {
             return new InflaterInputStream(in, new Inflater(true));
         }
         return in;
--- a/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
+++ b/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
@@ -251,16 +251,15 @@ public class HttpZipLocator implements AssetLocator {
     }
 
     private void readCentralDirectory() throws IOException{
-        InputStream in = readData(tableOffset, tableLength);
         byte[] header = new byte[tableLength];
-
-        // Fix for "PK12 bug in town.zip": sometimes
-        // not entire byte array will be read with InputStream.read()
-        // (especially for big headers)
-        fillByteArray(header, in);
+        try (InputStream in = readData(tableOffset, tableLength)) {
+            // Fix for "PK12 bug in town.zip": sometimes
+            // not entire byte array will be read with InputStream.read()
+            // (especially for big headers)
+            fillByteArray(header, in);
 
 //        in.read(header);
-        in.close();
+        }
 
         entries = new HashMap<String, ZipEntry2>(numEntries);
         int offset = 0;
@@ -287,10 +286,10 @@ public class HttpZipLocator implements AssetLocator {
         // In that case, we have to search for it.
         // Increase search space to 200 bytes
 
-        InputStream in = readData(Integer.MAX_VALUE, 200);
         byte[] header = new byte[200];
-        fillByteArray(header, in);
-        in.close();
+        try (InputStream in = readData(Integer.MAX_VALUE, 200)) {
+            fillByteArray(header, in);
+        }
 
         int offset = -1;
         for (int i = 200 - 22; i >= 0; i--){
@@ -322,12 +321,12 @@ public class HttpZipLocator implements AssetLocator {
         if (entry.nameLength == null && entry.extraLength == null) {
             // Need to fetch local file header to obtain file name length
             // and extra field length.
-            InputStream in = readData(entry.offset, ZipEntry.LOCHDR);
-            byte[] localHeader = new byte[ZipEntry.LOCHDR];
-            in.read(localHeader);
-            entry.nameLength = get16(localHeader, ZipEntry.LOCNAM);
-            entry.extraLength = get16(localHeader, ZipEntry.LOCEXT);
-            in.close();
+            try (InputStream in = readData(entry.offset, ZipEntry.LOCHDR)) {
+                byte[] localHeader = new byte[ZipEntry.LOCHDR];
+                in.read(localHeader);
+                entry.nameLength = get16(localHeader, ZipEntry.LOCNAM);
+                entry.extraLength = get16(localHeader, ZipEntry.LOCEXT);
+            }
         }
 
         // We want the offset in the file data:
--- a/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
+++ b/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
@@ -236,7 +236,7 @@ public class HttpZipLocator implements AssetLocator {
 
         // We want the offset in the file data:
         // move the offset forward to skip the LOC header.
-        entry.offset += ZipEntry.LOCHDR + nameLen + extraLen;
+        entry.offset += ZipEntry.LOCHDR + nameLen;
 
         entries.put(entry.name, entry);
         
--- a/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
+++ b/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java
@@ -42,7 +42,7 @@ import java.net.URL;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.CharacterCodingException;
-import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CoderResult;
 import java.util.HashMap;
@@ -80,14 +80,9 @@ public class HttpZipLocator implements AssetLocator {
     private int tableLength;
     private HashMap<String, ZipEntry2> entries;
     
-    private static final ByteBuffer byteBuf = ByteBuffer.allocate(250);
-    private static final CharBuffer charBuf = CharBuffer.allocate(250);
-    private static final CharsetDecoder utf8Decoder;
-    
-    static {
-        Charset utf8 = Charset.forName("UTF-8");
-        utf8Decoder = utf8.newDecoder();
-    }
+    private final ByteBuffer byteBuf = ByteBuffer.allocate(250);
+    private final CharBuffer charBuf = CharBuffer.allocate(250);
+    private final CharsetDecoder utf8Decoder = StandardCharsets.UTF_8.newDecoder();
 
     private static class ZipEntry2 {
         String name;
@@ -129,7 +124,7 @@ public class HttpZipLocator implements AssetLocator {
              (((long)(b[off]&0xff)) << 24);
     }
 
-    private static String getUTF8String(byte[] b, int off, int len) throws CharacterCodingException {
+    private String getUTF8String(byte[] b, int off, int len) throws CharacterCodingException {
         StringBuilder sb = new StringBuilder();
         
         int read = 0;
@@ -147,7 +142,7 @@ public class HttpZipLocator implements AssetLocator {
             byteBuf.flip();
             
             // decode data in byteBuf
-            CoderResult result = utf8Decoder.decode(byteBuf, charBuf, endOfInput); 
+            CoderResult result = utf8Decoder.decode(byteBuf, charBuf, endOfInput);
             
             // If the result is not an underflow, it's an error
             // that cannot be handled.


Ali-RS:
Fix #1843

tonihele:
Hmm, looking at the specs though... https://en.wikipedia.org/wiki/ZIP_(file_format)

It reads this 

42 | 4 | Relative offset of local file header.  This is the number of bytes between the start of the first disk on which the file occurs, and the  start of the local file header. This allows software reading the central  directory to locate the position of the file inside the ZIP file.
-- | -- | --

Adds the local file header size and the file name size (variable length) and then the extra field size (variable length). But probably the extra field size here now just refers to the Central Directory File Header extra field. Right? This is not clear.

What I think is that this is broken. And this fix wont fix it.

What I think is the correct solution is that we should just always store the local file header offset. And start the reading from there, since clearly it is of variable size and the variable size can't be figured out from the central directory.

Then of course we would need to read a bit extra again like with the end header. Or sort the entries found and figure out the size of file data + file header from subsequent entries.

tonihele:
Maybe clearer example of ZIP format:
https://users.cs.jmu.edu/buchhofp/forensics/formats/pkzip.html

Ali-RS:
>But probably the extra field size here now just refers to the Central Directory File Header extra field. Right? 

Right, this is what it says in java doc:

```
    /**
     * Central directory (CEN) header extra field length field offset.
     */
    static final int CENEXT = 30;
```

https://github.com/jMonkeyEngine/jmonkeyengine/blob/5ae543cc6f35cd93b53eb511cb912f91e210749f/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java#L206

Ali-RS:
Taking a quick look at the Wikipedia link I see there is local file name length and local extra field length

```
    /**
     * Local file (LOC) header filename length field offset.
     */
    static final int LOCNAM = 26;

    /**
     * Local file (LOC) header extra field length field offset.
     */
    static final int LOCEXT = 28;
```

seems they are not used anywhere in HttpZipLocator. Are they perhaps meant to be used instead of the central directory headers (nameLen, extraLen) at this line? 

`entry.offset += ZipEntry.LOCHDR + nameLen + extraLen; `

tonihele:
> seems they are not used anywhere in HttpZipLocator🤔 Are they perhaps meant to be added to local offset instead of the central directory headers CENNAM (`nameLen`) and CENEXT (`extraLen`) at this line?

Yeah, it doesn't read the local file header. That is my point. It guesses the length of the local file header from the central directory header and that is the bug. But you can't use those LOCNAM in the central directory header of course.

So what I am proposing that we just start to read from the local file header always. Meaning in code:
`entry.offset += ZipEntry.LOCHDR + nameLen + extraLen; ` <- this would be totally removed, the whole line. Store the offset as it is.

So what we then have in the hash map, is the starting location for each entry. Starting location to the local file header of the entry to be exact. Then we make a guess for reading the file data from the HTTP connection. What we want to read is ZipEntry.LOCHDR (size of the local file header static parts) + nameLen (from the central directory header, assume it is always the same in both) + file data length (from the central directory header) + GUESS the length of the extra field (given by LOCEXT from the local header).

This way we read a bit extra, and future bug maybe that someone wrote 7GB to the extra header. I would guess that initially 100 bytes extra is enough for a guess. And shouldn't break the idea of the optimization, hopefully. I would imagine assets are big, 100 bytes is nothing in comparison.

tonihele:
I can also code this in if you agree with me in this approach and you don't want to take a stab at this. In addition:

- Would be nice to utilize the new Java 11 HttpClient to support HTTP2 and all that
- The static charset thing can be removed as these common ones are already found as static entries

pspeed42:
Note that JME has a very long history of "not invented here" aversion style development... whether because the developer was unaware of alternatives (even in the JDK) or because of some specific requirement they had latched onto.

I do not know the history or requirements for this class (though I can guess on the latter) but it might be worth exploring whether "modern Java" can already do what this class is trying to do.  (And by "modern", I mean post Java 1 point 2 (written out for emphasis how old I mean)).

For example, Java (effectively forever) has supported URL jar references to specific files within the jar... and because it uses this for remote class loading, it's been pretty well optimized.

Meaning: we might be able to rip out all of this weird special code and convert it to some clever URL construction or maybe just create a URLClassLoader that points to the zip file and then use it to resolve resources like a regular classpath asset locator does.

Ali-RS:
>I can also code this in if you agree with me in this approach

Please feel free to submit your change. But please note that Java 11 HttpClient still is not supported on android so better be avoided if we want android compatibility.

>it might be worth exploring whether "modern Java" can already do what this class is trying to do.

There is ZipInputStream in java (since java 1.1) that does what HttpZipLocator is doing, but seems the issue is that for loading a specific ZipEntry we must travers through the all other entries using `getNextEntry()` inside the stream untill find the entry we want rather than just giving the file name and get to the location. That is what HttpZipLocator tried to solve I belive.

tonihele:
`Please feel free to submit your change. But please note that Java 11 HttpClient still is not supported on android (I have made a [request](https://issuetracker.google.com/issues/197880209) to support it) so better be avoided if we want android compatibility.`

Damn Android, why you have to be different... Let's forget the HttpClient then.

```
I do not know the history or requirements for this class (though I can guess on the latter) but it might be worth exploring whether "modern Java" can already do what this class is trying to do. (And by "modern", I mean post Java 1 point 2 (written out for emphasis how old I mean)).

For example, Java (effectively forever) has supported URL jar references to specific files within the jar... and because it uses this for remote class loading, it's been pretty well optimized.

Meaning: we might be able to rip out all of this weird special code and convert it to some clever URL construction or maybe just create a URLClassLoader that points to the zip file and then use it to resolve resources like a regular classpath asset locator does.
```

I came to that conclusion that if we use anything ready, we will lose the optimization of just downloading the parts that we are interested in. And that seems to be the whole point in the class. Surely Internet speeds are probably faster today, but so are asset sizes probably on average bigger. Although not supporting Zip64, we have a limit on 4Gb ZIP file.

This locator has very niche usage probably, like you expressed in the forum. So I'd just fix the thing, keeping the optimization in place even though it is the part causing all the headaches :D

pspeed42:
But I would think that a Java URLClassLoader accessing a jar through a URL would have to have solved all of these problems and used to be super common.  It has to jump all over the place loading one .class file or another .class file... all over the jar.  So either it's doing clever caching or something even more clever.

My point is, don't necessarily trust that the "old ones" were smarter than we are.  If you look through JME history then you can find plenty of cases where the original authors just didn't know any better... or on that day just thought it would be easier to roll their own than to learn how to do it a built-in way.

I also think that in modern Java usage, using this class is a bad idea in general.  Without any real modern environments where "all of the code and data" is on a web site (ie: Applets), this is going to be poor experience for the user over any alternative.  It's tough to decide what the proper use-case would be.  And without some use-cases, can we honestly say that the four-line URLClassLoader version would perform worse than our bodged together fragile code?

tonihele:
Well, I mean.. after the fix it wouldn't been that fragile, I'm super confident about that :D ZIP format is quite simple and doesn't change. I agree that the code is really complicated for what it is trying to do. Just because it doesn't want to download the whole ZIP file, just the part needed.

I can try it hopefully today, but I'm quite confident that URLClassLoader approach will download the whole file. It probably caches it. At least the API faintly suggests that there are caches but you never know without looking the code more clearly what cache they are referring to (could be just the HTTP headers where it declares that no cache plz). But do we call this a success then? That the whole file is downloaded once, is that acceptable? That will be a change of behavior and might be noticeable if someone has that kind of weird use case where somehow only few assets are required.

If it downloads the ZIP more than once, then I would at least argue that this would be too big of a change to fix this bug.

pspeed42:
If it does download the whole file then I maybe agree.  But I have doubts.  I'm old enough to have lived through the age of the ubiquitous Java applet and we somehow didn't have to wait 5-10 minutes every time we started one up.

When we assume how it's going to work so continue hand-rolling then we do exactly what the original authors did.  Which is fine... sometimes that's how things get done.  Just a little unfortunate.

Ali-RS:
>So what I am proposing that we just start to read from the local file header always. Meaning in code:
entry.offset += ZipEntry.LOCHDR + nameLen + extraLen; <- this would be totally removed, the whole line. Store the offset as it is.

>So what we then have in the hash map, is the starting location for each entry. Starting location to the local file header of the entry to be exact. Then we make a guess for reading the file data from the HTTP connection. What we want to read is ZipEntry.LOCHDR (size of the local file header static parts) + nameLen (from the central directory header, assume it is always the same in both) + file data length (from the central directory header) + GUESS the length of the extra field (given by LOCEXT from the local header).


@tonihele something like this?

```
    private InputStream openStream(HttpZipLocator.ZipEntry2 entry) throws IOException{
        if (entry.nameLength < 0 && entry.extraLength < 0) {
            // Need to fetch local file header to obtain local file name length
            // and extra field length.
            InputStream in = readData(entry.offset, ZipEntry.LOCHDR);
            byte[] localHeader = in.readAllBytes();
            entry.nameLength = get16(localHeader, ZipEntry.LOCNAM);
            entry.extraLength = get16(localHeader, ZipEntry.LOCEXT);
            in.close();
        }

        // We want the offset in the file data:
        // move the offset forward to skip the LOC header.
        int fileDataOffset = entry.offset + ZipEntry.LOCHDR + entry.nameLength + entry.extraLength;
        InputStream in = readData(fileDataOffset, entry.compSize);
        if (entry.deflate) {
            return new InflaterInputStream(in, new Inflater(true));
        }
        return in;
    }
```
The above code works successfully. The thing to note is that LOCNAM and CENNAM are the same lengths and LOCEXT is 0. I guess this is true for 99% (?) of the case when we create a zip using an archive creation program unless someone manually adds a local extra field.

Please let me know if this code looks good to you, then will submit a PR.

tonihele:
That looks correct yes. I would also guess that the name is always the same, whether you got it from the central or the local. But you better make sure and this is the correct way to do it.

What comes now to the guessing how much we need to read. I would say we are safe with 100 bytes extra, ZIP64 for example adds just 32 bytes to the local headers, in form of the extra headers. Surely we wont support it currently and well... And the name should be the same.. This is probably just fine.

What I also noted that the whole class...  I guess it can be stateful, I have seen this a lot in jME. But there might be a risk with multithreading, although I'm not sure how the AssetManager handles the requests, does it allow for multithreaded processing or not. So there is this:
```
    static {
        Charset utf8 = Charset.forName("UTF-8");
        utf8Decoder = utf8.newDecoder();
    }
```
`Charset utf8 = Charset.forName("UTF-8");` <- can be replaced with a constant in modern Java
`utf8Decoder = utf8.newDecoder();` <- should be a class variable, not constant, it breaks with multithreading

And lot of those streams could be handled with `try-with` code.

tonihele:
Of course the cleaning up and all that would be better of as its own PR, but this is just... How could I put it, to me it doesn't matter that much. I'm always guilty of combining formatting, fix and optimizations into same PR. Which is horrible I know.

I'm quite curious to add ZIP 64 support. For absolutely no reason at all :D

Ali-RS:
>What comes now to the guessing how much we need to read. I would say we are safe with 100 bytes extra,

Not sure if I am understanding this. What are these 100 bytes for?

tonihele:
> Not sure if I am understanding this. What are these 100 bytes for?

From the central directory you get the offset where the local header starts and the file size. But you don't exactly know how much of the ZIP you should read as the local header size varies. HTTP is expensive, relatively. So we probably want to read the file at one go. Thus we need to make a guess what is the range of bytes we need.

The formula to calculate the range of bytes to read is:
**ZipEntry.LOCHDR** (size of the local file header static parts) + **nameLen** (from the central directory header, assume it is always the same in both and even if it isn't, the is some margin) + **file data length** (from the central directory header) + **GUESS** (the length of the extra fields)

**GUESS** could be 100 bytes.

Ali-RS:
>GUESS (the length of the extra fields)

Hmm, but extra field length is already saved in LOCEXT so why need to guess it?

tonihele:
Yeah, you are reading it from there yes. You know it AFTER you have downloaded the file. The asset requested.

1. Downloads 200 bytes from the end of the file to get the central directory
2. Parse the central directory to get the files (names and locations)
3. Downloads the requested asset

The number 3 dissected:
- We know the location where the asset starts in the ZIP
- We do **NOT** know how much we need to read

To download the asset you need to either take the guess route. Or you need to download in 1-2 parts once you have read the local header.

This is how I understand it works currently too. And that is the bug. It guesses the file starting location from the central directory. This fails because the file starting location is depending on the file local header length, which is unknown. It can only be guessed.

Ali-RS:
>You know it AFTER you have downloaded the file.

If you take a look at the code I showed above, I'm reading it in 2 passes, in the first connection, I just read the local header to obtain the name length and extra fields length then close the connection, now I already know from where the file data starts ( fileDataOffset = entry.offset + ZipEntry.LOCHDR + entry.nameLength + entry.extraLength ) then I create a new connection to readd the file data.

tonihele:
> If you take a look at the code I showed above, I'm reading it in 2 passes, in the first connection, I just read the local header to obtain the name length and extra fields length then close the connection, now I already know from where the file data starts ( fileDataOffset = entry.offset + ZipEntry.LOCHDR + entry.nameLength + entry.extraLength ) then I create a new connection to readd the file data.

Ah yes sorry, that is the route where you don't need guesses yes. `readData` should be `downloadData`, not really. I just wasn't paying attention. Sorry.

Ali-RS:
>utf8Decoder = utf8.newDecoder(); <- should be a class variable, not constant, it breaks with multithreading

It is used from a static method `private static String getUTF8String` so if I want to make utf8Decoder non static I must also make this method non static as well. Should I procceed with this change?

pspeed42:
Probably just recreate the decoder every time.  JME bends over so far backwards to avoid creating objects that it ends up breaking its spine and having its head up its own... you get the picture.  And yes, quite often this leads to major and often-subtle threading bugs (see my previous issues about animation tracks).

Unless we have evidence that creating a new decoder will be expensive then just create one when needed.

tonihele:
Yeah, I would say too that re-create it. It is the safest option. That can't be the heaviest operation there... Then you can still have the method as static.

Ali-RS:
What about these:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/5ae543cc6f35cd93b53eb511cb912f91e210749f/jme3-core/src/plugins/java/com/jme3/asset/plugins/HttpZipLocator.java#L83-L84

Should they be non-static as well? They are also used in `private static String getUTF8String` and passed into `utf8Decoder.decode()`.

What about just making `getUTF8String` non-static? It is a private method anyway.

I do not think they are thread

pspeed42:
Yeah, seems like a threading problem.

Ali-RS:
Looks like I have to make `getUTF8String()` method non static. It is a private method so won't break anything I guess.

Ali-RS:
I'm done with the update. Please feel free to review the changes.

tonihele:
I think it is excellent! Good work!

tonihele (approved):

Ali-RS:
Thanks! I could not do it without your help, guys.

Ali-RS:

PR #1847, opened 2022-08-22
Added an smart stretching option to BlendAction

Ali-RS:
By default, BlendAction will stretch any action that doesn't have the same length. This will cause stretched actions to end up playing slowly so one needs to manually speed up action to resolve this. What smart stretching does is that it tries to loop the action when possible before stretching to resolve the slow animation effect.

Gives the best result when the max transition weight is set to something lower than 1 (e.g. 0.5)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
@@ -1,5 +1,7 @@
 package com.jme3.anim.tween.action;
 
+import com.jme3.anim.tween.Tween;
+import com.jme3.anim.tween.Tweens;
 import com.jme3.anim.util.HasLocalTransform;
 import com.jme3.math.Transform;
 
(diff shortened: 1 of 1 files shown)


Ali-RS:
Still in draft until I test it with some more examples.

Ali-RS:
By the way, if you think the term "smart stretch" is not suitable, please let me know if you know a better one.

Ali-RS:
I have a new idea that is simpler and visually looks better.

Closing this draft:D

MeFisto94:
Is this related to #1303 ?

Ali-RS:
Yes. 

Well, the looping idea generates some visual artifacts while blending from one to another. Artifacts can be greatly reduced by setting the [max transition weight](https://github.com/jMonkeyEngine/jmonkeyengine/pull/1845)  to 0.5 or something.

In the new idea instead of looping I want to dynamically change the speed. So getSpeed() will not return a static number, based on the current blend weight it will interpolate a speed factor. I will submit the PR soon, so you can take a closer look.

Ali-RS:
See my new PR here: https://github.com/jMonkeyEngine/jmonkeyengine/pull/1848

Ali-RS:
I made a demo video showing how blending looks with this approach:

https://youtu.be/DJHwDaTy79o

PR #1848, opened 2022-08-23
BlendAction: resolve slow-motion side effect caused by stretching actions

Ali-RS:
This resolves slow-motion side effects caused by BlendAction stretching any animation that doesn't have the same length. It generates speed factors for each child animation that are dynamically interpolated and applied to base speed based on the blend weight taken from blend space.

This is a breaking change, as the speed returned by `BlendAction.getSpeed()` used to be a static value but now it returns a dynamic value that is changed as blend weight is changed. Unless `speedFactors` is cleared out or factors are equal. (i.e. clips have the same length).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendAction.java
@@ -1,3 +1,34 @@
+/*
+ * Copyright (c) 2009-2022 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
 package com.jme3.anim.tween.action;
 
 import com.jme3.anim.util.HasLocalTransform;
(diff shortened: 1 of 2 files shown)


Ali-RS:
Please let me know if something is vague about Javadoc or method naming.

Ali-RS:
I made a demo video:

The old behavior: (no using dynamic speed)

https://www.youtube.com/watch?v=jZY2u9GU_L8

The new behavior: (using dynamic speed)

https://youtu.be/1TuevW1yick

capdevon:
Hi Ali, may I ask you why the `calculateSpeedFactors` method is _public static_ and not _private_? :)

Ali-RS:
Hi @capdevon :)

Thanks so much for your interest.

Well, this PR is something that I rewrote multiple times in different ways because I was not sure which one is good. I am interested to hear other ideas. I wanted to indicate that users can modify the default factors to something else or even clear it and **be able to reset them back to the default values** so I made `calculateSpeedFactors` a utility method for that case (flexibility).

I would also like to hear from @pspeed42 about this.

pspeed42:
I don't have a problem with the public static utility method.  I'm not sure I'd think to use it, either... but I think it's ok.

Ali-RS:
@MeFisto94 I would like to hear your opinion as well regarding this PR :)

capdevon:
@Ali
> users can modify the default factors to something else or even clear it and **be able to set them back to the default ones**

@pspeed
> I'm not sure I'd think to use it, either...

What do you think of this proposal? :)
how about highlighting the 'default' keyword?

```java
//before
public static double[] calculateSpeedFactors(BlendAction blendAction) {
	Action[] actions = blendAction.getActions();
	return Arrays.stream(actions)
			.mapToDouble(action -> blendAction.getLength() / action.getLength())
			.toArray();
}

//usage example
BlendAction ba = ...;
ba.clearSpeedFactors();
ba.setSpeedFactors(BlendAction.calculateSpeedFactors(ba));
```

```java
//after (more compact and self consistent version)
public void setDefaultSpeedFactors() {
	this.speedFactors = getActions();
	return Arrays.stream(actions)
			.mapToDouble(action -> getLength() / action.getLength())
			.toArray();
}

//usage example:
BlendAction ba = ...;
ba.clearSpeedFactors();
// I deleted the previous values ​​for some reason, and now I want to set the defaults again.
ba.setDefaultSpeedFactors();
```

I have an API on top of the current animation system, where I use BlendActions with animations of different lengths. To solve the slow motion problem, I dynamically vary the speed using an algorithm inspired by Unity and Godot. What impact can this PR have on those who have already implemented this type of solution before, since the `speedFactors` are calculated in the constructor?

I saw that in the `TestAnimMigration` class, you removed the instructions that change the speed.
```java
//action.setSpeed(blendValue);
```

Ali-RS:
>how about highlighting the 'default' keyword?

Yeah, I am fine with `setDefaultSpeedFactors()`. I am going to apply your suggestion. 


>I have an API on top of the current animation system, where I use BlendActions with animations of different lengths. To solve the slow motion problem, I dynamically vary the speed using an algorithm inspired by Unity and Godot. What impact can this PR have on those who have already implemented this type of solution before, since the speedFactors are calculated in the constructor?

Will break them. If you want to manage speed using your API you must unset the speed factors using (`clearSpeedFactors`). 

Again I would like to ask @pspeed42 what does he think about it?

pspeed42:
I'm doing my best to follow along but I don't have much experience in this area.  So while I can prevent "obviously wrong" general API things from happening, I don't have a strong opinion on how the details work out.

If the speed can vary based on blending and different people want to blend things different ways, I wonder if a strategy object/function would be appropriate?  At some point it becomes over-engineered for the problem space, though.

Ali-RS:
>different people want to blend things different ways

Well, I am not sure about this one. This is the meat of the code
```
            return super.getSpeed() * FastMath.interpolateLinear(blendWeight,
                    (float) speedFactors[firstActiveIndex],
                    (float) speedFactors[secondActiveIndex]);
```

they may want to either set a different factor or use a different interpolator instead a of linear one. For the first, the PR already provides a way to set the custom factor (and I believe @capdevon API can take use of it). For the second, I think the linear interpolation is good enough for this case and one may not really want to change it.

Ali-RS:
Changed calculateSpeedFactors() to be a non static method and renamed it to applyDefaultSpeedFactors.

capdevon:
I've done some testing with my API and everything looks ok. I have no problem modifying it. I'll do some more testing when version 3.6.0-alpha is released. Another step forward for the engine. Good job! ;)

Ali-RS:
That's cool.

Ali-RS:

MeFisto94:
So from what I remember, on the surface, there are three types of ways you can combine two animations in parallel:
1. "Non Looping", i.e. they just run with their different lengths, start at the same point, end different and then when the full cycle is complete, they start again.
2. Looping them with their natural speed, i.e. if one animation only has half the length, it is just played twice as often
3. Looping them by stretching to the slowest animation (i.e. they end at the same time/loop roll over is at the same time)

So in theory, being able to choose between those may be nice, however I think option 2 (that is implemented by that PR) is the most common, especially considering animations are developed for a specific length (so 3 doesn't make too much sense) and option 1 can probably be done by clever tweening?

For the implementation I don't know the animation API good enough any more, I just knew that I was thinking on how you could get them to the same speed by just seleting a clever interpolator value (t), but setting a "fake" speed is probably a lot easier here (or the only possible way)

So yeah, TLDR: Nothing relevant :D

Ali-RS:
Thanks, @MeFisto94 

Well, the looping idea did not turn out well unfortunately, otherwise it would definitely be my pick. The blending looks bad, you can see it in this demo video: 
https://youtu.be/DJHwDaTy79o

That's why I opted for my second solution, that is dynamically adjusting the speed.

PR #1849, opened 2022-08-24
Added Tweens.cycle() and Tweens.invert() methods

Ali-RS:
Cycle tween is used for running delegate tween back and force and Invert tween is used to run delegate tween backward.


Ali-RS:

Issue #1857, opened 2022-10-16
Broken "License" link in README file


capdevon:
Broken "License" link in README file
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/LICENSE

It should be:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/LICENSE.md

stephengold:
This is a direct result of PR #1855.

kaaquist:
Can I fix this here as part of the hacktoberfest?

PR #1858, opened 2022-10-21
docs(README) - fix: broken link in README.md

kaaquist:
### Docs - style
* Fixed broken link in the `README.md` file.

Related issue: #1857

Closes: #1857

If possible I would like to do this here fix as part of the "Hacktoberfest".  I guess that we need to add a tag that it is accepted as such if that is ok?


riccardobl:
well, i am merging this as usual, but i am not sure if this will count toward the Hacktoberfest goal since it is just a small change

PR #1865, opened 2022-11-14
Add instance culling function in InstancedGeometry

Ali-RS:
By default, InstancedGeometry was culling instances that were outside camera frustum and this was causing the shadow to disappear on instances away from the camera. See https://hub.jmonkeyengine.org/t/shadow-disapeared-on-instanced-objects-away-from-camera/46174

I introduced an instance culling function that can be switched or unset. The default implementation checks against camera frustum and does not consider shadow at all. Users wanting proper shadows would have to implement a function that took shadow light frustums into account before culling or otherwise they can unset the function to disable instance culling and it will resolve the shadow issue.


Ali-RS:

Issue #1192, opened 2019-09-20
TestCanvas crashes with LWJGL3


stephengold:
Current `master` branch on Windows:
```
INFO JmeDesktopSystem 10:35:35 AM Running on jMonkeyEngine 3.3-6918
 * Branch: master
 * Git Hash: 46a52af
 * Build Date: 2019-09-20
SEVERE JmeDesktopSystem 10:35:35 AM CRITICAL ERROR: Context class is missing!
Make sure jme3_lwjgl-ogl is on the classpath.
java.lang.ClassNotFoundException: com.jme3.system.lwjgl.LwjglCanvas
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at com.jme3.system.JmeDesktopSystem.newContextLwjgl(JmeDesktopSystem.java:196)
	at com.jme3.system.JmeDesktopSystem.newContext(JmeDesktopSystem.java:279)
	at com.jme3.system.JmeSystem.newContext(JmeSystem.java:159)
	at com.jme3.app.LegacyApplication.createCanvas(LegacyApplication.java:510)
	at jme3test.awt.TestCanvas.createCanvas(TestCanvas.java:217)
	at jme3test.awt.TestCanvas.main(TestCanvas.java:248)
Exception in thread "main" java.lang.NullPointerException
	at com.jme3.system.JmeDesktopSystem.newContext(JmeDesktopSystem.java:280)
	at com.jme3.system.JmeSystem.newContext(JmeSystem.java:159)
	at com.jme3.app.LegacyApplication.createCanvas(LegacyApplication.java:510)
	at jme3test.awt.TestCanvas.createCanvas(TestCanvas.java:217)
	at jme3test.awt.TestCanvas.main(TestCanvas.java:248)
```

The same test app works fine with LWJGL2 on Windows.

I'm not sure what ever happened to `jme3_lwjgl-ogl`.

stephengold:
Similar crash with LWJGL3 on Linux. How hard would it be to implement `LwjglCanvas` for LWJGL3?

tonihele:
[...]

PR #1868, opened 2022-11-21
Lwjgl3 canvas

tonihele:
Would this be acceptable solution for now? If we open another ticket to implement this for real?

This would fix https://github.com/jMonkeyEngine/jmonkeyengine/issues/1192 and also TestAwtPanels. The performance is not great.. not terrible...

I tried getting https://github.com/LWJGLX/lwjgl3-awt to work but didn't get it display anything yet.

The problem with this solution is that it is pretty much software rendering. The rendering is first done to a Framebuffer as it is recommended. But then everything goes bad, we just render it to an image (I guess slightly accelerated but...) for displaying it to the user.

This is basically copy paste from what we already had in the engine, the Offscreen rendering to an AWT panel.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -46,15 +46,6 @@ import com.jme3.system.JmeSystem;
 import com.jme3.system.NanoTimer;
 import com.jme3.util.BufferUtils;
 import com.jme3.util.SafeArrayList;
-import org.lwjgl.Version;
-import org.lwjgl.glfw.GLFWErrorCallback;
-import org.lwjgl.glfw.GLFWFramebufferSizeCallback;
-import org.lwjgl.glfw.GLFWImage;
-import org.lwjgl.glfw.GLFWVidMode;
-import org.lwjgl.glfw.GLFWWindowFocusCallback;
-import org.lwjgl.glfw.GLFWWindowSizeCallback;
-import org.lwjgl.system.Platform;
-
 import java.awt.Graphics2D;
 import java.awt.image.BufferedImage;
 import java.nio.ByteBuffer;
@@ -64,10 +55,17 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
+import org.lwjgl.Version;
 import static org.lwjgl.glfw.GLFW.*;
+import org.lwjgl.glfw.GLFWErrorCallback;
+import org.lwjgl.glfw.GLFWFramebufferSizeCallback;
+import org.lwjgl.glfw.GLFWImage;
+import org.lwjgl.glfw.GLFWVidMode;
+import org.lwjgl.glfw.GLFWWindowFocusCallback;
+import org.lwjgl.glfw.GLFWWindowSizeCallback;
 import static org.lwjgl.opengl.GL11.GL_FALSE;
 import static org.lwjgl.system.MemoryUtil.NULL;
+import org.lwjgl.system.Platform;
 
 /**
  * A wrapper class over the GLFW framework in LWJGL 3.
@@ -149,8 +147,8 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
     protected boolean allowSwapBuffers = false;
 
     // temp variables used for glfw calls
-    private int width[] = new int[1];
-    private int height[] = new int[1];
+    private final int width[] = new int[1];
+    private final int height[] = new int[1];
 
     // state maintained by updateSizes()
     private int oldFramebufferWidth;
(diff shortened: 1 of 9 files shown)


jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -291,6 +289,8 @@ public void invoke(int error, long description) {
             requestWidth = videoMode.width();
             requestHeight = videoMode.height();
         }
+        oldFramebufferHeight = requestHeight;
```
tonihele, line 292: By assuming that the scaling factor of 1 is correct we avoid extra resize at the start of application, also fixes TestAwtPanels. If you really don't have scaling... AwtContext wasn't build scaling in mind. It throws an exception when framebuffer resize is triggered, and that is how our scaling currently works.

riccardobl:
Thank you for the PR. 

> Would this be acceptable solution for now? If we open another ticket to implement this for real?

Yes, I think this is the least sketchy way to get this to work reliably, other solutions can be implemented as optimization for specific platforms (eg. i see lwjgl3-awt works only on linux systems that use X11).

Also this should probably work well  in most systems.

I have 2 questions:

1. What is the deal with all the changes to Lists?
2. Is it possible to remove the non-power-of-two requirement by rounding the sizes to the nearest power of two? (it will introduce some cropping, yes, but it should probably be fine i think)

tonihele:
> * What is the deal with all the changes to Lists?
> 
> * Is it possible to remove the non-power-of-two requirement by rounding the sizes to the nearest power of two? (it will introduce some cropping, yes, but it should probably be fine i think)

1. It is just cleanup, aren't we all taught to program with interfaces? :) Well, I intended to actual have the event queues be queues, but I changed my mind as it would achieve nothing. In sense of clarity or performance the same, so then I just left them as cleanups. Irrelevant changes.

2. Can be done, also in such matter that our Offscreen rendering would also be void of this requirement. Perhaps in such manner that the capability is checked against and a proper strategy is then used.

riccardobl:
> It is just cleanup, aren't we all taught to program with interfaces? :) Well, I intended to actual have the event queues be queues, but I changed my mind as it would achieve nothing. In sense of clarity or performance the same, so then I just left them as cleanups. Irrelevant changes.

Ok, i was questioning the use of Deque vs List, but i've seen now that you did that to use poll(). Ok then.

>  Can be done, also in such matter that our Offscreen rendering would also be void of this requirement. Perhaps in such manner that the capability is checked against and a proper strategy is then used.

Yep, should be easy enough, since we have already a capability registered for that. Do you plan to address it in this PR or would you like to do it at a later time? Non-power-of-two textures should be supported by the vast majority of GPUs anyway, so we could also decide to leave it like that for now.. and just merge this PR

tonihele:
> Yep, should be easy enough, since we have already a capability registered for that. Do you plan to address it in this PR or would you like to do it at a later time? Non-power-of-two textures should be supported by the vast majority of GPUs anyway, so we could also decide to leave it like that for now.. and just merge this PR

I would do additional issues. Not this PR. The following issues would be created:

1. The non-power of two issue is existing issue with the offscreen rendering, create issue for it
2. Implement HW accelerated Lwjgl3 Canvas

1 & 2 can be handled then separately without applying any pressure on one another. And if 2 is done before, well, that is all good then, canvas works.

Ali-RS:
What is the state of this PR?
Should this become a candidate for the next release?
Is anybody willing to review and possibly merge it?

tonihele:
> What is the state of this PR? Should this become a candidate for the next release? Is anybody willing to review and possibly merge it?

It is ready.

What it is, is a poor solution to something that currently doesn't work at all. With this it works. Usability depends largely on what it is being used for, i.e. gaming, not feasible; using LWJGL 3 and a small 3D canvas in your app, feasibleish. There is the test canvas test app for this. 

Even I'm on the fence with this one. My small hope was that we could use LWJGL 3 with the jME SDK with this. But I don't know if this even really makes it possible. SDK rendering at least would need so little power that in that way this would work.

Future development needed as stated above.

Sailsman63:
Would if be possible to add a short Javadoc comment explaining what/when this is used? IIUC, it's to allow lwjgl to render a scene to an AWT component, but lack of Javadoc is an ongoing pain-point for me with JME.

tonihele:
> Would if be possible to add a short Javadoc comment explaining what/when this is used? IIUC, it's to allow lwjgl to render a scene to an AWT component, but lack of Javadoc is an ongoing pain-point for me with JME.

The code is self-doc... :D As far as I know it is indeed used to render jME context inside an AWT component. Would that itself be a suffice explanation?

Sailsman63:
Sorry, I thought I had responded. Yeah, Just a one-liner stating the use. Most of the methods seem to be `@Overrides` anyway, so they are presumably documented further up the tree.

I am wondering if `drawFrameInThread()` and `destroyFrameBuffer()` are intended to be called from user code, or if they are marked `public` to allow some of the existing machinery to trigger them. If the later, a doc-comment warning that they should not be called directly might be in order.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
Ali-RS: Please see: https://github.com/jMonkeyEngine/jmonkeyengine/pull/1950/files#r1103859250

Ali-RS: > TestAwtPanels throws an exception

AwtPanelsContext is modified to log an error instead of throwing an exception which should make TestAwtPanels work. 

Please see https://github.com/jMonkeyEngine/jmonkeyengine/pull/1949

PR #1869, opened 2022-11-26
Implementation of a glTF extension loader for KHR_texture_transform

manuelrmo:
Implementation of a glTF extension loader for KHR_texture_transform, which allows for UV coordinates to be scaled/rotated/translated based on transformation properties from textures in the glTF model.

The extension will avoid the following warning messages:

Extension KHR_texture_transform is not supported, please provide your own implementation in the GltfModelKey
Extension KHR_texture_transform is mandatory for this file, the loaded scene result will be unexpected.
Could not find loader for extension KHR_texture_transform

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -96,9 +96,6 @@ public class GltfLoader implements AssetLoader {
     Map<SkinData, List<Spatial>> skinnedSpatials = new HashMap<>();
     IntMap<SkinBuffers> skinBuffers = new IntMap<>();
 
-    // Last created geometry
-    private Geometry lastCreatedGeom = null;
-    
     public GltfLoader() {
         defaultMaterialAdapters.put("pbrMetallicRoughness", new PBRMetalRoughMaterialAdapter());
     }
@@ -467,7 +464,10 @@ public class GltfLoader implements AssetLoader {
             // Read mesh extras
             mesh = customContentManager.readExtensionAndExtras("primitive", meshObject, mesh);
             Geometry geom = new Geometry(null, mesh);
-            lastCreatedGeom = geom;
+            // Cached data about the geometry needed by TextureTransformExtensionLoader.class
+            addToCache("lastCreatedGeom", 0, geom, 3); // Last created geometry
+            addToCache("lastCreatedGeom", 1, null, 3); // Last processed geometry
+            addToCache("lastCreatedGeom", 2, null, 3); // Last applied transformation matrix (inverted)
 
             Integer materialIndex = getAsInteger(meshObject, "material");
             if (materialIndex == null) {
@@ -1210,11 +1210,6 @@ public class GltfLoader implements AssetLoader {
     public Node getRootNode() {
         return rootNode;
     }
-    
-    // Getter for the last created geometry
-    public Geometry getLastCreatedGeom() {
-        return lastCreatedGeom;
-    }    
 
     private String decodeUri(String uri) {
         try {
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -382,6 +382,7 @@ public class GltfLoader implements AssetLoader {
         for (JsonElement primitive : primitives) {
             JsonObject meshObject = primitive.getAsJsonObject();
             Mesh mesh = new Mesh();
+            addToCache("mesh", 0, mesh, 1);
             Integer mode = getAsInteger(meshObject, "mode");
             mesh.setMode(getMeshMode(mode));
             Integer indices = getAsInteger(meshObject, "indices");
@@ -464,10 +465,6 @@ public class GltfLoader implements AssetLoader {
             // Read mesh extras
             mesh = customContentManager.readExtensionAndExtras("primitive", meshObject, mesh);
             Geometry geom = new Geometry(null, mesh);
-            // Cached data about the geometry needed by TextureTransformExtensionLoader.class
-            addToCache("lastCreatedGeom", 0, geom, 3); // Last created geometry
-            addToCache("lastCreatedGeom", 1, null, 3); // Last processed geometry
-            addToCache("lastCreatedGeom", 2, null, 3); // Last applied transformation matrix (inverted)
 
             Integer materialIndex = getAsInteger(meshObject, "material");
             if (materialIndex == null) {
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -96,6 +96,9 @@ public class GltfLoader implements AssetLoader {
     Map<SkinData, List<Spatial>> skinnedSpatials = new HashMap<>();
     IntMap<SkinBuffers> skinBuffers = new IntMap<>();
 
+    // Last created geometry
+    private Geometry lastCreatedGeom = null;
+    
     public GltfLoader() {
         defaultMaterialAdapters.put("pbrMetallicRoughness", new PBRMetalRoughMaterialAdapter());
     }
@@ -464,6 +467,7 @@ public class GltfLoader implements AssetLoader {
             // Read mesh extras
             mesh = customContentManager.readExtensionAndExtras("primitive", meshObject, mesh);
             Geometry geom = new Geometry(null, mesh);
+            lastCreatedGeom = geom;
 
             Integer materialIndex = getAsInteger(meshObject, "material");
             if (materialIndex == null) {
@@ -1206,6 +1210,11 @@ public class GltfLoader implements AssetLoader {
     public Node getRootNode() {
         return rootNode;
     }
+    
+    // Getter for the last created geometry
+    public Geometry getLastCreatedGeom() {
+        return lastCreatedGeom;
+    }    
 
     private String decodeUri(String uri) {
         try {
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
@@ -65,7 +65,7 @@ public class TextureTransformExtensionLoader implements ExtensionLoader {
      * @param transform The matrix containing the scale/rotate/translate transformations
      */    
     private void uvTransform(Mesh mesh, Matrix3f transform) {
-        if (!transform.isIdentity()) { // if m is the identity matrix, there's nothing to do
+        if (!transform.isIdentity()) { // if transform is the identity matrix, there's nothing to do
             VertexBuffer tc = mesh.getBuffer(VertexBuffer.Type.TexCoord);
             if (tc == null) {
                 throw new IllegalStateException("The mesh has no texture coordinates");
@@ -105,8 +105,8 @@ public class TextureTransformExtensionLoader implements ExtensionLoader {
             JsonObject jsonObject = extension.getAsJsonObject();
             if (jsonObject.has("offset")) {
                 JsonArray jsonArray = jsonObject.getAsJsonArray("offset");
-                translation.set(2, 0, jsonArray.get(0).getAsFloat());
-                translation.set(2, 1, jsonArray.get(1).getAsFloat());                    
+                translation.set(0, 2, jsonArray.get(0).getAsFloat());
+                translation.set(1, 2, jsonArray.get(1).getAsFloat());                    
             }
             if (jsonObject.has("rotation")) {
                 float rad = jsonObject.get("rotation").getAsFloat();
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
@@ -137,8 +137,12 @@ public class TextureTransformExtensionLoader implements ExtensionLoader {
                 // at this point, we're processing a new mesh or the same mesh as before but for a different UV set
                 if (mesh != meshLast) { // it's a new mesh
                     loader.addToCache("textureTransformData", 0, mesh, 2);
-                    transformMap = new HashMap<>(); // initialize transformMap
-                    loader.addToCache("textureTransformData", 1, transformMap, 2);
+                    if (transformMap == null) {
+                        transformMap = new HashMap<>(); // initialize transformMap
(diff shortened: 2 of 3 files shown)


Ali-RS:
Thanks for your contribution

Can you test if the extension loader works fine in multithreaded asset loading?

From a quick look, it seems it caches some state, which might get corrupted in multithreaded loading 

```
    private Geometry geomLast = null; // Last geometry created by the GltfLoader.class object
    private Matrix3f mInvLast = null; // Last transformation matrix (inverted) that was applied
```

AssetManager supports concurrent asset loading and in the case of Gltf models, it will instantiate a new instance of GltfLoader for each thread but I noticed CustomContentManager used inside GltfLoader registers the extension loaders statically this implies that extension loaders must be thread-safe or otherwise we should make the extension loaders map non-static as well. 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/CustomContentManager.java#L54-L60

@stephengold do you concur with me?

Ali-RS:
I also noticed `MaterialAdapter` that is used by other extension loaders (like PBRSpecGlossExtensionLoader & UnlitExtensionLoader) also is not thread-safe and keeps an internal state which will break if used across multiple threads.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/MaterialAdapter.java#L51-L56

So my suggestion is to simply make `defaultExtensionLoaders` inside CustomContentManager non-static so each GltfLoader per thread will have its own instance of extension loaders not shared with other threads. This is better done in a separate PR.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
```diff
@@ -0,0 +1,146 @@
...
+            Vector3f v = m.mult(new Vector3f(x, y, 1));
+            fb.put(v.getX()).put(v.getY());
+        }
+        fb.clear();
+        tc.updateData(fb);        
+    }
+    
+    @Override
+    public Object handleExtension(GltfLoader loader, String parentName, JsonElement parent, JsonElement extension, Object input) throws IOException {
+        if (input instanceof Texture2D) {
```
Ali-RS, line 97: Regarding `if (input instanceof Texture2D)` 
Can input be other than Texture2D?

I am thinking if the `input instanceof Mesh` is called before the `input instanceof Texture2D` then you may just grab the mesh instance there and cache it, this way you can get rid of the `GltfLoader.getLastCreatedGeom()`.

manuelrmo: Thanks for catching those issues and the insightful comments.

Indeed, the extension loader got corrupted in multithreaded loading tests. 

I've implemented a thread-safe version that fixes those issues, and uses the GltfLoader cache as per your advice.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
```diff
@@ -464,6 +464,10 @@ public Geometry[] readMeshPrimitives(int meshIndex) throws IOException {
             // Read mesh extras
             mesh = customContentManager.readExtensionAndExtras("primitive", meshObject, mesh);
             Geometry geom = new Geometry(null, mesh);
+            // Cached data about the geometry needed by TextureTransformExtensionLoader.class
+            addToCache("lastCreatedGeom", 0, geom, 3); // Last created geometry
+            addToCache("lastCreatedGeom", 1, null, 3); // Last processed geometry
+            addToCache("lastCreatedGeom", 2, null, 3); // Last applied transformation matrix (inverted)
```
Ali-RS, line 470: These should be removed. GltfLoader should know nothing about TextureTransformExtensionLoader.

Instead, you should grab the mesh in TextureTransformExtensionLoader

```
public Object handleExtension(GltfLoader loader, String parentName, JsonElement parent, JsonElement extension, Object input) throws IOException {
        if (input instanceof Mesh) {
            loader.addToCache("textureTransformData", 0, mesh, 2); // Last created mesh
            loader.addToCache("textureTransformData", 1, null, 2); // Inverted transform matrix
        } else if (input instanceof Texture2D) {
             Mesh mesh = loader.fetchFromCache("textureTransformData", 0, mesh, Mesh.class);
             // do things...
        }
}

```

you do not need  "Last processed geometry", just check with the transformation matrix, if it is null then calculate and apply it and if it is not null then it means this is from the next texture of the same material, you need to calculate matrix again and check with the previous one as you have done in `if(!m.mult(mInvLast).isIdentity())` and warn if they are different.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
```diff
@@ -0,0 +1,150 @@
...
+    // loads all textures of a given geometry/mesh before doing so 
+    // for the next geometry/mesh.    
+    @Override
+    public Object handleExtension(GltfLoader loader, String parentName, JsonElement parent, JsonElement extension, Object input) throws IOException {
+        if (input instanceof Texture2D) {
+            Geometry geom = loader.fetchFromCache("lastCreatedGeom", 0, Geometry.class);
+            if (geom != null) {
+                Matrix3f t = new Matrix3f();
+                Matrix3f r = new Matrix3f();
+                Matrix3f s = new Matrix3f();
```
Ali-RS, line 104: Better to name them "translation", "rotation", and "scale".

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
```diff
@@ -0,0 +1,150 @@
...
+                }                
+                if (jsonObject.has("scale")) {
+                    JsonArray jsonArray = jsonObject.getAsJsonArray("scale");
+                    s.set(0, 0, jsonArray.get(0).getAsFloat());
+                    s.set(1, 1, jsonArray.get(1).getAsFloat());
+                }     
+                if (jsonObject.has("texCoord")) {
+                    logger.log(Level.WARNING, "KHR_texture_transform extension: the texCoord property is not supported");                
+                }                 
+                Matrix3f m = t.mult(r).mult(s);
```
Ali-RS, line 126: Better to rename `m` to `transform` or `transformMatrix`.

manuelrmo: KHR_texture_transform is usually defined at the level of a texture in a glTF model, as it seems to be the case for Blender. Accordingly, GltfLoader calls the extension loader (readExtensionAndExtras) passing a Texture2D to the "input" parameter when reading a texture (readTexture). In this case, the "if" condition for the Mesh of the proposed code above will always be evaluated as false within the extension loader:

if (input instanceof Mesh) { // => always false for most common glTF models

So the extension loader needs an alternative way to get the mesh. The cache of the GltfLoader can be used for this purpose, but meshes are cached too late (created at line 384, cached at line 496 in geomArray) after textures are already read. Having a more complete state in the cache that extension loaders can use seems reasonable, either by modifying the way GltfLoader caches meshes or by adding an extra line in GltfLoader to cache the mesh before textures are read (as done in the latest commit).

However, I believe the proper way to handle this would be for the extension loader to store each texture transform within the corresponding Texture2D object for later processing, but JME does not seem to support such a design/implementation especially due to the fact that textures are meant to be shared among materials and meshes/geometries. 

Indeed, the original purpose of KHR_texture_transform is to support packing textures into a texture atlas, as explained at:
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_texture_transform

Also, different transforms might be specified for different textures. For example, see the UV Mapping section of the glTF documentation of Blender, explaining how Mapping nodes can be attached to Image Texture nodes:
https://docs.blender.org/manual/en/latest/addons/import_export/scene_gltf2.html#id18

Ali-RS: Thanks for the detailed explanation.

>the "if" condition for the Mesh of the proposed code above will always be evaluated as false

Ah, yes you are right, stupid me!

>Indeed, the original purpose of KHR_texture_transform is to support packing textures into a texture atlas

By the way, JME also has a utility to batch geometries and generate a texture atlas.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/417509eb17722a39a941de66d03e078d542d6e8a/jme3-core/src/tools/java/jme3tools/optimize/GeometryBatchFactory.java#L97

Ali-RS:
Looks fine to me. Thanks @manuelrmo

manuelrmo:
Great, thanks for your help.

manuelrmo:
The extension loader passes the Texture Transform Test from KhronosGroup/glTF:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureTransformTest

test

manuelrmo:
The extension loader passes the Texture Transform Multi Test from KhronosGroup/glTF:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureTransformMultiTest

Notes:
- JME has no KHR_materials_clearcoat extension loader
- JME renderer does not seem to support TEXCOORD_1 coordinates (UV 1)

CaptureMultiTest

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
```diff
@@ -121,19 +125,19 @@ public Object handleExtension(GltfLoader loader, String parentName, JsonElement
...
-                uvTransform(mesh, transform);
+                loader.addToCache("textureTransformData", 1, transform, 2);
+                uvTransform(mesh, transform, getVertexBufferType("TEXCOORD_" + (texCoord != null ? texCoord : 0)));
                 logger.log(Level.FINE, "KHR_texture_transform extension successfully applied."); 
             }
             else {
-                Matrix3f mInvLast = loader.fetchFromCache("textureTransformData", 1, Matrix3f.class);
-                if (!transform.mult(mInvLast).isIdentity()) {
+                Matrix3f transformLast = loader.fetchFromCache("textureTransformData", 1, Matrix3f.class);
+                if (!transform.toString().equals(transformLast.toString())) {
```
Ali-RS, line 140: Why using toString() instead of `Matrix3f.equals()`?

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
```diff
@@ -86,7 +89,7 @@ private void uvTransform(Mesh mesh, Matrix3f transform) {
                 fb.put(v.getX()).put(v.getY());
             }
             fb.clear();
-            tc.updateData(fb);   
+            tc.updateData(fb);
```
Ali-RS, line 92: Please remove those trailing whitespace.

Ali-RS:
>JME renderer does not seem to support TEXCOORD_1 coordinates (UV 1)

https://github.com/jMonkeyEngine/jmonkeyengine/blob/55574561cfcacf0b7b33358330528061b3fd4c42/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfUtils.java#L135-L136

I think it is used for lightmap

https://github.com/jMonkeyEngine/jmonkeyengine/blob/52568d76ffec063355bcff515066cb75887baf34/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.j3md#L68-L72

manuelrmo:
> I think it is used for lightmap

Correct, it is used for LightMap which is shared with AOMap, see the screenshot below after enabling SeparateTexCoord. 

So it seems that the JME loader ignores the glTF setting asking to use TEXCOORD_1 for the UVs, and the JME renderer will use TEXCOORD_0 by default. 
For LightMap/AOMap, the user still has a chance to edit the material later and enable SeparateTexCoord to use UVs from TEXCOORD_1.

CaptureMultiTest2

Ali-RS:
Can you please explain what is the latest change for?

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
```diff
@@ -0,0 +1,164 @@
...
+                texCoord = jsonObject.get("texCoord").getAsInt(); // it overrides the parent's texCoord value
+            }                 
+            Matrix3f transform = translation.mult(rotation).mult(scale);
+            Mesh meshLast = loader.fetchFromCache("textureTransformData", 0, Mesh.class);
+            Map transformMap = loader.fetchFromCache("textureTransformData", 1, HashMap.class);
+            if (mesh != meshLast || (transformMap != null && transformMap.get(texCoord) == null)) {
+                // at this point, we're processing a new mesh or the same mesh as before but for a different UV set
+                if (mesh != meshLast) { // it's a new mesh
+                    loader.addToCache("textureTransformData", 0, mesh, 2);
+                    if (transformMap == null) {
```
Ali-RS, line 140: Isn't `if (transformMap == null)` always true when it is a new mesh?

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
```diff
@@ -0,0 +1,164 @@
...
+                JsonArray jsonArray = jsonObject.getAsJsonArray("scale");
+                scale.set(0, 0, jsonArray.get(0).getAsFloat());
+                scale.set(1, 1, jsonArray.get(1).getAsFloat());
+            }     
+            if (jsonObject.has("texCoord")) {
+                texCoord = jsonObject.get("texCoord").getAsInt(); // it overrides the parent's texCoord value
+            }                 
+            Matrix3f transform = translation.mult(rotation).mult(scale);
+            Mesh meshLast = loader.fetchFromCache("textureTransformData", 0, Mesh.class);
+            Map transformMap = loader.fetchFromCache("textureTransformData", 1, HashMap.class);
```
Ali-RS, line 135: Please specify Map generic types. 
`Map<Integer, Matrix3f>`

And remove unnecessary cast to (Matrix3f) in line 153.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/TextureTransformExtensionLoader.java
manuelrmo: `transformMap` is `null` only during the first call to `handleExtension`. For subsequent calls, `transformMap` is always not `null`, irrespective of whether the mesh is the same as/different (new) from the one during the previous call.

Ali-RS:
@manuelrmo is this PR ready to be merged?

manuelrmo:
Yes, thank you

PR #1872, opened 2022-11-28
Fix #1867 (LightFilter gets applied even if not needed)

### LINKED ISSUE #1867: Performance: LightFilter gets applied even if not needed
Hi, during my investigation on my deferred performance i found that the LightFilter gets applied on each RenderManager.renderGeometry()

I looked up on how Jme internally works around this issue when no lights are needed and it seems it does not. 

It surely impacts the performance of the AbstractShadowRenderer. At least on objects outside of the view frustum but inside of the shadow frustum the light filtering is not needed. 

A quick fix would be to implement:
``
tmp = RenderManager.getLightFilter();
RenderManager.setLightFilter(null);
//Do your stuff
RenderManager.setLightFilter(tmp);
``

what do you think? applying the easy fix or implement a fix at global scale?

zzuegg:
I have added the null light filter to all shadow map generators. 
Further investigation is needed to see where else it can be applied


jme3-core/src/main/java/com/jme3/shadow/AbstractShadowRenderer.java
```diff
@@ -77,7 +79,7 @@
 public abstract class AbstractShadowRenderer implements SceneProcessor, Savable, JmeCloneable, Cloneable {
 
     protected static final Logger logger = Logger.getLogger(AbstractShadowRenderer.class.getName());
-
+    private static LightFilter NULL_LIGHT_FILTER = new NullLightFilter();
```
Ali-RS, line 82: Can you please add `final` modifier? (here and in the other classes you have used NullLightFilter?

zzuegg: sure.

Ali-RS: thanks

jme3-core/src/main/java/com/jme3/light/NullLightFilter.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.jme3.renderer.Camera;
+import com.jme3.scene.Geometry;
+/**
+ * NullLightFilter does nothing. Used when you want
+ * to disable the light filter
+ *
+ * @author Kirill Vainer
+ */
+public class NullLightFilter implements LightFilter{
```
Ali-RS, line 43: Looks like there is a missing whitespace in `LightFilter{`

Ali-RS:
Thanks, it looks good to me.

Ali-RS:
@zzuegg I plan to merge this but before that, have you tested this with some of the JME shadow examples? just to make sure nothing is broken.

zzuegg:
I have run all the tests that include *shadow* in the name at least 1trillion time in the last weeks. The only problem was that i accidentally had checked out 3.3 as my master. Since i forgot to update my github fork. Let me check if there where any relevant changes since 3.3. I do not expect any issues since after updating all works the same but i have to make sure. Tomorrow evening i have time for this

zzuegg:
Did not notice any regressions. Good to go from my side

Ali-RS:
Thanks

PR #1880, opened 2022-12-17
Fix #1412 (GltfLoader does not support AO packed in MetallicRoughnessMap)

### LINKED ISSUE #1412: GLTF Converter needs support for AO value in MetallicRoughnessMap
Now that we have added support for packing the AmbientOcclusion value into the Red channel of the MetallicRoughnessMap,
we need to update the GLTF converter so that auto-generated materials utilize this feature.

Currently, whenever the GLTF converter detects the AO packed into the MR map, it will just define the MR map to both material params: "MetallicRoughnessMap" (as expected) and also to "LightMap" (and then it also sets the boolean LightMapAsAoMap to true)

This resulted in an extra texture read from the light map in the PBR shader (which defeats the purpose of packing it into the MR map), which is why PR https://github.com/jMonkeyEngine/jmonkeyengine/pull/1403 was necessary to make it so AO is read directly from the red channel of the MR Map 

Now we need to update the converter to instead set "AoPackedInMrMap" to true whenever the GLTF file detects an AO map packed in the MR Map.

manuelrmo:
This is an update for the glTF loader to be able to detect that an occlusion texture is packed in the metallic-roughness texture, so the loader can enable the `AoPackedInMRMap` parameter from the material and avoid double-loading the same texture.

The update consists of few lines added to `GltfLoader` that retrieve the `index` (if any) of each texture from the glTF model and compare them: if they exist and are the same, `AoPackedInMRMap` will be enabled and no texture will be loaded for occlusion. 

See for example the WaterBottle test model:
https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/WaterBottle/glTF/WaterBottle.gltf

```
  "materials": [
    {
      "pbrMetallicRoughness": {
        "baseColorTexture": {
          "index": 0
        },
        "metallicRoughnessTexture": {
          "index": 1
        }
      },
      "normalTexture": {
        "index": 2
      },
      "occlusionTexture": {
        "index": 1
      },
      "emissiveFactor": [
        1.0,
        1.0,
        1.0
      ],
      "emissiveTexture": {
        "index": 3
      },
      "name": "BottleMat"
    }
  ],
```

In the glTF code, when AO map is packed in MR map, the `index` value for `"occlusionTexture"` is the same as the `index` value for `"metallicRoughnessTexture",` which means that they refer to the same texture image. Note that the clauses `"pbrMetallicRoughness"` and `"occlusionTexture"` can appear in any order.

This PR resolves Issue #1412.

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -635,6 +635,7 @@ public class GltfLoader implements AssetLoader {
             setDefaultParams(adapter.getMaterial());
         }
 
+        Integer metallicRoughnessIndex = null;
         if (pbrMat != null) {
             adapter.setParam("baseColorFactor", getAsColor(pbrMat, "baseColorFactor", ColorRGBA.White));
             adapter.setParam("metallicFactor", getAsFloat(pbrMat, "metallicFactor", 1f));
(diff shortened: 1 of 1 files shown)


Ali-RS:
Thank you @manuelrmo

PR #1881, opened 2022-12-19
BlendableAction: Fix JavaDoc for setMaxTransitionWeight & replace assert with IllegalArgumentException

mrxz:
I happened to spot a small mistake in the following commit: https://github.com/jMonkeyEngine/jmonkeyengine/commit/9d5eeee8923c27495c948be7e7c91276e739da51
Based on the assertion in the method and the fact that the default value is 1, it's clear that the 0 and 1 bounds are inclusive.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendableAction.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendableAction.java
@@ -114,10 +114,13 @@ public abstract class BlendableAction extends Action {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Ali-RS:
Nice catch, thank you @mrxz!

Aside from that, I am also not sure why I used `assert` instead of IllegalArgumentException back then.:man_facepalming: 

Must fix it!

mrxz:
> Would you mind please removing assert and instead threw an IllegalArgumentException?
Done.

Ali-RS:
Thanks

Ali-RS:

PR #1884, opened 2022-12-23
Fix #1882 (J3MLoader always generates mips ignoring MinFilter)

### LINKED ISSUE #1882: J3MLoader always generates mips ignoring MinFilter
J3MLoader always set generate mips flag to true regardless of whether it is required by MinFilter or not.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/befa930cf9075054ecff154e2e36a42947d2ced1/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java#L279

for example, `NearestNoMipMaps` and `BilinearNoMipMaps` do not require mips.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ac8c1012d2ded0812d91a393f935c9f0f7d0c192/jme3-core/src/main/java/com/jme3/texture/Texture.java#L87-L102

Shouldn't it check MinFilter and generate mips only if it is required?

Because of that, J3MLoader is broken, I noticed if I set the min filter to `BilinearNoMipMaps` on texture and export the material to a j3m file in the code then if I load it at runtime the min filter is set to `Trilinear`. It will work fine if I set it to something other than `BilinearNoMipMaps`.

That is because `BilinearNoMipMaps` is not written to the j3m file as it is the default min filter in the Texture class and is supposed to be used when nothing is specified for the min filter in the j3m file:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-core/src/main/java/com/jme3/texture/Texture.java#L319

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-plugins/src/main/java/com/jme3/material/plugin/export/material/J3MOutputCapsule.java#L142-L150

but it does not work as expected because J3MLoader will  set the mips generation flag to true and this causes TextureProcessor to change the min filter to `Trilinear` on texture when loaded with asset manager:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/ceb356462aeedf7769dced52e1d5572b01deca31/jme3-core/src/main/java/com/jme3/texture/TextureProcessor.java#L64-L68

*If someone manually adds the `BilinearNoMipMaps` inside j3m file then it will work fine and will not get changed to `Trilinear` after loading but the generate mips flag will still be set to true.* 

**Solution**:

Those can be fixed by checking this inside the min filter and setting the flag only if it is required.

We would need to override the `applyToTextureKey` method :

```

            @Override
            public void applyToTextureKey(final String option, final TextureKey textureKey) {
                Texture.MinFilter minFilter = Texture.MinFilter.valueOf(option);
                textureKey.setGenerateMips(minFilter.usesMipMapLevels());
            }
```

inside

https://github.com/jMonkeyEngine/jmonkeyengine/blob/befa930cf9075054ecff154e2e36a42947d2ced1/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java#L860-L868

forum post:
https://hub.jmonkeyengine.org/t/j3mloader-always-generates-mips-ignoring-minfilter-bug/46296

Ali-RS:
Updated J3MLoader to read texture mips generation flag from MinFilter. If no min filter is specified, texture will use the `BilinearNoMipMaps` by default.

Fixes #1882

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
@@ -222,6 +222,7 @@ public class J3MLoader implements AssetLoader {
         // If there is only one token on the value, it must be the path to the texture.
         if (textureValues.size() == 1) {
             textureKey = new TextureKey(textureValues.get(0), false);
+            textureKey.setGenerateMips(true);
         } else {
             String texturePath = value.trim();
 
@@ -255,6 +256,8 @@ public class J3MLoader implements AssetLoader {
             if (textureKey == null) {
                 textureKey = new TextureKey(textureValues.get(textureValues.size() - 1), false);
             }
+
+            textureKey.setGenerateMips(true);
 
             // Apply texture options to the texture key
             if (!textureOptionValues.isEmpty()) {
--- a/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/material/plugins/J3MLoader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2022 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


Ali-RS:
>If no min filter is specified, texture will use the BilinearNoMipMaps by default.

Ok, I changed this for backward compatibility reasons. If no min filter is specified inside the j3m file then the Trilinear filter will be used by default.

Read more about the issue here: https://hub.jmonkeyengine.org/t/j3mloader-always-generates-mips-ignoring-minfilter-bug/46296/32?u=ali_rs

Ali-RS:

PR #1885, opened 2022-12-24
Fix #1883 (Image class wrongly setting GL mips flags inside the constructor)

### LINKED ISSUE #1883: Image class wrongly set mips flags inside the constructor
These two constructors on `Image` class behave differently!

https://github.com/jMonkeyEngine/jmonkeyengine/blob/f92a73f91ce1a2852645cd1bb5a4c6b597faf7af/jme3-core/src/main/java/com/jme3/texture/Image.java#L749-L762

https://github.com/jMonkeyEngine/jmonkeyengine/blob/f92a73f91ce1a2852645cd1bb5a4c6b597faf7af/jme3-core/src/main/java/com/jme3/texture/Image.java#L805-L816

Look at `needGeneratedMips` and `mipsWereGenerated` flags, if `mipMapSizes` is **null** the first one won't modify the flags but the second one will do (I belive it is wrongly setting mipsWereGenerated = true)

Forum post:
https://hub.jmonkeyengine.org/t/issue-in-image-class-bug/46297

Ali-RS:
If we pass null for `int[] mipMapSizes` argument in the second constructor, this will wrongly set `mipsWereGenerated` flag to true. Fixed it to behave like the first constructor.

Fixes #1883

### DIFF
--- a/jme3-core/src/main/java/com/jme3/texture/Image.java
+++ b/jme3-core/src/main/java/com/jme3/texture/Image.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2022 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


Ali-RS:

PR #1886, opened 2022-12-24
Made Gltf extension loaders non-static to avoid concurrency issues

Ali-RS:
Now each GltfLoader instantiated via ThreadLocal will have its own instances of extension loaders.


Ali-RS:

PR #1893, opened 2023-01-05
Fix #1892 (TestChooser does not show classes list when run with java 8)

### LINKED ISSUE #1892: TestChooser does not display test class list when run with java 8
From the IDE it works just fine but when I run the JME examples jar file I extracted from jME3.6.0-alpha1.zip with java 8 it does not show the class list. Running with java 11 and 17 it works fine.

Screenshot_2023-01-05_14-19-26

Ali-RS:
Happens that java 8 adds "/" at the end of the directory name when reading from the jar filesystem. This patch checks if there is a "/" appended at the end of the directory name and removes it!

Resolves issue #1892


Ali-RS:

Issue #1891, opened 2023-01-04
incorrect Nifty GUI layout in full-screen LWJGL v3 context


stephengold:
I believe I found a regression in JME 3.6. It occurs when Nifty GUI is running in full-screen LWJGL v3 context. Nifty seems to believe the mouse cursor is about 100 pixels above where it appears onscreen.

I've observed this phenomenon in several apps. It is particularly noticeable in the `TestIssue1013` app of jme3-examples.  Clicking on the "Restart Context" button has no effect. In order to highlight the button, the cursor must be placed about 100 px above the button. Clicking in that area of the GUI will trigger a restart.

JME 3.5.2 doesn't have this issue, but I've found it in "master" branch as far back as May 2022. I'm attempting to determine which commit introduced the regression.

stephengold:
The regression occurred with PR #1750 .

Ali-RS:
I can not reproduce this.

stephengold:
I'll attempt to characterize it further and determine the root cause.

stephengold:
Something in PR 1750 altered how the TestIssue1013 GUI gets laid out. Prior to 1750 it looked like this:

![image](https://user-images.githubusercontent.com/2287609/210670670-e1952b1f-3ab3-4037-aaa1-2470f5ecedab.png)

After 1750 it looked like this:

![image](https://user-images.githubusercontent.com/2287609/210670732-5263e1c3-1c92-423c-8801-d8d22c25f1a9.png)

stephengold:
Since PR 1750, there's a situation during initialization when the height of the viewport changes from 1002 to 1080. My suspicion is that the GUI gets laid out for a display 1002-pixels tall.

stephengold:
Yes, in `NiftyJmeDisplay.initialize()` line 279, `this.height` gets set to 1002.

Ali-RS:
I randomly could reproduce this issue once!

stephengold:
I suspect this is related to #1793. If the display size changes, JME is supposed to notify Nifty so it can re-do the layout.

stephengold:
The workaround for #1793 was integrated to "master" branch on 5 April 2022. Yet even after that commit, this issue still occurs. Here's why:

+ `NiftyJmeDisplay` has two `initialize()` methods, which are invoked at different times.
+ The 4-argument method gets invoked first. It sets the `h` field (cached window height) to 1002.
[...]

PR #1895, opened 2023-01-08
jme3-niftygui:  solve issue #1891 (incorrect fullscreen layout)

stephengold:
I've tested this with windowed and full-screen contexts, with LWJGL v2 and v3.


Ali-RS:
Thanks

Ali-RS:

Issue #1215, opened 2019-11-14
UnsatisfiedLinkError: libjawt.so: version `SUNWprivate_1.1' not found (lwjgl2)


vushu:
when trying to run using java 13 this error occurs:
```
INFO: Running on jMonkeyEngine 3.3-alpha5
 * Branch: HEAD
 * Git Hash: bb32b88
 * Build Date: 2019-09-25
Exception in thread "jME3 Main" java.lang.UnsatisfiedLinkError: /home/vushu/coding/java/jmonkey/magicman/liblwjgl64.so: /home/vushu/.sdkman/candidates/java/13.0.1-open/lib/libjawt.so: version `SUNWprivate_1.1' not found (required by /home/vushu/coding/java/jmonkey/magicman/liblwjgl64.so)
        at java.base/java.lang.ClassLoader$NativeLibrary.load0(Native Method)
        at java.base/java.lang.ClassLoader$NativeLibrary.load(ClassLoader.java:2440)
        at java.base/java.lang.ClassLoader$NativeLibrary.loadLibrary(ClassLoader.java:2497)
        at java.base/java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2694)
        at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2627)
        at java.base/java.lang.Runtime.load0(Runtime.java:744)
        at java.base/java.lang.System.load(System.java:1873)
        at org.lwjgl.Sys$1.run(Sys.java:70)
        at java.base/java.security.AccessController.doPrivileged(AccessController.java:312)
        at org.lwjgl.Sys.doLoadLibrary(Sys.java:66)
        at org.lwjgl.Sys.loadLibrary(Sys.java:87)
        at org.lwjgl.Sys.<clinit>(Sys.java:117)
        at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:210)
        at java.base/java.lang.Thread.run(Thread.java:830)
```

vushu:
a work around is to use 13.0.1-zulu instead of open the java 13 openjdk

stephengold:
Does it also occur with earlier JDKs, such as java11 JDKs?

vushu:
I just tested the earlier version and It works fine using openjdk 12 and 11

Ali-RS:
Have you also tried with LWJGL3?

Ali-RS:
[...]

Issue #947, opened 2018-11-04
fullscreen mode in LWJGL2 on Linux crashes during startup


RossComputerGuy:
I get this exception while trying to put my game into any fullscreen resolution. Is there any way to go into fullscreen from the dialog that doesn't have this problem?
```
java.lang.RuntimeException: Unable to find fullscreen display mode matching settings
	at com.jme3.system.lwjgl.LwjglDisplay.createContext(LwjglDisplay.java:79)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:113)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:211)
	at java.lang.Thread.run(Thread.java:748)
```

stephengold:
This appears to be a troubleshooting question. Such questions are best answered at the Forum:
    https://hub.jmonkeyengine.org/

MeFisto94:
I guess Linux and lwjgl3?
Then it's something we should really take care of. The Clue was that the settings dialog is falsely setting the frequency (setFrequency) as it's somehow differently reported by AWT/whatever we use to get the resolutions

Ev1lbl0w:
I had this problem too some time ago and reported it [here](https://hub.jmonkeyengine.org/t/error-cant-launch-jme3-projects-in-fullscreen/39625). It has not yet been fixed AFAIK but for anyone willing to dig deeper, here's what I found out:
- As @MeFisto94 said, this problem happens because LWJGL and AWT are reporting different refresh rates for the same resolution. Forcing the refresh rate to 60Hz in my case makes jMonkey apps able to go fullscreen.
- This problem is happening with the nvidia proprietary driver for linux, in my case. Switching to `nouveau` fixed these problems to me, so I encourage other people to test this as well. Back then I had a legacy driver (340.X), but even with the latest driver (415.X) this problem persists.

louhy:
I see these strange refresh rates coming from AWT too. Linux, NVidia proprietary drivers. I don't know if there's a fix, maybe a workaround, but that workaround (ex. force 60Hz on Linux) may just create a problem for someone else, no?

riccardobl:
If this happens on linux with lwjgl2, it might be related to this https://hub.jmonkeyengine.org/t/fix-for-unable-to-find-fullscreen-display-mode-matching-settings-for-multimonitor-configurations-in-linux/41510 .
[...]

PR #1902, opened 2023-01-10
jme3-lwjgl:updated to lwjgl v2.9.4 hosted by org.jmonkeyengine.

### LINKED ISSUE #1247: Linux: Native Crash lwjgl2 OpenJDK 11.0.5
So, originally this has been discovered by @jayfella but I don't think we've been tracking this officially already, thus:

Just trying to run the most simple example (here app.start()) using lwjgl2 fails as follows:
```
Jan. 11, 2020 3:11:12 NACHM. com.jme3.system.JmeDesktopSystem initialize
INFORMATION: Running on jMonkeyEngine 3.3.0-beta1
 * Branch: HEAD
 * Git Hash: bd1b6d2
 * Build Date: 2019-12-22
Inconsistency detected by ld.so: dl-lookup.c: 111: check_match: Assertion `version->filename == NULL || ! _dl_name_match_p (version->filename, map)' failed!
```
However opposing to the claim this should be working with 11.0.4, I've seen reports about 11.0.3 [here](https://unix.stackexchange.com/q/532054). 

Either way this is unfortunate as 11.0.5 is the one available and I think will replace Java 8 soon.
I am not sure if we can do anything about it, but it should be tracked so people who encounter this know it.

Edit: https://github.com/lattera/glibc/blob/master/elf/dl-lookup.c#L111
this is why it fails, and I am uncertain how openJDK could break that other than by changing the way they invoke glibc to load symbols.

Can anyone manage to find a tracking bug for openJDK? Maybe they don't know it as everyone is just downgrading to openJDK8?

Ali-RS:
We now maintain our own lwjgl2 repo: https://github.com/jMonkeyEngine/lwjgl2

See the release page for what is changed in v2.9.4

Fixes #1247, #1215, and partially https://github.com/jMonkeyEngine/jmonkeyengine/issues/947

### DIFF
--- a/jme3-lwjgl/build.gradle
+++ b/jme3-lwjgl/build.gradle
@@ -1,7 +1,10 @@
 dependencies {
     api project(':jme3-core')
     api project(':jme3-desktop')
-    api 'org.lwjgl.lwjgl:lwjgl:2.9.3'
+
+    api 'org.jmonkeyengine:lwjgl:2.9.4'
+    runtimeOnly 'org.jmonkeyengine:lwjgl-platform:2.9.4'
+
     /*
      * Upgrades the default jinput-2.0.5 to jinput-2.0.9 to fix a bug with gamepads on Linux.
      * See https://hub.jmonkeyengine.org/t/linux-gamepad-input-on-jme3-lwjgl-splits-input-between-two-logical-gamepads


Ali-RS:

PR #1915, opened 2023-01-15
jme3-core: make PrimitiveAllocator default

Ali-RS:
The ReflectionAllocator uses unsafe and needs to add special flags to work on java 11+ ("--add-opens=java.base/jdk.internal.ref=ALL-UNNAMED", "--add-exports=java.base/sun.nio.ch=ALL-UNNAMED"). It might break again in the future.

Related forum post:
https://hub.jmonkeyengine.org/t/deprecating-reflectionallocator/46350

Related issue: #1674

### DIFF
--- a/jme3-core/src/main/java/com/jme3/util/BufferAllocatorFactory.java
+++ b/jme3-core/src/main/java/com/jme3/util/BufferAllocatorFactory.java
@@ -1,3 +1,34 @@
+/*
+ * Copyright (c) 2009-2023 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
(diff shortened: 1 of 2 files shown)


theMinka:
How is the status of this pull request? We have faced the same issues in our game and looking forward for your fix. :)

Ali-RS:
Hi,

I am not sure when/if this would be merged but you can switch it to the primitive allocator by passing a system property:
`-Dcom.jme3.BufferAllocatorImplementation=com.jme3.util.PrimitiveAllocator`

theMinka:
Oh, I wasn't aware of the lwjgl3 option. I switched our projet to it and everything was working as before, just without the Error from the ReflectionAllocator. Thank you for this nice little hint. :D

riccardobl:
Closing this since we have a new set of native allocators

PR #1916, opened 2023-01-15
main.yml: removed java-8 runner image

Scrappers-glitch:
### This PR attempts to remove the java-8 runner images since we are using gradle-7.6 with a release option 8:
- [x] Removed jdk 8 runner image.
- [ ] Migrate deploy to the jdk 17 image (wip).
  
For more refer to the gradle docs on building jvm applications:
https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -98,8 +98,6 @@ jobs:
             osName: windows
           - os: macOS-latest
             osName: mac
-          - jdk: 11
-            deploy: false
           - jdk: 17
             deploy: false
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -89,7 +89,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest,windows-2019,macOS-latest]
-        jdk: [8, 11, 17]
+        jdk: [11, 17]
         include:
           - os: ubuntu-latest
             osName: linux


Ali-RS:
I converted this to draft until it is completed.

.github/workflows/main.yml
```diff
@@ -98,8 +98,6 @@ jobs:
             osName: windows
           - os: macOS-latest
             osName: mac
-          - jdk: 11
-            deploy: false
```
Ali-RS, line 102: You should revert this, and instead set `deploy: true` on jdk 17.

Ali-RS: By the way, what will happen if we just set deploy: true on jdk 17 instead of removing it from "include". Will it deploy with all OS?

Scrappers-glitch: The include just adds additional flag for each runner image.

Scrappers-glitch: I am going to explicitly evaluate the jdk11 to `deploy: false` and jdk17 to `deploy: true` and update the comments, I think it will be much better if we specify flags explicitly, do we agree on these changes ?

Ali-RS: >and jdk17 to deploy: true

I wonder if it will also deploy with mac and windows as well if you set it to `true`? I need to take a look at the docs to see how `include` works.

Scrappers-glitch: And I will also create deploy: false explicitly for other systems (Mac and windows) to avoid confusion.

Ali-RS: >And I will also create deploy: false explicitly for other systems (Mac and windows) to avoid confusion.

[mac:deploy=false, jdk17: deploy=true] 
[windows:deploy=false, jdk17: deploy=true] 

Will the `deploy` defined for jdk overwrite the one in os?

Scrappers-glitch: It shouldn't, the old code currently in the master defines only jdk 8 with true and Ubuntu with true, so the runner which runs Ubuntu jdk8 will deploy, deploy is always false unless explicitly defined by the runner as true.

Scrappers-glitch: I will see.

Scrappers-glitch: Long story short, we shouldn't set `deploy: true` on a jdk runner because it will override the previously defined flag.

Ali-RS: I see.

.github/workflows/main.yml
```diff
@@ -98,8 +98,6 @@ jobs:
             osName: windows
           - os: macOS-latest
             osName: mac
```
Ali-RS, line 100: I think we need to also explicitly specify `deploy: false` on windows and mac os.

Ali-RS:
I finished this in #1922

Ali-RS:
@Scrappers-glitch, I'm sorry about this! I agree with you that I must have asked before I wanted to close this and submit a new PR, my fault. Sorry again!

PR #1922, opened 2023-01-18
main.yml: deploy with jdk17

Ali-RS:
This should not break java 8 compatibility because of the "release = 8" compile option specified in the Gradle build. I hope!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
           name: android-natives
           path: build/native
 
-  # Build the engine, we only deploy from ubuntu-latest jdk8
+  # Build the engine, we only deploy from ubuntu-latest jdk17
   BuildJMonkey:
     needs: [BuildAndroidNatives]
     name: Build on ${{ matrix.osName }} jdk${{ matrix.jdk }}
@@ -96,11 +96,13 @@ jobs:
             deploy: true
           - os: windows-2019
             osName: windows
+            deploy: false
           - os: macOS-latest
             osName: mac
-          - jdk: 11
             deploy: false
-          - jdk: 17
+          - jdk: 8
+            deploy: false
+          - jdk: 11
             deploy: false
 
     steps:


.github/workflows/main.yml
```diff
@@ -96,11 +96,13 @@ jobs:
             deploy: true
           - os: windows-2019
             osName: windows
+            deploy: false
           - os: macOS-latest
             osName: mac
-          - jdk: 11
             deploy: false
-          - jdk: 17
+          - jdk: 8
```
Scrappers-glitch, line 103: I wonder why we still use a java 8 runner here, if all the builds use java 11+ ? I think it will be misleading if the runner title has java 8 but gradle doesn't use java-8.

Ali-RS: I downloaded the [jMonkeyEngine 3.6.0-alpha3 artifacts uploaded to GitHub release](https://github.com/jMonkeyEngine/jmonkeyengine/releases/tag/v3.6.0-alpha3) which was built and uploaded by jdk8 runner and also downloaded the maven.zip and release.zip artifacts from [jMonkeyEngine 3.6.0-alpha3 GitHub action](https://github.com/jMonkeyEngine/jmonkeyengine/actions/runs/3922595564) page. All of them are built with java 8, and there is a "Created-By: 1.8 (Temurin)" in the MANIFEST.MF of all the jars.

So they do not use java 11, which is why I am still keeping the jdk8 job.

I still do not understand why the jars on the [Maven Central Repo]( https://repo1.maven.org/maven2/org/jmonkeyengine/jme3-core/3.6.0-alpha3/) are built with java 11 and have "Created-By: 11 (Eclipse Adoptium)" in their MANIFEST.MF.

I doubt if the below task is somehow being screwed:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/458e9ec210c131f92f4d8e5f93f01f1dc93b0a5b/.github/workflows/main.yml#L326-L340

Ali-RS: Oops, It looks like DeployRelease job has no java installed on it so gradle must have installed the JDK 11 by default using toolchain.
https://github.com/jMonkeyEngine/jmonkeyengine/blob/458e9ec210c131f92f4d8e5f93f01f1dc93b0a5b/.github/workflows/main.yml#L293-L341

Ali-RS: I opened a new issue for that: https://github.com/jMonkeyEngine/jmonkeyengine/issues/1923
Let's continue the discussion there.

Issue #1928, opened 2023-01-29
FBX import throws OutOfMemoryError


stephengold:
I'm testing current "master" branch with FBX assets downloaded from CGtrader, Mixamo, and Sketchfab.
Attempting to load any file with version > 7400 results in an `OutOfMemoryError`.
For example:
```console
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.ArrayList.<init>(ArrayList.java:156)
	at com.jme3.scene.plugins.fbx.file.FbxElement.<init>(FbxElement.java:61)
	at com.jme3.scene.plugins.fbx.file.FbxReader.readFBXElement(FbxReader.java:92)
	at com.jme3.scene.plugins.fbx.file.FbxReader.readFBXElement(FbxReader.java:102)
	at com.jme3.scene.plugins.fbx.file.FbxReader.readFBX(FbxReader.java:77)
	at com.jme3.scene.plugins.fbx.FbxLoader.loadData(FbxLoader.java:149)
	at com.jme3.scene.plugins.fbx.FbxLoader.load(FbxLoader.java:110)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:439)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:444)
	at jme3test.helloworld.HelloJME3.loadModel(HelloJME3.java:83)
	at jme3test.helloworld.HelloJME3.simpleInitApp(HelloJME3.java:57)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:240)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:139)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:221)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

The proximate cause of the crash is that the importer reads a huge number of properties (`propCount`) for an element --- as in more than 10^9 properties!

Apparently there were fundamental changes to how FBX elements are written between file versions 7400 and 7500.
In particular:
[...]

PR #1929, opened 2023-01-29
solve issue #1928 (OutOfMemoryError in FBX importer)

stephengold:
Here's a proposed fix for issue #1928.

+ The constants for `BLOCK_SENTINEL_LENGTH` and `BLOCK_SENTINEL_DATA` were removed, since these values are no longer constant.
+ Two new methods were added to `FbxFile` to provide values that depend on the FBX file version.
+ An `FbxFile` instance was added to the arguments of `readFBXElement()`, giving the method access to the values.
+ A utility method was added, to test whether a byte array contains all zeroes.


stephengold:

Issue #1930, opened 2023-01-29
NPE in FbxLayerElement.fromElement()


stephengold:
I'm testing current "master" branch with FBX assets downloaded from CGtrader, Mixamo, and Sketchfab.
Attempting to import certain FBX files causes the following crash:
```console
Jan 28, 2023 11:10:18 PM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.NullPointerException: Cannot read the array length because "layerElement.dataIndices" is null
	at com.jme3.scene.plugins.fbx.mesh.FbxLayerElement.fromElement(FbxLayerElement.java:202)
	at com.jme3.scene.plugins.fbx.mesh.FbxMesh.fromElement(FbxMesh.java:91)
	at com.jme3.scene.plugins.fbx.obj.FbxObjectFactory.createObject(FbxObjectFactory.java:192)
	at com.jme3.scene.plugins.fbx.FbxLoader.loadObjects(FbxLoader.java:195)
	at com.jme3.scene.plugins.fbx.FbxLoader.loadData(FbxLoader.java:161)
	at com.jme3.scene.plugins.fbx.FbxLoader.load(FbxLoader.java:110)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:439)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:444)
	at jme3test.helloworld.HelloJME3.loadModel(HelloJME3.java:88)
	at jme3test.helloworld.HelloJME3.simpleInitApp(HelloJME3.java:67)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:240)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:139)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:221)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

The comment at FbxLayerElement.java:200 indicates that the intent was to copy `dataIndices` to `data`.
If `dataIndices` is null,the entire block should probably be skipped.

[...]

PR #1931, opened 2023-01-29
solve issue #1930 (NPE in FbxLayerElement)

stephengold:
Skip the copy if `dataIndices` is null.

This enables import of 2 models that used to crash due to issue #1930:
+ "Old_house_block_01.fbx"  (file version 7400) and
+ "Urban_communications.fbx" (file version 6100)


stephengold:

Issue #1933, opened 2023-01-29
UnsupportedOperationException in FbxNode.setWorldBindPose()


stephengold:
I'm testing current "master" branch with FBX assets downloaded from CGtrader, Mixamo, and Sketchfab.
Attempting to import a certain FBX file causes the following crash:
```console
Jan 28, 2023 11:58:05 PM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.UnsupportedOperationException: Bind poses don't match
	at com.jme3.scene.plugins.fbx.node.FbxNode.setWorldBindPose(FbxNode.java:163)
	at com.jme3.scene.plugins.fbx.FbxLoader.applyBindPoses(FbxLoader.java:282)
	at com.jme3.scene.plugins.fbx.FbxLoader.load(FbxLoader.java:113)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:439)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:444)
	at jme3test.helloworld.HelloJME3.loadModel(HelloJME3.java:88)
	at jme3test.helloworld.HelloJME3.simpleInitApp(HelloJME3.java:67)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:240)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:139)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:221)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

The failing code tests for exact equality between two bind-pose matrices:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/7a5d085ff5274b651401d1460a7bf72abd006ac4/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/node/FbxNode.java#L160-L165

For the file I'm studying, the actual differences between the matrices are less than 10^-19, so I assume they're caused by roundoff error somewhere upstream. There ought to be some tolerance in the comparison.

PR #1936, opened 2023-01-29
solve issue #1933 (unsupported operation in FbxNode)

stephengold:
Added the `isSimilar()` method to `Matrix4f` since that seems the best place for it. The implementation is closely based on that of `Vector3f.isSimilar()`.

Hardcoded the tolerance to `1e-6f`, even though the largest difference seen so far was `< 1e-19`. I suspect that differences smaller than `1e-6f` will be imperceptible.


stephengold:

Issue #1937, opened 2023-01-29
NPE in FbxObject.unsupportedConnectObjectProperty()


stephengold:
I'm testing current "master" branch with FBX assets downloaded from CGtrader, Mixamo, and Sketchfab.
Attempting to import a certain FBX file causes the following crash:
```console
Jan 29, 2023 10:11:35 AM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.NullPointerException: Cannot invoke "com.jme3.scene.plugins.fbx.obj.FbxObject.getFullClassName()" because "object" is null
	at com.jme3.scene.plugins.fbx.obj.FbxObject.unsupportedConnectObjectProperty(FbxObject.java:134)
	at com.jme3.scene.plugins.fbx.anim.FbxAnimCurveNode.connectObjectProperty(FbxAnimCurveNode.java:129)
	at com.jme3.scene.plugins.fbx.FbxLoader.connectObjects(FbxLoader.java:261)
	at com.jme3.scene.plugins.fbx.FbxLoader.loadData(FbxLoader.java:163)
	at com.jme3.scene.plugins.fbx.FbxLoader.load(FbxLoader.java:110)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:439)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:444)
	at jme3test.helloworld.HelloJME3.loadModel(HelloJME3.java:88)
	at jme3test.helloworld.HelloJME3.simpleInitApp(HelloJME3.java:80)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:240)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:139)
	at com.jme3.system.lwjgl.LwjglAbstractDisplay.run(LwjglAbstractDisplay.java:221)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

The crash occurs because in `FbxLoader.connectObjects()`, the child object (with ID 67443872) doesn't exist in `objectMap`:

[...]

PR #1938, opened 2023-01-29
solve issue #1937 (NPE in FbxObject)

stephengold:
The solution avoids the immediate crash and logs a warning with clues to the cause.


stephengold:

Issue #1939, opened 2023-02-01
NullPointerException in FbxMesh.applyCluster()


stephengold:
Using the current "master" branch of JME, I was attempting to load a skinned FBX model without any animation tracks (T-pose only):
```console
Jan 31, 2023 6:43:07 PM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.NullPointerException: Cannot read the array length because "positionIndices" is null
	at com.jme3.scene.plugins.fbx.mesh.FbxMesh.applyCluster(FbxMesh.java:127)
	at com.jme3.scene.plugins.fbx.mesh.FbxMesh.toJmeObject(FbxMesh.java:212)
	at com.jme3.scene.plugins.fbx.mesh.FbxMesh.toJmeObject(FbxMesh.java:58)
	at com.jme3.scene.plugins.fbx.obj.FbxObject.getJmeObject(FbxObject.java:113)
	at com.jme3.scene.plugins.fbx.node.FbxNode.toJmeObject(FbxNode.java:399)
	at com.jme3.scene.plugins.fbx.node.FbxNode.toJmeObject(FbxNode.java:71)
	at com.jme3.scene.plugins.fbx.obj.FbxObject.getJmeObject(FbxObject.java:113)
	at com.jme3.scene.plugins.fbx.node.FbxNode.createScene(FbxNode.java:496)
	at com.jme3.scene.plugins.fbx.node.FbxNode.createScene(FbxNode.java:503)
	at com.jme3.scene.plugins.fbx.node.FbxNode.createScene(FbxNode.java:503)
	at com.jme3.scene.plugins.fbx.FbxLoader.constructSceneGraph(FbxLoader.java:412)
	at com.jme3.scene.plugins.fbx.FbxLoader.load(FbxLoader.java:124)
	at com.jme3.asset.DesktopAssetManager.loadLocatedAsset(DesktopAssetManager.java:272)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:388)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:439)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:444)
	at jme3test.helloworld.HelloJME3.loadModel(HelloJME3.java:90)
[...]

PR #1940, opened 2023-02-01
solve issue #1939 [NPE in FbxMesh.applyCluster()]

stephengold:
Cleanly handle the absence of animation data in 2 places:
+ in `FbxCluster.fromElement()`, create empty arrays of weights and indices
+ in `FbxLoader.constructAnimations()`, skip over any animations without keyframes


stephengold:

Issue #1879, opened 2022-12-11
Skinning.glsllib fails compilation when NUM_WEIGHTS_PER_VERT = 1


zzuegg:
Line 53:
`tangent = m_BoneMatrices[int(inHWBoneIndex.x)] * tangent;`
is 
`vec3 = mat4*vec3`

I  will fix it when someone more capable at math says how. i guess:
``tangent = (m_BoneMatrices[int(inHWBoneIndex.x)] * vec4(tangent,1.0f)).xyz;``

should work?

stephengold:
I'm not very knowledgeable regarding GLSL.

One thing I don't understand: 
+ Line 7 has `#define NUM_WEIGHTS_PER_VERT 4` so how could `NUM_WEIGHTS_PER_VERT` ever be 1?

Mathematically, I don't think `tangent = (m_BoneMatrices[int(inHWBoneIndex.x)] * vec4(tangent,1.0f)).xyz;` is correct, since `m_BoneMatrices[int(inHWBoneIndex.x)]` probably includes a translation component, which shouldn't be applied to direction vectors such as normals and tangents.

In light of line 66, I would expect rotation of tangents to be analogous with the rotation of normals. Something like:
```glsl
        tangent = (mat3(m_BoneMatrices[int(inHWBoneIndex.x)][0].xyz,
                        m_BoneMatrices[int(inHWBoneIndex.x)][1].xyz,
                        m_BoneMatrices[int(inHWBoneIndex.x)][2].xyz) * tangent);
```
should work.

Or, if we're concerned with readability, compute a `rotMat` that can be applied to both normals and tangents:
```glsl
        mat3 rotMat = mat3(m_BoneMatrices[int(inHWBoneIndex.x)][0].xyz,
                           m_BoneMatrices[int(inHWBoneIndex.x)][1].xyz,
                           m_BoneMatrices[int(inHWBoneIndex.x)][2].xyz);
        tangent = rotMat * tangent;
        normal = rotMat * normal;
```

PR #1942, opened 2023-02-08
solve issue #1879 (compile-time error in Skinning.glsllib)

stephengold:
A straightforward fix for issue #1879, rotating tangents using the same matrix that's used to rotate normals.

I've run "TestJaime.java" as a sanity check, however I don't see a way to test this fix. That's because I don't see how the `#if NUM_WEIGHTS_PER_VERT == 1` code in "Skinning.glsllib" ever gets compiled.

@zzuegg could you please test this fix or explain how it should be tested?


zzuegg:
You are correct. I only noticed it as compilation error without having an actual test case. I suspected it is only used in super rare edge cases like single bone models.  Regarding the doc in the animation class, there should always be 4 bones and 4 weights assigned to each vertex. 

I am not sure whats the reason for the code to be there at all. The code in question was added in 2013. At this point i suspect there was the idea to support a variable amount of weights at some point but 4 it is since then. 

I also did not find a reference to NUM_WEIGHTS_PER_VERT outside of Skinning.glsllib.

Ali-RS:

Issue #1806, opened 2022-05-03
New animation system: shared stateful static objects cause unexpected threading issues


pspeed42:
In general, JME operates in a "threading not supported" mode but in a "you are fine to do whatever until you attach it to the scene graph" sort of way.  Meaning that you can load and manipulate Spatials from separate threads but once you attach them to the render-thread managed scene, then hands off.

The animation system internally uses several global objects that keep state.  By default, these are shared across every loaded animation, no matter where it was loaded.

Examples at the lowest level are AnimInterpolators.NLerp and AnimInterpolators.LinearVec3f:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/interpolator/AnimInterpolators.java#L46

...which are then used by one shared FrameInterpolator instance: 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/interpolator/FrameInterpolator.java#L41
...that is the default for MorphTrack and TransformTrack:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java#L55
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/TransformTrack.java#L53

And while you could meticulously set your own:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/TransformTrack.java#L302
...it won't be serialized:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/TransformTrack.java#L332

At a high level, the right solution would be to share a single FrameInterpolator instance (and its own instances of NLerp, LinearVector3f, etc.) across the AnimComposer/SkinningControl and not across the whole Java virtual machine.  I'm still trying to figure out the simplest/least-impactful way to make that happen.  (Locally I've modified all of these cases to create their own instances and it solves the problems.)

As it stands, if you load Jaime.j3o into a scene and let him run around while meanwhile loading a totally separate model in another thread... posing or modifying that animation from that other thread will cause Jaime to glitch or crash.  Which breaks the mental threading model under which JME normally operates.

Or in my case, when having a game load thread animate physics collision shapes while the rendering thread is trying to play animations:
[...]

PR #1943, opened 2023-02-08
solve issue #1806 (global FrameInterpolator violates threading model)

stephengold:
Instead of a `FrameInterpolator` instance for each `AnimComposer`/`SkinningControl`, this solution provides an instance for each thread. The per-thread instances are allocated using `ThreadLocal.withInitial()`.

To maximize compatibility with JME v3.5, the public `FrameInterpolator.DEFAULT` field is retained, however it is deprecated and no longer used in the Engine.

Similarly, the `setFrameInterpolator()` methods in `MorphTrack` and `TransformTrack` are retained. However, a `null` interpolator now has special meaning to these classes, namely that the per-thread default instance should be used.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
+++ b/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2023 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


Ali-RS:
When a thread gets removed does its cache also get cleared from ThradLocal?

stephengold:
https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html

"after a thread goes away, all of its copies of thread-local instances are subject to garbage collection (unless other references to these copies exist)"

Ali-RS (approved):

Ali-RS:
Thanks

Unless there's something to discuss, I plan to integrate this PR in about 24 hours.

pspeed42:
I was asked to review this one and so I glanced at the code.  I think this solves my original problem but I don't have time to test it.

It's probably not the "correct" solution from the standpoint of "state shared between models when you don't expect it to" but the correct solution (make sure each anim/skin has its own instance) is really difficult and the provided solution will probably fix all cases a user will encounter.

Just something to keep in the back of our minds if there is ever another future case where JME allows tandem access to the state of multiple spatials (for example, looping in a tween that relies on access the animation state of a separate spatial or something... just trying to think of ways it could possibly happen.)  But in that case, an argument could also be made that frame interpolator should never be shared.

Anyway, short of being able to test it myself, I think this solution is fine.

PR #1944, opened 2023-02-09
Fix issue #1558 (TestAWTPanels crashes with LWJGL v3 on Linux)

### LINKED ISSUE #1558: TestAWTPanels crashes with LWJGL v3 on Linux

Ali-RS:
This resolves the JVM crash happening when running TestAWTPanels on Linux and LWJGL 3. Seems setting the Look And Feel on Swing UIManager prevents the crash.

Fix #1558 

Still need to figure out a solution for IllegalStateException happening on JME 3.6.0-beta1. See https://github.com/jMonkeyEngine/jmonkeyengine/issues/1558#issuecomment-1423951041 
But I think this should be in a separate PR.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/awt/TestAwtPanels.java
+++ b/jme3-examples/src/main/java/jme3test/awt/TestAwtPanels.java
@@ -1,3 +1,34 @@
+/*
+ * Copyright (c) 2009-2023 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


Ali-RS:
Unless there's something to discuss, I plan to self-integrate this PR in about 24 hours and close issue #1558.

PR #1949, opened 2023-02-11
Fix #1945 (IllegalStateException when running TestAWTPanels with LWJGL 3)

Ali-RS:
Log a "not supported" error on AwtPanelsListener.reshape/rescale method instead of throwing an exception. 

Fix #1945

PR #1950, opened 2023-02-11
LwjglWindow: initialize oldFramebufferWidth & oldFramebufferHeight (LWJGL 3)

### LINKED ISSUE #1945: IllegalStateException when running TestAWTPanels with 3.6.0-beta1and and LWJGL 3
Testing with JME 3.6.0-beta1 and LWJGL 3 on Linux I am getting this exception in:

```
> Task :TestAwtPanels.main()
Feb 09, 2023 1:45:24 PM com.jme3.system.JmeDesktopSystem initialize
INFO: Running on jMonkeyEngine 3.6.0-beta1
 * Branch: HEAD
 * Git Hash: ee91fb8
 * Build Date: 2023-01-24
Feb 09, 2023 1:45:25 PM com.jme3.app.LegacyApplication handleError
java.lang.IllegalStateException
	at com.jme3.system.awt.AwtPanelsContext$AwtPanelsListener.reshape(AwtPanelsContext.java:67)
	at com.jme3.system.lwjgl.LwjglWindow.updateSizes(LwjglWindow.java:393)
	at com.jme3.system.lwjgl.LwjglWindow.createContext(LwjglWindow.java:370)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:582)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:704)
	at java.lang.Thread.run(Thread.java:750)
```

https://github.com/jMonkeyEngine/jmonkeyengine/blob/925ff4561b29fff22fe2e47ecc63d129cb3a4000/jme3-desktop/src/main/java/com/jme3/system/awt/AwtPanelsContext.java#L65-L68

I think this has something to do with changes made in PR https://github.com/jMonkeyEngine/jmonkeyengine/pull/1794 
>This results in some redundant callbacks (multiple callbacks for a single createContext())

Ali-RS:
… to the values from AppSettings. 

And fixed a typo from PR #1794

Should also, fix #1945
I plan to apply PR #1949 as well.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -287,7 +287,7 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
         final GLFWVidMode videoMode = glfwGetVideoMode(glfwGetPrimaryMonitor());
         int requestWidth = settings.getWindowWidth();
         int requestHeight = settings.getWindowHeight();
-        if (requestWidth <= 0 || requestWidth <= 0) {
+        if (requestWidth <= 0 || requestHeight <= 0) {
             requestWidth = videoMode.width();
             requestHeight = videoMode.height();
         }
@@ -361,6 +361,8 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
         });
 
         allowSwapBuffers = settings.isSwapBuffers();
+        oldFramebufferWidth = settings.getWidth();
+        oldFramebufferHeight = settings.getHeight();
 
         // Create OpenCL
         if (settings.isOpenCLSupport()) {


Ali-RS:

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -361,6 +361,8 @@ public void invoke(final long window, final int width, final int height) {
         });
 
         allowSwapBuffers = settings.isSwapBuffers();
+        oldFramebufferWidth = settings.getWidth();
+        oldFramebufferHeight = settings.getHeight();
```
stephengold, line 365: As I mentioned at #1945, I'm not convinced this is the right thing to do.

I'd rather omit this delta and leave `oldFrameBufferHeight` and `oldFrameBufferWidth` at their initial values of `0`.

stephengold (changes requested):

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
Ali-RS: what about initializing them to the width and height values read from `glfwGetFramebufferSize(window, width, height);`?

stephengold: That would be better than using `settings`, but I still prefer initializing to zero.

That way, if the app registers a `SystemListener` early enough, it will get a `reshape()` callback after the window is created. That could be useful. It's the ideal time to lay out the GUI, for instance.

Ali-RS: I see, then let's close this PR and I will submit a new one for the`if (requestWidth <= 0 || requestWidth <= 0)`.

Thank you so much for reviewing

PR #1951, opened 2023-02-11
Add WaterFilter.getReflectionView method

Ali-RS:
I use this to grab and hook the water reflection ViewPort into SkyControl for automatically updating the day/night color.

```
WaterFilter water = fpp.getFilter(WaterFilter.class);
if (water != null) {
     skyControl.getUpdater().addViewPort(water.getReflectionView());
}
```

What do you think, is this a good addition? 

Edit:
In my case, I am setting an empty node as a "reflection scene" (I am doing a cartoon-style water and realistic reflection is not needed also it is a bit expensive) but I still need to be able to reflect the day/night color which I control by the viewport color via SkyControl.


Ali-RS:
May this be a candidate for v3.6?

stephengold:
A worthy addition to the Engine, but too late for version 3.6 .

Ali-RS:
Ok, thanks

PR #1952, opened 2023-02-12
Catch and log all uncaught exceptions propagated from simpleUpdate().

danielperano:
I recently encountered an issue where exceptions were propagating out of my code and silently being swallowed without the usual "uncaught exception" window being shown or any logging. Since jME already tries to handle this case, I was surprised after a couple of hours of debugging to find that the origin was an exception coming from within `simpleUpdate()` (in my case it happened in a subtle and rather difficult to debug scenario).

The solution I added in my codebase was wrapping the body of `simpleUpdate()` in a try-catch block and logging the stack trace. Since `SimpleApplication` aims to be a consistent and easy to use one-stop-shop for app-level concerns, it would be nice if it handled this situation as well.

I'm not entirely convinced that this is the best way to do this (especially given jME's existing uncaught exception handling), but it seems a reasonable starting point for further discussion - feedback is welcomed.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
+++ b/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
@@ -267,6 +267,7 @@ public abstract class SimpleApplication extends LegacyApplication {
             simpleUpdate(tpf);
         }catch(Exception e){
             logger.log(Level.WARNING, "Uncaught exception in simpleUpdate():", e);
+            throw e;
         }
 
         if (prof != null)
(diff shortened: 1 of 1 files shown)


Ali-RS:
>without the usual "uncaught exception" window being shown

In JME 3.6? If so, you need to add `jme3-awt-dialogs` into the Gradle dependencies. See https://hub.jmonkeyengine.org/t/next-engine-release/46314/71?u=ali_rs

danielperano:
This is on `3.5.1-stable`, and I see the old-style AWT exception dialogs for every other uncaught exception that I'm aware of.

Ali-RS:
May you please post the exception with the full stack trace?

Issue #1955, opened 2023-02-17
Can not play vorbis audio on Android API 31+


Ali-RS:
Apparently, JME Vorbis (Ogg) audio plugin is broken on the Android API 31+.

Here is the crash log:

```
02-17 11:49:41.417 I/com.jme3.audio.openal.ALAudioRenderer(3852): INFO Audio Renderer Information
02-17 11:49:41.417 I/com.jme3.audio.openal.ALAudioRenderer(3852):  * Device: OpenAL Soft
02-17 11:49:41.417 I/com.jme3.audio.openal.ALAudioRenderer(3852):  * Vendor: OpenAL Community
02-17 11:49:41.417 I/com.jme3.audio.openal.ALAudioRenderer(3852):  * Renderer: OpenAL Soft
02-17 11:49:41.417 I/com.jme3.audio.openal.ALAudioRenderer(3852):  * Version: 1.1 ALSOFT 1.21.1
02-17 11:49:41.417 I/com.jme3.audio.openal.ALAudioRenderer(3852):  * Supported channels: 64
02-17 11:49:41.417 I/com.jme3.audio.openal.ALAudioRenderer(3852):  * ALC extensions: ALC_ENUMERATE_ALL_EXT ALC_ENUMERATION_EXT ALC_EXT_CAPTURE ALC_EXT_DEDICATED ALC_EXT_disconnect ALC_EXT_EFX ALC_EXT_thread_local_context ALC_SOFT_device_clock ALC_SOFT_HRTF ALC_SOFT_loopback ALC_SOFT_loopback_bformat ALC_SOFT_output_limiter ALC_SOFT_pause_device
[...]

PR #1956, opened 2023-02-17
resolve issue #1955 (Can not play vorbis audio on Android API 31+)

Scrappers-glitch:
### This PR fixes the following:
- [x] Issue [#1955] by delegating the file closure to the android [`ParcelFileDescriptor`](https://github.com/jMonkeyEngine/jmonkeyengine/blob/493195eae810c4a38744e6c34ea0e7e8c48542a9/jme3-android/src/main/java/com/jme3/audio/plugins/NativeVorbisLoader.java#L71), the original resource that opens the audio asset.

### This PR adds the following: 
- [x] Java documentation.
- [x] Native.

### This PR refactors the following: 
- [x] `NativeVorbisFile#open(int, long, long)` to `NativeVorbisFile#init(int, long, long)`.
- [x] `NativeVorbisFile#close()` to `NativeVorbisFile#clearResources()`. 
- [x] `NativeVorbisFile#read(byte[], int, int)` to `NativeVorbisFile#readIntoArray(byte[], int, int)`.
- [x] `NativeVorbisFile#readFully(ByteBuffer)` to `NativeVorbisFile#readIntoBuffer(ByteBuffer)`.

### DIFF
--- a/jme3-android-native/src/native/jme_decode/com_jme3_audio_plugins_NativeVorbisFile.c
+++ b/jme3-android-native/src/native/jme_decode/com_jme3_audio_plugins_NativeVorbisFile.c
@@ -110,13 +110,18 @@ static int FileDesc_seek(void *datasource, ogg_int64_t offset, int whence)
     wrapper->current = actual_offset;
 }
 
-static int FileDesc_close(void *datasource)
+static int FileDesc_clear(void *datasource)
 {
     FileDescWrapper* wrapper = (FileDescWrapper*)datasource;
     
-    LOGI("FD close");
-    
-    return close(wrapper->fd);
+    LOGI("Clear resources -- delegating closure to the Android ParcelFileDescriptor");
+
+    /* release the file descriptor wrapper buffer */
+    free(wrapper);
+
+    wrapper = NULL;
+
+    return 0;
 }
 
 static long FileDesc_tell(void *datasource)
@@ -139,7 +144,7 @@ static long FileDesc_tell(void *datasource)
 static ov_callbacks FileDescCallbacks = {
     FileDesc_read,
     FileDesc_seek,
-    FileDesc_close,
+    FileDesc_clear,
     FileDesc_tell
 };
 
@@ -157,10 +162,10 @@ static jfieldID nvf_field_bitRate;
 static jfieldID nvf_field_totalBytes;
 static jfieldID nvf_field_duration;
 
-JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_nativeInit
+JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_preInit
   (JNIEnv *env, jclass clazz)
 {
-    LOGI("nativeInit");
+    LOGI("preInit");
     
     nvf_field_ovf = (*env)->GetFieldID(env, clazz, "ovf", "Ljava/nio/ByteBuffer;");;
     nvf_field_seekable = (*env)->GetFieldID(env, clazz, "seekable", "Z");
@@ -171,7 +176,7 @@ JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_nativeInit
     nvf_field_duration = (*env)->GetFieldID(env, clazz, "duration", "F");
 }
 
-JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_open
+JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_init
   (JNIEnv *env, jobject nvf, jint fd, jlong off, jlong len)
 {
     LOGI("open: fd = %d, off = %lld, len = %lld", fd, off, len);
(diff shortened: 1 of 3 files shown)


jme3-android/src/main/java/com/jme3/audio/plugins/NativeVorbisFile.java
```diff
@@ -16,22 +16,33 @@ public class NativeVorbisFile {
...
+     * @param offset an integer representing the start of the 
+     * @param length an integer representing the length of the 
+     * @throws IOException
+     */
+    public NativeVorbisFile(int fd, long offset, long length) throws IOException {
+        init(fd, offset, length);
     }
     
-    private native void open(int fd, long off, long len) throws IOException;
+    private native void init(int fd, long offset, long length) throws IOException;
```
Ali-RS, line 34: what about naming it "load"? just a thought

Scrappers-glitch: Hi thanks for the review, yeah i thought of this too, multiple things can work here, for example `getVorbisFromFile`, `createVorbisFromFile`, `loadVorbisFromFile`, but we have to consider the function/method original source code: 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/493195eae810c4a38744e6c34ea0e7e8c48542a9/jme3-android-native/src/native/jme_decode/com_jme3_audio_plugins_NativeVorbisFile.c#L179-L226

What it does here, is it neither opens nor loads something ("load" can mean different things, including something precreated, cached and now scheduled to be loaded by a loader), the method here basically creates and initializes a new `OggVorbis_File` from an asset file (it doesn't create only and it doesn't load from cache), i consider it as a low constructor function, so the best name is `init()` as it is also called inside a constructor.

Scrappers-glitch: In java architecture, constructors are the `<init>()` methods, so this was the idea as the original author calls it inside the constructor.

Scrappers-glitch: The `nativeInit` also, it doesn't initialize C or native buffers, it retrieves some java fields into the native environment to be used as data manipulators later between java code and C code, so it prepares the java fields into the native environment.

Ali-RS: I see, I am fine with this, thanks

Scrappers-glitch:
I modified some names (describing their sources better now), and added java docs.

jme3-android-native/src/native/jme_decode/com_jme3_audio_plugins_NativeVorbisFile.c
```diff
@@ -171,7 +176,7 @@ JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_nativeInit
     nvf_field_duration = (*env)->GetFieldID(env, clazz, "duration", "F");
 }
 
-JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_open
+JNIEXPORT void JNICALL Java_com_jme3_audio_plugins_NativeVorbisFile_init
   (JNIEnv *env, jobject nvf, jint fd, jlong off, jlong len)
 {
     LOGI("open: fd = %d, off = %lld, len = %lld", fd, off, len);
```
Ali-RS, line 182: Should this be `LOGI("init:...`

Scrappers-glitch: Yes, as i mentioned above, i will refactor old native logs soon..

jme3-android/src/main/java/com/jme3/audio/plugins/NativeVorbisFile.java
```diff
@@ -16,22 +57,69 @@ public class NativeVorbisFile {
...
+     * Seeks to a playback time relative to the decompressed pcm (Pulse-code modulation) stream.
+     * 
+     * @param time the playback seek time
+     * @throws IOException if the seek is not successful
+     */
     public native void seekTime(double time) throws IOException;
     
-    public native int read(byte[] buf, int off, int len) throws IOException;
+    /**
+     * Reads the vorbis file into a primitive byte buffer [buf] with an [offset] indicating the read start and a [length] indicating the read end on the output buffer.
```
Scrappers-glitch, line 84: Notice the documentation here is provisional and still subjected to modifications...

Ali-RS (approved):

Scrappers-glitch:
@Ali-RS For now, this is satisfying for me, i added javadocs improvements and some documentation to the native code, i will keep the rest (including the public java fields on the `NativeVorbisFile` class) until i have some time to understand things better, let  me know if there is something else to add.

Ali-RS (approved):

Ali-RS:
Thank you @Scrappers-glitch 
It looks fine to me

Ali-RS:

Issue #1958, opened 2023-02-20
It seems jmf library coming with j-ogg-all is not required?


Ali-RS:
I noticed if I exclude the jmf library coming with [j-ogg-all](https://github.com/stephengold/j-ogg-all), I can still play ogg sound in JME. Looks like jmf is not required for decoding ogg files(?) if so, we may safely exclude it in `jme3-jogg`. (the jmf jar size is 1.8 MB!)

From a quick glance at [ogg](https://github.com/stephengold/j-ogg-all/tree/master/src/main/java/de/jarnbjo/ogg) package, I could not find any class using `jmf` package.

@stephengold can you please confirm if jmf library is required for decoding ogg files in JME?

stephengold:
> if jmf library is required for decoding ogg files in JME?

I don't see an easy way to check whether JMF is required. The runtime checks you did are encouraging but they didn't cover all code paths. "jmf-2.1.1e.jar" contains 993 class files in over 100 packages with a wide variety of names, so grepping for names would be a huge undertaking.

I'm sure jme3-jogg utilizes only a fraction of j-ogg-all's capabilities. If the goal is to minimize application size, one might copy code from j-ogg-all to jme3-jogg on an as-required basis and see where the process leads. That would be intellectually satisfying, but my experience with shrinking JME apps has convinced me that that big wins are in reducing texture sizes, not deleting unused code.

Scrappers-glitch:
Jmf has a lot of functions rather than just code used by j-ogg-all, so i think the only way to reduce the size of the jmf library is by maintaining the jmf library and re-distributing it into multiple modules and so the jogg-all can selectively implement the needed code only....

Ali-RS:
>If the goal is to minimize application size, one might copy code from j-ogg-all to jme3-jogg on an as-required basis and see where the process leads.

The goal is to remove unrequired libraries. It is a good idea to copy only the required classes to jme3-jogg but until someone does that I would rather go with the simplest path by excluding the unrequired library.

```
   api ('com.github.stephengold:j-ogg-all:1.0.2') {
        exclude module: "jmf"
    }
```

stephengold:
> one might copy code from j-ogg-all to jme3-jogg on an as-required basis and see where the process leads.

I performed this experiment and found that jme3-jogg needs only 3 packages from j-ogg-all in order to compile:
 + de.jarnbjo.ogg
 + de.jarnbjo.util.io and
[...]

PR #1959, opened 2023-02-20
resolve issue #1958 (jmf library coming with j-ogg-all is not required)

Ali-RS:
This PR excludes `javax.media:jmf` library from `jme3-jogg` module as it is not required for decoding ogg files in JME.

Fix issue #1958

### DIFF
--- a/jme3-jogg/build.gradle
+++ b/jme3-jogg/build.gradle
@@ -1,4 +1,6 @@
 dependencies {
     api project(':jme3-core')
-    api 'com.github.stephengold:j-ogg-all:1.0.2'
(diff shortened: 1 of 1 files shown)


stephengold:
I plan to release a new library "j-ogg-vorbis" which will be like "j-ogg-all" but will exclude the flac and Theora decoders. It won't depend on JMF. That will make this PR redudant, so perhaps you should hold off on this...

Ali-RS:
Ok, no problem. I will discard this PR then and will submit a new one  that uses `j-ogg-vorbis`.

Thanks

Ali-RS:
Closing this for https://github.com/jMonkeyEngine/jmonkeyengine/pull/1962

PR #1966, opened 2023-02-23
Replace Exception with warning in TerrainPatch

yaRnMcDonuts:
PR following up on the discussion in this thread:  https://hub.jmonkeyengine.org/t/terrain-collision-exception/46491/6

### DIFF
--- a/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
+++ b/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
@@ -801,22 +801,22 @@ public class TerrainPatch extends Geometry {
 
     @Override
     public int collideWith(Collidable other, CollisionResults results) throws UnsupportedCollisionException {
-        if (refreshFlags != 0){
+        if (refreshFlags != 0) {
             logger.warning("Scene graph must be updated before checking collision");
             return 0;
         }            
 
-        if (other instanceof BoundingVolume){
-            if (!getWorldBound().intersects((BoundingVolume)other)){
+        if (other instanceof BoundingVolume) {
+            if (!getWorldBound().intersects((BoundingVolume)other)) {
                 return 0;
             }
         }
 
-        if(other instanceof Ray){
+        if (other instanceof Ray) {
             return collideWithRay((Ray)other, results);
-        }else if (other instanceof BoundingVolume){
+        } else if (other instanceof BoundingVolume) {
             return collideWithBoundingVolume((BoundingVolume)other, results);
-        }else {
+        } else {
             throw new UnsupportedCollisionException("TerrainPatch cannot collide with "+other.getClass().getName());
         }
     }
--- a/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
+++ b/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
@@ -804,14 +804,14 @@ public class TerrainPatch extends Geometry {
         if (refreshFlags != 0) {
             logger.warning("Scene graph must be updated before checking collision");
             return 0;
-        }            
-
+        }
+        
         if (other instanceof BoundingVolume) {
             if (!getWorldBound().intersects((BoundingVolume)other)) {
                 return 0;
             }
         }
-
+        
         if (other instanceof Ray) {
             return collideWithRay((Ray)other, results);
         } else if (other instanceof BoundingVolume) {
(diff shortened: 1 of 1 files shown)


jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
```diff
@@ -799,8 +799,7 @@ protected void setLodBottom(int lodBottom) {
     @Override
     public int collideWith(Collidable other, CollisionResults results) throws UnsupportedCollisionException {
         if (refreshFlags != 0)
-            throw new IllegalStateException("Scene graph must be updated" +
-                                            " before checking collision");
+            System.err.println("Scene graph must be updated before checking collision");
```
Ali-RS, line 802: Can you use Logger instead of `System.err.println`?
Also, I think we should immediately return 0 when if the refreshFlags are modified.

yaRnMcDonuts: I added logger code following the same format of other JME classes, let me know if it looks alright or if I should change anything.

I also added brackets around the if statements to make the code cleaner and easier to understand, hopefully this change is okay as well.

jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
```diff
@@ -56,6 +56,8 @@
 import java.nio.FloatBuffer;
 import java.util.HashMap;
 import java.util.List;
+import java.util.logging.Level;
```
Ali-RS, line 59: It looks like `import java.util.logging.Level;` is not used? if so please remove the unused import.

jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
Ali-RS: >I also added brackets around the if statements

Please fix the formatting issues. There should be a whitespace added in '){' and `if(`

Ali-RS: also in `}else`

yaRnMcDonuts: It should be formatted correctly now

Ali-RS: thanks

Ali-RS:
Can you please update the copyright date?

Ali-RS (approved):

jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
```diff
@@ -798,19 +801,22 @@ protected void setLodBottom(int lodBottom) {
 
     @Override
     public int collideWith(Collidable other, CollisionResults results) throws UnsupportedCollisionException {
-        if (refreshFlags != 0)
-            throw new IllegalStateException("Scene graph must be updated" +
-                                            " before checking collision");
-
-        if (other instanceof BoundingVolume)
-            if (!getWorldBound().intersects((BoundingVolume)other))
+        if (refreshFlags != 0) {
```
Ali-RS, line 804: I think only `RF_BOUND` and `RF_TRANSFORM` should be considered similar to what Geometry does.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-core/src/main/java/com/jme3/scene/Geometry.java#L465

so it should be

```
if ((refreshFlags & (RF_BOUND | RF_TRANSFORM)) != 0) {
....
}
```

Ali-RS:
Unless there's something to discuss, I plan to integrate this PR to master and v3.6 branch in about 24 hours.

jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
```diff
@@ -798,19 +801,22 @@ protected void setLodBottom(int lodBottom) {
...
     @Override
     public int collideWith(Collidable other, CollisionResults results) throws UnsupportedCollisionException {
-        if (refreshFlags != 0)
-            throw new IllegalStateException("Scene graph must be updated" +
-                                            " before checking collision");
-
-        if (other instanceof BoundingVolume)
-            if (!getWorldBound().intersects((BoundingVolume)other))
+        if ((refreshFlags & (RF_BOUND | RF_TRANSFORM)) != 0) {
+            logger.warning("Scene graph must be updated before checking collision");
```
capdevon, line 805: Hi guys, just 2 cents. Throughout the engine the log is used like this:
`logger.log(Level.WARNING, ...)` instead of `logger.warning(...)`. Maybe it's better to keep the same rule here too, just for a matter of style. What do you think?

Ali-RS: Good idea!

yaRnMcDonuts: I updated the file with this change

Issue #1971, opened 2023-02-25
NativeLibraryLoader should not extract to working directory


Ali-RS:
>Extracting natives from the classpath into the working directory is a risky hack. In the long term, JME should probably use a native-library loader analogous with that of LWJGL v3.

>JME first looks [working directory](https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L255-L272) if it is writable then chooses that else it looks for [user cache directory](https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L293-L320) inside user.home directory, if found, will extract there, if not found then it will extract to `userHome/.jme3`.

>_Originally posted by @Ali-RS in https://github.com/jMonkeyEngine/jmonkeyengine/issues/1965#issuecomment-1444258551_
            

We should omit the check for "working directory" step above.

pspeed42:
What happens if a user routinely runs applications that use different versions of JME? (like me)

Ali-RS:
The application should still work as before. It will just extract and load natives from a different place.

pspeed42:
It seemed like you would be extracting all of the natives to the same place instead of the application-specific location.

...so if I were running my mapper tool in one window and tried to run jmec in another window then one of those applications might be corrupting the native libraries of another.

Ali-RS:
Ah, now I see what you mean. 

>one of those applications might be corrupting the native libraries of another

Yeah, that is possible...


Probably that is why JME generates a hash for naming the native directory(?)

https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L329

Ali-RS:
* to prevent collisions, a special subfolder is used
     * called <code>natives_&lt;hash&gt;</code> where &lt;hash&gt;
     * is computed automatically as the XOR of the classpath hash code
     * and the last modified date of this class.
[...]

Issue #1968, opened 2023-02-24
NativeLibraryLoader uses some hardcoded hacks


Ali-RS:
While dealing with NativeLibraryLoader in another issue I notices NativeLibraryLoader uses some hardcoded hacks that need to be fixed.

for example, in the below code, when it can not find the file inside the jar instead of immediately failing it tries to load it as a system library using `System.loadLibrary(libName);` which shouldn't in my opinion. 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L573-L600

or here, for lwjgl and jinput, it does not load them actually just extracts them and sets the path in a system property to be loaded later on by the library itself.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L666-L678

Ali-RS:
> I can not see why we need to redesign it.

I don't prefer large functions with too many blocks, there are a lot of hidden bugs in them, plus yeah the hardcorded directories and names may fail in some other places (according to the condition, the names may change or written wrongly, so you will have to go over each dependency and refactor it), I haven't glanced the NativeLibraryLoader in details yet, give me sometime to glance it and I will reply back on what should be done from my opinion on the newly created issue, but it indeed looking at it from here, it needs a lot of work.

_Originally posted by @Scrappers-glitch in https://github.com/jMonkeyEngine/jmonkeyengine/issues/1965#issuecomment-1442901956_

Ali-RS:
>hardcorded directories

Do you mean the native path in the library jar?
But that is inevitable if we want to support customer library loading. I mean how else we can load a native if we do not know the path to file inside the jar? 
The best we can do is to load them from a config file instead which I think it not worth because changing to the native path in a library is rare anyway and some libraries like lwjgl3 have their own loaders anyway. 
I think only jme3-lwjgl and bullet use it for loading natives.

Ali-RS:
Here is a preview of how workaround those hacks:

https://github.com/jMonkeyEngine/jmonkeyengine/commit/a23467d8e6a93c62a75dbe93edd1e3a437a8f931

Please let me know what you think about it.

Scrappers-glitch:
[...]

Issue #1422, opened 2020-11-14
bulletjme: Add Bitness/Architecture Suffix


MeFisto94:
For all other native libs we have like `OpenAL64.dll` and `lwjgl32.dll`, not so for bulletjme.
When switching jvms between 32bit and 64bit, you get an UnsatisfiedLinkError due to the DLL requiring a manual deletion.

stephengold:
1. Changing the filenames of Bullet native libraries will break Minie, which relies on the current naming scheme

2. Even with this change, file deletion will still be required when switching between different versions of jme3-bullet-native, or when switching between jme3-bullet-native and Minie

Ali-RS:
Apparently, Minie depends on `bulletjme` natives that are loaded via NativeLibraryLoader on JME side.

Ali-RS:
>Extracting natives from the classpath into the working directory is a risky hack. In the long term, JME should probably use a native-library loader analogous with that of LWJGL v3.

>_Originally posted by @stephengold in https://github.com/jMonkeyEngine/jmonkeyengine/issues/1965#issuecomment-1444233100_
            

I concur with this. We should not extract natives into the working.


LWJGL will try the following paths and the first successful will be used:
- System.getProperty("java.io.tmpdir")
- `<working directory>`
- System.getProperty("user.home")
- Files.createTempDirectory("lwjgl", "")


>JME first looks [working directory](https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L255-L272) if it is writable then chooses that else it looks for [user cache directory](https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L293-L320) inside user.home directory, if found, will extract there, if not found then it will extract to `userHome/.jme3`.

_Originally posted by @Ali-RS in https://github.com/jMonkeyEngine/jmonkeyengine/issues/1965#issuecomment-1444258551_
            
I think we should omit the "check for the working directory" from the above steps, and add another step at the very end to check "java.io.tmpdir" if none of the previous steps succeed.

Ali-RS:
Fixed in #1973 

Added Architecture Suffix to all `bulletjme` natives when extracting. 
[...]

Issue #1965, opened 2023-02-22
NativeLibraryLoader wrong libraries path


Ali-RS:
- All the native libraries registered by default for lwjgl 2 and 3 in NativeLibraryLoader have the wrong path! (it still works for lwjgl2 and JInput because of this [fallback](https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L570) because lwjgl2 and JInput natives are located in the root of the jar)

https://github.com/jMonkeyEngine/jmonkeyengine/blob/375015f14c5f3cbda18302b3aa990e5f023e9da1/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L123-L200

- NativeLibraryLoader is never used by jme3-lwjgl3. LWJGL 3 uses its own [SharedLibraryLoader](https://github.com/LWJGL/lwjgl3/blob/b44ba7af8030200bee34d7dfb77edf5617589328/modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java#L35) so we should remove lwjgl3 libraries from there as it is misleading. Also trying to load lwjgl3 libraries via NativeLibraryLoader won't work anyway and will produce an UnsatisfiedLinkError as it wouldn't find the libraries because of a wrong path. 

```
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.UnsatisfiedLinkError: The required native library 'lwjgl' was not found in the classpath via 'native/linux/liblwjgl.so'. Error message: no lwjgl in java.library.path
	at com.jme3.system.NativeLibraryLoader.loadNativeLibrary(NativeLibraryLoader.java:590)
	at jme3test.helloworld.HelloAudio.simpleInitApp(HelloAudio.java:27)
	at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:240)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:601)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:704)
	at java.lang.Thread.run(Thread.java:750)
```

- It does not register the ARM natives for macOS and Linux.

Thoughts?

stephengold:
I'm fine with jme3-lwjgl not supporting ARM architectures.

[...]

PR #1973, opened 2023-02-26
Improve NativeLibraryLoader

Ali-RS:
Solves https://github.com/jMonkeyEngine/jmonkeyengine/issues/1968, https://github.com/jMonkeyEngine/jmonkeyengine/issues/1422 and https://github.com/jMonkeyEngine/jmonkeyengine/issues/1971

@stephengold and @Scrappers-glitch please feel free to review this and let me know what do you think

### DIFF
--- a/jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
+++ b/jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright (c) 2009-2023 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.jme3.system;
+
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Defines default native libraries used by jMonkeyEngine.
+ *
+ * @author Ali-RS
+ */
+public enum NativeLibraries {
+
+    // Note: LWJGL 3 handles its native library extracting & loading using
+    // its own SharedLibraryLoader.
+
+    // LWJGL 2
+    Lwjgl(new LibraryInfo("lwjgl", libPath ->
+            // lwjgl handle loading by itself.
+            System.setProperty("org.lwjgl.librarypath",
+                    Paths.get(libPath).getParent().toAbsolutePath().toString()))
+            .addNativeVariant(Platform.Windows32, "lwjgl.dll")
+            .addNativeVariant(Platform.Windows64, "lwjgl64.dll")
+            .addNativeVariant(Platform.Linux32, "liblwjgl.so")
+            .addNativeVariant(Platform.Linux64, "liblwjgl64.so")
+            .addNativeVariant(Platform.MacOSX32, "liblwjgl.dylib")
+            .addNativeVariant(Platform.MacOSX64, "liblwjgl.dylib")
+    ),
+
+    // OpenAL for LWJGL 2
+    // For OSX: Need to add lib prefix when extracting
+    Openal(new LibraryInfo("openal")
+            .addNativeVariant(Platform.Windows32, "OpenAL32.dll")
+            .addNativeVariant(Platform.Windows64, "OpenAL64.dll")
+            .addNativeVariant(Platform.Linux32,   "libopenal.so")
+            .addNativeVariant(Platform.Linux64,   "libopenal64.so")
+            .addNativeVariant(Platform.MacOSX32,  "openal.dylib", "libopenal.dylib")
+            .addNativeVariant(Platform.MacOSX64,  "openal.dylib", "libopenal.dylib")
+    ),
+
+    // BulletJme
+    BulletJme(new LibraryInfo("bulletjme")
+            .addNativeVariant(Platform.Windows32, "native/windows/x86/bulletjme.dll", "bulletjme-x86.dll")
+            .addNativeVariant(Platform.Windows64, "native/windows/x86_64/bulletjme.dll", "bulletjme-x86_64.dll")
+            .addNativeVariant(Platform.Windows_ARM64, "native/windows/arm64/bulletjme.dll", "bulletjme-arm64.dll")
+            .addNativeVariant(Platform.Linux32, "native/linux/x86/libbulletjme.so", "libbulletjme-x86.so")
+            .addNativeVariant(Platform.Linux64, "native/linux/x86_64/libbulletjme.so", "libbulletjme-x86_64.so")
+            .addNativeVariant(Platform.Linux_ARM32, "native/linux/arm32/libbulletjme.so", "libbulletjme-arm32.so")
+            .addNativeVariant(Platform.Linux_ARM64, "native/linux/arm64/libbulletjme.so", "libbulletjme-arm64.so")
+            .addNativeVariant(Platform.MacOSX32, "native/osx/x86/libbulletjme.dylib", "libbulletjme-x86.dylib")
+            .addNativeVariant(Platform.MacOSX64, "native/osx/x86_64/libbulletjme.dylib", "libbulletjme-x86_64.dylib")
+            .addNativeVariant(Platform.MacOSX_ARM64, "native/osx/arm64/libbulletjme.dylib", "libbulletjme-arm64.dylib")
+    ),
+
+    // JInput
+    // For OSX: Need to rename extension jnilib -> dylib when extracting
+    JInput(new LibraryInfo("jinput", libPath ->
+            // jinput handle loading by itself.
+            System.setProperty("net.java.games.input.librarypath",
+                    Paths.get(libPath).getParent().toAbsolutePath().toString()))
+            .addNativeVariant(Platform.Windows32, "jinput-raw.dll")
+            .addNativeVariant(Platform.Windows64, "jinput-raw_64.dll")
+            .addNativeVariant(Platform.Linux32, "libjinput-linux.so")
+            .addNativeVariant(Platform.Linux64, "libjinput-linux64.so")
+            .addNativeVariant(Platform.MacOSX32, "libjinput-osx.jnilib", "libjinput-osx.dylib")
+            .addNativeVariant(Platform.MacOSX64, "libjinput-osx.jnilib", "libjinput-osx.dylib")
+    ),
+
+    // JInput Auxiliary (only required on Windows)
+    JInputDX8(new LibraryInfo("jinput-dx8")
+            .addNativeVariant(Platform.Windows32, "jinput-dx8.dll", null)
+            .addNativeVariant(Platform.Windows64, "jinput-dx8_64.dll", null)
+            .addNativeVariant(Platform.Linux32, null)
+            .addNativeVariant(Platform.Linux64, null)
+            .addNativeVariant(Platform.MacOSX32, null)
+            .addNativeVariant(Platform.MacOSX64, null)
+    );
+
+    private final LibraryInfo library;
+
+
+    NativeLibraries(LibraryInfo library) {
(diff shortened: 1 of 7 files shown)


Ali-RS:
> About the different "versions", they will have different lastModified dates so the new one should overwrite the old one.

>Sometimes one actually wants the older version.

_Originally posted by @stephengold in https://github.com/jMonkeyEngine/jmonkeyengine/issues/1965#issuecomment-1444582118_
            
This should resolve that

https://github.com/jMonkeyEngine/jmonkeyengine/blob/b80117df23bc1d053edbae8f5bbc25628beb0638/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L504-L515

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Defines default native libraries used by jMonkeyEngine.
+ *
+ * @author Ali-RS
+ */
+public enum NativeLibraries {
```
Scrappers-glitch, line 44: Yeah, that is what i were talking about, thanks for providing the clean-up.

Please, add the missing java docs on the public methods and the constructors.

jme3-desktop/src/main/java/com/jme3/system/NativeLibrary.java
```diff
@@ -76,6 +79,31 @@ public boolean equals(Object obj) {
             return true;
         }
     }
+
```
Scrappers-glitch, line 82: Could you add discriminative javadocs for those constructors documenting the constructor job and parameters usage and what [null] input means to the internal processing, please ?

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
```diff
@@ -0,0 +1,170 @@
...
+
+    public String getName() {
+        return library.getName();
+    }
+
+    private void registerLibrary() {
+        library.getNativeVariants().forEach(NativeLibraryLoader::registerNativeLibrary);
+    }
+
+    public static class LibraryInfo {
```
Scrappers-glitch, line 138: Could you please define javadocs for this class constructors and the public methods, please ?

Please, provide a full discriminative documentation including how the internal code process the [null] input at some parameters in brief.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
```diff
@@ -0,0 +1,170 @@
...
+            .addNativeVariant(Platform.Linux64, "native/linux/x86_64/libbulletjme.so", "libbulletjme-x86_64.so")
+            .addNativeVariant(Platform.Linux_ARM32, "native/linux/arm32/libbulletjme.so", "libbulletjme-arm32.so")
+            .addNativeVariant(Platform.Linux_ARM64, "native/linux/arm64/libbulletjme.so", "libbulletjme-arm64.so")
+            .addNativeVariant(Platform.MacOSX32, "native/osx/x86/libbulletjme.dylib", "libbulletjme-x86.dylib")
+            .addNativeVariant(Platform.MacOSX64, "native/osx/x86_64/libbulletjme.dylib", "libbulletjme-x86_64.dylib")
+            .addNativeVariant(Platform.MacOSX_ARM64, "native/osx/arm64/libbulletjme.dylib", "libbulletjme-arm64.dylib")
+    ),
+
+    // JInput
+    // For OSX: Need to rename extension jnilib -> dylib when extracting
```
Scrappers-glitch, line 88: What is the `.jnilib` extension ?

Scrappers-glitch: Is jinput outputing their native binaries with the `jnilib` extensions ?

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
```diff
@@ -0,0 +1,170 @@
...
+    JInput(new LibraryInfo("jinput", libPath ->
```
Scrappers-glitch, line 89: I don't understand what should the consumer loadFunction provide here ? other than the system load, but if the jinput handles the loading themselves, why would we even care about adding a loader function ?

Scrappers-glitch:
For a complete clean-up, could you break the [`NativeLibraryLoader`](https://github.com/jMonkeyEngine/jmonkeyengine/blob/b80117df23bc1d053edbae8f5bbc25628beb0638/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java#L427-L547) method down into small private static methods ? 
Thank you.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
```diff
@@ -571,88 +478,69 @@ public static void loadNativeLibrary(String name, boolean isRequired) {
...
             if (targetFile.exists()) {
                 // OK, compare last modified date of this file to 
                 // file in jar
                 long targetLastModified = targetFile.lastModified();
                 long sourceLastModified = conn.getLastModified();
 
                 // Allow ~1 second range for OSes that only support low precision
-                if (targetLastModified + 1000 > sourceLastModified) {
-                    logger.log(Level.FINE, "Not copying library {0}. " + 
+                if (Math.abs(sourceLastModified - targetLastModified) < 1000) {
```
Scrappers-glitch, line 505: This could be separated into a small private static function that returns a boolean indicating whether an operation should be executed, for example: 
```java
/**
* Tests whether 2 files are identical by comparing their last-date of modifications metadata, the last date of modification should be 1 second or more for the source and target to be considered as non-identical.
*
* @param source the source file
* @param target the target file 
* @return true if the difference in ldm is less than 1 second, false otherwise
*/
private static boolean hasIdenticalLdm(final URLConnection source, final File target) { 
     return Math.abs(source.getLastModified() - target.lastModified()) < 1000;
}
```
Where; `Ldm` stands for Last-Date-Of-Modification.

Then, remove `extract = false` and use the newly helper method in the proceeding checks.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Ali-RS: Done

jme3-desktop/src/main/java/com/jme3/system/NativeLibrary.java
Ali-RS: Done

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Ali-RS: Done

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Ali-RS: Yes, for macOS it is named `libjinput-osx.jnilib`. I do not know what it is.
The comment is from the old code.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
Scrappers-glitch: You may also change the `hasIdenticalLdm(..., ...)` with `areIdentical(..., ...)` too, choose your favourite.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Scrappers-glitch: Looks good thanks.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Scrappers-glitch: I see, no problem, you can resolve this.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Ali-RS: From what I understand both jinput and lwjgl2 do not support extracting the natives from the jar. They expect that the natives are extracted by application and the path to natives  is set via a system property. If the `net.java.games.input.librarypath` property is set then jinput will load them from there else it will use `System.loadLibrary()` to load from system-installed libraries.

I could not come up with a better solution so I added the load function style for custome load handling.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
Ali-RS: Done

jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
```diff
@@ -571,89 +478,80 @@ public static void loadNativeLibrary(String name, boolean isRequired) {
...
+    /**
+     * Checks if library extraction is required by comparing source and target
+     * last modified date. Returns true if target file does not exist.
+     *
+     * @param conn the source file
+     * @param targetFile the target file
+     * @return false if target file exist and the difference in last modified date is
+     *          less than 1 second, true otherwise
+     */
+    private static boolean isLibraryExtractionRequired(URLConnection conn, File targetFile) {
```
Scrappers-glitch, line 540: Nice work, btw when a function has more than one return, i guess ordering those returns ascendingly by complexity will give the reader a better idea of  the function job, so for example: 
```java
private static boolean isLibraryExtractionRequired(URLConnection conn, File targetFile) {
     // extract anyway if the file doesn't exist 
     if (!targetFile.exists()) {
         return true;  
     }
    // if the file exists then:
    // compare last modified date of this file to file in jar
    // Allow ~1 second range for OSes that only support low precision
    return Math.abs(conn.getLastModified() - targetFile.lastModified()) >= 1000;
}
```

Scrappers-glitch: If you are using intellij, it may show you an equivalent shortcut to my proposed change, like this:
```java
private static boolean isLibraryExtractionRequired(URLConnection conn, File targetFile) {
     // extract anyway if the file doesn't exist 
     // if the file exists then:
    // compare last modified date of this file to file in jar
    // Allow ~1 second range for OSes that only support low precision
    return !targetFile.exists() || Math.abs(conn.getLastModified() - targetFile.lastModified()) >= 1000;
}
```
You may choose to keep it like that too, it's the same, but i don't prefer this syntax...

Ali-RS: Done, 

Thanks for the feedback :)

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Ali-RS: LWJGL3 also uses a similar load function but for different purpose I guess

https://github.com/LWJGL/lwjgl3/blob/b44ba7af8030200bee34d7dfb77edf5617589328/modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java#L56-L60

Ali-RS:
Modified `getExtractionFolder()` to use `System.getProperty("java.io.tmpdir")` instead of extracting to the working directory.

Fix https://github.com/jMonkeyEngine/jmonkeyengine/issues/1971

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Scrappers-glitch: Hello there @Ali-RS,
Which class has the `System.load` to load the native libraries ?

Ali-RS: Sorry, I do not understand the question

Scrappers-glitch: Nope, i mean the jme3 load function that loads the native libaries ?

Scrappers-glitch: My original objection was on the loadFunction consumer name, the code uses some custom loading consumers which aren't actually real loading function, so i guess this is a kinda of injecting an action into a command-state pattern, since, the LibraryInfo has no idea of what the actual assigned function is supposed to do, i suggest renaming all the occurrences of the `loadFunction` to `onLoad` and documenting that this is called on the behalf of the `NativeLibraryLoader#loadNativeLibrary(..., ...)`.

Here is a list of good names, pick what you think is best: 
- onLoad
- nativeLoaderCallback
- nativeLoaderConsumer

Scrappers-glitch: The `onLoad` may be also a little bit vague; because it is a listener for the caller of the stack `NativeLibraryLoader#loadNativeLibrary(..., ...)` and it is not a listener of the actual `System.load(String)`, but you can also consider the name `onExtractionCompleted` for example too, since you are extracting before loading and loading can be considered a listener, i haven't glanced the full code, so i cannot fully decide, but you already know most of the "under-the-hood" work, so i think you will select the best in this case, here is the list again: 
- onLoad
- nativeLoaderCallback
- nativeLoaderConsumer
- onExtractionCompleted

Scrappers-glitch: I guess onExtractionCompleted will be the best, it is fully descriptive of the actual pattern, but idk, we don't extract every time we try to load a library too.......

Ali-RS: >and documenting that this is called on the behalf of the NativeLibraryLoader#loadNativeLibrary(..., ...).

Note NativeLibraryLoader#loadNativeLibrary(..., ...) internally calls the `loadFunction` of NativeLibrary it does not call System.load() directly.

Scrappers-glitch: Yup, i know, it calls the consumer function...

Scrappers-glitch: > ) or do something else.

And so that why i gave my objection to the name `loadFunction`.

Ali-RS: so I believe the name `loadFunction` suits more closely of what it does. (i.e it is not a listener)

Scrappers-glitch: I see, no problem.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
```diff
@@ -0,0 +1,223 @@
...
+ *
+ * @author Ali-RS
+ */
+public enum NativeLibraries {
+
+    // Note: LWJGL 3 handles its native library extracting & loading using
+    // its own SharedLibraryLoader.
+
+    // LWJGL 2
+    Lwjgl(new LibraryInfo("lwjgl", libPath ->
```
Scrappers-glitch, line 51: Sorry, i realized that we forgot javadocs here on those public static final Objects, could you add them, please ?

Scrappers-glitch, line 51: Sorry, i realized that we forgot javadocs here on those public static final Objects, could you add them, please ?

Ali-RS: Done.

Ali-RS:
Unless there's substantial discussion, I plan to self-integrate this in about 24 hours.

jme3-desktop/src/main/java/com/jme3/system/NativeLibraries.java
Scrappers-glitch: Gj, thanks for the improvement work !

PR #1979, opened 2023-03-06
Allow use of Emissive color as a multiplier with EmissiveMap in PBRLighting

Ali-RS:
Similar to PR #1444 but for PBR.

Issued on the forum: https://hub.jmonkeyengine.org/t/khr-transform/46524/37?u=ali_rs


stephengold:
I believe it's too late in the release cycle to be adding features to 3.6

PR #1980, opened 2023-03-06
Add support for NormalScale in PBRLighting & GltfLoader

Ali-RS:
Based on Gltf specs 
https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/material.normalTextureInfo.schema.json

and as issued on the forum https://hub.jmonkeyengine.org/t/khr-transform/46524/39?u=ali_rs 

I added a `NormalScale` factor in PBRLighting. The scalar parameter is applied to each normal vector of the normal map. This value scales the normal vector in X and Y directions using the formula: `scaledNormal = normalize((<sampled normal texture value> * 2.0 - 1.0) * vec3(<normal scale>, <normal scale>, 1.0))`. (from the above gltf spec)

and updated GltfLoader to support this.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -89,6 +89,9 @@ varying vec3 wPosition;
   uniform sampler2D m_NormalMap;   
   varying vec4 wTangent;
 #endif
+#ifdef NORMALSCALE
+  uniform float m_NormalScale;
+#endif
 varying vec3 wNormal;
 
 #ifdef DISCARD_ALPHA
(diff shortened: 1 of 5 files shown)


stephengold:
I believe it's too late in the release cycle to be adding features to 3.6

stephengold:
This PR appears to have caused the regression that's holding up the v3.7.0-stable release. @Ali-RS are you available to help troubleshoot the regression?

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
```diff
@@ -170,7 +173,11 @@ void main(){
       //as it's compliant with normal maps generated with blender.
       //see http://hub.jmonkeyengine.org/forum/topic/parallax-mapping-fundamental-bug/#post-256898
       //for more explanation.
-      vec3 normal = normalize((normalHeight.xyz * vec3(2.0, NORMAL_TYPE * 2.0, 2.0) - vec3(1.0, NORMAL_TYPE * 1.0, 1.0)));
+      #ifdef NORMALSCALE
+        vec3 normal = normalize((normalHeight.xyz * vec3(2.0, NORMAL_TYPE * 2.0, 2.0) - vec3(1.0, NORMAL_TYPE * 1.0, 1.0)) * vec3(m_NormalScale, m_NormalScale, 1.0));
```
codex128, line 177: I'm not sure what normal scale is supposed to achieve exactly, but it seems very strange to have `vec3(m_NormalScale, m_NormalScale, 1.0)` with the z component left as 1.0. That isn't how scalars are normally applied.

scenemax3d: @codex128 you mean that usually the m_NormalScale should also apply to the Z component like this?
vec3(m_NormalScale, m_NormalScale, m_NormalScale)

BTW, the link in the comment (http://hub.jmonkeyengine.org/forum/topic/parallax-mapping-fundamental-bug/#post-256898) is broken

codex128: Yeah, having all components be `m_NormalScale` seems like it should be correct.

PR #1981, opened 2023-03-07
Add support for AoStrength in PBRLighting & GltfLoader

Ali-RS:
Added a scalar multiplier controlling the amount of occlusion applied and updated GltfLoader to read it if present on "occlusionTexture".

Based on specs https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/material.occlusionTextureInfo.schema.json

the formula should be for final occlusion value as: `1.0 + strength * (<sampled occlusion texture value> - 1.0)` but in JME I used `ao *= strength` based on forum suggestion.

Forum topic: https://hub.jmonkeyengine.org/t/gltf-ao-strenght/46553

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -236,6 +236,7 @@ void main(){
 
     #ifdef AO_STRENGTH
        ao = 1.0 + m_AoStrength * (ao - 1.0);
+       ao = max(ao, 0.0);
     #endif
 
     float ndotv = max( dot( normal, viewDir ),0.0);
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -236,7 +236,8 @@ void main(){
 
     #ifdef AO_STRENGTH
        ao = 1.0 + m_AoStrength * (ao - 1.0);
-       ao = max(ao, 0.0);
+       // sanity check
+       ao = clamp(ao, 0.0, 1.0);
     #endif
 
     float ndotv = max( dot( normal, viewDir ),0.0);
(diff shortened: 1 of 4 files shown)


Ali-RS:
>the formula should be for final occlusion value as: 1.0 + strength * (<sampled occlusion texture value> - 1.0) but in JME I used ao *= strength based on forum suggestion.

Fixed it to follow the gltf specs: `ao = 1.0 + m_AoStrength * (ao - 1.0);` as requested [here](https://hub.jmonkeyengine.org/t/gltf-ao-strenght/46553/5?u=ali_rs).

yaRnMcDonuts:
We might also want to max ao here to make sure the final ao value is never less than 0

```
ao = 1.0 + m_AoStrength * (ao - 1.0);
ao = max(ao, 0.0);
```

Since right now I think the value you get after applying AoStrength can be below 0 at high AO strengths, and I am not sure whether or not the pbr lighting method would work with a negative ao paramater sent in.

Ali-RS:
>at high AO strengths, and I am not sure whether or not the pbr lighting function would work with a negative ao paramater sent in.

Based on gltf specs occlusion strength must be in [0, 1] so I believe the `ao` value should never go below zero. I will document this on material definition. Thanks

https://github.com/KhronosGroup/glTF/blob/6753012ec446efa747177e2fcc08aa5534fc27e4/specification/2.0/schema/material.occlusionTextureInfo.schema.json#L14-L15

yaRnMcDonuts:
So it sounds like their idea is that the AO Strength can only reduce the AO intensity, and can't increase the intensity at all since its supposed to be in 0-1 range and 1.0 is the default.

We could follow their format with this, but we could also allow values over 1.0 (and then clamp the final ao value to 0-1) so that AO strength could also allow you to increase AO intensity with the strength value.

I notice the PBR shader also doesn't force you to keep the m_Roughness and m_Metallic paramaters in 0-1 range, which I personally like because then I can make minor adjustments to both increase or decrease the value without having to change the textures. So it likely that setting values slightly over 1.0 for AoStrength will be useful for slightly intensifying aoMaps to be more noticeable, so it could still be useful to clamp the final aoValue between 0-1 to make sure it never ends up negative in the case a user does set aoStrength higher than 1.0

Ali-RS:
Ok, done.

PR #1983, opened 2023-03-08
Deploy master branch commits as snapshot

Sailsman63:
In response to this discussion:
https://hub.jmonkeyengine.org/t/adding-nightly-builds-to-engine/46560

This is mostly just a copy-paste of the relevant parts of the existing release deployment mechanism. It should not require manual intervention at any step.

I do not have a deployment account for Sonatype, so this branch is only partially tested. (The gradle script call returns "401: Unauthorized" when run from the command line. This is what I would expect without an account.)

Users who want to use the snapshot will need to add `https://oss.sonatype.org/content/repositories/snapshots/` to their build scripts' repository list, and then target version `<Major_Version>-SNAPSHOT`. Not sure where to document that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -296,7 +296,7 @@ jobs:
     needs: [BuildJMonkey]
     name: Deploy Java Snapshot
     runs-on: ubuntu-latest
-    if: github.ref_name == 'master'
+    if: github.event_name == 'push' && github.ref_name == 'master'
     steps:
 
       # We need to clone everything again for uploadToMaven.sh ...
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -356,6 +356,13 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
 
+      # Download all the stuff...
+      - name: Download maven artifacts
+        uses: actions/download-artifact@master
+        with:
+          name: maven
+          path: dist/maven
+
       - name: Download release
         uses: actions/download-artifact@master
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -312,13 +312,6 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
 
-      # Download all the stuff...
-      - name: Download maven artifacts
-        uses: actions/download-artifact@master
-        with:
-          name: maven
-          path: dist/maven
-
       - name: Download natives for android
         uses: actions/download-artifact@master
         with:
@@ -362,13 +355,6 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '17'
-
-      # Download all the stuff...
-      - name: Download maven artifacts
-        uses: actions/download-artifact@master
-        with:
-          name: maven
-          path: dist/maven
 
       - name: Download release
         uses: actions/download-artifact@master
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -296,7 +296,7 @@ jobs:
     needs: [BuildJMonkey]
     name: Deploy Java Snapshot
     runs-on: ubuntu-latest
-    if: GITHUB_EVENT_NAME == 'push' && GITHUB_REF_NAME == 'master'
+    if: github.ref_name == 'master'
     steps:
 
       # We need to clone everything again for uploadToMaven.sh ...
@@ -318,12 +318,6 @@ jobs:
         with:
           name: maven
           path: dist/maven
-
-      - name: Download release
-        uses: actions/download-artifact@master
-        with:
-          name: release
-          path: dist/release
 
       - name: Download natives for android
         uses: actions/download-artifact@master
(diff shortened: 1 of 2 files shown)


Ali-RS:
>Not sure where to document that.

Maybe here https://wiki.jmonkeyengine.org/docs/3.4/getting-started/maven.html

Sailsman63:
Hold on this. I did some poking around to test this, and I found that I mixed up how the workflow system operates. This PR will *not* function as intended at this point, and getting it running may lead to some refactoring of the existing workflow.

Sailsman63:
Okay, I've done a bit more testing & fixed a couple of things.

The only thing left is to verify that the correct files actually get uploaded, and the only way to test that is to trigger a run in an appropriate branch under a repository that actually provides a set of sonatype credentials.

(If someone wants to test without merging into master, it could be done by merging to another branch and modifying line 299 of the workflow to reference *that* branch.)

.github/workflows/main.yml
```diff
@@ -291,6 +291,57 @@ jobs:
             fi
           fi
 
+  # This job deploys snapshots on the master branch
+  DeployJavaSnapshot:
+    needs: [BuildJMonkey]
+    name: Deploy Java Snapshot
+    runs-on: ubuntu-latest
+    if: github.ref_name == 'master'
```
Ali-RS, line 299: Will this job be gnored on pull requests to master branch? 
because that is what we want. It should only run on commit push.

Ali-RS: probably we want 
```
if: github.event_name == 'push' && github.ref_name == 'master'
```
right?

Sailsman63: Probably. I knew I was losing track of something here.

.github/workflows/main.yml
```diff
@@ -363,13 +356,6 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
 
-      # Download all the stuff...
-      - name: Download maven artifacts
-        uses: actions/download-artifact@master
-        with:
-          name: maven
-          path: dist/maven
-
```
Ali-RS, line 321: Hmm, I noticed `Deploy to github package registry` uses artifacts from dist/maven

https://github.com/jMonkeyEngine/jmonkeyengine/blob/44a50de4d7797dc3dd8d46cd1f51308e9a05d627/.github/workflows/main.yml#L368-L373

Ali-RS (approved):

Ali-RS:
Thanks, @Sailsman63 

>The only thing left is to verify that the correct files actually get uploaded

I plan to merge this in 12H and we can then verify this works as expected.

Ali-RS:
Seems it failed to upload snapshots

https://github.com/jMonkeyEngine/jmonkeyengine/actions/runs/4390946051/jobs/7690244506#step:5:488

Ali-RS:
Fixed in https://github.com/jMonkeyEngine/jmonkeyengine/commit/a0c2c5ee74f87f588e539531a5c535b77b9ea4b3

let's see if it will work now!

Ali-RS:
Seems it works!

Snapshots are uploaded to 
https://s01.oss.sonatype.org/content/repositories/snapshots/org/jmonkeyengine/

Sailsman63:
Thanks for finishing this up. I've opened a documentation PR for the Wiki: https://github.com/jMonkeyEngine/wiki/pull/167

(not all users can edit directly.)

stephengold:
Note that I've now set up the "v3.6" branch to deploy snapshots in a similar matter.

I notice that the current scripts don't make efficient use of SonaType's resources. Each successful CI build causes a snapshot deployment which uploads 450 MBytes of data to s01.oss.sonatype.org . As far as I know, the uploaded files are permanent: never overwritten or deleted. So now we have 30+ snapshots deployed, and the number is steadily growing.

Is this intentional?

I don't see any reason to retain snapshot builds for more than a week or so. In fact, I when I heard about deploying snapshots to SonaType, I assumed each snapshot deployment would overwrite the previous one.

Ali-RS:
>Is this intentional?

I might be wring, but I think back then we opted for the easiest solution and let other concerns aside until it become an issue. But yeah not efficient!

Is it possible to set it up to remove the old snapshots? If so that might be a good idea to remove the old ones and only keep the new ones, for example, we can keep the last 10 builds. 

I have no idea how we can implement it, though.

stephengold:
I agree there's no evidence that the old files are causing an issue ... yet.

When I browse the "Snapshots" repository in my web browser via the Nexus Repository Manager, there's a "Delete" button for each file. (Open the "Browse Storage" tab in the lower left quadrant then the "Artifact" tab in the lower right.) Assuming those buttons work, removal of old snapshots is theoretically possible. In practice, doing this through the web interface would require thousands of mouse clicks, since a single snapshot contains about 900 files (18 libraries x 5 artifacts/lib x 10 files/artifact).

What I'd ideally like is a tool to remove all snapshot artifacts we own except for the latest 3.6.0 snapshot and the latest 3.7.0 snapshot. Apparently such a feature exists in the Nexus software:

https://support.sonatype.com/hc/en-us/articles/213465308-Can-I-delete-releases-from-Nexus-Repository-2-or-3-after-they-have-been-published-

I don't know how to create a "scheduled task". That might be something we can request via a Jira ticket.

Ali-RS:
Looks like this requires administrator access.

PR #1986, opened 2023-03-14
Some javadoc cleanup

Sailsman63:
The first two commits fix some issues that were actually interfering with correct rendering in the javadoc.

The third really is just some typo and reflow to be more compliant with project code standards. Note that the file in question has several comment blocks that are still too long, (~125 characters average) but I didn't feel like messing with them on an API that is not really currently maintained. I only picked the truly low-hanging fruit.


Ali-RS (approved):

Ali-RS:
@Sailsman63, thank you!

Issue #1982, opened 2023-03-07
IllegalArgumentException in registerObject()


stephengold:
This issue was reported at the Discourse hub/forum:  https://hub.jmonkeyengine.org/t/java-lang-illegalargumentexception-object-id-must-be-greater-than-zero-on-mac/46559

https://github.com/jMonkeyEngine/jmonkeyengine/blob/44a50de4d7797dc3dd8d46cd1f51308e9a05d627/jme3-core/src/main/java/com/jme3/util/NativeObjectManager.java#L104-L107

While JME specifically reserves the value `-1` for invalid IDs:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/44a50de4d7797dc3dd8d46cd1f51308e9a05d627/jme3-core/src/main/java/com/jme3/util/NativeObject.java#L47

I believe negative IDs in general don't justify throwing an exception.

The OpenGL documentation doesn't specify range of values for a texture name, only that they are `GLuint`:

+ https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGenBuffers.xhtml
+ https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGenTextures.xhtml

According to the OpenGL spec, `GLuint` is at least 32 bits, so values > 0x7fffffff will produce negative values in a Java `int`.

I think the tests in lines 105 and 132 should be specifically for ` == INVALID_ID` instead of ` <= 0`.

stephengold:
The `GLRenderer` class appears to have similar issues. In some places, it checks against -1; in others, it requires `> 0`.

+ https://github.com/jMonkeyEngine/jmonkeyengine/blob/44a50de4d7797dc3dd8d46cd1f51308e9a05d627/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java#L1254
+ https://github.com/jMonkeyEngine/jmonkeyengine/blob/44a50de4d7797dc3dd8d46cd1f51308e9a05d627/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java#L1364
+ https://github.com/jMonkeyEngine/jmonkeyengine/blob/44a50de4d7797dc3dd8d46cd1f51308e9a05d627/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java#L1455-L1456
[...]

PR #1991, opened 2023-03-30
correctly handle negative IDs in getUniqueId() methods

stephengold:
As I mentioned at issue #1982, one issue with negative native IDs is the encoding scheme used to implement `NativeObject.getUniqueId()`. To generate a unique ID, the object type is stored in the upper 32 bits. For a negative ID, sign extension during conversion from `int` to `long` obliterates the type information.

This PR masks off the upper 32 bits before adding the object type. It is merely the first of *many* changes that will be needed to handle negative native IDs in JMonkeyEngine.


stephengold:
Unless there's substantive discussion, I plan to integrate this PR in about 24 hours.

Issue #1992, opened 2023-04-09
Spatial assertions do not quote the name of the spatial involved


richardTingle:
Assertions within the spatial class have no message associated with them. This makes them much less useful for debugging problems (they effectively just say; there is a problem, somewhere in your entire scene graph), By including the spatial name in the assertion message they will help track down the problem much quicker.

(I intend to submit a PR for this)

pspeed42:
Note that if you are actually seeing assertions then there are bugs in the engine.  Assertions should not be used for user-level errors but double-checking engine state... the cases where if something is wrong then it is really wrong.

Adding messages is good... I've always been a little frustrated that they were never added.  Just wanted to point out that if there are cases where we expect the user to care about these then maybe they need a different kind of check.

...we will certainly care, though, and that's where the messages are super-handy.

richardTingle:
Interesting. I am definitely seeing them. The refreshFlags ones. I'll try and see if I can figure out a pattern (and these messages may help), they are very rare though, maybe only twice in months of development

pspeed42:
They are either a sign of engine bugs or engine abuse.  Either way, yeah, it would be good to track down why it's happening... and if you could be more specific then I might even speculate.

stephengold:
Looking at the changes in #1993, my best guess is that the assertion failures are caused by an application updating the scene graph asynchronously. Since there's little JME can do to prevent such abuse, perhaps the assertions should be replaced by `IllegalStateException` throws like this one:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/76ebd1443ccc18249e9f082d528cd39baf59da84/jme3-core/src/main/java/com/jme3/scene/Spatial.java#L364-L367

richardTingle:
Yeah, that was what my money was on as well. I use a lot of threading and while I obviously know I'm not supposed to modify the scenegraph on anything but the main thread it's easy to do it accidentally (it's why I religiously give my spatials names so that that IllegalStateException you quote gives helpful answers).

On IllegalStateException vs asserts. Is there an issue with performance, in that the asserts are free in deployed applications but test in development. Whereas an IllegalStateException's checks have to run every time

pspeed42:
To me,  if the cause turns out to be thread-abuse then I saw leave the assertions.  IllegalStateExceptions are a lot more expensive than compiled-out assertions.  Even the act of adding a throw new Exception to your code wires in a bunch of infrastructure during compilation that wouldn't normally be there.

PR #1993, opened 2023-04-09
solve issue #1992: better messages in spatial assertions

richardTingle:
Assertions within the spatial class have no message associated with them. This makes them much less useful for debugging problems (they effectively just say; there is a problem, somewhere in your entire scene graph), By including the spatial name in the assertion message they will help track down the problem much quicker.

I have some of these very intermittently pop up in my own application but I can't really fix them without more information as to what spatial is causing them

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -368,7 +368,7 @@ public abstract class Spatial implements Savable, Cloneable, Collidable,
         }
 
         CullHint cm = getCullHint();
-        assert cm != CullHint.Inherit : "CullHint should never be inherit. " + getName();
+        assert cm != CullHint.Inherit : "CullHint should never be inherit. Problem spatial name: " + getName();
         if (cm == Spatial.CullHint.Always) {
             setLastFrustumIntersection(Camera.FrustumIntersect.Outside);
             return false;
@@ -586,7 +586,7 @@ public abstract class Spatial implements Savable, Cloneable, Collidable,
             worldLights.update(localLights, null);
             refreshFlags &= ~RF_LIGHTLIST;
         } else {
-            assert (parent.refreshFlags & RF_LIGHTLIST) == 0 : "Illegal light list update. " + getName();
+            assert (parent.refreshFlags & RF_LIGHTLIST) == 0 : "Illegal light list update. Problem spatial name: " + getName();
             worldLights.update(localLights, parent.worldLights);
             refreshFlags &= ~RF_LIGHTLIST;
         }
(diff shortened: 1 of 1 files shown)


Ali-RS (approved):

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -811,7 +811,7 @@ public void addControlAt(int index, Control control) {
 
         if (index < numControls) { // re-arrange the list directly
             boolean success = controls.remove(control);
-            assert success;
+            assert success : "Surprising control remove failure. " + control.getClass().getSimpleName() + " from spatial " + getName();
```
Sailsman63, line 814: and `Surprising \"control.remove()\" failure`

Sailsman63 (commented):
Thought here would be to make sure that code references don't get accidentally read as normal text. Small change, perhaps, but might make the results more readable.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
pspeed42: I would only offer to beware of using string-notation in messages when the value is not actually a string.  In these cases they are both unnecessary but neither would be confusing... but that is not always the case.

Sometimes there are messages like:
Foodebla blah blah bad "This string" deblah blah...

And in that case the quoted string has meaning... because if you instead see:
Foodebla blah blah bad "This string " deblah blah... (note extra space)
Or:
Foodebla blah blah bad ""This string"" deblah blah...

...you have an indication of your error already.

I would argue that none of your examples need strings around them but I could see an argument for wrapping getName() in quotes for all the reasons I suggest the other is confusing.

Sailsman63: Good point. I'm not married to using quotes; I'm just wondering if there's a good, simple way to differentiate identifiers from the ordinary English words that they resemble, and that are almost - but not quite - grammatically correct in context.

pspeed42: No getting away from looking in the code for an exception like this... then it's obvious.

Sailsman63: Okay, I think you convinced me. Thanks for the discussion.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -952,7 +952,7 @@ public void updateGeometricState() {
         if ((refreshFlags & RF_MATPARAM_OVERRIDE) != 0) {
             updateMatParamOverrides();
         }
-        assert refreshFlags == 0;
+        assert refreshFlags == 0 : "Illegal refresh flags state: " + refreshFlags + " for spatial " + getName();
```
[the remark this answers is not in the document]
stephengold: I think this is fine as it stands. Double quotes don't add any clarity and might mislead the reader into thinking it's a String.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -368,7 +368,7 @@ public boolean checkCulling(Camera cam) {
         }
 
         CullHint cm = getCullHint();
-        assert cm != CullHint.Inherit;
+        assert cm != CullHint.Inherit : "CullHint should never be inherit. Problem spatial name: " + getName();
```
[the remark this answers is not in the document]
stephengold: I think this is fine as it stands. Double quotes don't add any clarity and might mislead the reader into thinking it's a String.

Ali-RS:

Issue #2003, opened 2023-05-07
ParticleDepositionHeightMap.load() returns false for no apparent reason


capdevon:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/25a3407c9e403dbc0250430a304031c169636a03/jme3-terrain/src/main/java/com/jme3/terrain/heightmap/ParticleDepositionHeightMap.java#L325

PR #2005, opened 2023-05-07
solve issue #2003 (ParticleDepositionHeightMap.load return value)

stephengold:
As issue #2003 says, there's no reason for `load()` to return `false` in this case.

One reason the issue went undetected so long:  all uses of the `load()` method in jme3-terrain and jme3-examples ignore the return value. However, Minie (taking the javadoc literally) *does* check the value, so returning the correct value is important.

I quickly checked the other known implementations of `load()` and none of them seemed to have similar issues.


stephengold:

stephengold:
Existing applications might depend on the return value. I have changed my mind about including this in v3.6.1

Issue #2002, opened 2023-05-07
TerrainGridTestData.zip: Missing assets. TerrainGrid/testgrid_x_y_z.j3o.


capdevon:
The `TerrainGridTileLoaderTest `class cannot load the j3o terrain files because they are not present in the .zip file
I downloaded the zip file locally and verified that indeed the j3o files are not inside it.

package `jme3test.terrain.TerrainGridTileLoaderTest`

StackTrace:
```java
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_-1_0_-1.j3o
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_-1_0_0.j3o
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_-1_0_1.j3o
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_2_0_1.j3o
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_-1_0_2.j3o
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_0_0_2.j3o
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_1_0_2.j3o
mag 07, 2023 6:44:34 PM com.jme3.terrain.geomipmap.grid.AssetTileLoader getTerrainQuadAt
WARNING: Could not load terrain grid tile: TerrainGrid/testgrid_2_0_2.j3o
```

stephengold:
[...]

PR #2006, opened 2023-05-07
 solve issue #2002 (TerrainGridTileLoaderTest fails to load tiles)

stephengold:
As discussed at issue #2002.


stephengold:

Issue #2007, opened 2023-05-10
Instanced objects are culled when using WaterFilter


foxcc2021:
It seems there is a bug in culling checks of instanced objects.

![20230510202939](https://github.com/jMonkeyEngine/jmonkeyengine/assets/93635496/ffeda85e-90c8-4d36-bcbf-bc657fca3b42)
The tree model should not be culled.

![20230510203009](https://github.com/jMonkeyEngine/jmonkeyengine/assets/93635496/143f7e2f-4a75-467f-bd45-673dc5261e85)
The correct scene expected．

After some debugging, I found that, it seems the instanceCullingFunction used the wrong camera to cull checks in the InstancedGeometry。
Because the WaterFilter has a reflectionCam, the default camera will be replaced when waterFilter rendering.
![20230510202701](https://github.com/jMonkeyEngine/jmonkeyengine/assets/93635496/f99bffdb-23e1-4bec-bfeb-9615320d7e29)

![20230510210756](https://github.com/jMonkeyEngine/jmonkeyengine/assets/93635496/8641fb3b-c566-4eb0-be0f-005773c6270e)
I've solved my problem using this way,pass the render camera from the instanced node to the instanced geometry for culling checks.
Hope this can helps.

JME version:3.6.0-stable
System: win10

Ali-RS:
I can not reproduce this issue, can you provide a simple test case demonstrating the issue?

Ali-RS:
I can not reproduce this issue, can you provide a simple test case demonstrating the issue?

foxcc2021:
I made a video to show the issue.

https://youtu.be/fbHqHQu6js8

The simple test case is:
`
public class TestWaterAndInstanced extends SimpleApplication {
    
    public static void main(String[] args) {
        TestWaterAndInstanced test = new TestWaterAndInstanced();
        test.start();
    }
    
    private Box mesh;
    private Material mat;
    private InstancedNode instanceNode;
    
    @Override
    public void simpleInitApp() {
        flyCam.setMoveSpeed(10);
        
        DirectionalLight light = new DirectionalLight();
        light.setDirection(new Vector3f(-1, -1, -1));
        rootNode.addLight(light);
        
        mat = new Material(assetManager, "Common/MatDefs/Light/Lighting.j3md");
        mat.setBoolean("UseInstancing", true);
        
        mesh = new Box(0.5f, 0.5f, 0.5f);
        
[...]

PR #2008, opened 2023-05-11
solve issue #2007 (instanced objects are culled when using WaterFilter)

Ali-RS:
Solves https://github.com/jMonkeyEngine/jmonkeyengine/issues/2007

The issue is that InstancedGeometry uses the wrong camera for "instance culling" check when rendering the scene from multiple viewports (e.g. main viewport and water reflection viewport) and this causes objects to get culled wrongly.

It is because `InstancedGeometry.updateInstances()` use the old camera (from the previous viewport) before it is updated in the below method:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/25a3407c9e403dbc0250430a304031c169636a03/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java#L420-L423

this PR changed the code so it passes the camera into `updateInstances()` by the caller method.

Because `updateInstances()` is a `public` method on InstancedGeometry (not sure why it is so?) this PR should be considered a breaking change I believe because the method signature is changed!

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2022 jMonkeyEngine
+ * Copyright (c) 2009-2023 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -75,7 +75,6 @@ public class InstancedGeometry extends Geometry {
 
     private int firstUnusedIndex = 0;
     private int numVisibleInstances = 0;
-    private Camera cam;
 
     public InstancedGeometry() {
         super();
@@ -276,7 +275,7 @@ public class InstancedGeometry extends Geometry {
         }
     }
 
-    public void updateInstances() {
+    public void updateInstances(Camera cam) {
         FloatBuffer fb = (FloatBuffer) transformInstanceData.getData();
         fb.limit(fb.capacity());
         fb.position(0);
@@ -414,12 +413,6 @@ public class InstancedGeometry extends Geometry {
             allData.addAll(Arrays.asList(globalInstanceData));
         }
         allInstanceData = allData.toArray(new VertexBuffer[allData.size()]);
-    }
-
-    @Override
-    public boolean checkCulling(Camera cam) {
-        this.cam = cam;
-        return super.checkCulling(cam);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


stephengold:
Yes it's a breaking change. No, I don't know why the method is `public`.

Ali-RS:

Ali-RS:
We might want to consider this for 3.6.1 (even if a breaking change), because

1- the bug is important enough 
2- this won't affect 99% of JME apps in a bad way I believe, because `updateInstances()` is not meant to be called by user code.

stephengold:
I'm willing to consider this for 3.6.1, but I'm not convinced. To me, the issue looks like a corner case. What percentage of JME apps do you imagine combine instanced objects with a WaterFilter?

Ali-RS:
>What percentage of JME apps do you imagine combine instanced objects with a WaterFilter?

Note, this is not just about the water filter, but any case that a scene is rendered by multiple viewports. The water filter is just an example of such a case.

stephengold:
In that case I would expect shadow filters to be affected, which greatly raises the priority of this fix.

Still, I can't see myself listing "breaking changes" in the release notes for a patch. Isn't there *any* way to solve #2007 without introducing a breaking change?

Ali-RS:
>I would expect shadow filters to be affected,

Seems this does not affect the shadow filter because it does not call `RenderManager.renderScene()` and so `Spatial.checkCulling(Camera cam)` is not called.

stephengold:
It seems to me that with the following additional changes, this PR would no longer be a breaking change:

```diff
--- a/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/instancing/InstancedGeometry.java
@@ -75,6 +75,7 @@ public class InstancedGeometry extends Geometry {
 
     private int firstUnusedIndex = 0;
     private int numVisibleInstances = 0;
+    private Camera cam;
 
     public InstancedGeometry() {
         super();
@@ -275,6 +276,13 @@ public class InstancedGeometry extends Geometry {
         }
     }
 
+    /**
+     * @Deprecated use {@link #updateInstances(com.jme3.renderer.Camera)
+     */
+    public void updateInstances() {
+        updateInstances(cam);
+    }
+
     public void updateInstances(Camera cam) {
         FloatBuffer fb = (FloatBuffer) transformInstanceData.getData();
         fb.limit(fb.capacity());
@@ -415,6 +423,12 @@ public class InstancedGeometry extends Geometry {
         allInstanceData = allData.toArray(new VertexBuffer[allData.size()]);
     }
 
+    @Override
+    public boolean checkCulling(Camera cam) {
+        this.cam = cam;
+        return super.checkCulling(cam);
+    }
+
     @Override
     public int collideWith(Collidable other, CollisionResults results) {
         return 0; // Ignore collision
```

stephengold:
PR #2036 makes this not a breaking change, so it will be included in v3.6.1 .

PR #2022, opened 2023-05-27
com.jme3.anim.tween.action.BlendSpace: basic javadoc

pavly-gerges:
This PR adds a basic javadoc for the BlendSpace class.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -38,7 +38,7 @@ package com.jme3.anim.tween.action;
  * <p>
  * Notes:
  * <li> Blending is the action of mixing between 2 successive animation {@link BlendableAction}s by interpolating their transforms and 
- * then applying them on a {@link HasLocalTransform} object, the {@link BlendSpace} provides this blending action with a blend weight value. </li>
+ * then applying the result on the assigned {@link HasLocalTransform} object, the {@link BlendSpace} provides this blending action with a blend weight value. </li>
  * <li> The blend weight is the value for the interpolation for the target transforms. </li>
  * <li> The blend weight value should lie in this interval [0, 1]. </li>
  * </p>
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -69,7 +69,8 @@ public interface BlendSpace {
      * Provides the blend weight value to the assigned {@link BlendAction} instance,
      * this value will be used for interpolating a collection of actions' transformations (keyframes).
      * 
-     * @return the blending weight value in the range from 0 to 1.
+     * @return the blending weight value in the range from 0 to 1, 
+     *         negative values and values above 1 aren't allowed.
      * @see LinearBlendSpace#getWeight()
      */
     public float getWeight();
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -68,9 +68,8 @@ public interface BlendSpace {
     /**
      * Provides the blend weight value to the assigned {@link BlendAction} instance,
      * this value will be used for interpolating a collection of actions' transformations (keyframes).
-     * The Blend weight value must be in the range from 0 to 1.
      * 
-     * @return the blending weight value.
+     * @return the blending weight value in the range from 0 to 1.
      * @see LinearBlendSpace#getWeight()
      */
     public float getWeight();
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -49,7 +49,7 @@ package com.jme3.anim.tween.action;
  * a blend value of 1 and the second action will use the blend space weight as a value for the interpolation. </li>
  * <li> In case of (Blending weight = 0), the blending hasn't started yet, only the first action will be interpolated at (weight = 1). </li>
  * <li> In case of (Blending weight = 1), the blending is finished and only the second action will continue to run at (weight = 1). </li>
- * <li> Negative values and values greater than 1 aren't allowed (extrapolations are disabled). </li>
+ * <li> Negative values and values greater than 1 aren't allowed (extrapolations aren't allowed). </li>
  * <li> Find more at {@link BlendAction#doInterpolate(double)} and {@link BlendAction#collectTransform(HasLocalTransform, Transform, float, BlendableAction)}. </li>
  * </p>
  * 
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -49,7 +49,7 @@ package com.jme3.anim.tween.action;
  * a blend value of 1 and the second action will use the blend space weight as a value for the interpolation. </li>
  * <li> In case of (Blending weight = 0), the blending hasn't started yet, only the first action will be interpolated at (weight = 1). </li>
  * <li> In case of (Blending weight = 1), the blending is finished and only the second action will continue to run at (weight = 1). </li>
- * <li> Negative values and values greater than 1 aren't allowed. </li>
+ * <li> Negative values and values greater than 1 aren't allowed (extrapolations are disabled). </li>
  * <li> Find more at {@link BlendAction#doInterpolate(double)} and {@link BlendAction#collectTransform(HasLocalTransform, Transform, float, BlendableAction)}. </li>
  * </p>
  * 
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -68,7 +68,7 @@ public interface BlendSpace {
     /**
      * Provides the blend weight value to the assigned {@link BlendAction} instance,
      * this value will be used for interpolating a collection of actions' transformations (keyframes).
-     * Blend weight value must be in the range from 0 to 1.
+     * The Blend weight value must be in the range from 0 to 1.
      * 
      * @return the blending weight value.
      * @see LinearBlendSpace#getWeight()
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -66,7 +66,7 @@ public interface BlendSpace {
 
     /**
      * Used for passing the desired weight to the selected {@link BlendAction} instance,
-     * this value will be used for interpolating a collection of actions' transformations.
+     * this value will be used for interpolating a collection of actions' transformations (keyframes).
      * 
      * @return the blending weight value.
      * @see LinearBlendSpace#getWeight()
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
@@ -40,7 +40,7 @@ package com.jme3.anim.tween.action;
  * <li> Blending is the action of mixing between 2 successive animation {@link BlendableAction}s by interpolating their transforms and 
  * then applying the result on the assigned {@link HasLocalTransform} object, the {@link BlendSpace} provides this blending action with a blend weight value. </li>
  * <li> The blend weight is the value for the interpolation for the target transforms. </li>
- * <li> The blend weight value should lie in this interval [0, 1]. </li>
+ * <li> The blend weight value must be in this interval [0, 1]. </li>
  * </p>
  * 
  * <p>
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
```diff
@@ -1,10 +1,84 @@
...
+ * then applying them on a {@link HasLocalTransform} object, the {@link BlendSpace} provides this blending action with a blend weight value. </li>
+ * <li> The blend weight is the value for the interpolation for the target transforms. </li>
+ * <li> The blend weight value should lie in this interval [0, 1]. </li>
+ * </p>
+ * 
+ * <p>
+ * Different blending weight case scenarios managed by {@link BlendAction} internally:
+ * <li> In case of (0 < Blending weight < 1), the blending is executed each update among 2 actions, the first action will use 
+ * a blend value of 1 and the second action will use the blend space weight as a value for the interpolation. </li>
+ * <li> In case of (Blending weight <= 0), the blending hasn't started yet, only the first action will be interpolated at (weight = 1). </li>
```
Scrappers-glitch, line 50: I hope I am correct about this; because the code is not clear about the negative values!

So, Are the negative values treated the same as the {0}?

Ali-RS: BlendSpace should never return a negative weight. You only need to document that `getWeight()` must return a value in [0, 1].

Scrappers-glitch: BlendAction has a conditional statements that handle different blend weight values for the user, I guess this should be documented in the root origin of blend weight which is the blend space?

Scrappers-glitch: I am trying to document the blend weight values behavior.

Scrappers-glitch: Okay, I still donot fully agree with you, I think every side effect should be documented, as a user if I am going to build a new blend space on my game, it would be nice if the API tells me everything about the blend weight behavior.

Note: if you think that there is a defective behavior regarding the negative values, then it should be corrected before we go over documenting this API.

Ali-RS: I think there is a missing check here:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/7025310b8f767e0b5956468f9ae2ef6ca8dd8604/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendableAction.java#L93

In the above method, we can throw an illegal argument exception if it is not in [0, 1].

Like what I did here:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/7025310b8f767e0b5956468f9ae2ef6ca8dd8604/jme3-core/src/main/java/com/jme3/anim/tween/action/BlendableAction.java#L120-L126

Scrappers-glitch: Which is better, throwing a business exception or printing a warning message denoting that the interpolation is skipped? 

I know we must be consistent, but I am just asking.

Scrappers-glitch: I think we should discuss these missing checks on forums, to keep this PR clean from unrelated contents.

Ali-RS: >interpolation is skipped

I do not how we can skip it, but it feels wrong to me. We should either throw an exception or clamp it into [0, 1]. 

And 99% of the time, people will use LinererBlendSpace which will never return a value other than [0, 1].

If someone wants to create a BlendSpace that returns a negative weight that is their issue.

Scrappers-glitch: According to `BlendAction#collectTransform`, negative values are treated the same as zero, so I guess they are skipped.

But you are right regarding people's blame 🤣, I agree throwing an exception will be the righteous choice.

Scrappers-glitch: We can then document it here as "Negative values are not allowed".

Ali-RS: >"Negative values are not allowed".

and values greater than 1, or just saying "weight must be in [0, 1]" which indicates both.

Scrappers-glitch: Looks great, will you work on the PR to add this restriction, please?

Ali-RS: Yes, I can submit a PR.

Scrappers-glitch: I have added the requested changes, if you think that this will satisfy the new behavior, please resolve this conversation.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
```diff
@@ -1,10 +1,84 @@
...
+     * 
+     * @param action the blend action instance that will utilize this blend-space.
+     */
     public void setBlendAction(BlendAction action);
 
+    /**
+     * Provides the blend weight value to the assigned {@link BlendAction} instance,
+     * this value will be used for interpolating a collection of actions' transformations (keyframes).
+     * 
+     * @return the blending weight value.
```
Ali-RS, line 72: I think we should also document it here:  "weight must be in [0, 1] range"

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
```diff
@@ -1,10 +1,84 @@
...
+ * <li> The blend weight value should lie in this interval [0, 1]. </li>
+ * </p>
+ * 
+ * <p>
+ * Different blending weight case scenarios managed by {@link BlendAction} internally:
+ * <li> In case of (0 < Blending weight < 1), the blending is executed each update among 2 actions, the first action will use 
+ * a blend value of 1 and the second action will use the blend space weight as a value for the interpolation. </li>
+ * <li> In case of (Blending weight = 0), the blending hasn't started yet, only the first action will be interpolated at (weight = 1). </li>
+ * <li> In case of (Blending weight = 1), the blending is finished and only the second action will continue to run at (weight = 1). </li>
+ * <li> Negative values and values greater than 1 aren't allowed and will result in an IllegalArgumentException. </li>
```
Ali-RS, line 52: ><li> Negative values and values greater than 1 aren't allowed and will result in an IllegalArgumentException. </li>

The execption is not in this class scope and should not be documented here I belive. I think this should rather be documented on BlendableAction. I would sugest to remove it.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
Scrappers-glitch: Oh didn't notice that you refer to 'here' meaning in this method entry, sorry, will add this.

jme3-core/src/main/java/com/jme3/anim/tween/action/BlendSpace.java
```diff
@@ -1,10 +1,85 @@
...
+ * <li> The blend weight is the value for the interpolation for the target transforms. </li>
+ * <li> The blend weight value must be in this interval [0, 1]. </li>
+ * </p>
+ * 
+ * <p>
+ * Different blending weight case scenarios managed by {@link BlendAction} internally:
+ * <li> In case of (0 < Blending weight < 1), the blending is executed each update among 2 actions, the first action will use 
+ * a blend value of 1 and the second action will use the blend space weight as a value for the interpolation. </li>
+ * <li> In case of (Blending weight = 0), the blending hasn't started yet, only the first action will be interpolated at (weight = 1). </li>
+ * <li> In case of (Blending weight = 1), the blending is finished and only the second action will continue to run at (weight = 1). </li>
```
Ali-RS, line 51: This looks fine, however, I think the below is a bit nicer and simplified (just my opinion)

"If weight is 0, only the first action will run, if it is 1, only the second action will run, and for anything in between, it will blend actions by the weight factor".

Scrappers-glitch: The way I write documentation is the same way I study Maths and programming in general, I would like things to be summarized and not use too much English. 

I think it's easier to read this way, in a list.

Scrappers-glitch: Btw, in fact, the first action doesn't really use the blend weight, it always runs at (weight = 1) when it's triggered, this fact should be at the user's hands through the documentation.

I really cannot see a better way to summarize the code in a readable format better than this list of different value scenarios, it is a proven good mathematical way to explain things as well, and a lot of Maths textbooks use this method to elaborate the information in a human-readable and testable format, far better than raw English.

pspeed42: Taken to the logical conclusion: just include the code itself.

Scrappers-glitch: > Taken to the logical conclusion: just include the code itself.

It's my maximum capability, I think I cannot do it better than this.

stephengold:
Javadoc is important, and I appreciate the considerable effort that's gone into this PR. Unfortunately, it's still very confusing. @Scrappers-glitch are you willing to continue refining it?

Scrappers-glitch:
@stephengold In this PR, I chose to stick to describing the behavior of the pattern used, or maybe the interaction of the API components, if you want to understand more, please use this paradigm designed by Remy (Most of this documentation is based on this paradigm, the talks on forums and [my own techdemos](https://github.com/Monkey-Droid/DBTraining/tree/master/app/src/main/java/com/scrappers/dbtraining/mainScreens/prefaceScreen/renderer/animationWrapper)): 
[Monkanim: new animation system in the works](https://hub.jmonkeyengine.org/t/monkanim-new-animation-system-in-the-works/36359/1)

stephengold:
Figuring out why the javadoc is confusing and writing a clear explanation seems like a lot of effort. I don't want to spend more time on this PR.

Scrappers-glitch:
BlendSpace is all about a provider interface. It provides the blend weight value to blend between 2 actions. Blend weight is then used as the value of interpolation. That simple. I like to think of it from a higher perspective as an interface that enables the user to adjust the behavior of blending, and that blend weight value is a read-only value. It could be manipulated using the mediator `BlendSpace#setValue(...)`.

Scrappers-glitch:
For me, this PR has completed. Please let me know if there are any changes you want to add, otherwise it's ready for merging.

pavly-gerges:
It would be great if you consider this PR on jme-3.8 if possible. It has been a year since no one has reviewed this PR and requested any changes. I guess its complete. However, if clarification is required, I will happily try to provide.

pavly-gerges:
I would like to hear from @stephengold; as there were some conflicts about the way things got to be explained so far. In the meantime, I will update the copyright and see if anything could be added.

stephengold:
> I would like to hear from @stephengold; as there were some conflicts about the way things got to be explained so far.

As I said a year ago, I don't want to spend any more time on this PR.

pavly-gerges:
I appreciate your comment Mr. Wyatt @capdevon. However, sometimes the AI Models introduce polished solutions which are really annoying.

For example, what is the core difference between:
```java
* A provider interface which provides a value {@link BlendSpace#getWeight()} to control the blending between 2 successive actions in a {@link BlendAction}.
```
And this:
```java
A provider interface that supplies a value to control the blending between
 * two successive actions in a {@link BlendAction}. 
```
Really?

I will go over them anyway, and see if there are things to add, but core changes, I don't think so.

yaRnMcDonuts:
I believe @pavly-gerges did a decent job addressing some of the initial concerns that were raised. 

I wouldn't say that your wording was the wording I would have use, but it did seem to be understandable enough and objectively correct in its descripion of methods when I reviewed it.

But I also agree that the suggested changes from @capdevon improve upon it even further. 

I'm typically skeptical of using AI for jme related things (it seems to get things wrong about jme quite a lot), but I think for javadocs it is one place where Ai can do a goodjob to help refine the word choice and make things more understandable. I wouldn't trust AI to write javadocs all on its own, but it certainly has done a good job refining and finalizing the existing javadoc that you gave it.

yaRnMcDonuts:
> Really?
> 
> I will go over them anyway, and see if there are things to add, but core changes, I don't think so.

I think most of your wording is acceptable, there's only a few places where I feel that the AI generated changes offered some improvements.

The more I compare, the less big differences I see.

But, for example, 1 place that I do prefer the wordage of the AI generated javadoc would be here:

` <li> The blend weight value must be in this interval [0, 1]. </li>`..

The proposed changes swap at the word "interval" with "range", and I'd say that the word "range" is somewhat better of a choice here. "Inteval" is still understandable but the word "range" is probably a better choice of the two.

yaRnMcDonuts:
@pavly-gerges Do you have plans to make any more adjustments to this PR?

If not, then I will merge it in about 24 hours.

I do not think that any of the suggested wording changes are major enough to warrant holding up a PR for a basic javadoc any longer. Especially considering that there are language differences amongst jme users and english is not the first language for many, so the perceived "best wording" is not always going to be consistent and agreed upon. 

In its current state, your javadoc is understandable does a satisfactory job explaining the class' functionality and ultimately does a better job than having no javadoc at all. And if anyone wants to improve some wording in the future, then that can always be adjusted in a future PR.

pavly-gerges:
> @pavly-gerges Do you have plans to make any more adjustments to this PR?

I am willing to do some refinements, just for the benefit of the community. Coming up today, keep updated!

pavly-gerges:
For me, this is ready to be merged. Please note that, if you find anything that you would like to change, I have no problems with that; you can just change them directly before merging.

Issue #2023, opened 2023-05-28
getRotationColumn() may return incorrect results


stephengold:
`Quaternion.getRotationColumn()` is supposed to handle non-normalized quaternions. According to the javadoc:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/7025310b8f767e0b5956468f9ae2ef6ca8dd8604/jme3-core/src/main/java/com/jme3/math/Quaternion.java#L635-L639

However, this claim is untrue. For `q = new Quaternion(0f, -0.2f, 0f, 0.6f)`:
 + `q.getRotationColumn(0)` returns `(0.8735089, 0.0, 0.37947333)`, while
 + `q.normalizeLocal().getRotationColumn(0)` returns `(0.8, 0.0, 0.6)`

The root cause is at line 654:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/7025310b8f767e0b5956468f9ae2ef6ca8dd8604/jme3-core/src/main/java/com/jme3/math/Quaternion.java#L653-L655

The correct assignment would be `norm = 1f / norm;` .

PR #2024, opened 2023-05-28
solve issue #2023 (getRotationColumn() returns wrong result)

stephengold:
An automated testcase is included, to detect future regressions.


stephengold:
Oops! Now `LightFilterTest` is failing. The camera orientation gets set to an impossible value at line 137: look direction and up direction are the same! Based on the comments, I don't believe this was intentional, so I will alter the test.

stephengold:
Unless there's substantial discussion, I plan to integrate this change in about 24 hours.

stephengold:
On further consideration, it's quite possible existing apps rely on the incorrect result. Postpone this until 3.7 or 4.0 .

PR #2026, opened 2023-05-28
jme3-core:  document a few places where normalization is assumed

stephengold:
The PR documents a few places where normalization is assumed.

JME assumes that the local rotations of scene-graph spatials are normalized. A violation of this assumption caused the issue @yaRnMcDonuts recently reported [at the Discourse Forum/Hub](https://hub.jmonkeyengine.org/t/solved-static-rigidbodycontrol-is-wrongly-offsetting-location-of-spatial/46792).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -1152,6 +1152,9 @@ public abstract class Spatial implements Savable, Cloneable, Collidable,
     /**
      * <code>setLocalTransform</code> sets the local transform of this
      * spatial.
+     *
+     * @param t the desired local transform (not null, {@code t.rot.norm() == 1}
+     *     or nearly so, unaffected)
      */
     @Override
     public void setLocalTransform(Transform t) {
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -1063,8 +1063,8 @@ public abstract class Spatial implements Savable, Cloneable, Collidable,
     /**
      * <code>setLocalRotation</code> sets the local rotation of this node.
      *
-     * @param quaternion
-     *            the new local rotation (not null, norm=1 or nearly so)
+     * @param quaternion the new local rotation (not null,
+     *     {@code quaternion.norm()} approximately equal to 1, unaffected)
      */
     public void setLocalRotation(Quaternion quaternion) {
         localTransform.setRotation(quaternion);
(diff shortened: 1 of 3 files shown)


jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -1064,7 +1064,7 @@ public void setLocalRotation(Matrix3f rotation) {
      * <code>setLocalRotation</code> sets the local rotation of this node.
      *
      * @param quaternion
-     *            the new local rotation.
+     *            the new local rotation (not null, norm=1 or nearly so)
```
Scrappers-glitch, line 1067: Hello, I think the word 'nearly so' needs some elaboration, does that mean the method can work with norm<1?

stephengold: The situation is as follows:
+ Due to floating-point round-off, it's often impractical to make the norm exactly equal to 1.
+ Any deviation from norm=1 is going to distort the world transform. The further the local rotation norm deviates from 1, the greater the distortion will be.

Would "norm approximately equal to 1" be clearer?

Scrappers-glitch: Yes, I think the word 'approximately equal to' is clearer. Thank you.

stephengold:

stephengold:
I assume that the `Quaternion.mult()` and `multLocal()` methods involving vectors were named by analogy with `Matrix3f`. Both sets of methods serve analogous purposes: for restricted values (unit quaternions and 3x3 rotation matrices) the methods perform 3-D rotations on vectors.

Unfortunately, mathematicians defined multiplication of a quaternion with a vector to suit their purposes, and that definition conflicts with what `Quaternion.mult()` actually does to a 3-D vector, namely: rotate (not multiply!) the vector. Seeing these methods named `mult()` caused me to add javadoc disclaimers at #1725, lest someone using `Quaternion` for something other than rotation be misled by the names. (Interestingly, the `mult()` methods that *don't* involve vectors *do* agree with the mathematical definition of multiplication, which only magnifies the confusion.)

While it *seems* expedient for rotation methods to assume their inputs are normalized and (for such inputs) produce normalized outputs, it leads to subtle bugs. For instance, if you start with a normalized value and then apply enough math operations, floating-point round-off may cause the norms to diverge substantially from 1.

Furthermore, naive users and buggy tools (such as the one that bit Ryan) generate unintentional non-normalized quaternions with surprising frequency.

I realize the disclaimers added in this PR are unlikely to mean much to naive users. They're mainly to sooth my conscience at the lack of run-time checks.

In my own projects, I prefer to check for a normal quaternion *before* each rotation. The cost seems slight compared to the time I spend debugging issues. But since such run-time checks seem contrary to the JME Way, so I have not added them here.

Issue #1990, opened 2023-03-28
AndroidNativeBufferAllocator memory leak


Ali-RS:
For regular Java-managed direct buffers, JME does not need to clean them up as the GC will clean them up.

For any memory that Java is not allocating, special support (like PhantomReference) will have to be used to make sure to free the native memory.

I think until we switch to using `jme3-alloc` in (in JME 3.7) for JME 3.6.1 we should switch to PrimitiveAllocator on Android.

Ali-RS:
In the below link you can find out why direct byte buffers allocated via JNI env->NewDirectByteBuffer() are not released when GCed but the ones created with ByteBuffer.allocateDirect() are freed with GC.

https://stackoverflow.com/a/35364247

stephengold:
By "cleaned" you mean that the corresponding native objects are freed?

Ali-RS:
I have never used JNI so I do not know much about the details so at this point my knowledge is limited to what was mentioned in the above link. It says a DirectByteBufer reterned by JNI has no `Cleaner` associated with that to free the native memory but `ByteBuffer.allocateDirect()` has a `Cleaner` object associated with it:

The relevant source from DirectByteBuffer I get from OpenJDK 20:

```
       private static class Deallocator
        implements Runnable
    {

        private long address;
        private long size;
        private int capacity;

        private Deallocator(long address, long size, int capacity) {
            assert (address != 0);
            this.address = address;
            this.size = size;
            this.capacity = capacity;
        }

        public void run() {
            if (address == 0) {
                // Paranoia
                return;
            }
            UNSAFE.freeMemory(address);
            address = 0;
            Bits.unreserveMemory(size, capacity);
        }

    }

    private final Cleaner cleaner;

    public Cleaner cleaner() { return cleaner; }

```

```
// Primary constructor
    //
    DirectByteBuffer(int cap) {                   // package-private

        super(-1, 0, cap, cap, null);
        boolean pa = VM.isDirectMemoryPageAligned();
        int ps = Bits.pageSize();
        long size = Math.max(1L, (long)cap + (pa ? ps : 0));
        Bits.reserveMemory(size, cap);

        long base = 0;
        try {
            base = UNSAFE.allocateMemory(size);
        } catch (OutOfMemoryError x) {
            Bits.unreserveMemory(size, cap);
            throw x;
        }
        UNSAFE.setMemory(base, size, (byte) 0);
        if (pa && (base % ps != 0)) {
[...]

PR #2029, opened 2023-06-04
By default use PrimitiveAllocator on Android

Ali-RS:
This PR makes PrimitiveAllocator the default buffer allocator due to a memory leak issue with AndroidNativeBufferAllocator.

See issue https://github.com/jMonkeyEngine/jmonkeyengine/issues/1990

In the below link you can find out why direct byte buffers allocated via JNI (`env->NewDirectByteBuffer()`) are not cleaned when GCed but the ones created with `ByteBuffer.allocateDirect()` are cleaned with GC. 

https://stackoverflow.com/a/35364247


Ali-RS:

Issue #2030, opened 2023-06-05
Add support for Multiple Monitors in jme-LWJGL3


bob0bob:
Add support for multiple monitors.
Add a feature so that when a "Full Screen" window is created, that you can tell it which monitor to create the window on.
Add a feature so that the application can call context to get a list of monitors that OPENGL found.   It returns them in an ArrayList so that the programmers can select a monitor from the list.  JME will take the pos of the monitor from the arraylist to get its handle.  So if you have 2 monitors, you will have 2 in the list.  So to tell JME which monitor to create the window on it would be 0 or 1.  The array position in the list.

The thought behind this is the program gets a list of monitors and then they can use that list in their settings for the user to select which monitor to us.   
Since the ID of the monitor changes between each launch, I went with the position from the arraylist that it returned.  So many if user changes the order of the monitors then the program will launch on a different screen.  Minor.

Added in AppSettings a way to get/set Monitor.   Monitor value is used only when creating a Full Screen.

pavly-gerges:
Has this issue been resolved? Why is it still open?

yaRnMcDonuts:
I have been waiting to close this until the PR gets tested more, since I ran into some possible issues with multiple monitor support on one of my devices. I also am not sure if it has gotten significant testing, no one has reported any success nor failure while running this test yet, unfortunately.

And I personally was not able to get the Example to work to display fullscreen on the secondary monitor on my primary device.

At first I thought it still worked in non fullscreen mode, but then I realized that it was drawing a single wide viewport that stretched across both of my monitors when not in full screen mode. So it seems to be treating both of my monitors as if they are one very wide monitor.

And when in full screen mode, it appears to render properly at first. But the default fullscreen resolution is equal to the value of both my screens added together (I.E. isntead of being ~1500 width resolution, my resolution is ~3000 because I have 2 each about 1500 pixels in width). So the display on the primary monitor ends up having a resolution in fullscreen mode that is double what it shouuld be, which also cases the sizes for all gui components to be shrinked in half as well.

I also have not heard if anyone else (other than the author of the PR) has had success using Multiple Monitor support.

If this feature can be fixed to not cause any resolution issues, then I'd like to try to keep this PR for 3.8. I assume that the issues I'm experiencing are device related, since the author of the PR has been using this in their own project for a while. So hopefully this can all be fixed.

@bob0bob does everything with multiple monitor suppor tstill work okay on your device that you developed this on?

[...]

PR #2031, opened 2023-06-05
 Add support for Multiple Monitors in jme-LWJGL3 #2030 

bob0bob:
Add support for multiple monitors.
Add a feature so that when a "Full Screen" window is created, that you can tell it which monitor to create the window on.
Add a feature so that the application can call context to get a list of monitors that OPENGL found. It returns them in an ArrayList so that the programmers can select a monitor from the list. JME will take the pos of the monitor from the arraylist to get its handle. So if you have 2 monitors, you will have 2 in the list. So to tell JME which monitor to create the window on it would be 0 or 1. The array position in the list.

The thought behind this is the program gets a list of monitors and then they can use that list in their settings for the user to select which monitor to us.
Since the ID of the monitor changes between each launch, I went with the position from the arraylist that it returned. So many if user changes the order of the monitors then the program will launch on a different screen. Minor.

Added in AppSettings a way to get/set Monitor. Monitor value is used only when creating a Full Screen.


tonihele:
> I don't have a multi-monitor setup myself, so it didn't occur to me that this might be wanted. I assumed it would usually be easier to create windows in the primary display and use a window manager to move them to where they're wanted. Is that not the case?

I think it is impossible in many windowing systems. This is the way to go.

jme3-core/src/main/java/com/jme3/system/AppSettings.java
```diff
@@ -1479,4 +1480,32 @@ public int getWindowYPosition() {
...
+    public int getMonitor() {
+        return getInteger("Monitor");
+    }
+
+    /**
+     * Sets the monitor number used when creating a window.  The position
+     * number is the number in the list of monitors GlfwGetMonitors returns.
+     *
+     * <p>This setting is used only with LWJGL3, it defines which monitor
+     * to use when creating a OpenGL window. its default value is -1.
```
Ali-RS, line 1503: >its default value is -1.

But looks like it is 0.

Ali-RS:
Does any game let you select the monitor when running the game?

jme3-core/src/main/java/com/jme3/system/AppSettings.java
bob0bob: Need to update comment.

bob0bob:
Yes, this is for full screen only.  You can't select a monitor for "WINDOW", it is forced to "primary OS monitor".   When doing full screen, if you don't select it will do it on primary.  If your primary is the window you want, then it is fine.  But for it to launch and then move a FULL SCREEN is not nice.  When you can cause it to launch on the correct window.

I'll update the copyright.

bob0bob:
Forgot..  TestResizableMonitor was failing after creating a window in full screen when using a different monitor.
"reshape" function was called before "SimpleInitApp()" was called when creating a window other than "NULL".  I put a check in there to not update label.   So 'txt' variable was "NULL" because it had not called simpleInitApp(), yet. 

I didn't look into way JME was calling reship before it did the initialization.

bob0bob:
I updated the comments.  Let me know if you find anything else.

bob0bob:
I missed the LegacyApplication update for getMonitors()

Scrappers-glitch (changes requested):
Hello Kevin, thanks for this feature, I would like to request some changes as a part of consistency with jMonkeyEngine, please correct me if you think there are untrue statements or you disagree with something, many of these requested changes have duplicates across the PR that I am not referring to it here, so please pay attention to them as far as you can.

jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -502,4 +503,18 @@ protected void createContext(AppSettings settings) {
             // TODO: Fix deadlock that happens after the error (throw runtime exception?)
         }
     }
+
+	@Override
+	public Monitors getMonitors()
+	{
+		// TODO Auto-generated method stub
```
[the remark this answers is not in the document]
Scrappers-glitch: I think it doesn't really matter if you return a null or an empty object, what matters here is there should be a transparent message indicating that this feature is not supported.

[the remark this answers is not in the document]
Scrappers-glitch: And you don't really need a return statement, it's an unreachable code anyway.

jme3-core/src/main/java/com/jme3/system/AppSettings.java
```diff
@@ -1479,4 +1480,32 @@ public int getWindowYPosition() {
...
+     * Gets the monitor number used when creating a window.
+     *
+     * <p>This setting is used only with LWJGL3, it defines which monitor
+     * to use when creating a OpenGL window.
+     *
+     * @return the desired monitor used when creating a OpenGL window
+     * @see #setMonitor(long)
+     */
+    public int getMonitor() {
+        return getInteger("Monitor");
```
Scrappers-glitch, line 1495: It's better to wrap the `"Monitor"` string key in a constant object for future refactoring and testing purposes, you can use the `MonitorInfo` to define default Keys there.

jme3-core/src/main/java/com/jme3/system/MonitorInfo.java
```diff
@@ -0,0 +1,73 @@
...
+package com.jme3.system;
+
+/**
+ * This class holds information about the monitor that was
+ * returned by glfwGetMonitors() calls in the context
+ * class
+ * 
+ * @author Kevin Bales
+ */
+public class MonitorInfo {
```
Scrappers-glitch, line 41: This should be marked as `final`, and I don't know if it needs to use the conventional style of getters and setters since their advantage here is not very obvious, but I think encapsulation should be the rule here, in case direct changes need to be applied to the member variables in the future.

pspeed42:
As a general comment, this new API feels a little weird to me.  Having a Monitors class that is essentially just a list of monitor infos with one odd control method feels off.  Even weirder if this is ever extended to provide display information about the monitors other than "what it is right now".

What I would have expected from this feature is some way to ask JME what monitors are available that returns a List of info objects... then some method to set the current monitor on the same thing I asked for the list of monitors.

something like:
```
JmeSystem.getMonitorInfos() : List<MonitorInfo>
JmeSystem.setCurrentMonitor(MonitorInfo)
```
And even that might be a dead-end design if we ever want to support changing things about a particular monitor (like current display settings or whatever).

Instead of MonitorInfo we could have a Monitor interface with the getters (no setters necessary) then different system implementations can implement the Monitor interface as they see fit.  (An abstract version could be provided to make this easier for the system implementations that only support one display... and perhaps Display is a better name than Monitor.)

So:
```
JmeSystem.getAvailableDisplays() : List<Display>
JmeSystem.setCurrentDisplay(Display)
```

Regardless of all that, a MonitorInfo class with all public fields becomes a maintenance burden forever and should be avoided.

jme3-core/src/main/java/com/jme3/app/LegacyApplication.java
```diff
@@ -852,4 +852,28 @@ public Object call() {
...
+    
+    /**
+     * This call will return a list of Monitors that glfwGetMonitors()
+     * returns and information about the monitor, like width, height, 
+     * and refresh rate.
+     * 
+     * @return returns a list of monitors and their information.
+     */
+    public Monitors getMonitors()
+    {
```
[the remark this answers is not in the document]
bob0bob: sure no problem

bob0bob:
@pspeed42, 
I thought about just capturing the list of all the available mode that monitor support and add them to the list.   I didn't do this because when I run it on my both of my 2k monitors,  it shows the max being 800x600.   when it is 2k.   It doesn't report back (At least my monitors) accurately.  
Also, the monitor names come from Windows and even though I have Acer monitors, the driver comes up generic monitor.

The monitors do 2k 170hz on both of them, but they report is not even close.  So I thought if we listed resolution, people might think that is the only option because it doesn't report correctly.

I can change the Monitor classes, I could create getter/setters  and even a interface class.  That would all be easy.

Doing a setCurrentDisplay()  would be very hard to do.  The monitor selection is a "WINDOW CREATION"  so this is before you get into the App simpleInitApp().   So the only option to set which monitor in full screen is through the ini.  A restart of the application is require after you allow user to select a monitor.  I didn't see an LWJGL command to move monitor to another monitor, just during window creation.

Let me know if someone knows something I don't and I can look into it and change it.

pspeed42:
The point about the resolution settings is that someday the OS/drivers/etc. may fix those issues and those APIs would work like we expect... so it's better not to automatically design dead-ends into JME's API.

As to setCurrentDisplay(), I was confused by this which seemed to be essentially doing the same thing:
public void setPrimaryMonitor(int monPos)

But maybe change what I said to setPrimaryDisplay() and it makes more sense?  Maybe it should be on AppSettings?

Either way, I prefer a "request the available options" and "set the option I want" API versus a "get the available options and somehow call something on those options to set my option"

And having setters on MonitorInfo is probably confusing when they don't really do anything except let callers mess up the values.

Thanks for the work so far.  I don't mean to be picking on every thing.  I'm just trying to forestall some future maintenance burden.

bob0bob:
@pspeed42,
Those things make sense,   The Monitors::setPrimaryMonitor() is misleading.  It is more "Mark monitor as the active monitor".   I can change it to be setActiveMonitor().  Which is more accurate.

There is no way to do a "setPrimaryDisplay()"  sense it is before the simpleInitApp().

All the calls to getMonitors() and such is "AFTER" the fact.  It only use is for programmers to get a list of monitors if they wish to use it in "settings" gui to list the monitor to use on "Next" launch.

I have no problem with you making suggestions or even demands to fit JME.  I don't write JME and don't know the true layout.

So I can change
Monitors:setPrimaryMonitor()  to  Monitors::setActiveMonitor().  That way if the game grabs monitors again, it will reflect the current monitor that game uses on boot and they could display that if they wish.


Calling AppSettings::setMonitor()   updates the "App Ini settings".   The game could then capture the close and write out ini so it is updated,  or could write the ini in the gui control when it updates the monitor.

pspeed42:
The objects returned from getMonitors/Displays/Whatever should be immutable objects with no setters on them.  There should be no Monitors class as it is unnecessary.  A list is fine.  "Monitor" is also a limiting name since a Monitor is a Display but not all Displays are "monitors".

We should find a way to fix JME so that we can make platform specific queries before starting JME.  Else we end up with cases where the user never gets to see the app because the OS chose to put it on a display that isn't even turned on.

Setting which display to use should be done on AppSettings.  That is the "JME way", I guess... but we need a platform-independent way of reliably querying the available displays (and even the available resolutions... which we currently do not have either).  That last bit is kind of a larger issue with how JME is setup.

bob0bob:
I used "Monitor" because that is the reference inside OPENGL.   It is called a monitor in the code.

I could change everything related to "Monitor" to "Display"  to fix that idea of "Not a Monitor"
Selecting which display to use is through "AppSettings",  calling setMonitor()  (change it to setDisplay).   Since it is part of the AppSettings those values can be saved and reused on launch.

I have a check inside jme3-lwjgl3::LwjglWindow::createContext().   it checks if the number of the display doesn't show up in the list of available displays then use the "primaryMonitor" instead.

Now since OPENGL lwjgl3  call send a new number each time that application is launch we will never be able to guarantee it is the same monitor.   So if a person changed their primary monitor.  Then next time, it will open on another monitor.  We only get a list of monitor and the ID changes all the time.  So we can't determine which monitor it is.

bob0bob:
OpenGL does call it monitors.  
[https://www.glfw.org/docs/3.3/group__window.html#ga3555a418df92ad53f917597fe2f64aeb](url)

image

bob0bob:
The comment is related to you comments about a window opening up on a monitor that is not there anymore or turned on.  The best we could do is if you select monitor 1 (0-1)  and when game launches,  we could default it to 0.   But if you have 3 monitors 0-2 and you select 1.  Then remove your 2nd monitor, so the 3rd monitor becomes the 2nd monitor.  There is no way in code, to determine that is a different monitor now.

pspeed42:
GLFW = "for windows" not mobile.  So is still going to be very PC-monitor specific.  FWIW, OpenGL-ES calls it a "surface".

bob0bob:
Interesting,  but this is not for Android.   The only support is for PC.   I do not code for mobile and don't know enough to write something that would support mobile.

But again, I do not see support for multiple monitors in mobiles.  So this feature is basically for PC not other devices.  That is why if you call monitor functions, you get either NULL or 0, because it is not supported.

pspeed42:
What interface are you putting this method on?  I thought it was shared across all platforms JME supports and so it would be nice if it didn't have PC-specific stuff hard-coded onto it.

If you want to make some specific windows-style-os-specific new class to handle "the window-bearing subset" of platforms that JME supports then so be it.  Otherwise, we should code the "all platforms" API as if someone someday will do something other than UnsupportedOperationException for the platforms you don't develop implementations for.

As API developers, we need to think about the future.  Maybe in two years flip phones have two screens again and treat them as separate displays.

Then again, as already alluded to, JME provides no way to query anything about the current display settings, available display settings, or anything prior to starting the application... and that would be a really useful thing to have and maybe IS windows-style-os-specific.

Scrappers-glitch:
> But again, I do not see support for multiple monitors in mobiles. So this feature is basically for PC not other devices.

Android already supports [Multiple Displays](https://source.android.com/docs/core/display/multi_display/displays#more_displays), but it's a part of Android core hardware not GLES specifically, you have to be careful developing this API to build it across the JME Cross platform interface, I suggest we continue on forums.

bob0bob:
@Scrappers-glitch,  just got to ask.  How is handling it the way Mobile devices more compatible, when desktops call it Monitors under OpenGL.   There is not a common verbiage.

How about "screen"  it something that is not really in either.  Desktop uses Monitor, Mobile uses Displays.

pspeed42:
Let's come at this from a different direction: why do you have such a huge problem with the word "display"?

(Mobile calls them 'surfaces', by the way.  The trick is to come up with a word that can apply to all things that 'display' graphics... whether monitor, screen, surface, projector, laser array, HMD, whatever.)

Scrappers-glitch:
> @Scrappers-glitch, just got to ask. How is handling it the way Mobile devices more compatible, when desktops call it Monitors under OpenGL. There is not a common verbiage.

Display is an abstract keyword, both Graphics Monitor and Screen are types of Display, that's why I suggested Display too at the first place.

bob0bob:
Give me a couple of days, life,  and I'll get it changed to display.

bob0bob:
Sorry, been away for sometime.  Work and the farm taking up all the spare time.  I just uploaded the changes to "Devices".

bob0bob:
I didn't mean to close this.

bob0bob:
My branch was old, and merged it to current, but not sure if I missed something.  Please let me know, and I'll fix it.

tonihele:
What is the difference between Display and Monitor? They look copy paste to me.

pspeed42:
A Display is something capable of displaying graphics.  A Monitor is the big giant display sitting on your desk... which is by definition a subset of Display.

Was this about code or the naming conversation?

pspeed42:
Did github autoformat everything again?  Going to take some extra time to review with all of the non-changes mixed in.

tonihele:
> A Display is something capable of displaying graphics. A Monitor is the big giant display sitting on your desk... which is by definition a subset of Display.
> 
> Was this about code or the naming conversation?

Sorry, I'll try top be more specific. My question is that why both Java code classes are needed, `Displays `& `DisplayInfo `and `Monitors `& `MonitorInfo`:
image

What is the difference between the two in that context? They look exactly the same to me. If both classes and terms are needed, should there be some interface/abstract class rather than just copy paste code. Design-wise.

bob0bob:
Sorry about this.   When we decided to change from Monitor to Display the 2 old files related to Monitors was not deleted.  I removed them.

tonihele:
Looks ok to me. We should finally either accept, refuse or otherwise make this go forward. It is not fair to keep this hanging for so long, rather trivial addition too.

stephengold:
> It is not fair to keep this hanging for so long

@tonihele We have a new release manager in training. Please be patient.

scenemax3d:
@pspeed42 @stephengold @tonihele 
reading the comments I saw concerns about the structure of the API (Paul's comments). Are we OK with the API then?
If yes, I will add it to v3.7.0
10X :)

jme3-core/src/main/java/com/jme3/system/DisplayInfo.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package com.jme3.system;
+
+/**
+ * This class holds information about the display that was returned by glfwGetMonitors() calls in
+ * the context class
+ *
+ * @author Kevin Bales
+ */
+public class DisplayInfo {
```
capdevon, line 34: is it correct that the variables of the `DisplayInfo` class all have the public access modifier? I would like to know what the core members think about this?

Scrappers-glitch: It really comes down to the style of code, and this style is almost the same as the C-struct-style. While, jMonkeyEngine and Java style doesn't use public fields as outside accessors, I recommend to stick to the style too. I am really now bothered more with thread-safety.

capdevon: I would suggest privatizing the variables and adding getter/setter methods. Don't use 'record' otherwise you break the engine's compatibility with java 8 and all applications currently out there.

yaRnMcDonuts:
> @pspeed42 @stephengold @tonihele
> reading the comments I saw concerns about the structure of the API (Paul's comments). Are we OK with the API then?
> If yes, I will add it to v3.7.0
> 10X :)

It looks like this was not quite ready in time for 3.7, but I'd like to try to get this into the next 3.8 release, if possible. 

It appears that the last 2 things holding up this PR were :

1. A final approval on the API from @pspeed42 , since he had some API changes requested from @bob0bob. I haven't looked too deeply into this code yet, but from what I can see, the requested changes were made, so if no other issues are brought up, then it sounds like this PR should be ready.

2. Changing some public variables to be private/protected as suggested in the recent review, and creating getters/setters if needed. This is less important, so in the even that the author of this PR does not return to make these changes in time for the next release, then that is something that I could easily address in another PR following this one. Or the author can turn on "allow edits from maintainers" in which case I (or anyone else who wants to) can make those final changes prior to passing the PR.

pspeed42:
I haven't looked.  It's probably fine.  I used to see github stuff more 'in my face' but now I have to drill in and look around for it... so don't count on me for properly following up here these days.

yaRnMcDonuts:
Unless anyone else contests, I will merge this PR in the next day or two.

I will also resolve the existing conflict (it appears to just be duplicate imports, an easy fix)
And I will open a new issue/PR to address @capdevon 's review and to implement the code he graciously supplied as a solution.

Issue #2032, opened 2023-06-05
DefineList needs refactored to support >64 defines


yaRnMcDonuts:
I've opened this issue following up the discussion from this thread:

https://hub.jmonkeyengine.org/t/optimize-terrain-shaders-for-tri-planar-mode-per-layer-remove-fog-defines-to-make-room/46816/9

The DefineList class currently sets an arbitrary limit of 64 defines per technique.

As suggested by @pspeed42 the best way to fix this is to refactor the DefineList class is to allow the Array to expand rather than pre-allocating it with a limited size of 64. And he also mentioned some other things that could be cleaned up or done better in that class, so we can continue discussing the specifics of what needs refactored here. 

I'm not a core dev nor an official contributor so I admittedly do not feel comfortable making an attempt at refactoring this area of the engine without first having more guidance or a clear outline to follow ahead of time. So if anyone wants to give me some guidance and specifications then I can work on refactoring DefineList, or if anyone more experienced wants to steal the thunder and submit your own PR for this issue then you are more than welcome to do so.

pspeed42:
I make take a stab at starting some unit tests this week if I get a minute.

stephengold:
Resolved in "master" branch at 58365c4b8a3929954718ac3e292ca77675dd5540 by PR #2116.

PR #2033, opened 2023-06-05
Increase MAX_DEFINES in DefineList

yaRnMcDonuts:
Partial fix for this issue:
https://github.com/jMonkeyEngine/jmonkeyengine/issues/2032

### DIFF
--- a/jme3-core/src/test/java/com/jme3/renderer/OpaqueComparatorTest.java
+++ b/jme3-core/src/test/java/com/jme3/renderer/OpaqueComparatorTest.java
@@ -1,0 +1,338 @@
+/*
+ * Copyright (c) 2009-2021 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.jme3.renderer;
+
+import com.jme3.asset.AssetManager;
+import com.jme3.material.Material;
+import com.jme3.material.TechniqueDef;
+import com.jme3.math.ColorRGBA;
+import com.jme3.renderer.queue.GeometryList;
+import com.jme3.renderer.queue.OpaqueComparator;
+import com.jme3.scene.Geometry;
+import com.jme3.scene.Mesh;
+import com.jme3.scene.shape.Box;
+import com.jme3.system.TestUtil;
+import com.jme3.texture.Image;
+import com.jme3.texture.Image.Format;
+import com.jme3.texture.Texture;
+import com.jme3.texture.Texture2D;
+import com.jme3.texture.image.ColorSpace;
+import com.jme3.util.BufferUtils;
+import java.nio.ByteBuffer;
+import java.util.HashSet;
+import java.util.Set;
+import org.junit.Before;
+import org.junit.Test;
+
+public class OpaqueComparatorTest {
+    
+    private final Mesh mesh = new Box(1,1,1);
+    final private Camera cam = new Camera(1, 1);
+    private RenderManager renderManager;
+    private AssetManager assetManager;
+    final private OpaqueComparator comparator = new OpaqueComparator();
+    
+    @Before
+    public void setUp() {
+        assetManager = TestUtil.createAssetManager();
+        renderManager = TestUtil.createRenderManager();
+        comparator.setCamera(cam);
+    }
+    
+    /**
+     * Given a correctly sorted list of materials, check if the 
+     * opaque comparator can sort a reversed list of them.
+     * 
+     * Each material will be cloned so that none of them will be equal to each other
+     * in reference, forcing the comparator to compare the material sort ID.
+     * 
+     * E.g. for a list of materials A, B, C, the following list will be generated:
+     * <pre>C, B, A, C, B, A, C, B, A</pre>, it should result in
+     * <pre>A, A, A, B, B, B, C, C, C</pre>.
+     * 
+     * @param materials The pre-sorted list of materials to check sorting for.
+     */
+    private void testSort(Material ... materials) {
+        GeometryList gl = new GeometryList(comparator);
+        for (int g = 0; g < 5; g++) {
+            for (int i = materials.length - 1; i >= 0; i--) {
+                Geometry geom = new Geometry("geom", mesh);
+                Material clonedMaterial = materials[i].clone();
+                
+                if (materials[i].getActiveTechnique() != null) {
+                    String techniqueName = materials[i].getActiveTechnique().getDef().getName();
+                    clonedMaterial.selectTechnique(techniqueName, renderManager);
+                } else {
+                    clonedMaterial.selectTechnique(TechniqueDef.DEFAULT_TECHNIQUE_NAME, renderManager);
+                }
+                
+                geom.setMaterial(clonedMaterial);
+                gl.add(geom);
+            }
+        }
+        gl.sort();
+        
+        for (int i = 0; i < gl.size(); i++) {
+            Material mat = gl.get(i).getMaterial();
+            String sortId = Integer.toHexString(mat.getSortId()).toUpperCase();
+            System.out.print(sortId + "\t");
+            System.out.println(mat);
+        }
+        
+        Set<String> alreadySeen = new HashSet<>();
+        Material current = null;
+        for (int i = 0; i < gl.size(); i++) {
+            Material mat = gl.get(i).getMaterial();
+            if (current == null) {
+                current = mat;
+            } else if (!current.getName().equals(mat.getName())) {
+                assert !alreadySeen.contains(mat.getName());
+                alreadySeen.add(current.getName());
+                current = mat;
+            }
+        }
+        
+        for (int i = 0; i < materials.length; i++) {
+            for (int g = 0; g < 5; g++) {
+                int index = i * 5 + g;
+                Material mat1 = gl.get(index).getMaterial();
+                Material mat2 = materials[i];
+                assert mat1.getName().equals(mat2.getName()) : 
+                       mat1.getName() + " != " + mat2.getName() + " for index " + index;
+            }
+        }
+    }
+    
+    @Test
+    public void testSortByMaterialDef() {
+        Material lightingMat = new Material(assetManager, "Common/MatDefs/Light/Lighting.j3md");
+        Material particleMat = new Material(assetManager, "Common/MatDefs/Misc/Particle.j3md");
+        Material unshadedMat = new Material(assetManager, "Common/MatDefs/Misc/Unshaded.j3md");
+        Material skyMat      = new Material(assetManager, "Common/MatDefs/Misc/Sky.j3md");
+        
+        lightingMat.setName("MatLight");
+        particleMat.setName("MatParticle");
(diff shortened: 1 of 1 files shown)


stephengold:
Based on my reading of the codebase, `isSet()` seems essential to the functionality of `DefineList`. Ryan, would you please amend this PR so that `isSet()` works for more than 64 defines?

pspeed42:
It's using the 64 bits to specify which defined ID is set.  bit 0 set then define ID 0 is set.  bit 1 is set then define ID 1 is set.

Java provides a BitSet class that can be any size.  I'm not sure why that wasn't used originally.  There is a 50/50 chance that the answers is "Author didn't think of it."

yaRnMcDonuts:
I made an attempt to update DefineList to use BitSet instead now, let me know if anything needs fixed.

The BitSet appeared much easier to use than bitshifting multiple longs would've been, but let me know if I did anything wrong with it. I was also unsure about 3 things in particular:

- The hashCode method previously used the isSet long. So for now I just made it isSet.hashCode() so it returns the hashCode of the new BitSet. 
- The equals method uses isSet, but the code still works now that isSet is a BitSet instead of long, so I left it alone. 
- I made MaxDefines 128 instead of 64. Should I raise this more or should we look into ways to let this be changed in an engine setting?

pspeed42:
Just answering the last question: I would think once we use BitSet that maybe we no longer need a "max defines" concept at all. I wonder if that is used for anything useful elsewhere.

yaRnMcDonuts:
I did a search of "MAX_DEFINES" in the jmonkey repo on github, and it appears it's used in two places in the TchniqueDef class, but only to throw an error to prevent going over the previous limit due to using a long:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/02a006d04e660df889141ac0329bd8ed3c5f5008/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java#L425

So it looks like we should be able to remove these checks and get rid of MAX_DEFINES entirely now.

yaRnMcDonuts:
I tried updating the hashCode() method since it looked like some of the failed checks were related to hashCode and equals, and the equals methods looked correct from as far as I can tell. But the checks still failed it looks like. I have very little experience with the tests/checks that get done on github since this is the first and only project I've ever created PRs for, so any guidance or advice is greatly appreciated.

stephengold:
I suspect that the `OpaqueComparatorTest` tests are failing because the `DefineList.equals()` method is now incorrect. When the `isSet` field was a primitive type, it was fine to compare fields using `if (isSet == o.isSet) {`, but now that the field is an object, we should of course use something like `if (isSet.equals(o.isSet) {` (assuming it's never `null`).

stephengold:
Also the copy constructor `private DefineList(DefineList original)` needs some rework, to avoid multiple `DefineList` instances sharing the same `BitSet` instance.

yaRnMcDonuts:
I updated the equals method, and it looks like a null check for isSet shouldn't be necessary since isSet is prviate and final, but I could go back and add a null check if you think that is best. I also considered doing an instanceof check to make sure the object paramater in the equals method is indeed a DefineList, but I wasn't sure if that is necessary to add or not since this is mostly an internally used class and has never encountered an error from that before.

And I also changed the copy constructor so that it clones the BitSet now. Is this enough or should I instead create a new BitSet and loop through to set the same values as the one thats being copied from?

And of course let me know if there's anything else you see that could be done better.

pspeed42:
I don't know if we want to fix it as part of this since equals() was already broken  before... but the idiom should be:
(pseudo code):
```
if( this == obj ) {
    return true;
}
if( obj ==  null || obj.getClass() != getClass() ) {
    return false;
}
DefineList other = (DefineList)obj;
...do the comparison stuff... usually using Object.equals() for the non-primitive fields
if( !Object.equals(isSet, other.isSet) ) {
    return false;
}
return true;
```

yaRnMcDonuts:
I updated the equals() method as you suggested, let me know if it looks like I understood your suggestion correctly or if I messed up or should change anything else.

I also am still fairly confused about a few things regarding DefineList and this PR in genera:

First I am curious what is the reason that we are overriding the equals and hashCode method here, and what is the reason for the unit tests that are looking for a specific hash code value? Is this just to make things more optimized, and if so what is the reason that it matters in this class but not in others? Is DefineList just a very low level class that benefits from these minor optimizations? Or is the code overriding the equals() all overkill and not really necessary for things to work? (sorry if this is a naïve question, but most of my experience as a coder has been very high level stuff where minor optimizations like this are negligible, so I guess that's why I managed to get so far without learning much about all of this...)

As I mentioned earlier, overriding equals() is something I have frequently seen in other Java apps and have googled why/when to do it, but it's not something I have done in my app and it doesn't appear as though I've encountered any errors as a result (possibly a huge blunder on my end?) and anytime I google when/why to override the equals method to try to understand better, I feel like the examples I've found are very poor and overly basic, and I've never had that "eureka" moment where I realized I oughta go back into my own app and override a bunch of equals() method to fix everything... and this worries me since I still seem to feel like there's no reason to do so in my own app and have no apparent problems related to this, but I have come to realize that it is very common in other java apps and that worries me and makes me cast

Any insight on this stuff is greatly appreciated.

stephengold:
It's not an optimization; it's a correctness issue.

A good rule of thumb is that if you want to use a class in a hashing context (such as in a `HashSet` or as the key in a `HashMap`) then you need to override `Object.equals()` and `Object.hashCode()`. Otherwise you risk creating 2 distinct objects that have all fields the same, yet `equals()` returns false and they have different hash codes. Does that help?

In the case of `DefineList`, it's used here: https://github.com/jMonkeyEngine/jmonkeyengine/blob/4e6a7f14086e8a0f83689d65e88eb39cc6f9f5b7/jme3-core/src/main/java/com/jme3/material/TechniqueDef.java#L147

I haven't looked at the failing test(s), but I did take a quick glance at the revised `equals()` and `hashCode()` methods, and they both seem excessively complex. Since `BitSet` provides both `equals()` and `hashCode()`, you should rely on those methods. There's no need for `equals()` or `hashCode()` to examine at the individual bits.

yaRnMcDonuts:
Yes that helps, thanks. I had always understood how equals() checks value and differs from == but somehow I never realized that you have to override that method and write the code correctly for a class to actually work that way... I thought that was just the default way it worked for all objects, but I think I see how I was mistaken now..

> Otherwise you risk creating 2 distinct objects that have all fields the same, yet `equals()` returns false
I think I managed to go so long without realizing this because all of the classes I've made for game objects (like npcs or items) luckily never have a situation where 2 copies of something get created with identical values. So I just never had a problem and could get by not overriding it without any issues. But I see how not overriding equals() would break things for something like Vector3f where having multiple copies with identical values is very common.

Definitely a slightly embarrassing thing for me to have have overlooked for so long. I'm primarily self taught aside from very minor schooling (my university didn't have a Software Engineering program, the best option was Computer Information Systems which taught me very little about the software side of computers unfortunately, and JME is the only real coding experience I've had since graduating a few years ago). So I frequently find that there are alot of things like this this that I never learned properly since I am limited to coding and learning by trial/error alone in my own bubble. But that's the great things about trying to make a PR to the engine, I usually learn things that I otherwise wouldn't on my own. So I appreciate all the help and patience.

yaRnMcDonuts:
I reverted the hashCode() method to how I originally did it using isSet.hashCode(), I had only tried making it more complex because I thought one of the failing tests was related to an insufficient hashCod() method. So I guess I will still need insight on how to get some of those checks to pass.

I also took a look at the equals() method as of my most recent update based on pspeed's suggestion, however I don't quite see how I could simplify it further. Maybe I am still misunderstanding something then 😕

pspeed42:
Your equals() looks fine to me.  I think maybe in all the talk of bitsets folks forget we are actually keeping the values as well.

I don't know why someone wrote unit tests for the hashCode() function.  You might search around the code to see if anything is using this raw hash code for anything.  However, it may also be someone was trying to increase test coverage and did it poorly.  By pure unit test standards, the only way to unit test a hashCode() method is to check for exact values (presuming a predictable algorithm) but in almost no cases is this actually a useful test. (and I'd argue that in the cases that it IS a useful test then hashCode() is being abused somewhere.)

This particular hashCode() is stunted because it's calculating a hash from only the bitset and not the values  (rather than just using Objects.hash(isSet, values))  The only way to reasonable unit test it would be to create two different DefineLists and compare their hash codes.  In this case, if the "only using bitset for the hash" is desirable then we could construct the DefineList with two different sets of values (but the same positions) and see if their hash matches.

...but I'm also inclined not to require that behavior because I think it was an overzealous optimization.  To know for sure would require some detailed searching of the engine to see how exactly DefineList is used everywhere.

...and perhaps DefineList.equals()/hashCode() are not used directly anywhere but are used by Material.equals()/hashCode() and maybe _those_ are called often.  Some profiling might show how often these methods are called and in what context.

AND: if it turns out that these methods are never called or rarely called... then this is a good lesson on how much time can be wasted when things are over-engineered without comments explaining why.

yaRnMcDonuts:
> ...and perhaps DefineList.equals()/hashCode() are not used directly anywhere but are used by Material.equals()/hashCode() and maybe _those_ are called often. Some profiling might show how often these methods are called and in what context.

I did a search of every occurrence of "DefineList" on the github repo and looked through all potential usages (DefineList had multiple occurrences across 12 files according to the github search), and it appears that the hashCode() and equals() method are never directly called anywhere in jme's codebase. 

I was about to go ahead and remove these unit tests, however I am wondering what is the best way to do that.

I see that all but 1 failing test are in a file titled DefineListTest. Can I just remove these tests individually from that file? Or is there anywhere else I need to remove the calls to these methods as well?

And I see one of the failing tests is in a separate test file here : 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/177f7d34491a933a2c5213d5be6cc2e8d6247023/jme3-core/src/test/java/com/jme3/renderer/OpaqueComparatorTest.java#L227

Is that related to something I am doing wrong, or should this also be removed? It appears to be related to alot more than just the DefineList class (unlike the other failing tests) so I wonder if there is still something I need to fix in DefineList?

yaRnMcDonuts:
> I'm ok with removing those DefineList tests as long as we are sure there isn't any funny code in Material.equals() or something that is depending on some specific behavior (unlikely).

I took a look at the Material class (forgot to mention it in my previous reply) and DefineList isn't used anywhere there or in its equals() method. 

It looks like the Material class also doesn't override a standard equals() method:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/material/Material.java#L227

and instead that class uses a contentEquals() with a comment saying "    Since materials are mutable objects, implementing equals() properly is not possible, hence the name contentEquals()." Not sure if that's relevant to this discussion though.

But regardless, I removed the 5 failing tests, and left the other 1 alone for now. 
    

> As to the other, I'd need to know more about why it's failing to comment. Without diving in too deeply, it seems like some code somewhere is using the hash code is a sort discriminator. This is ok but it does mean that assumptions made in this particular test may be incorrect and the materials passed to the testSort() method need to be rearranged.

I don't have any clue how to start going about fixing that test since I don't quite know what its doing 😆 so I may need to wait until you or someone else has time to dive in deeper to help figure out how to fix the last remaining test.

pspeed42:
I believe the test is making sure that sorting is stable (because sometimes that's important).

As such, it's probably easy enough to just see what the order difference is and change the ordering of arguments passed to testSort() so that it matches what's expected.  But I'm making assumptions because I don't even know why the test is failing exactly.

yaRnMcDonuts:
I was sick for the past month so I procrastinated finishing this PR, but I finally got around to it, and the test is finally passing when I run the build locally, but it fails on github after ive uploaded it here

> As such, it's probably easy enough to just see what the order difference is and change the ordering of arguments passed to testSort() so that it matches what's expected

Yes I was able to see the exception when I built the engine locally, and then I just kept swapping the order for each param and building until it stopped failing which (I thought) meant I had finally found the correct sort order.

But it certainly was a struggle figuring out this sort order 1 param at a time with repetitive abbreviated named DefineLists params such as "DefVL, DefVC, DefTC, DefVCVL, DefVCVLTC" that didn't match their unabbreviated variable name either, so that made this take 10x longer than it probably should have. Definitely not the best thing for me to have tried working on while I'm just getting over a bad head cold lol...

I uploaded DefineList.java from my local branch that passed again just in case, and. I ensured the local branch I downloaded is the branch I am using for this PR titled yaRnMcDonuts:patch-5 which I downloaded after my last change, so it should be identical to this branch on github. But still the most recent build fails on github.

Any ideas what could be wrong still? I could still be making a mistake but right now it looks like the failing branch I've uploaded is identical to the one that passes locally, so I don't know what else could be the issue.

yaRnMcDonuts:
Tried running checks again on here, but the test still fails.

So I re-downloading this exact github branch and built with gradle in windows command prompt and the test passes and the build is successful 😕

jme3-core/src/main/java/com/jme3/shader/DefineList.java
```diff
@@ -142,21 +143,27 @@ public int getInt(int id) {
...
         }
-        return false;
+        if(object ==  null || object.getClass() != getClass()) {
+            return false;
+        }
+        DefineList otherDefineList = (DefineList) object;
+        if (!isSet.equals(otherDefineList.isSet)) {
+            return false;
+        }            
+        for (int i = 0; i < values.length; i++) {
```
stephengold, line 161: What if the `values` arrays are of different lengths? In that case, I believe some elements won't be compared!

yaRnMcDonuts: that is a good point, i think that would also make it possible to throw an array out of bounds exception in the case that they are different sizes.

Although I also think the code right above the values check `if (!isSet.equals(otherDefineList.isSet)`) { that is checking for equal bitSets should already cause the equals() method to return false if the values array have different sizes (since values and isSet both appear to get initialized with the same size in the class' constructor)

But it could still be a good idea to add a length check for the size of the values array and return false if they are not equal lengths prior to looping through the arrays, just to be on the safe side.

yaRnMcDonuts:
I've just noticed that the failure report on github is showing a new error in the sortAll test (whereas previously the only error was thrown in the test for sortDefineLists).

So by changing the sort order of params to be correct in the one test, I broke the order in another test. And i suspect if I fix the order for that test, it will break the order for the test I just fixed.... Which further confirms my suspicion that it is not possible to pass this test in any way now that I changed hashCode() and scrambled the exact values expected from these sort tests.

jme3-core/src/main/java/com/jme3/shader/DefineList.java
yaRnMcDonuts: i added the changes, although unfortunately that did not have any effect on the failing tests.

pspeed42:
I think we are an impasse until someone with time actually learns what these unit tests were testing and fixes them to work.  I think these tests are testing something real and something important... but I think they are doing it very BADLY.

stephengold:
After studying `DefineListTest` more closely, I believe some of its uses of `hashCode()` are valid. In other places, however, the test author relied on an undocumented property of `DefineList.hashCode()` when they should've used `isSet()` instead.

`OpaqueComparatorTest` is poorly designed and needs a rewrite, but that's way outside the scope of this PR.

PR #2038, opened 2023-06-21
Application#stop(boolean): fix javadoc args issue

Scrappers-glitch:
This PR fixes the javadoc of the `Application#stop(boolean)` argument.


stephengold:
Good catch, Pavl! Looks good to me.

 Unless there's further discussion, I will integrate this in 24 hours.

PR #2040, opened 2023-07-04
remove runtime dependency to lwjgl-platform:2.9.5 so mvn package does not fail

neuweiler:
As mentioned [here](https://hub.jmonkeyengine.org/t/solved-failure-to-find-org-jmonkeyenginejar-2-9-5-after-upgrading-to-3-6-0/46838), the mvn package failed with the runtime dependency.


stephengold:
Thank you for providing the fix.

Ali-RS:
Unless there's something to discuss, I plan to integrate this PR in about 24 hours.

Issue #1011, opened 2019-01-26
inconsistent implementations of JmeExporter.save(Savable, File)


stephengold:
In `com.jme3.export.binary.BinaryExporter`, the method for saving to a `File` creates the parent directory if it doesn't already exist. However, in `com.jme3.export.xml.XMLExporter`, the corresponding method throws a `java.io.FileNotFoundException` if the parent directory is missing.

Since both classes implement the same interface, it would be convenient if their behaviors were more similar.

stephengold:
Changing the semantics of `BinaryExporter` is a lot more likely to break existing code.

pspeed42:
Yes, sometimes we are slaves to the bad decisions of the past.

pspeed42:
A compromise might be to modify the interface to also have a version that has a boolean createDirectories argument... then the old version can call the new version with true... but folks can do it the safe way if they want to.

PR #2041, opened 2023-07-23
Issue 1011: Make JmeExport save consistent in its path handling

andygibson:
Add a new save method that takes a flag to indicate if the file parent directories should be created.

Make the current save() method default in the interface and use the new method with createDirectories true.

Added tests for JmeExport using all current implementations of it.


stephengold:
Thank you for working on issue #1011. I'm busy at present and unsure when I'll get around to reviewing this.

stephengold:
Thank you for your contribution to JME, @andygibson !

PR #2043, opened 2023-07-24
Adds SingleLayerInfluenceMask

codex128:
This is my first pull request ever, so take it easy on me :)

The main element of this class (~~AlertArmatureMask~~ SingleLayerInfluenceMask) is the ability to check if a joint is currently being used by a higher layer before approving it for personal use via `contains(Object target)`. This can help smooth some interpolation issues that may arise between layers with looped/sequenced animations (see [here](https://hub.jmonkeyengine.org/t/solved-animation-transition-not-seamless-as-expected/46941/8) for more details on the problem).

This class also provides several helper methods to make mask creation easier.


pspeed42:
So is this basically a filter?  Or is it like firing off "Hey, look at me! look at me! look at me! look at me!" events?

codex128:
Yes, I guess you could call it a filter.  It's pretty much the same as ArmatureMask except it has an extra criteria for masking joints.

PR #2044, opened 2023-07-24
1977: Fix AppSettings samples javadoc

andygibson:
* Indicate the default is 0
* Ammended to explain 0 or 1 is single sampled
* Might seem odd without explaining it.


stephengold:

PR #2046, opened 2023-07-25
Removed redundant joint looping in ArmatureMask

codex128:
Removed a redundant loop in ArmatureMask

https://github.com/jMonkeyEngine/jmonkeyengine/blob/5155f56616497a4087ffc43af53f968c565dd065/jme3-core/src/main/java/com/jme3/anim/ArmatureMask.java#L89-L96

Now as:
```
public static ArmatureMask createMask(Armature armature, String... joints) {
    ArmatureMask mask = new ArmatureMask();
    mask.addBones(armature, joints);
    return mask;
}
```
Edit: I made a mistake on my git handling, so there are more commits displayed than necessary. The only commit that matters is the last one, which negates the effects of all the others and adds this fix. :P


codex128:
I've had to close this because of a git mistake, sorry!

PR #2051, opened 2023-07-27
1982: Negative Native Object Id Issue

andygibson:
Prototype that seems to solve the problem with negative numbers being a valid object id and an indicator of whether the id/object is valid or not. If interested, I can polish it up a bit. 

The changes start in NativeObject, and most of the other changes are in response to those changes.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/util/NativeObject.java
+++ b/jme3-core/src/main/java/com/jme3/util/NativeObject.java
@@ -44,8 +44,9 @@ import java.nio.Buffer;
  */
 public abstract class NativeObject implements Cloneable {
 
-    public static final int INVALID_ID = -1;
-    
+    //can be anything since its not used for validity
+    private static final int INVALID_ID = 0;
+
     protected static final int OBJTYPE_VERTEXBUFFER = 1,
                                OBJTYPE_TEXTURE      = 2,
                                OBJTYPE_FRAMEBUFFER  = 3,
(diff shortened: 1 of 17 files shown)


riccardobl:
This is probably the right way to resolve this issue, but it can be a breaking change for third party libraries that relies on getId() to be == INVALID_ID

jme3-core/src/main/java/com/jme3/util/NativeObject.java
```diff
@@ -249,4 +259,17 @@ public <T> WeakReference<T> getWeakRef() {
         }
         return (WeakReference<T>) weakRef;
     }
+
+    public boolean isValid() {
+        return valid;
+    }
+
+    public boolean isInvalid() {
```
scenemax3d, line 267: I propose removing this method and using !isValid() whenever needed. 
I think it's generally good practice to avoid "negative" method names.
WDYT?

Scrappers-glitch: Yes, you are right, and it's redundant anyway.

neuweiler:
If this fix resolves this problem: 
com.jme3.renderer.RendererException: Invalid ID (0) received when trying to create shader program. at com.jme3.renderer.opengl.GLRenderer.updateShaderData(GLRenderer.java:1595)

Is there any chance, this PR makes it into 3.7 ? Because 3.6.1 is currently not usable for my on my Mac 2015.

stephengold:
On behalf of the JMonkeyEngine project, I apologize for the delay and confusion.
@scenemax3d is release manager for v3.7.0; the decision to include a PR or not is up to him.
You could improve the odds by responding to his review (above) and resolving the merge conflict in BufferObject.java .

Regarding issue #1982, I use v3.6.1 on a Mac Mini M1 running Ventura 13.2.1 and have yet to encounter it.
I'm wondering if it is specific to a particular device driver.
Do you have a test that can reliably reproduce the issue on your system?

Issue #2039, opened 2023-06-27
DOMInputCapsule.readIntSavableMap does not check exceptions


tohidemyname:
The code is as follows:
`public IntMap<? extends Savable> readIntSavableMap(String name, IntMap<? extends Savable> defVal) throws IOException {
...

                                if (n instanceof Element && n.getNodeName().equals("MapEntry")) {
                                        Element elem = (Element) n;
                                        currentElem = elem;
                                        int key = Integer.parseInt(currentElem.getAttribute("key"));
                                        Savable val = readSavable("Savable", null);
                                        ret.put(key, val);
                                }                  
         ...
    }`
Here Integer.parseInt(currentElem.getAttribute("key")) can throw NumberFormatException. The other methods of this class catch and rethrow IOException. An example is as follows:

`public byte[] readByteArray(String name, byte[] defVal) throws IOException {
  try {...
      for (int i = 0; i < strings.length; i++) {
              tmp[i] = Byte.parseByte(strings[i]);
       }...
  }catch (IOException ioe) {
            throw ioe;
        } catch (NumberFormatException nfe) {
            IOException io = new IOException(nfe.toString());
            io.initCause(nfe);
            throw io;
        } catch (DOMException de) {
            IOException io = new IOException(de.toString());
            io.initCause(de);
            throw io;
        }
    }
 `

stephengold:
Thanks for bringing this inconsistency to our attention. Would you be interested in authoring a pull request to solve this issue?

tohidemyname:
Here is the pull request:https://github.com/jMonkeyEngine/jmonkeyengine/pull/2052
Thx a lot.

PR #2052, opened 2023-07-28
solve issue #2039 (DOM readIntSavableMap() doesn't check exceptions)

tohidemyname:
Integer.parseInt(currentElem.getAttribute("key")) can throw NumberFormatException, which is inconsistent with other parsing methods.

### DIFF
--- a/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMInputCapsule.java
+++ b/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMInputCapsule.java
@@ -1245,9 +1245,16 @@ public class DOMInputCapsule implements InputCapsule {
                                 if (n instanceof Element && n.getNodeName().equals("MapEntry")) {
                                         Element elem = (Element) n;
                                         currentElem = elem;
-                                        int key = Integer.parseInt(currentElem.getAttribute("key"));
-                                        Savable val = readSavable("Savable", null);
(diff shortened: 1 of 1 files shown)


jme3-plugins/src/xml/java/com/jme3/export/xml/DOMInputCapsule.java
```diff
@@ -1245,9 +1245,16 @@ public IntMap<? extends Savable> readIntSavableMap(String name, IntMap<? extends
...
-                                        int key = Integer.parseInt(currentElem.getAttribute("key"));
-                                        Savable val = readSavable("Savable", null);
-                                        ret.put(key, val);
+                                        try {
+	                                        int key = Integer.parseInt(currentElem.getAttribute("key"));
+	                                        Savable val = readSavable("Savable", null);
+	                                        ret.put(key, val);
+                                        }catch (NumberFormatException nfe) {
+                                            IOException io = new IOException(nfe.toString());
+                                            io.initCause(nfe);
```
Scrappers-glitch, line 1254: I don't think it's a good idea to throw a checked exception from an unchecked exception, if you want to propagate an unchecked runtime exception to the user application use a `throws NumberFormatException` to give a clue to catch the exception or else use a strategy pattern with a compatible JDK interface to handle this exception from the user application, but this shouldn't throw a business checked exception.

PR #2053, opened 2023-07-28
solve issue #2045 (ArmatureMask.createMask() duplicate code)

### LINKED ISSUE #2045: ArmatureMask.createMask() duplicate code
From a quick glance, the for-loop in the below method looks redundant to me and should be removed because `affectedJoints` is already set by the `mask.addBones(armature, joints);`

https://github.com/jMonkeyEngine/jmonkeyengine/blob/5155f56616497a4087ffc43af53f968c565dd065/jme3-core/src/main/java/com/jme3/anim/ArmatureMask.java#L89-L96

refrence for `addBones()` method:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/5155f56616497a4087ffc43af53f968c565dd065/jme3-core/src/main/java/com/jme3/anim/ArmatureMask.java#L104-L109

Ali-RS:
Fix #2045

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/ArmatureMask.java
+++ b/jme3-core/src/main/java/com/jme3/anim/ArmatureMask.java
@@ -1,3 +1,34 @@
+/*
+ * Copyright (c) 2009-2023 jMonkeyEngine
+ * All rights reserved.
+ *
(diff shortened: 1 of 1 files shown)


Ali-RS:

PR #2054, opened 2023-07-28
1877: Fix style issues on jme3-core/com.jme3.app package

andygibson:
One remaining issue regarding local var distance from declaration to use. 
Partial Implementation of #1877

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/AppTask.java
+++ b/jme3-core/src/main/java/com/jme3/app/AppTask.java
@@ -31,7 +31,12 @@
  */
 package com.jme3.app;
 
-import java.util.concurrent.*;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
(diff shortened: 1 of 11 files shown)


Scrappers-glitch (approved):
Looks fine, thanks.

jme3-core/src/main/java/com/jme3/app/AppTask.java
```diff
@@ -100,7 +106,8 @@ public V get() throws InterruptedException, ExecutionException {
     }
 
     @Override
-    public V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+    public V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException,
```
stephengold, line 109: "The prime directive of line-wrapping is: prefer to break at a higher syntactic level."  (Google Java Style Guide 4.5.1)

Here I recommend breaking the line before the `throws` keyword, so that the exceptions all appear on a single line.

jme3-core/src/main/java/com/jme3/app/DetailedProfilerState.java
```diff
@@ -413,7 +424,8 @@ public void layout(int indent) {
             int y = -(height * LINE_HEIGHT + HEADER_HEIGHT);
 
             label.setLocalTranslation(PADDING + indent * PADDING, y, 0);
-            float gpuPos = PANEL_WIDTH - font.getLineWidth(gpuText.getText()) - PADDING * (maxLevel - indent + 1);
+            float gpuPos = PANEL_WIDTH - font.getLineWidth(gpuText.getText()) - PADDING
```
stephengold, line 427: I suggest breaking the line before `- PADDING` so that the final term is not split across 2 lines.

stephengold (commented):
The readability of the code is improved by the PR. Thank you for your contributions!

I've quibbled with a couple line-break choices.  Let us know whether you're willing to rework those.

stephengold:
Thanks again for your contributions, @andygibson!

Issue #2056, opened 2023-08-07
PointLight.setRadius(POSITIVE_INFINITY) silently breaks the engine


riccardobl:
Setting PointLight radius to POSITIVE_INFINITY as done in the gltf loader

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java#L130

causes unexpected behavior across the engine by causing some values to become NaN.

PR #2058, opened 2023-08-08
Fix issue #2056 . Check for NaN uniforms

riccardobl:


pspeed42:
Note regarding the protected void isValidNumber(float v) methods.

...because they are protected it is harder for Java to inline them but even if asserts are turned off they will be called.

Either way, I wonder if it would be better if isValidNumber() simply returned true/false and then the calls are replaced with asserts... then when asserts are turned off no extra calls are made.

Else, isValidNumber() should be renamed because "is" implies a boolean return.

pspeed42:
My understanding is that if assertions are disabled then the entire part after "assert" is removed... including the string concatenation.

pspeed42:
...and by the way, I think this kind of instrumenting is really great for those "special times" when it's necessary.

riccardobl:
I added the values to the assert messages

riccardobl:
If there aren't objections i will merge this

capdevon:
Hi Riccardo, please take a look at my revision before merging the code, there are formatting errors in the code and double semicolons at the end of the instructions.

riccardobl:
> Hi @riccardobl, please take a look at my revision before merging the code, there are code formatting errors, extra lines and double semicolons at the end of the instructions. Could you please correct them?

Mhh i don't see your revision anywhere, but i think i've fixed the issues you pointed out, see the last commit

Ali-RS:
Looks like he forgot to "submit" the reviews, so they won't be displayed to others. 
This also happened to me before!

capdevon:
> Looks like he forgot to "submit" the reviews, so they won't be displayed to the public. This also happened to me before!

@Ali-RS Correct, I have submitted the revision which should now be visible. 
Thanks for your patience @riccardobl . I think there is still a little something to correct ;)

riccardobl:
Done

PR #2059, opened 2023-08-09
Replace direct usage of ClassLoader with ResourcesLoader

riccardobl:
To enhance compatibility with JVMs that deviate from standard norms and may lack complete or partial support for classloaders (such as teavm), this pull request introduces a new encapsulating class called ResourcesLoader. Functioning analogously to BufferAllocators, this addition empowers developers to specify a personalized implementation. The default implementation, named ResourcesLoaderJImpl, operates by interfacing with the classloader.

Throughout the core, all instances where classloaders were used have been substituted with the utilization of ResourcesLoader, excluding specific parts in GL debug classes and jme3-networking which involve proxies and might need more careful consideration.

The methods within AssetManager that depend on ClassLoader usage have been deprecated and could potentially be marked for removal, as they are not utilized by the engine itself.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/util/res/DefaultResourceLoader.java
+++ b/jme3-core/src/main/java/com/jme3/util/res/DefaultResourceLoader.java
@@ -36,8 +36,48 @@ import java.io.InputStream;
 import java.net.URL;
 import java.util.Enumeration;
 
-public interface ResourcesLoaderImpl {
-    public URL getResource(String path, Class<?> clazz);
-    public InputStream getResourceAsStream(String path, Class<?> clazz);
-    public Enumeration<URL> getResources(String path, Class<?> clazz) throws IOException;
+/**
+ * Default java-like implementation of ResourceLoaderImpl. Loads from classpath.
+ */
+public class DefaultResourceLoader implements ResourceLoader {
+
+    public InputStream getResourceAsStream(String path, Class<?> parent) {
+        if (parent == null) {
+            return Thread.currentThread().getContextClassLoader().getResourceAsStream(path);
+        } else {
+            return parent.getResourceAsStream(path);
+        }
+    }
+
+    @Override
+    public URL getResource(String path, Class<?> parent) {
+        if (parent == null) {
+            return Thread.currentThread().getContextClassLoader().getResource(path);
+        } else {
+            return parent.getResource(path);
+        }
+    }
+
+    @Override
+    public Enumeration<URL> getResources(String path, Class<?> parent) throws IOException {
+        if (parent == null) {
+            return Thread.currentThread().getContextClassLoader().getResources(path);
+        } else {
+            if (!path.startsWith("/")) {
+                Class<?> c = parent;
+                while (c.isArray()) {
+                    c = c.getComponentType();
+                }
+                String baseName = c.getName();
+                int index = baseName.lastIndexOf('.');
+                if (index != -1) {
+                    path = baseName.substring(0, index).replace('.', '/') + "/" + path;
+                }
+            } else {
+                path = path.substring(1);
+            }
+            return parent.getClassLoader().getResources(path);
+        }
+    }
+
 }
--- a/jme3-core/src/main/java/com/jme3/util/res/DefaultResourceLoader.java
+++ b/jme3-core/src/main/java/com/jme3/util/res/DefaultResourceLoader.java
@@ -65,25 +65,8 @@ class DefaultResourceLoader implements ResourceLoader {
     }
 
     @Override
-    public Enumeration<URL> getResources(String path, Class<?> parent) throws IOException {
-        if (parent == null) {
-            return Thread.currentThread().getContextClassLoader().getResources(path);
-        } else {
-            if (!path.startsWith("/")) {
-                Class<?> c = parent;
-                while (c.isArray()) {
-                    c = c.getComponentType();
-                }
-                String baseName = c.getName();
-                int index = baseName.lastIndexOf('.');
-                if (index != -1) {
-                    path = baseName.substring(0, index).replace('.', '/') + "/" + path;
-                }
-            } else {
-                path = path.substring(1);
-            }
-            return parent.getClassLoader().getResources(path);
-        }
+    public Enumeration<URL> getResources(String path) throws IOException {
+        return Thread.currentThread().getContextClassLoader().getResources(path);        
     }
 
 }
--- a/jme3-core/src/main/java/com/jme3/util/res/DefaultResourceLoader.java
+++ b/jme3-core/src/main/java/com/jme3/util/res/DefaultResourceLoader.java
@@ -37,10 +37,16 @@ import java.net.URL;
 import java.util.Enumeration;
 
 /**
- * Default java-like implementation of ResourceLoaderImpl. Loads from classpath.
+ * Default implementation of {@link ResourceLoader}. 
+ * Loads from classpath.
  */
-public class DefaultResourceLoader implements ResourceLoader {
+class DefaultResourceLoader implements ResourceLoader {
+    
+    DefaultResourceLoader() {
+        
+    }
 
+    @Override
     public InputStream getResourceAsStream(String path, Class<?> parent) {
         if (parent == null) {
             return Thread.currentThread().getContextClassLoader().getResourceAsStream(path);
--- a/jme3-core/src/main/java/com/jme3/util/res/ResourceLoader.java
+++ b/jme3-core/src/main/java/com/jme3/util/res/ResourceLoader.java
@@ -36,49 +36,10 @@ import java.io.InputStream;
 import java.net.URL;
 import java.util.Enumeration;
 
-/**
- * Default java-like implementation of ResourceLoaderImpl.
- * Loads from classpath.
- */
-public class ResourcesLoaderJImpl implements ResourcesLoaderImpl {
+public interface ResourceLoader {
+    public URL getResource(String path, Class<?> clazz);
 
-    public InputStream getResourceAsStream(String path, Class<?> parent) {
-        if (parent == null) {
-            return Thread.currentThread().getContextClassLoader().getResourceAsStream(path);
-        } else {
-            return parent.getResourceAsStream(path);
-        }
-    }
+    public InputStream getResourceAsStream(String path, Class<?> clazz);
 
-    @Override
-    public URL getResource(String path, Class<?> parent) {
-        if (parent == null) {
-            return Thread.currentThread().getContextClassLoader().getResource(path);
-        } else {
-            return parent.getResource(path);
-        }
-    }
-
-    @Override
-    public Enumeration<URL> getResources(String path, Class<?> parent) throws IOException {
-        if (parent == null) {
-            return Thread.currentThread().getContextClassLoader().getResources(path);
-        } else {             
-            if (!path.startsWith("/")) {
-                Class<?> c = parent;
-                while (c.isArray()) {
-                    c = c.getComponentType();
-                }
-                String baseName = c.getName();
-                int index = baseName.lastIndexOf('.');
-                if (index != -1) {
-                    path = baseName.substring(0, index).replace('.', '/') + "/" + path;
-                }
-            } else {
-                path = path.substring(1);
-            }
-            return parent.getClassLoader().getResources(path);
-        }
-    }
-
+    public Enumeration<URL> getResources(String path, Class<?> clazz) throws IOException;
 }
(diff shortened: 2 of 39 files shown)


jme3-core/src/main/java/com/jme3/util/res/ResourcesLoaderImpl.java
```diff
@@ -0,0 +1,43 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.jme3.util.res;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Enumeration;
+
+public interface ResourcesLoaderImpl {
```
Ali-RS, line 39: I think it is common to use the "Impl" suffix name on the classes that implement the interface and is not used for naming the interface itself, but I might be wrong!

pspeed42: Yes, *Impl = concrete class

Ali-RS:
Curious to know how a non-JVM-based resource loader would look?

jme3-core/src/main/java/com/jme3/util/res/ResourcesLoader.java
```diff
@@ -0,0 +1,101 @@
...
+    
+
+   public static Enumeration<URL> getResources(String path ) throws IOException {
+        return getImpl().getResources(path,null);        
+    }
+
+
+   public static Enumeration<URL> getResources(String path,Class<?> clazz) throws IOException {
+        return getImpl().getResources(path,clazz);        
+    }
```
Ali-RS, line 100: Please fix the formatting issues in the above methods. (add required whitespace and remove extras)

riccardobl:
> Do you have an implementation of ResourceLoader that is compatible with Teavm?
Yes, the resource loader for teavm loads from the site root

```java
package com.jme3.web.filesystem;

import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.InputStream;
import java.net.MalformedURLException;
import java.net.URL;
import java.nio.charset.Charset;
import java.util.ArrayList;
import java.util.Collections;
import java.util.Enumeration;
import java.util.List;
import java.util.logging.Level;
import java.util.logging.Logger;

import org.teavm.jso.browser.Window;

import com.jme3.util.res.ResourcesLoaderImpl;

public class WebResourceLoaderImpl implements ResourcesLoaderImpl {
    private static final Logger logger = Logger.getLogger(WebResourceLoaderImpl.class.getName());

    private String getFullPath(Class<?> clazz, String path) {
        String resourcePath = path;
        if (clazz != null) {
            String className = clazz.getName();
            String classPath = className.replace('.', '/') + ".class";
            classPath = classPath.substring(0, classPath.lastIndexOf('/'));
            resourcePath = classPath + "/" + path;
        }
        try {
            URL baseURL = new URL(Window.current().getLocation().getFullURL());
            resourcePath = new URL(baseURL, resourcePath).toString();
        } catch (MalformedURLException e) {
            e.printStackTrace();
        }
        return resourcePath;
    }

    @Override
    public URL getResource(String path, Class<?> clazz) {
        path = getFullPath(clazz, path);
        logger.log(Level.FINE, "Fetch resource {0}", path);
        try {
            URL url = new URL(path);
            return url;
        } catch (MalformedURLException e) {
            e.printStackTrace();
        }
        return null;
    }

    @Override
    public InputStream getResourceAsStream(String path, Class<?> clazz) {
        path = getFullPath(clazz, path);
        logger.log(Level.FINE, "Fetch resource as stream {0}", path);
        try {
            URL url = new URL(path);
            InputStream is = url.openStream();
            return is;
        } catch (IOException e) {
            e.printStackTrace();
        }
        return null;
    }

    @Override
    public Enumeration<URL> getResources(String path, Class<?> clazz) throws IOException {
        List<URL> urls = new ArrayList<URL>();
        try {
            String indexPath = "/index.dat";
            logger.log(Level.FINE, "Fetch resources index {0}", indexPath);
            URL resourcesIndex = getResource(indexPath, null);
            ByteArrayOutputStream resourcesBo = new ByteArrayOutputStream();
            byte[] buffer = new byte[1024];
            int read = 0;
            InputStream is = resourcesIndex.openStream();
            while ((read = is.read(buffer)) > 0) {
                resourcesBo.write(buffer, 0, read);
            }
            is.close();
            logger.log(Level.FINE, "Listing resources in {0}", path);
            String resources[] = new String(resourcesBo.toByteArray(), Charset.forName("UTF-8")).split("\n");
            for (String resource : resources) {
                resource = resource.trim();
                if (resource.length() > 0 && resource.startsWith(path)) {
                    URL url = getResource(resource, null);
                    urls.add(url);
                }
            }
            logger.log(Level.FINE, "Found {0} resources in {1}", new Object[] { urls.size(), path });
        } catch (Exception e) {
            e.printStackTrace();
        }
        return Collections.enumeration(urls);
    }

}
```

jme3-core/src/main/java/com/jme3/util/res/ResourcesLoaderImpl.java
riccardobl: Renamed as you proposed

jme3-core/src/main/java/com/jme3/util/res/ResourcesLoader.java
riccardobl: Done

jme3-core/src/main/java/com/jme3/util/res/DefaultResourceLoader.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+package com.jme3.util.res;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Enumeration;
+
+/**
+ * Default java-like implementation of ResourceLoaderImpl. Loads from classpath.
```
Ali-RS, line 40: ResourceLoaderImpl in the Javadoc should be renamed to ResourceLoader.

jme3-core/src/plugins/java/com/jme3/asset/plugins/ClasspathLocator.java
```diff
@@ -87,20 +89,10 @@ public AssetInfo locate(AssetManager manager, AssetKey key) {
...
-            for (final ClassLoader classLoader : classLoaders) {
-                url = classLoader.getResource(name);
-                if(url != null) {
-                    break;
-                }
-            }
+        if (JmeSystem.isLowPermissions()) {
+            url = Resources.getResource("/" + name, ClasspathLocator.class);    
+        } else {
+            url = Resources.getResource( name);
```
Ali-RS, line 95: please remove the trailing space in `Resources.getResource( name); `

jme3-core/src/main/java/com/jme3/util/res/ResourcesLoaderImpl.java
Ali-RS: Thanks

jme3-core/src/main/java/com/jme3/util/res/ResourcesLoader.java
Ali-RS: Thanks

jme3-core/src/main/java/com/jme3/util/res/Resources.java
```diff
@@ -0,0 +1,95 @@
...
+            try {
+                impl = (ResourceLoader) Class.forName(DEFAULT_IMPL).getDeclaredConstructor().newInstance();
+            } catch (final Throwable e1) {
+                throw new RuntimeException("Unable to access default resources loader implementation", e1);
+            }
+        }
+        return impl;
+    }
+
+    public static void setImpl(ResourceLoader impl) {
```
Ali-RS, line 68: Can you please rename this method to `setResourceLoader()`?

jme3-core/src/main/java/com/jme3/util/res/Resources.java
```diff
@@ -0,0 +1,95 @@
...
+ * classpath.
+ */
+public class Resources {
+    public static final String PROPERTY_RESOURCE_LOADER_IMPLEMENTATION = "com.jme3.ResourceLoaderImplementation";
+    private static final String DEFAULT_IMPL = "com.jme3.util.res.DefaultResourceLoader";
+
+    private static final Logger LOGGER = Logger.getLogger(Resources.class.getName());
+    private static ResourceLoader impl = null;
+
+    private static ResourceLoader getImpl() {
```
Ali-RS, line 52: Can you please rename this method to `getResourceLoader()`?

jme3-core/src/main/java/com/jme3/util/res/ResourceLoader.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Enumeration;
+
+public interface ResourceLoader {
+    /**
+     * Finds the resource with the given name relative to the given parent class
+     * or to the root of the parent is null.
```
Ali-RS, line 42: >or to the root of the parent is null.

Typo? (`of` -> `if`)

jme3-core/src/main/java/com/jme3/util/res/ResourceLoader.java
```diff
@@ -0,0 +1,81 @@
...
+     *            The resource name
+     * @param parent
+     *            Optional parent class
+     * @return The resource URL or null if not found
+     */
+    public URL getResource(String path, Class<?> parent);
+
+    /**
+     * Finds the resource with the given name relative to the given parent class
+     * or to the root of the parent is null.
```
Ali-RS, line 54: Same here

jme3-core/src/main/java/com/jme3/util/res/ResourceLoader.java
riccardobl: Thanks, fixed

MeFisto94:
When reviewing, I've asked myself what the actual problem/purpose here is and to me I could come up with 2-3 actual different use-cases:
1. We seem to have quite some code that loads ominous .cfg files and by doing so, it doesn't even leverage the asset manager but uses class.getResourceAsStream(). This in itself may be questionable, because you have a config file, but can maybe hardly overwrite it? And if you're supposed to override it, why can't we use the AssetManager for such, perspectively?
2. There are some cases that have a legitimate class loader, especially around the `Savable` interface. In order to properly deserialize, we need class instances and it looks like the class loader just tries to locate and instantiate those (instead of manually registering implementations or `Class<?>`s.
3. Then there's the ClassLoader AssetLocator, probably the default for DesktopAssetManager, but do we really need to replace it with a ResourceLoader, if we could just swap in another AssetLocator?

So basically I just have the gut feeling that we could solve each of those even better, given good design/refactorings.

riccardobl:
> We seem to have quite some code that loads ominous .cfg files and by doing so, it doesn't even leverage the asset manager but uses class.getResourceAsStream(). This in itself may be questionable, because you have a config file, but can maybe hardly overwrite it? And if you're supposed to override it, why can't we use the AssetManager for such, perspectively?

These files are used to initialize the assetManager , so it doesn't exist yet at that point. Also it is a good thing they don't use the asset manage features, because they aren't assets, they don't need caching or to be overwritten by things in the asset path.


> 2\. There are some cases that have a legitimate class loader, especially around the `Savable` interface. In order to properly deserialize, we need class instances and it looks like the class loader just tries to locate and instantiate those (instead of manually registering implementations or `Class<?>`s.

This change is only related to resources in the classpath, not classes.





> Then there's the ClassLoader AssetLocator, probably the default for DesktopAssetManager, but do we really need to replace it with a ResourceLoader, if we could just swap in another AssetLocator?

If the problem was only with the asset locator, i would have written a new one of my teavm implementation, but this affect various parts of the engine outside of the asset locator and the side effect of this patch is that the default classpath locator becomes automatically compatible.




> So basically I just have the gut feeling that we could solve each of those even better, given good design/refactorings.

Maybe, but redesign all this would take way too much time vs this simple patch that doesn't break anything.

jme3-core/src/main/java/com/jme3/util/res/Resources.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This class is used to load resources from the default location usually the
+ * classpath.
+ */
+public class Resources {
+    public static final String PROPERTY_RESOURCE_LOADER_IMPLEMENTATION = "com.jme3.ResourceLoaderImplementation";
+    private static final String DEFAULT_IMPL = "com.jme3.util.res.DefaultResourceLoader";
```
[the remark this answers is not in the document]
riccardobl: That's done on purpose so that "com.jme3.util.res.DefaultResourceLoader"  doesn't need to be included in the build.

PR #2061, opened 2023-08-09
GL ES 3.0 megapatch

riccardobl:
This PR improves GL ES 3.0 support in jme:
- add glBindVertexArray, glDeleteVertexArrays and  glGenVertexArrays
- enable coreprofile  (except on Ios due to limited support) and other supported capabilities when GL ES 3.0 is used
- detect webgl platforms
- detect when RGB floating point draw buffers are supported vs RGBA
- replace use of RGB floating attachment with RGBA when not supported
- reimplement glDrawBuffer shim to workaround the limits of GL ES 3 and reproduce the exact same behavior of OpenGL (when GLSLCompat is included in shaders)
- adapt all materials to explicitly support GL ES 3 and new glDrawBuffer shim.


riccardobl:
> Do we need to test Android for these new changes?

Probably yes, before the next release at least.

riccardobl:
New issues have arisen: 

1. the current implementation of glDrawBuffer workaround (since not available in GL ES 3) was incomplete ( see old implementation discussion here: https://hub.jmonkeyengine.org/t/opengl-es-3-0-and-newer-on-android/41913/61 ).
The problem here is that jme allows to set the target attachment of a multi target framebuffer, this is not possible in any way in GL ES since glDrawBuffers can only enable or disable attachments but cannot change their order (eg. color0 must always draw on attachment0). My workaround uses multi target rendering and some macro tricks to define the outFragColor in GLSLCompat to output to the right attachment as per framebuffer target, basically a transparent fix for all those shaders that use GLSLCompat that requires no code change from the shader developer. All materials are updated to support this fix with the inclusion of BoundDrawBuffer param and BOUND_DRAW_BUFFER define (similar to the way we support NUM_SAMPLES of MSAA)

2. All shaders langs declarations in matdefs have been reordered with the proper order

3. A capability to detect if RGB float attachments (nb. without alpha) are supported (normally they aren't in standard GL ES 3 implementations) and the relevant code has been changed to use RGBA when RGB is not supported (ie filter post processor)

riccardobl:
The last commit removes the hardcoded precision from GLRenderer and puts it in the optional GLSLCompat, it also checks if highp is supported and sets to mediump if it isn't.

PR #2062, opened 2023-08-09
Fix contrast adjustment matdef

riccardobl:
Some fixes to the contrast adjustment filter:
- Include GLSLCompat so that there is no need to have two default techniques and two fragment shaders.
- Remove versions > GLSL150 from the declaration since they aren't required
- add NUM_SAMPLES


Ali-RS:
Probably this is related 

https://hub.jmonkeyengine.org/t/solved-error-with-contrastadjustmentfilter-in-android/46215

riccardobl:
This patch has been merged with https://github.com/jMonkeyEngine/jmonkeyengine/pull/2061

PR #2071, opened 2023-09-15
make GltfLoader use node names for geometries instead of mesh names

JosiahGoeman:
https://hub.jmonkeyengine.org/t/solved-getting-the-names-of-objects-not-meshes-into-jme/45511

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -261,7 +261,7 @@ public class GltfLoader implements AssetLoader {
                 }
                 spatial = node;
             }
-            spatial.setName(readMeshName(meshIndex));
+            spatial.setName(getAsString(nodeData, "name"));
 
         } else {
             // no mesh, we have a node. Can be a camera node or a regular node.
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -31,21 +31,22 @@
  */
 package com.jme3.scene.plugins.gltf;
 
-import com.jme3.plugins.json.JsonArray;
-import com.jme3.plugins.json.JsonObject;
-import com.jme3.plugins.json.JsonPrimitive;
-import com.jme3.plugins.json.JsonElement;
+import static com.jme3.scene.plugins.gltf.GltfUtils.*;
+
 import com.jme3.anim.*;
 import com.jme3.asset.*;
 import com.jme3.material.Material;
 import com.jme3.material.RenderState;
 import com.jme3.math.*;
+import com.jme3.plugins.json.JsonArray;
+import com.jme3.plugins.json.JsonElement;
+import com.jme3.plugins.json.JsonObject;
+import com.jme3.plugins.json.JsonPrimitive;
 import com.jme3.renderer.Camera;
 import com.jme3.renderer.queue.RenderQueue;
 import com.jme3.scene.*;
 import com.jme3.scene.control.CameraControl;
 import com.jme3.scene.mesh.MorphTarget;
-import static com.jme3.scene.plugins.gltf.GltfUtils.*;
 import com.jme3.texture.Texture;
 import com.jme3.texture.Texture2D;
 import com.jme3.util.IntMap;
@@ -128,7 +129,11 @@ public class GltfLoader implements AssetLoader {
             String version = getAsString(asset, "version");
             String minVersion = getAsString(asset, "minVersion");
             if (!isSupported(version, minVersion)) {
-                logger.log(Level.SEVERE, "Gltf Loader doesn''t support this gltf version: {0}{1}", new Object[]{version, minVersion != null ? ("/" + minVersion) : ""});
+                logger.log(
+                    Level.SEVERE,
+                    "Gltf Loader doesn''t support this gltf version: {0}{1}",
+                    new Object[] { version, minVersion != null ? ("/" + minVersion) : "" }
+                );
             }
 
             scenes = docRoot.getAsJsonArray("scenes");
@@ -262,7 +267,6 @@ public class GltfLoader implements AssetLoader {
                 spatial = node;
             }
             spatial.setName(getAsString(nodeData, "name"));
-
         } else {
             // no mesh, we have a node. Can be a camera node or a regular node.
             Integer camIndex = getAsInteger(nodeData, "camera");
@@ -341,24 +345,29 @@ public class GltfLoader implements AssetLoader {
         JsonArray translation = nodeData.getAsJsonArray("translation");
         if (translation != null) {
             transform.setTranslation(
-                    translation.get(0).getAsFloat(),
-                    translation.get(1).getAsFloat(),
-                    translation.get(2).getAsFloat());
+                translation.get(0).getAsFloat(),
+                translation.get(1).getAsFloat(),
+                translation.get(2).getAsFloat()
+            );
         }
         JsonArray rotation = nodeData.getAsJsonArray("rotation");
         if (rotation != null) {
-            transform.setRotation(new Quaternion(
+            transform.setRotation(
+                new Quaternion(
                     rotation.get(0).getAsFloat(),
                     rotation.get(1).getAsFloat(),
                     rotation.get(2).getAsFloat(),
-                    rotation.get(3).getAsFloat()));
+                    rotation.get(3).getAsFloat()
+                )
+            );
         }
         JsonArray scale = nodeData.getAsJsonArray("scale");
         if (scale != null) {
             transform.setScale(
-                    scale.get(0).getAsFloat(),
-                    scale.get(1).getAsFloat(),
-                    scale.get(2).getAsFloat());
+                scale.get(0).getAsFloat(),
+                scale.get(1).getAsFloat(),
+                scale.get(2).getAsFloat()
+            );
         }
 
         return transform;
@@ -403,16 +412,20 @@ public class GltfLoader implements AssetLoader {
                 String bufferType = entry.getKey();
                 if (bufferType.startsWith("JOINTS")) {
                     SkinBuffers buffs = getSkinBuffers(bufferType);
-                    SkinBuffers buffer
-                            = readAccessorData(entry.getValue().getAsInt(), new JointArrayPopulator());
+                    SkinBuffers buffer = readAccessorData(
+                        entry.getValue().getAsInt(),
+                        new JointArrayPopulator()
+                    );
                     buffs.joints = buffer.joints;
                     buffs.componentSize = buffer.componentSize;
                 } else if (bufferType.startsWith("WEIGHTS")) {
                     SkinBuffers buffs = getSkinBuffers(bufferType);
                     buffs.weights = readAccessorData(entry.getValue().getAsInt(), new FloatArrayPopulator());
                 } else {
-                    VertexBuffer vb = readAccessorData(entry.getValue().getAsInt(),
-                            new VertexBufferPopulator(getVertexBufferType(bufferType)));
+                    VertexBuffer vb = readAccessorData(
+                        entry.getValue().getAsInt(),
+                        new VertexBufferPopulator(getVertexBufferType(bufferType))
+                    );
                     if (vb != null) {
                         mesh.setBuffer(vb);
                     }
@@ -454,8 +467,10 @@ public class GltfLoader implements AssetLoader {
                     for (Map.Entry<String, JsonElement> entry : target.getAsJsonObject().entrySet()) {
                         String bufferType = entry.getKey();
                         VertexBuffer.Type type = getVertexBufferType(bufferType);
-                        VertexBuffer vb = readAccessorData(entry.getValue().getAsInt(),
-                                new VertexBufferPopulator(type));
+                        VertexBuffer vb = readAccessorData(
+                            entry.getValue().getAsInt(),
+                            new VertexBufferPopulator(type)
+                        );
                         if (vb != null) {
                             morphTarget.setBuffer(type, (FloatBuffer) vb.getData());
                         }
@@ -474,8 +489,10 @@ public class GltfLoader implements AssetLoader {
             } else {
                 useNormalsFlag = false;
                 geom.setMaterial(readMaterial(materialIndex));
-                if (geom.getMaterial().getAdditionalRenderState().getBlendMode()
-                        == RenderState.BlendMode.Alpha) {
+                if (
+                    geom.getMaterial().getAdditionalRenderState().getBlendMode() ==
+                    RenderState.BlendMode.Alpha
+                ) {
                     // Alpha blending is enabled for this material. Let's place the geom in the transparent bucket.
                     geom.setQueueBucket(RenderQueue.Bucket.Transparent);
                 }
@@ -533,8 +550,14 @@ public class GltfLoader implements AssetLoader {
         return data;
     }
 
-    public Object readBuffer(Integer bufferViewIndex, int byteOffset, int count, Object store,
-            int numComponents, VertexBuffer.Format format) throws IOException {
+    public Object readBuffer(
+        Integer bufferViewIndex,
+        int byteOffset,
+        int count,
+        Object store,
+        int numComponents,
+        VertexBuffer.Format format
+    ) throws IOException {
         JsonObject bufferView = bufferViews.get(bufferViewIndex).getAsJsonObject();
         Integer bufferIndex = getAsInteger(bufferView, "buffer");
         assertNotNull(bufferIndex, "No buffer defined for bufferView " + bufferViewIndex);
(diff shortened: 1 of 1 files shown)


Ali-RS:
@JosiahGoeman thanks for your contribution.

@riccardobl can we adjust autoformatting to only apply to the changed lines and not the whole class?

tonihele:
At least the formatting is in its own commit. So it is still easy to check what actually changed.

pspeed42:
Auto-formatting = "being exactly the same is more important than being readable."  And several of these changes are good examples.  Oh, well.

riccardobl:
> @riccardobl can we adjust auto-formatting to only apply to the changed lines and not the whole class?
That's not possible afaik, the best we can do is run the formatter once  for the entire repo.

Considering that the commits are separated and the pr can be merged without squash, is the current approach really bad?

stephengold:
Anything that harms readability is bad, unless it's required for correctness.

JosiahGoeman:
Oof, it really did a number on that file.  Sorry, I didn't notice the extra commit.

Ali-RS:
>Sorry, I didn't notice the extra commit.

That is added by the auto-formatting plugin runs via our GitHub actions. Nothing wrong on your side.

riccardobl:
Ok, so we can disable with workflow for now, and re-enable when (if) we get to jme4 after a full formattation of the legacy code

riccardobl:
> https://hub.jmonkeyengine.org/t/solved-getting-the-names-of-objects-not-meshes-into-jme/45511

I am not sure about this PR.
The loader already creates a parent node with the node name, and that's the correct approach, since gltf meshes map to jme geometries and they should be named after the mesh name.

Consider that one gltf mesh with multiple materials can be converted into multiple jme geometries.

Eg. if you have a car with 3 materials, it will look like this in gltf

```
car : Node ->
   carMesh(mat1, mat2, mat3) : Mesh
```
but like this in jme

```
car : Node -> 
   carMesh : Node ->
     carMesh1 (mat1) : Geometry
     carMesh2 (mat2) : Geometry
     carMesh3 (mat3) : Geometry
```

and that's expected.

With your patch it will look like this

```
car : Node -> 
   car : Node ->
     carMesh1 (mat1) : Geometry
     carMesh2 (mat2) : Geometry
     carMesh3 (mat3) : Geometry
```

This might lead to unexpected results when looking for the "car" node.


If anything i think we should get rid lines 252 to 256 since they introduce inconsistency in the node tree

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java#L252-L256

Ali-RS:
>Eg. if you have a car with 3 materials, it will look like this in gltf

I tried this but I could not reproduce the issue you mentioned

I made a simple test in Blender:

Screenshot_2023-09-16_00-49-07

this is the gltf file:

```
{
    "asset" : {
        "generator" : "Khronos glTF Blender I/O v3.4.50",
        "version" : "2.0"
    },
    "scene" : 0,
    "scenes" : [
        {
            "name" : "Scene",
            "nodes" : [
                0
            ]
        }
    ],
    "nodes" : [
        {
            "mesh" : 0,
            "name" : "Car"
        }
    ],
    "materials" : [
        {
            "doubleSided" : true,
            "name" : "Material-Green",
            "pbrMetallicRoughness" : {
                "baseColorFactor" : [
                    0.026959503069519997,
                    0.8000000715255737,
                    0.02678806707262993,
                    1
                ],
                "metallicFactor" : 0,
                "roughnessFactor" : 0.5
            }
        },
        {
            "doubleSided" : true,
            "name" : "Material-Blue",
            "pbrMetallicRoughness" : {
                "baseColorFactor" : [
                    0.0382981039583683,
                    0.02713284268975258,
                    0.8000000715255737,
                    1
                ],
                "metallicFactor" : 0,
                "roughnessFactor" : 0.5
            }
        }
    ],
    "meshes" : [
        {
            "name" : "Car-Mesh",
            "primitives" : [
                {
                    "attributes" : {
                        "POSITION" : 0,
                        "TEXCOORD_0" : 1,
                        "NORMAL" : 2
                    },
                    "indices" : 3,
                    "material" : 0
                },
                {
                    "attributes" : {
                        "POSITION" : 4,
                        "TEXCOORD_0" : 5,
                        "NORMAL" : 6
                    },
                    "indices" : 3,
                    "material" : 1
                }
            ]
        }
    ],
    "accessors" : [
        {
            "bufferView" : 0,
            "componentType" : 5126,
            "count" : 24,
            "max" : [
                1,
                1,
                1
            ],
            "min" : [
                -1,
                -1,
                -1
            ],
            "type" : "VEC3"
        },
        {
            "bufferView" : 1,
            "componentType" : 5126,
            "count" : 24,
            "type" : "VEC2"
        },
        {
            "bufferView" : 2,
            "componentType" : 5126,
            "count" : 24,
            "type" : "VEC3"
        },
        {
            "bufferView" : 3,
            "componentType" : 5123,
            "count" : 36,
            "type" : "SCALAR"
        },
        {
            "bufferView" : 4,
            "componentType" : 5126,
            "count" : 24,
            "max" : [
                3.7098071575164795,
                1,
                1
            ],
            "min" : [
                1.7098071575164795,
                -1,
                -1
            ],
            "type" : "VEC3"
        },
        {
            "bufferView" : 5,
            "componentType" : 5126,
            "count" : 24,
            "type" : "VEC2"
        },
        {
            "bufferView" : 6,
            "componentType" : 5126,
            "count" : 24,
            "type" : "VEC3"
        }
    ],
    "bufferViews" : [
        {
            "buffer" : 0,
            "byteLength" : 288,
            "byteOffset" : 0,
            "target" : 34962
        },
        {
            "buffer" : 0,
            "byteLength" : 192,
            "byteOffset" : 288,
            "target" : 34962
        },
        {
            "buffer" : 0,
            "byteLength" : 288,
            "byteOffset" : 480,
            "target" : 34962
        },
        {
            "buffer" : 0,
            "byteLength" : 72,
            "byteOffset" : 768,
            "target" : 34963
        },
        {
            "buffer" : 0,
            "byteLength" : 288,
            "byteOffset" : 840,
            "target" : 34962
        },
        {
            "buffer" : 0,
            "byteLength" : 192,
            "byteOffset" : 1128,
            "target" : 34962
        },
        {
            "buffer" : 0,
            "byteLength" : 288,
            "byteOffset" : 1320,
            "target" : 34962
        }
    ],
    "buffers" : [
        {
            "byteLength" : 1608,
            "uri" : "Cube.bin"
        }
    ]
}
```

This is the result in JME before and after the changes:

```

1) Before
Scene (Node) 
	Car-Mesh (Node) 
		Car-Mesh_0 (Geometry) / Material[name=Material-Green, def=PBR Lighting, tech=null]
		Car-Mesh_1 (Geometry) / Material[name=Material-Blue, def=PBR Lighting, tech=null]

2) After
Scene (Node) 
	Car (Node) 
		Car-Mesh_0 (Geometry) / Material[name=Material-Green, def=PBR Lighting, tech=null]
		Car-Mesh_1 (Geometry) / Material[name=Material-Blue, def=PBR Lighting, tech=null]

```

The second one looks correct to me.

Here is the model if you want to try: 
[TestGltf.zip](https://github.com/jMonkeyEngine/jmonkeyengine/files/12639880/TestGltf.zip)

riccardobl:
You are right, i was confused by the code, for multi material mesh the node tree seems correct, but for single material (one-geometry) it looks like this with the patch:

single:
```
Root Node (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
        autoRotate (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                probeNode (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                        Scene (Node) / (0.0, -1.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                Cube (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                
```

vs this for multi:

```
Root Node (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
        autoRotate (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                probeNode (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                        Scene (Node) / (0.0, -1.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                Cube (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        CubeMesh_0 (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        CubeMesh_1 (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
```


As you can see the mesh name is lost entirely in single material meshes.
I think that for consistency sake it should look like this instead:

multi;
```
Root Node (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
        autoRotate (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                probeNode (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                        Scene (Node) / (0.0, -1.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                Cube (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        CubeMesh (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                                CubeMesh_0 (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                                CubeMesh_1 (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
```

single:

```
Root Node (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
        autoRotate (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                probeNode (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                        Scene (Node) / (0.0, -1.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                Cube (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        CubeMesh (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                                CubeMesh_0 (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)

                                                                         
  ```


To get this node tree i've replaced the lines from 252 to 264  in  GltfLoader.java
https://github.com/jMonkeyEngine/jmonkeyengine/blob/2a1e87e5a8050ee4e48f3baa9abdd421918ff061/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java#L252-L264

 with

```java
           Geometry[] primitives = readMeshPrimitives(meshIndex);

            Node node = new Node();
            for (Geometry primitive : primitives) {
                node.attachChild(primitive);
            }

            node.setName(readMeshName(meshIndex));
            
            spatial = new Node();
            ((Node)spatial).attachChild(node);
```

and line 489 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/2a1e87e5a8050ee4e48f3baa9abdd421918ff061/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java#L488-L490

with
```
                    geom.setName(name +  "_" + index);
```


This gives a static structure to work with, that doesn't change with the content of the scene.
What do you think?

Ali-RS:
>As you can see the mesh name is lost entirely in single material meshes.

Yeah, and that is fine because in JME a mesh can not have a name anyway. If a mesh had a name in JME then we could use the gltf mesh name as well.

>What do you think?

Not sure, to me your approach unnesesarly adds redundant nodes.

JosiahGoeman:
> > As you can see the mesh name is lost entirely in single material meshes.
> 
> Yeah, and that is fine because in JME a mesh can not have a name anyway. If a mesh had a name in JME then we could use the gltf mesh name as well I think.

Yes, that was my reasoning too.  Since JME meshes don't have names, it makes sense for that information should be what gets lost on import.  Would it make sense to add a name field to the Mesh class?  For gltf at least, it would be cleaner than having the extra node.  I don't know how the FBX and ogre loaders work though.

riccardobl:
In jme the concept of blender/gltf mesh is expressed by Geometry, Mesh are at lower level and are just a collection of buffers.

eg. in jme you assign materials to Geometries, in blender you assign them to meshes, in gltf to primitives that are a subpart of a mesh and are loaded as geometries by the loader.

I don't think this approach is redundant because it  provides a single predictable way to represent  the scene tree independently on its content. 

### Let's explore a common use case: i want to replace the CubeMesh with Cube2Mesh for whatever reason (lod, customization, etc..), my Cube spatial is used by my game logic (eg. has control attached, rigidbody etc) how do i replace the mesh with the minimal change to the scene?

Let's see this in code
## with the proposed patch, i need a whole flowchart:
Untitled Diagram drawio(1)

You could say, just remove "Cube", and replace the entire node. Yes that would work, but you will need to transfer all the controls and replace the references in your app.

## with my proposal:
```
Node parent = (Node) s;
Spatial meshContainer = parent.getChild("CubeMesh");
meshContainer.getParent().attachChild(cube2Mesh);
meshContainer.removeFromParent();
```
done, references are not touched, no need to copy transforms or controls.

riccardobl:
I've found another corner case that is even worse:
image

With the proposed patch it will load like this

```
Root Node (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
        autoRotate (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                probeNode (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                        Scene (Node) / (0.0, -1.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                CubeParent (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        CubeMesh (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        Cube2 (Geometry) / (1.2840239, -2.1808097, -0.21893275), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        
```
So if i want to select the first cube i will need to use getChild("CubeMesh"), while if i want to select the second cube getChild("Cube2") ?

With my proposal it will load like this:
```
Root Node (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
        autoRotate (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                probeNode (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                        Scene (Node) / (0.0, -1.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                CubeParent (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        CubeMesh (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                                CubeMesh_0 (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                        Cube2 (Node) / (1.2840239, -2.1808097, -0.21893275), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                                CubeMesh2 (Node) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
                                                        CubeMesh2_0 (Geometry) / (0.0, 0.0, 0.0), (0.0, 0.0, 0.0, 1.0), (1.0, 1.0, 1.0)
```
and i can just use getChild("CubeMesh") getChild("CubeMesh2") that is incidentally what i would do in blender.

Gltf attached 
[parent.zip](https://github.com/jMonkeyEngine/jmonkeyengine/files/12641131/parent.zip)

riccardobl:
Fixed by https://github.com/jMonkeyEngine/jmonkeyengine/pull/2103

PR #2072, opened 2023-09-17
update LWJGL to v3.3.3

stephengold:
I've reviewed the release notes and begun using LWJGL 3.3.3 in my own projects, both on Linux and Windows 11. So far I haven't uncovered any regressions or new issues.

I think this change shouldn't be controversial.
I'm submitting this PR as a reminder to upgrade our Engine at some convenient time.

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -28,7 +28,7 @@ tasks.withType(JavaCompile) { // compile-time options:
 }
 
 ext {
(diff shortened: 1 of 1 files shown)


stephengold:
Unless there's substantive discussion, I plan to integrate this PR in about 72 hours.

PR #2074, opened 2023-09-22
Mesh:  add constants for initial/default values of fields

MelVimL:
Changed default values to "0" instead of "-1" and moved them to private constant class variables.
I guess "-1" was used as an fail/error-state but i didn't seen any checks besides ">=0", "!=0" and "==0".

Also after reading the request which also fixes #334 I am not sure if this a useful solution. 

Unittests -> ok
Handpicked examples(Scenes, Animation) ->  #ok

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Mesh.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Mesh.java
@@ -153,16 +153,17 @@ public class Mesh implements Savable, Cloneable, JmeCloneable {
     /**
      * Default Variables
      */
+    private static final int DEFAULT_VERTEXT_ARRAY_ID = -1;
     private static final CollisionData DEFAULT_COLLECTION_TREE = null;
 
     private static final float DEFAULT_POINT_SIZE = 1.0f;
     private static final float DEFAULT_LINE_WIDTH = 1.0f;
 
-    private static final int DEFAULT_VERT_COUNT = 0;
-    private static final int DEFAULT_ELEMENT_COUNT = 0;
-    private static final int DEFAULT_INSTANCE_COUNT = 0;
+    private static final int DEFAULT_VERT_COUNT = -1;
+    private static final int DEFAULT_ELEMENT_COUNT = -1;
+    private static final int DEFAULT_INSTANCE_COUNT = -1;
     private static final int DEFAULT_PATCH_VERTEX_COUNT = 3;
-    private static final int DEFAULT_MAX_NUM_WEIGHTS = 0;
+    private static final int DEFAULT_MAX_NUM_WEIGHTS = -1;
     
     /**
      * The bounding volume that contains the mesh entirely.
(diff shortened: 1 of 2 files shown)


stephengold:
This isn't a good solution for #313, because the render statistics will still be incorrect.

If the statistics are -1, that indicates that the mesh hasn't been completely initialized. If such a mesh were passed to the renderer, it might be best to fail quickly.

MelVimL:
Okay. I should read more thoroughly: **"If the stats value is -1 the mesh should be considered invalid and crash with exception."**

stephengold:
We could either close this PR or else turn it into a refactoring PR that simply adds a bunch of constants. @MelVimL what's your preference?

If we pursue the latter course, I request that the constants not be public unless there's a good case for using them outside the package.

MelVimL:
I like constants for defaults/inits, even the private ones.

jme3-core/src/main/java/com/jme3/scene/Mesh.java
```diff
@@ -148,28 +149,44 @@ public boolean isListMode() {
             return listMode;
         }
     }
+    
+    /**
+     * Default Variables
+     */
+    private static final int DEFAULT_VERTEXT_ARRAY_ID = -1;
+    private static final CollisionData DEFAULT_COLLECTION_TREE = null;
```
stephengold, line 157: This should be `COLLISION` not `COLLECTION`.

jme3-core/src/main/java/com/jme3/scene/Mesh.java
```diff
@@ -148,28 +149,44 @@ public boolean isListMode() {
             return listMode;
         }
     }
+    
+    /**
+     * Default Variables
+     */
+    private static final int DEFAULT_VERTEXT_ARRAY_ID = -1;
```
stephengold, line 156: This should be `VERTEX` not `VERTEXT`.

stephengold (changes requested):
with apologies for the slow review

jme3-core/src/test/java/com/jme3/scene/mesh/MeshTest.java
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.Test;
+
+import com.jme3.scene.Mesh;
+
+public class MeshTest {
+       @Test
+    public void testEmptyMesh() {
+        final Mesh mesh = new Mesh();
+
+        assertEquals(-1, mesh.getVertexCount());;
```
capdevon, line 14: double semicolon ';;'  at the end of the instruction

MelVimL: Changed it! Thank you for the Feedback!

jme3-core/src/main/java/com/jme3/scene/Mesh.java
MelVimL: Fixed it! Sorry for the typos.

stephengold (approved):
Looks good.

stephengold:
Unless there's further substantive discussion, I plan to integrate this in 24 hours.

stephengold:
Oops! I just noticed the added file "MeshTest.java". Probably it should be deleted from this PR. If not deleted, it needs javadoc and a copy of the license.

@MelVimL would you please take care of it?

MelVimL:
@stephengold  Added the requesting javadoc and license. I used the TestSpatial as a reference.

stephengold:
Thank you @MelVimL . I plan to integrate this in a few hours.

MelVimL:
@stephengold You're Welcome. Love the project!

Issue #2078, opened 2023-10-04
IllegalArgumentException in Camera.setFov()


stephengold:
A crash that's easy to reproduce using `FlyByCamera` on a non-perspective camera in JMonkeyEngine v3.6.1-stable:
```console
Oct 04, 2023 12:08:02 AM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.IllegalArgumentException: Cannot set field of view on orthogonal camera
	at com.jme3.renderer.Camera.setFov(Camera.java:664)
	at com.jme3.input.FlyByCamera.zoomCamera(FlyByCamera.java:388)
	at com.jme3.input.FlyByCamera.onAnalog(FlyByCamera.java:467)
	at com.jme3.input.InputManager.invokeAnalogsAndActions(InputManager.java:285)
	at com.jme3.input.InputManager.onMouseMotionEventQueued(InputManager.java:412)
	at com.jme3.input.InputManager.processQueue(InputManager.java:869)
	at com.jme3.input.InputManager.update(InputManager.java:923)
	at com.jme3.app.LegacyApplication.update(LegacyApplication.java:785)
	at com.jme3.app.SimpleApplication.update(SimpleApplication.java:248)
	at com.jme3.system.lwjgl.LwjglWindow.runLoop(LwjglWindow.java:628)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:717)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

An easy fix would be for `FlyByCamera` to zoom only when the camera is in perspective mode. Another approach would be to implement zoom differently when the camera is in orthogonal mode.

stephengold:
This might be related to PR #1492.

stephengold:
I reproduced the issue in current "master" branch.
Here's a minimal test app. Turning the scroll wheel on the mouse triggers the crash.
```java
import com.jme3.app.SimpleApplication;

public class TestIssue2078 extends SimpleApplication {
    public static void main(String[] args) {
        new TestIssue2078().start();
    }

    @Override
    public void simpleInitApp() {
        cam.setParallelProjection(true);
    }
}
```

stephengold:
I verified that this issue appeared between JME v3.3.2-stable and JME v3.4.0-stable .
That's consistent with PR #1492 being the cause.

PR #2079, opened 2023-10-04
bugfix: IllegalArgumentException in Camera.setFov() (issue #2078)

stephengold:
A proposed fix for issue #2078.
This solution enables `FlyByCamera` to zoom even when the camera is in orthographic mode.
Not exactly how it worked in JME v3.3.2, but similar.


stephengold:
Unless there's substantial discussion, I plan to integrate this pull request in about 72 hours.

Issue #2070, opened 2023-09-09
GltfModelKey and SceneKey should override hashCode() and equals()


stephengold:
`AssetKey` is used as a key in various hash maps, notably the `ConcurrentHashMap` in `WeakRefCloneAssetCache`. That's why it overrides the `hashCode()` and `equals()` methods of `Object`.

For the same reason, any subclass of `AssetKey` that adds parameters should also override `hashCode()` and `equals()`. That way, keys that differ only in those extra parameters can load different versions of the asset. `OgreMeshKey` serves as a good example of how to override the methods.

`GltfModelKey` and `SceneKey` are subclasses that add parameters, but they do not override `hashCode()` and `equals()`. As a result, if you (for instance) load the same glTF file using keys with different values of `keepSkeletonPose`, your 2nd load will get a clone of the cached model instead of a fresh load.

This is perhaps an unusual situation, but worth fixing in my opinion.

PR #2081, opened 2023-10-06
Fixes Issue #2070 (AssetKey overrides)

codex128:
This commit fixes the minor #2070 asset key issue by adding equals and hashcode methods to GltfModelKey and SceneKey.


stephengold:
Thank you, @codex128

stephengold (changes requested):

stephengold:
Thanks for the changes @codex128 .  Unless there's further substantial discussion, I plan to integrate this fix in about 24 hours.

stephengold (approved):

stephengold:
I intended a squashed integration but apparently I goofed and did a straight rebase, so now the fix is split into 2 commits in the "master" branch.

Issue #2082, opened 2023-10-06
jme3-plugins imports Alpha Blend Mode Test incorrectly


stephengold:
I noticed a defect in `PBRMaterialAdapter` while developing the MonkeyWrench library.

The test data is [the Alpha Blend Mode Test from Khronos's glTF sample models](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AlphaBlendModeTest). The model includes 5 rectangles with diagonal striping.

Here's how it looks in `CompareLoaders` when loaded using jme3-plugins:

![AlphaBlendMode-231006](https://github.com/jMonkeyEngine/jmonkeyengine/assets/2287609/3183a818-1cd6-4a1f-8814-97b95e883589)

The issue is that the center bar ("Cutoff 0.25") and the one immediately to its right ("Cutoff Default") aren't opaque all the way up to their cutoff values. Instead, they blend gradually: up to the cutoff value (in the case of "Cutoff 0.25") or else to the top of the scale (in the case of "Cutoff Default").  Please compare the screenshot embedded above with the image from the Khronos repo: https://github.com/KhronosGroup/glTF-Sample-Models/raw/master/2.0/AlphaBlendModeTest/screenshot/screenshot_large.jpg

I suspect this issue has to do with the `PBRMaterialAdapter` class.

To render the correct appearance, I think the "MatCutoffDefault" and "TestCutoff25Mesh" materials should both have `BlendMode.Off` in their additional render states (currently they both have `BlendMode.Alpha`).

Furthermore, I think the "MatCutoffDefault" material should set the "AlphaDiscardThreshold" material parameter to 0.5 (currently the parameter is not set).

PR #2084, opened 2023-10-08
Fixes Issue #2082 (AlphaBlendModeTest)

codex128:
Fixes issues loading blend modes and alpha discard levels on gltf models as described in #2082.


stephengold:
Thanks for your work. A couple changes, please:
+ The "AlphaBlendModeTest" data is licensed CC-BY 4.0, which is not compatible with JME's license.  Either create new test data or delete the test.
+ There should be no line break after a brace that is followed by `else`. Delete those line breaks.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/PBRMaterialAdapter.java
```diff
@@ -60,30 +65,32 @@ protected MatParam adaptMatParam(MatParam param) {
...
+                case "MASK":
+                    // "MASK" -> BlendMode.Off
+                    getMaterial().setFloat("AlphaDiscardThreshold", MASK_ALPHA_DISCARD);
+                    break;
                 case "BLEND":
                     getMaterial().getAdditionalRenderState().setBlendMode(RenderState.BlendMode.Alpha);
                     break;
             }
             // Alpha is a RenderState not a Material Parameter, so return null
             return null;
```
stephengold, line 77: I think the method should return a `MatParam` here.

stephengold:
Unless there is further substantive discussion, I plan to integrate this PR in about 48 hours.

stephengold:
Thanks for the contribution, @codex128 !

PR #2088, opened 2023-10-09
Fixes Issue #2085 (KHR_materials_emissive_strengh extension)

### LINKED ISSUE #2085: jme3-plugins imports EmissiveStrengthTest incorrectly
I noticed a defect in GltfLoader while developing the MonkeyWrench library.

The test data is [EmissiveStrengthTest from Khronos's glTF sample models](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/EmissiveStrengthTest). The model includes 5 colored cubes, each with a different emissive strength.

Here's how it looks in CompareLoaders when "EmissiveStrengthTest/glTF/EmissiveStrengthTest.gltf" is loaded using jme3-plugins:
EmissiveStrength-231008

Please compare the screenshot embedded above with the image from the Khronos repo: https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/EmissiveStrengthTest/screenshot/screenshot_large_plain.jpg

The cubes are supposed to be rendered in different hues.

Preliminary investigation revealed that GltfLoader created a distinct PBR material for each cube, but all the materials have identical parameters, including "EmissiveIntensity" of 2. In order to render properly, some of the cubes should have different values for the "EmissiveIntensity" material parameter, including 4, 8, and 16.

codex128:
Adds `KHR_materials_emissive_strength` GLTF extension loader that fixes #2085.

On a side note, I noticed PBRLighting.j3md does not have `GlowMap` or `GlowColor` material parameters. Is that on purpose?


stephengold:
The "EmissiveStrengthTest" model is licensed CC-BY 4.0, which is not compatible with the BSD license of jme3-examples.
Moving it to jme3-testdata won't help, since that's also licensed BSD. So please remove the 3 model files from this PR.

If you don't feel like creating a new test model from scratch, I suggest using jme3test.model.TestGltfLoading for testing.

codex128:
>If you don't feel like creating a new test model from scratch, I suggest using jme3test.model.TestGltfLoading for testing.

Oh, yeah, that looks much easier! :smile:

pspeed42:
re: "I believe PBRLighting.j3md uses EmissiveMap and Emissive in place of GlowMap and GlowColor"
I may be speaking on topics I don't understand, but: both Lighting.j3md and PBRLighting.j3md seem to use the same shaders for the glow pass.  So I'm not sure what adapts the Emissive* stuff to GLow* stuff.  Perhaps we need a PBRGlow.frag that uses EmissiveMap/Emissive.

stephengold:
Looks good to me. Unless there's further substantive discussion of this PR, I plan to integrate it in about 72 hours.

PR #2091, opened 2023-10-10
Partial fix for Issue #2087 (NegativeScaleTest)

### LINKED ISSUE #2087: jme3-plugins imports NegativeScaleTest incorrectly
I noticed a defect in GltfLoader while developing the MonkeyWrench library.

The test data is [NegativeScaleTest from Khronos's glTF sample models](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/NegativeScaleTest).

Here's how it looks in CompareLoaders when "NegativeScaleTest/glTF/NegativeScaleTest.gltf" is loaded using jme3-plugins:

NegativeScale

According to the documentation for NegativeScaleTest, this indicates "back-face culling being inappropriately applied to negatively-scaled meshes" plus normal vectors not flipped.

I haven't investigated to see how difficult the solution might be.

codex128:
This PR partially fixes #2087 by "flipping" the face cull mode as the render state is applied to the renderer if the geometry world scale is found to be negative. This does not alter the normal vectors, so shaders still receive backwards normals and models still may not appear quite right (especially with normal maps), but this fixes the face culling issue, at least.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/Material.java
+++ b/jme3-core/src/main/java/com/jme3/material/Material.java
@@ -46,7 +46,6 @@ import com.jme3.renderer.Renderer;
 import com.jme3.renderer.TextureUnitException;
 import com.jme3.renderer.queue.RenderQueue.Bucket;
 import com.jme3.scene.Geometry;
-import com.jme3.scene.Spatial.CullHint;
 import com.jme3.shader.*;
 import com.jme3.texture.Image;
 import com.jme3.texture.Texture;
--- a/jme3-core/src/main/java/com/jme3/material/Material.java
+++ b/jme3-core/src/main/java/com/jme3/material/Material.java
@@ -46,6 +46,7 @@ import com.jme3.renderer.Renderer;
 import com.jme3.renderer.TextureUnitException;
 import com.jme3.renderer.queue.RenderQueue.Bucket;
 import com.jme3.scene.Geometry;
+import com.jme3.scene.Spatial.CullHint;
 import com.jme3.shader.*;
 import com.jme3.texture.Image;
 import com.jme3.texture.Texture;
@@ -890,16 +891,46 @@ public class Material implements CloneableSmartAsset, Cloneable, Savable {
         return type == VarType.BufferObject;
     }
 
-    private void updateRenderState(RenderManager renderManager, Renderer renderer, TechniqueDef techniqueDef) {
+    private void updateRenderState(Geometry geometry, RenderManager renderManager, Renderer renderer, TechniqueDef techniqueDef) {
         if (renderManager.getForcedRenderState() != null) {
-            renderer.applyRenderState(renderManager.getForcedRenderState());
+            if (!renderManager.getForcedRenderState().isFaceCullFlippable() || !isNormalsBackward(geometry.getWorldScale())) {
+                renderer.applyRenderState(renderManager.getForcedRenderState());
+            }
+            else {
+                RenderState rs = renderManager.getForcedRenderState();
+                rs.flipFaceCull();
+                renderer.applyRenderState(rs);
+            }
         } else {
             if (techniqueDef.getRenderState() != null) {
-                renderer.applyRenderState(techniqueDef.getRenderState().copyMergedTo(additionalState, mergedRenderState));
+                techniqueDef.getRenderState().copyMergedTo(additionalState, mergedRenderState);
             } else {
-                renderer.applyRenderState(RenderState.DEFAULT.copyMergedTo(additionalState, mergedRenderState));
+                RenderState.DEFAULT.copyMergedTo(additionalState, mergedRenderState);
+            }
+            if (!mergedRenderState.isFaceCullFlippable() || !isNormalsBackward(geometry.getWorldScale())) {
+                renderer.applyRenderState(mergedRenderState);
+            }
+            else {
+                RenderState rs = mergedRenderState.clone();
+                rs.flipFaceCull();
+                renderer.applyRenderState(rs);
             }
         }
+    }
+    
+    /**
+     * Returns true if the given scalar indicates that normals will be backward.
+     * @param scalar
+     * @return 
+     */
+    private boolean isNormalsBackward(Vector3f scalar) {
+        // count number of negative scalar vector components
+        int n = 0;
+        if (scalar.x < 0) n++;
+        if (scalar.y < 0) n++;
+        if (scalar.z < 0) n++;
+        // an odd number of components means the vectors are backward
+        return n == 1 || n == 3;
     }
     
     /**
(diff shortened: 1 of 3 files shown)


jme3-examples/src/main/java/jme3test/model/TestGltfLoading.java
```diff
@@ -133,7 +133,7 @@ public void simpleInitApp() {
         //loadModel("Models/gltf/manta/scene.gltf", Vector3f.ZERO, 0.2f);
         //loadModel("Models/gltf/bone/scene.gltf", Vector3f.ZERO, 0.1f);
 //        loadModel("Models/gltf/box/box.gltf", Vector3f.ZERO, 1);
-        loadModel("Models/gltf/duck/Duck.gltf", new Vector3f(0, -1, 0), 1);
+        loadModel("Models/gltf/duck/Duck.gltf", new Vector3f(0, 1, 0), 1);
```
stephengold, line 136: What's the purpose of this change?

stephengold:
Did you test these changes on the "NegativeScaleTest" model?

jme3-examples/src/main/java/jme3test/model/TestGltfLoading.java
codex128: While testing scale, I had to move the model to view it properly.

codex128:
Oops, no, I didn't. Tested it out just now, and yes, the changes fixed the test model so there are no Xs anymore. I think it may have also fixed the shading on the spheres, but I couldn't tell for certain.

stephengold (approved):
Looks good to me.

jme3-core/src/main/java/com/jme3/material/Material.java
```diff
@@ -890,18 +890,51 @@ private boolean isBO(final VarType type) {
...
 
-    private void updateRenderState(RenderManager renderManager, Renderer renderer, TechniqueDef techniqueDef) {
+    private void updateRenderState(Geometry geometry, RenderManager renderManager, Renderer renderer, TechniqueDef techniqueDef) {
         if (renderManager.getForcedRenderState() != null) {
-            renderer.applyRenderState(renderManager.getForcedRenderState());
+            if (!renderManager.getForcedRenderState().isFaceCullFlippable() || !isNormalsBackward(geometry.getWorldScale())) {
+                renderer.applyRenderState(renderManager.getForcedRenderState());
+            }
+            else {
+                RenderState flipped = renderManager.getForcedRenderState().clone();
```
riccardobl, line 899: This creates a new object every time updateRenderState is called, that is every frame.
We need a different solution to avoid overloading the garbage collector.

jme3-core/src/main/java/com/jme3/material/Material.java
```diff
@@ -890,18 +890,51 @@ private boolean isBO(final VarType type) {
...
             } else {
-                renderer.applyRenderState(RenderState.DEFAULT.copyMergedTo(additionalState, mergedRenderState));
+                RenderState.DEFAULT.copyMergedTo(additionalState, mergedRenderState);
+            }
+            // test if the face cull mode should be flipped before render
+            if (!mergedRenderState.isFaceCullFlippable() || !isNormalsBackward(geometry.getWorldScale())) {
+                renderer.applyRenderState(mergedRenderState);
+            }
+            else {
+                RenderState flipped = mergedRenderState.clone();
```
riccardobl, line 915: Same as the other clone

riccardobl (changes requested):

riccardobl (approved):

stephengold:
Hey @codex128 ! Thanks for your contribution to JMonkeyEngine.

Trass3r:
According to git bisect this broke our rendering as soon as shadows are enabled.
There's an incorrect glDepthMask(false) for some of the rendered opaque geometry, I guess from the PreShadow material.
Can anyone confirm?

codex128:
My first guess would be that `RenderState#set` is problematic, since my change uses `set`, while the previous code did not. Are you using a shadow filter or renderer?

Trass3r:
Yeah from my debugging might be ForcedRenderState related.

codex128:
I've opened #2130 which should fix this issue.

stephengold:
Thank you @codex128 for following up on issue #2129.

JohnLKkk:
I just found a problem locally. This modification seems to have caused an exception with ShadowFilter (running TestDirectionalLightShadow.java results in a black screen). I found this issue at commitID=5975b3f791d18fae311a10e209b3ed40b5e5c60f.

JohnLKkk:
`
private void updateRenderState(Geometry geometry, RenderManager renderManager, Renderer renderer, TechniqueDef techniqueDef) {
        if (renderManager.getForcedRenderState() != null) {
            mergedRenderState.set(renderManager.getForcedRenderState());
        } else if (techniqueDef.getRenderState() != null) {
            // copyMergedTo writes to mergedRenderState
            techniqueDef.getRenderState().copyMergedTo(additionalState, mergedRenderState);
        } else {
            RenderState.DEFAULT.copyMergedTo(additionalState, mergedRenderState);
        }
        // test if the face cull mode should be flipped before render
        if (mergedRenderState.isFaceCullFlippable() && isNormalsBackward(geometry.getWorldScale())) {
            mergedRenderState.flipFaceCull();
        }
        renderer.applyRenderState(mergedRenderState);
    }
`
It should be changed to:
`
private void updateRenderState(Geometry geometry, RenderManager renderManager, Renderer renderer, TechniqueDef techniqueDef) {
        if (renderManager.getForcedRenderState() != null) {
            mergedRenderState.set(renderManager.getForcedRenderState());
        } else if (techniqueDef.getRenderState() != null) {
            // copyMergedTo writes to mergedRenderState
            mergedRenderState.set(techniqueDef.getRenderState().copyMergedTo(additionalState, mergedRenderState));
        } else {
            mergedRenderState.set(RenderState.DEFAULT.copyMergedTo(additionalState, mergedRenderState));
        }
        // test if the face cull mode should be flipped before render
        if (mergedRenderState.isFaceCullFlippable() && isNormalsBackward(geometry.getWorldScale())) {
            mergedRenderState.flipFaceCull();
        }
        renderer.applyRenderState(mergedRenderState);
    }
`

stephengold:
@JohnLKKK Thank you for reporting the issue! Two things:

1. When quoting multiple lines of code at GitHub, please use triple backquotes so line breaks and indentation will be preserved, like so:
```java
public static int main(String...args) {
    return 0;
}
```
2. Please open a new PR with your fix so it can be reviewed and integrated ASAP.

codex128:
I believe this problem was fixed in #2130.

JohnLKkk:
> @JohnLKkk Thank you for reporting the issue! Two things:
> 
> 1. When quoting multiple lines of code at GitHub, please use [triple backticks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) so line breaks and indentation will be preserved, like so:
> 
> ```java
> public static int main(String... args) {
>     return 0;
> }
> ```
> 
> 2. Please open a new PR with your fix so it can be reviewed and integrated ASAP.

Thank you for your guidance.

PR #2095, opened 2023-10-12
Prevent failed LWJGL 2 init to fail again

### LINKED ISSUE #2077: LWJGL 2 threading problem with failed init
Take a look what happens in:
https://hub.jmonkeyengine.org/t/help-with-jme-v3-6-1-stable-sdk3-assetbrowser-alpha/47140

Obviously the initialization fails and there is very little that we can do about that. But if we can handle that error more gracefully, it would be great. I'm talking about the failed synchronization that happens here:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-lwjgl/src/main/java/com/jme3/system/lwjgl/LwjglAbstractDisplay.java#L135

Need to create an error from similar source seen on the hub post. And see if adding sync block is enough or does that result in dead lock etc.

tonihele:
I raised the same exact error at the same exact code line than the OP in the issue. That resulted in getting 2 errors. By properly using synchronized block, the other (coding error) was gone and I observed no dead locks.

The fix in TestChooser is for fixing these failed launches preventing the Chooser from shutting down properly. Failed launches cause the thread to loop in sleep waiting for some kind of confirmation I guess. Wait to stay alive to call the GC. They will stay there forever. Using `shutdownNow` interrupts the threads hanging and properly shuts them down. Note that this doesn't change the behavior otherwise. If you have a test running and close the TestChooser, the test doesn't close. Just the chooser exits and stops waiting. Also `shutdownNow` doesn't block, neither did the `shutdown`.

Resolves #2077

### DIFF
--- a/jme3-examples/src/main/java/jme3test/TestChooser.java
+++ b/jme3-examples/src/main/java/jme3test/TestChooser.java
@@ -72,14 +72,13 @@ import javax.swing.event.DocumentListener;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
 
-
 /**
  * Class with a main method that displays a dialog to choose any jME demo to be
  * started.
  */
 public class TestChooser extends JFrame {
-    private static final Logger logger = Logger.getLogger(TestChooser.class
-            .getName());
+
+    private static final Logger logger = Logger.getLogger(TestChooser.class.getName());
 
     private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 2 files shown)


riccardobl:
> Auto-format strikes again.

Mh it seems having it disabled from the settings isn't enough to prevent it from running on forks. 
I've pushed a commit to disable it entirely.

riccardobl (approved):

PR #2100, opened 2023-10-13
Disable shadows for sky

riccardobl:
Set shadow mode to Off instead of Inherit for sky boxes to prevent them from messing with the shadow renderer if attached to a rootNode that has shadows enabled.


stephengold:
This change is long overdue.
Thank you, @riccardobl !

PR #2101, opened 2023-10-13
Expand support for GL ES / WebGL compatible formats.

riccardobl:
This pr expand the compatible format list for gl es 3 and adds ETC2 support.
The DDS loader can now load ETC DDSs converted with [compressonator](https://gpuopen.com/compressonator/) (basically old header dds with custom pixel formats).

limitations:
- *ETC flipping is not supported*: i couldn't figure out how to flip etc compressed textures without decompressing them first, even though it is probably possible (similar to dxt flipping). I don't this this is very important, since compressed textures are usually considered in their final form and they can be flipped during compression if needed, also we didn't even had dxt flipping for a long time. I plan to open an issue for this and hopefully someone that is knowledgeable about etc formats can contribute with the missing code one day.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/texture/plugins/DXTFlipper.java
+++ b/jme3-core/src/plugins/java/com/jme3/texture/plugins/DXTFlipper.java
@@ -36,6 +36,7 @@ import com.jme3.texture.Image.Format;
 import com.jme3.util.BufferUtils;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.util.logging.Logger;
 
 /**
  * DXTFlipper is a utility class used to flip along Y axis DXT compressed textures.
@@ -43,6 +44,7 @@ import java.nio.ByteOrder;
  * @author Kirill Vainer
  */
 public class DXTFlipper {
+    private static final Logger logger = Logger.getLogger(DXTFlipper.class.getName());
 
     private static final ByteBuffer bb = ByteBuffer.allocate(8);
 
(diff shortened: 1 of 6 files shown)


stephengold:
The new file (ETCFlipper.java) needs some javadoc describing its purpose.

jme3-core/src/plugins/java/com/jme3/texture/plugins/DXTFlipper.java
```diff
@@ -202,7 +204,11 @@ private static void flipDXT1orDXTA3Block(byte[] block, int h){
         }
     }
 
-    public static ByteBuffer flipDXT(ByteBuffer img, int w, int h, Format format){
+    public static ByteBuffer flipDXT(ByteBuffer img, int w, int h, Format format) {
+        if (format == Format.ETC1 || format == Format.ETC2 || format == Format.ETC2_ALPHA1) {
+            logger.warning("This is not a DXT format, but ETC. Use flipETC instead.");
+            return ETCFlipper.flipETC(img, w, h, format);
```
scenemax3d, line 210: Why call ETCFlipper.flipETC(img, w, h, format); if it doesn't do anything? why not just return img?

riccardobl: > ETC flipping is not supported: i couldn't figure out how to flip etc compressed textures without decompressing them first, even though it is probably possible (similar to dxt flipping). I don't this this is very important, since compressed textures are usually considered in their final form and they can be flipped during compression if needed, also we didn't even had dxt flipping for a long time. I plan to open an issue for this and hopefully someone that is knowledgeable about etc formats can contribute with the missing code one day.


Basically i don't know how to implement compressed etc flipping, so i wrote all the supporting code in the hope that someone can go into ETCFlipper some day and write the missing piece without the need to understand and modify the dds loader code that has some jme specific logic inside. 
We could open a bounty for it.

PR #2106, opened 2023-10-14
Add userdata loader to gltf

riccardobl:
Adds a default user data loader from Simsilica's JmeConvert to the GLTF loader to make it load custom properties/extras as user data.

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/CustomContentManager.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/CustomContentManager.java
@@ -47,6 +47,8 @@ import java.util.logging.Logger;
  * Created by Nehon on 20/08/2017.
  */
 public class CustomContentManager {
+    private final static ThreadLocal<ExtrasLoader> defaultExtraLoader = ThreadLocal
+            .withInitial(() -> new UserDataLoader());
 
     private final static Logger logger = Logger.getLogger(CustomContentManager.class.getName());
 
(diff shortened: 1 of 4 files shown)


jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/CustomContentManager.java
```diff
@@ -130,14 +130,16 @@ private <T> T readExtension(String name, JsonElement el, T input) throws AssetLo
...
     private <T> T readExtras(String name, JsonElement el, T input) throws AssetLoadException {
-        if (key == null) {
-            return input;
-        }
         ExtrasLoader loader;
-        loader = key.getExtrasLoader();
-        if (loader == null) {
-            return input;
+        if (key == null) {
+            loader = new UserDataLoader();
```
codex128, line 135: I think a static final UserDataLoader instance should be used here to avoid instantiating a new instance each time.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfModelKey.java
```diff
@@ -53,7 +53,7 @@ public class GltfModelKey extends ModelKey {
 
     private Map<String, MaterialAdapter> materialAdapters = new HashMap<>();
     private static Map<String, ExtensionLoader> extensionLoaders = new HashMap<>();
-    private ExtrasLoader extrasLoader;
+    private ExtrasLoader extrasLoader = new UserDataLoader();
```
codex128, line 56: I think a static final UserDataLoader instance should be used here instead.

codex128 (approved):
Just a couple small adjustments I think should be made if possible. Otherwise, I think these changes are fine.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/CustomContentManager.java
riccardobl: I refactored it to have a behavior similar to extension registration in  #2107.
The default extras loader is now configurable and instanced only once for each  GLTFLoader instance, making it also thread safe by design.

PR #2107, opened 2023-10-14
Make default  gltf extensions  registrable statically

riccardobl:
Makes default extensions registrable statically, to allows specific modules (think about (j)bullet, minie etc) to statically register their own extension in the gltf loader. 
Similar to what is done in other contexts (eg. blender) that allow to extend their gltf importer/exporter.


capdevon:
It seems that this change is introducing a problem with multithreading that has been solved before.

[Made Gltf extension loaders non-static to avoid concurrency](
https://github.com/jMonkeyEngine/jmonkeyengine/pull/1886)

capdevon:
OK, thank you for clarifying my doubts.

stephengold:
Unless there's additional substantive discussion, I plan to integrate this PR in about 36 hours.

PR #2109, opened 2023-10-14
Fixes GLTF Import of Morph Animations

### LINKED ISSUE #2092: jme3-plugins imports MorphStressTest incorrectly
I noticed a defect in `GltfLoader` while developing the MonkeyWrench library.

The test data is [MorphStressTest from Khronos's glTF sample models](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MorphStressTest).

When the test data is loaded using `GltfLoader`, the resulting `AnimComposer` contains 3 clips, but none of the clips contains any animation tracks. Without tracks, none of the animation clips do anything.

The expected behavior is that each clip should contain 2 morph tracks. Loading a clip should cause the heights of the colored bars to change as described in the model documentation.

codex128:
This PR fixes #2092 where morph animation clips were loaded without tracks. The problem was occurs because the animation channels were assigned to a node instead of a geometry in the gltf file, which excluded those channels from being properly loaded.

My solution is if the target spatial is a node, traverse over that node's children and apply the morphs to the geometries found, otherwise apply morphs as before. This allows the engine to pass the [Morph Stress Test](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MorphStressTest) where it couldn't before.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/model/TestGltfLoading.java
+++ b/jme3-examples/src/main/java/jme3test/model/TestGltfLoading.java
@@ -31,6 +31,7 @@
  */
 package jme3test.model;
 
+import com.jme3.anim.AnimClip;
 import com.jme3.anim.AnimComposer;
 import com.jme3.anim.SkinningControl;
 import com.jme3.app.*;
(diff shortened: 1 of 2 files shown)


riccardobl:
Looks great. Thanks!

PR #2111, opened 2023-10-15
Fix #1997 (shadows on hardware-accelerated morph animations)

tonihele:
Yep, just some errors on the shaders.

I modified the TestGltfMorph to have the shadows instead of duplicating it. I hope this is ok. I can change it if not of course.

Resolves #1997

PR #2112, opened 2023-10-15
Prevent updateShadowCamera from corrupting shadowCam matrix.

riccardobl:
Prevent updateShadowCamera from corrupting shadowCam matrix when there isn't any visible shadow caster in scene. 
Discovered in #2111

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
@@ -1337,32 +1337,33 @@ public final class GLRenderer implements Renderer {
         switch (uniform.getVarType()) {
             case Float:
                 Float f = (Float) uniform.getValue();
-                assert isValidNumber(f) : "Invalid float value " + f;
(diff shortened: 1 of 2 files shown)


stephengold:
Shouldn't the changes to asserts and diagnostics have been a separate PR?

Also, what's the significance of the hacktoberfest labels? (Just curious.)

riccardobl:
> Shouldn't the changes to asserts and diagnostics have been a separate PR?

Well i wouldn't dedicate a PR to this small change that i made just to debug this issue and I merged the pr without squash to keep the commits separated, so it shouldn't cause any issue for the git history or if we need to revert it.

> Also, what's the significance of the hacktoberfest labels? (Just curious.)

Two years ago, I assumed the responsibility of moderating [Hacktoberfest](https://hacktoberfest.com/) for our repositories. This year, in order for a pull request to be considered valid for the quota, it must either be approved, merged, or have a specific label applied for a continuous 7-day period. As we are in the second half of October, I've added the label to mark valid pull requests as soon as they are submitted, avoiding the need for rushed approvals in case the submitter is participating.
I added the label to this PR by mistake.

stephengold:
I see. Thanks for the answers.

PR #2113, opened 2023-10-15
Geometry constructor with material.

llzen44:
Convenient Geometry constructor with material, in order to avoid systematic subsequent setMaterial setter.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Geometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Geometry.java
@@ -144,13 +144,7 @@ public class Geometry extends Spatial {
      * @param material The material for this geometry
      */
     public Geometry(String name, Mesh mesh, Material material) {
-        this(name);
-
-        if (mesh == null) {
-            throw new IllegalArgumentException("mesh cannot be null");
-        }
-
-        this.mesh = mesh;
+        this(name, mesh);
         setMaterial(material);
     }
 
--- a/jme3-core/src/main/java/com/jme3/scene/Geometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Geometry.java
@@ -136,6 +136,24 @@ public class Geometry extends Spatial {
         this.mesh = mesh;
     }
 
+    /**
+     * Create a geometry node with mesh data and material.
+     *
+     * @param name The name of this geometry
+     * @param mesh The mesh data for this geometry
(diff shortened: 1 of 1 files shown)


riccardobl:
I have reservations about this PR. Since setMaterial() is a method of Spatial, it does make sense to require developers to call it explicitly for each type of spatial. Introducing a new constructor for Geometry makes it convenient, but it also introduces inconsistency.

I'm not entirely opposed on principle, but I'd prefer to hear other opinions.

jme3-core/src/main/java/com/jme3/scene/Geometry.java
```diff
@@ -136,6 +136,24 @@ public Geometry(String name, Mesh mesh) {
...
     }
 
+    /**
+     * Create a geometry node with mesh data and material.
+     *
+     * @param name The name of this geometry
+     * @param mesh The mesh data for this geometry
+     * @param material The material for this geometry
+     */
+    public Geometry(String name, Mesh mesh, Material material) {
```
riccardobl, line 146: This new constructor should call 
` public Geometry(String name, Mesh mesh) `
instead of copying its code.

riccardobl (changes requested):

jme3-core/src/main/java/com/jme3/scene/Geometry.java
llzen44: Totally right about constructor code duplication. To be changed
Regarding setMaterial as method of Spatial, it is true. Idea here is to have a fully functional Geometry instance , as soon as the constructor is called.
Let's hear from others.

pspeed42:
Spatial.setMaterial() is actually a surprise for almost everyone who uses it the first time.  Geometry is the only thing that requires a material.  Geometry is the only thing that accepts a material. Spatial.setMaterial() actually finds the child geometries and sets that material on all of them, sensible or not.  (It's a method I've thought should be removed since the first time I saw it.)

It is really weird that there is no way to create a functional geometry in one call.  I've always thought so.

But... "doing anything about it" has always been tempered with how ugly the code looks to put it on one giant line: because there is no way to trivially create a material.  It's always like half a screen-width of code.

I don't mind Geometry taking material as a parameter.  It's otherwise weird that we require setting the name (99% of the time useless) but not the material.

riccardobl (approved):

PR #2115, opened 2023-10-16
Increase Max Defines in DefineList (Attempt 2 since other PR may be bugged?)

yaRnMcDonuts:
Following up on this issue: https://hub.jmonkeyengine.org/t/optimize-terrain-shaders-for-tri-planar-mode-per-layer-remove-fog-defines-to-make-room/46816/32


yaRnMcDonuts:
Not successful sadly... 

Although I suspect this was not the best way to try duplicating my PR, since Github made reset and then reuse my same master branch that the bugged PR was derrived from. So if my JME fork is bugged, then I don't think this did anything to work around that... but I also am taking wild guesses here since we're in some strange and uncharted territory considering the build passes locally for us still..

@stephengold let me know if you have any luck duplicating the PR an your own branch (if you still have time to do so of course) and maybe that will prove some better results. 

Otherwise (unless another jme user can run this build and prove that the test doesn't pass locally on their device) then I am going to conclude that this may be bugged in some way that is better investigated within its own PR, and I will temporarily comment-out the failling tests unti it can be fixed or preferably a new replacement test is created (since the failing test is not actually indicative of any bugs in the  PR's code, and it also has been thoroughly discussed that the test is badly written in the first place and should be replaced by a better sorting test). And in this case, I will also open another issue related to replacing that test, this way the PR can go forward without the test hindering anything, but we can still work on replacing that test in the future.

Does anyone object to this idea, and if so do you have any other proposed solutions? (also, if anyone does manage to build this branch locally and trigger a failing test, then please let me know as well, as that will mean I am wrong in many of my assumptions. But for now, 2 different users have built this fork sucessfuly locally with no failing test, so unless at least 1 other user can produce a locally failling test then the odds are in favor that github is being buggy)

yaRnMcDonuts:
Oops it looks like you were already ontop of htis, and your PR passed! Yay! I will delete this PR and my other one.

Issue #2114, opened 2023-10-15
LodGenerator: null in returning VertexBuffer.


MelVimL:
During prepeartions for Issue #751 I found a Bug.
The result array can have "null" elements in it. The LodGenerator counts the successfully computed lods(`bakedLods`), failed Lods are stored as **null** in `lods.` 
In the end the lods coppied to the result:

`System.arraycopy(lods, 0, bakedLods, 0, numBakedLods);
return bakedLods;`

The Bug happens when lods in between failed so an null get copied.

![grafik](https://github.com/jMonkeyEngine/jmonkeyengine/assets/51533828/0feb551a-5a87-450f-9eb8-bb61ce796740)

How to reproduce:

```
   float[] REDUCTION_VALUES = { 0.5f, 0.55f, 0.6f, 0.65f, 0.7f, 0.75f, 0.80f };
   @Test
   public void testSphereReductionCollapsCost() {
        LodGenerator lod = new LodGenerator(sphere()); // sphere() -> new Sphere(12, 12, 1, false, false);
        VertexBuffer[] buffer = lod.computeLods(LodGenerator.TriangleReductionMethod.COLLAPSE_COST, REDUCTION_VALUES);
        
        int[] expected = { ... };
        int[] actual = getBufferSizes(buffer);  // Generates a List of buffer sizes.
        assertArrayEquals(expected, actual );

```

riccardobl:
Are you working on submitting a patch for this?
If i understand the issue correctly, we just need to replace System.arraycopy with a manual copy that filters out nulls. Is that right?

MelVimL:
Yes but I work currently on a refactoring of the LodGenerator. 
It looks like this bug get fixed by it but I am not sure how long it will take me to finish it. 

But if its easier I can create a pull request with the fix amd tests now I would do it.

riccardobl:
Thank you for the patch @MelVimL

PR #2117, opened 2023-10-16
Fixes 2114

MelVimL:
This patch fixes 2114 and adds simple tests.

Removes the check if some lod failed. 
The performance gain is negligible, most of the time it copies <10 VertexBuffer refs unless I overlook something.

### DIFF
--- a/jme3-core/src/tools/java/jme3tools/optimize/LodGenerator.java
+++ b/jme3-core/src/tools/java/jme3tools/optimize/LodGenerator.java
@@ -576,13 +576,22 @@ public class LodGenerator {
                 numBakedLods++;
             }
         }
-        if (numBakedLods <= lodCount) {
-            VertexBuffer[] bakedLods = new VertexBuffer[numBakedLods];
-            System.arraycopy(lods, 0, bakedLods, 0, numBakedLods);
-            return bakedLods;
-        } else {
-            return lods;
+
+        return cleanBuffer(lods, numBakedLods);
+    }
+
+    private VertexBuffer[] cleanBuffer(VertexBuffer[] lods, int lodLevel) {
+        int index = 0;
+        VertexBuffer[] result = new VertexBuffer[lodLevel];
+
+        for (VertexBuffer lod : lods) {
+            if (lod != null) {
+                result[index] = lod;
+                index++;
+            }
         }
+
+        return result;
     }
 
     /**
--- a/jme3-core/src/test/java/com/jme3/tools/LodGeneratorTest.java
+++ b/jme3-core/src/test/java/com/jme3/tools/LodGeneratorTest.java
@@ -1,0 +1,131 @@
+package com.jme3.tools;
+
+import static org.junit.Assert.assertArrayEquals;
+
+import java.net.URL;
+import java.util.ArrayList;
+
+import org.junit.Test;
+
+import com.jme3.asset.AssetManager;
+import com.jme3.scene.Geometry;
+import com.jme3.scene.Mesh;
+import com.jme3.scene.Node;
+import com.jme3.scene.Spatial;
+import com.jme3.scene.VertexBuffer;
+import com.jme3.scene.shape.Box;
+import com.jme3.scene.shape.Sphere;
+import com.jme3.system.JmeSystem;
+
+import java.util.List;
+import jme3tools.optimize.LodGenerator;
+
+public class LodGeneratorTest {
+    URL assetCfgUrl = JmeSystem.getPlatformAssetConfigURL();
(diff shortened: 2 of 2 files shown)


MelVimL:
Fixes #2114

riccardobl (changes requested):
The following unit tests are currently failing in this PR:
    com.jme3.tools.LodGeneratorTest > testMonkeyReductionProportional
    com.jme3.tools.LodGeneratorTest > testMonkeyReductionConstant
    com.jme3.tools.LodGeneratorTest > testMonkeyReductionCollapsCost

This could mean that the tests need to be updated or that the PR introduces issues.

MelVimL:
@riccardobl I removed the Test decorator. I guess it is an issue with loading the Jaime.j3o asset while building.

stephengold:
@riccardobl Do you support disabling the failing junit tests?

MelVimL:
@stephengold @riccardobl I added javadoc and the licenses.

riccardobl:
These tests were added in this pr correct? In that case i have nothing against disabling them if there are technical challenges in making them work.
I think this can be merged

riccardobl (approved):

MelVimL:
@riccardobl 
"These tests were added in this pr correct?":
Yes.

jme3-core/src/tools/java/jme3tools/optimize/LodGenerator.java
```diff
@@ -576,13 +576,22 @@ public VertexBuffer[] computeLods(TriangleReductionMethod reductionMethod, float
...
-            VertexBuffer[] bakedLods = new VertexBuffer[numBakedLods];
-            System.arraycopy(lods, 0, bakedLods, 0, numBakedLods);
-            return bakedLods;
-        } else {
-            return lods;
+
+        return cleanBuffer(lods, numBakedLods);
+    }
+
+    private VertexBuffer[] cleanBuffer(VertexBuffer[] lods, int lodLevel) {
```
capdevon, line 583: Hi guys, I have a doubt, I have not tried the code so correct me if I am wrong. I have a feeling that this method moves any null values to the end of the array.

 If the old code generated this output `[x, x, null, null, x ,x]`, this new code should produce the following output `[x, x, x, x, null, null]`. 
Is this the purpose of the change?

MelVimL: @capdevon  Jup `numBakedLods` is the number of successfully build lods. Maybe changing the lodLevel varibale @riccardobl ?

riccardobl: maybe rename it to numComputedLodLevels ?

capdevon: maybe you could do it this way:


```java
/**
 * @param lods ...
 * @param numBakedLods the num of computed lod levels
 */
private VertexBuffer[] cleanBuffer(VertexBuffer[] lods, int numBakedLods) {
	...
}
```

MelVimL: Documentation is a bit overkill for a method used once which is located two lines under the usage. BUT the name change makes sense.

capdevon: With the little javadoc I suggested, the changes introduced by your PR would be immediately clear to anyone, without having to read the whole thread or analyze all the complex code in the class, and no one would have any doubts about your solution. I don't have a strong position on this point, since it is a private method, but you would certainly leave important info for those who might come across the analysis of this class again in the future ;)

MelVimL: In gerneraly yes. But in this case would be the javadoc as big as the function. You have to read the code anyways, it could be that I've had used the `lodCount ` instead of the `numBakedLods ` and the doc woundn't made anything easier. 
My gut feeling tells me that in this case it would contributes more noise to the code which is as dense as the function itself.  

In addtion I am working on a refactoring of the LodGenerator and currently don't need the cleanBuffer function. Maybe I need this in the futur but it seams unlikly. On the other hand I have the feeling that this refactoring will be hard to get through if I look at the change/comment ratio :)

In the end it comes down to tast. So @riccardobl  javadoc or no javadoc.

riccardobl: I don't think there the need to add javadoc, the new variable name is enough. :+1:

capdevon: OK, everything looks fine to me now.  Thank you guys.

PR #2128, opened 2023-10-23
Solve issue #2127 (excessive memory use while loading GLB)

### LINKED ISSUE #2127: GLBLoader load much more into memory than GLTF equivalent.
related to post:
- https://hub.jmonkeyengine.org/t/enhancements-to-the-jme3-rendering-engine/47145/132
and:
- https://github.com/jMonkeyEngine/jmonkeyengine/issues/2125

Tonihele were able to load Bistro scene as GLTF, but GLB caused to exceed 32 GB memory.

Worth to investigate.

As Tonihele suggested, issue might be related to this file(that extends working GLTF):
https://github.com/jMonkeyEngine/jmonkeyengine/blob/c9d10eb6ffcebbf18a7709de15be1eac67de9c1f/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GlbLoader.java

tonihele:
Ok, GLB with the textures included easily causes LWJGL to run out of memory. As we might upload the same texture many times (we also read it many times...). 32Gb of RAM is nothing then. Also slowdowns and if one is into geometry batching.. no can do.

So this caches all the materials read to re-use them. Note that this affects both GLTF and GLB.

Resolves #2127

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -754,12 +754,17 @@ public class GltfLoader implements AssetLoader {
         Integer textureIndex = getAsInteger(texture, "index");
         assertNotNull(textureIndex, "Texture has no index");
         assertNotNull(textures, "There are no textures, yet one is referenced by a material");
+
+        Texture2D texture2d = fetchFromCache("textures", textureIndex, Texture2D.class);
+        if (texture2d != null) {
+            return texture2d;
+        }
         
         JsonObject textureData = textures.get(textureIndex).getAsJsonObject();
         Integer sourceIndex = getAsInteger(textureData, "source");
         Integer samplerIndex = getAsInteger(textureData, "sampler");
 
-        Texture2D texture2d = readImage(sourceIndex, flip);
+        texture2d = readImage(sourceIndex, flip);
 
         if (samplerIndex != null) {
             texture2d = readSampler(samplerIndex, texture2d);
@@ -768,6 +773,8 @@ public class GltfLoader implements AssetLoader {
         }
 
         texture2d = customContentManager.readExtensionAndExtras("texture", texture, texture2d);
+
+        addToCache("textures", textureIndex, texture2d, textures.size());
 
         return texture2d;
     }
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -470,7 +470,7 @@ public class GltfLoader implements AssetLoader {
                     geom.setMaterial(defaultMat);
                 } else {
                     useNormalsFlag = false;
-                    geom.setMaterial(readMaterial(materialIndex));
+                    geom.setMaterial(getMaterial(materialIndex));
                     if (geom.getMaterial().getAdditionalRenderState().getBlendMode()
                             == RenderState.BlendMode.Alpha) {
                         // Alpha blending is enabled for this material. Let's place the geom in the transparent bucket.
@@ -614,6 +614,21 @@ public class GltfLoader implements AssetLoader {
             throw new AssetLoadException("Buffer " + bufferIndex + " has no uri");
         }
         return data;
+    }
+
+    public Material getMaterial(int materialIndex) throws IOException {
+
+        // Get from cache
+        Material material = fetchFromCache("Material", materialIndex, Material.class);
+        if (material != null) {
+            return material;
+        }
+
+        material = readMaterial(materialIndex);
+
+        addToCache("Material", materialIndex, material, materials.size());
+
+        return material;
     }
 
     public Material readMaterial(int materialIndex) throws IOException {
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -621,7 +621,7 @@ public class GltfLoader implements AssetLoader {
 
         Material material = fetchFromCache("materials", materialIndex, Material.class);
         if (material != null) {
-            return material;
+            return material.clone();
         }
 
         JsonObject matData = materials.get(materialIndex).getAsJsonObject();
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -619,14 +619,14 @@ public class GltfLoader implements AssetLoader {
     public Material getMaterial(int materialIndex) throws IOException {
 
         // Get from cache
-        Material material = fetchFromCache("Material", materialIndex, Material.class);
+        Material material = fetchFromCache("materials", materialIndex, Material.class);
         if (material != null) {
             return material;
         }
 
         material = readMaterial(materialIndex);
 
-        addToCache("Material", materialIndex, material, materials.size());
+        addToCache("materials", materialIndex, material, materials.size());
 
         return material;
     }
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
@@ -470,7 +470,7 @@ public class GltfLoader implements AssetLoader {
                     geom.setMaterial(defaultMat);
                 } else {
                     useNormalsFlag = false;
-                    geom.setMaterial(getMaterial(materialIndex));
+                    geom.setMaterial(readMaterial(materialIndex));
                     if (geom.getMaterial().getAdditionalRenderState().getBlendMode()
                             == RenderState.BlendMode.Alpha) {
                         // Alpha blending is enabled for this material. Let's place the geom in the transparent bucket.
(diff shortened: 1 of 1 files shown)


jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
```diff
@@ -605,9 +605,9 @@ protected byte[] getBytes(int bufferIndex, String uri, Integer bufferLength) thr
                 BinDataKey key = new BinDataKey(info.getKey().getFolder() + decoded);
                 InputStream input = (InputStream) info.getManager().loadAsset(key);
                 data = new byte[bufferLength];
-                DataInputStream dataStream = new DataInputStream(input);
-                dataStream.readFully(data);
-                dataStream.close();
+                try (DataInputStream dataStream = new DataInputStream(input)) {
+                    dataStream.readFully(data);
```
oxplay2, line 609: if i see correctly dataStream.close(); before but now removed.

tonihele: Yeah, I removed it. If there was ever an exception. It was never called. So better to use try-with resources.

oxplay2: ok :) so its not required, just wanted to be sure.

tonihele:
Hmm, maybe I'll change the cache to the material read method. As it seems that those others do the same. They cache and no-one needs to know.

oxplay2:
yes, seems like it will be better. 

cache is per load as i see, then cleared.

tonihele:
One question though. Should we return material.clone() from the cache? Because a geometry might have vertex colors set and that would affect the material. Returning a clone still fixes the original issue...

tonihele:
> Wait a moment.
> 
> But then why GLTF had no 32 GB memory issue?
> 
> Both GLB and GLTF use readMaterial right?
> While only GLB had issue with multipled materials.

GLTF/GLB read materials multiple times to create a material. This is fine, until there is also a texture embedded. Like real binary, not a file reference. Then this texture is read to the `assetManager `with `loadAssetFromStream `which doesn't cache. This is the source of the problem. So there are going to be so much textures in the memory.

It is easier to cache the material in this case in my opinion. It doesn't interfere with non embedded textures (normal GLTF). It is less hassle in this case too. I didnt measure the perfomance of our Material cloning vs reading it all the time again from bytes (latter looks like more expensive operation).

oxplay2:
Ok, so i understand that GLTF have "sourceindex" that can know what material texture index it is, while GLB do not?

tonihele:
> Ok, so i understand that GLTF have "sourceindex" that can know what material texture index it is, while GLB do not?

GLB/GLTF are identical. But GLB can contain embedded binary textures. GLTF contains only URLs to textures. Both have indices for textures and materials.

I just guess that original author didn't realize that the assetManager used like that doesn't cache. Creates a new texture with the same key...

oxplay2:
```
        if (uri == null) {
            assertNotNull(bufferView, "Image " + sourceIndex + " should either have an uri or a bufferView");
            assertNotNull(mimeType, "Image " + sourceIndex + " should have a mimeType");
            byte[] data = (byte[]) readBuffer(bufferView, 0, -1, null, 1, VertexBuffer.Format.Byte);
            String extension = mimeType.split("/")[1];
            TextureKey key = new TextureKey("image" + sourceIndex + "." + extension, flip);
            result = (Texture2D) info.getManager().loadAssetFromStream(key, new ByteArrayInputStream(data));

        } else if (uri.startsWith("data:")) {
            // base64 encoded image
            String[] uriInfo = uri.split(",");
            byte[] data = Base64.getDecoder().decode(uriInfo[1]);
            String headerInfo = uriInfo[0].split(";")[0];
            String extension = headerInfo.split("/")[1];
            TextureKey key = new TextureKey("image" + sourceIndex + "." + extension, flip);
            result = (Texture2D) info.getManager().loadAssetFromStream(key, new ByteArrayInputStream(data));
        } else {
            // external file image
            String decoded = decodeUri(uri);
            TextureKey key = new TextureKey(info.getKey().getFolder() + decoded, flip);
            Texture tex = info.getManager().loadTexture(key);
            result = (Texture2D) tex;
        }
```

Ok i think i understand.

So the main reason why GLTF were working is that "loadTexture" cache, while every other case like "loadAssetFromStream" do not cache. 

Then why not make GLB work same as GLTF by enabling assetmanager cache?

tonihele:
Yep, you got it.

> Then why not make GLB work same as GLTF by enabling assetmanager cache?
> is it possible to cache in assetManager by name or anything or only uri?

I guess it would be doable by:
```
if(info.getManager(getFromCache(key)) != null) {
...}
texture = (Texture2D) info.getManager().loadAssetFromStream(key, new ByteArrayInputStream(data));
info.getManager().addToCache(key, texture);
```
But I don't know if it is the same cache. I think normally they go to the weak cache, but this might not be it. I am not an expert on this matter.

oxplay2:
Im also unsure if it will be same cache.

Then i guess we should ask on HUB maybe :) or mention someone here to help.

tonihele:
> Im also unsure if it will be same cache. (well it should, but we need be sure, yes)
> 
> Then i guess we should ask on HUB maybe :) or mention someone here to help.

Well, I read the code. It is the same cache. It is governed by the `assetKey ``cacheType `property, which for `textureKey`s defaults to `WeakRefCloneAssetCache`.

tonihele:
So I guess that is all doable. My personal preference is the material cloning though. But I can go either way. Someone just make the call then :)

oxplay2:
i rethink this a little.

Lets say we have 2 Models.

AssetManager cache textures per "game run" not per "model load"

So lets say Both of them use "MyTexture.png", but one as file and second as GLB "binary embeded"

Now if binary embeded one will load first, and will be little different than file based one, the next GLTF model will have "mismatch texture"

So i think we should implement "per-load" cache, for texture for both:

if (uri == null) {

and

} else if (uri.startsWith("data:")) {

only.

No material cache, No assetmanager cache, but local cache only for binary textures in this exactly places.

tonihele:
It is all about the asset keys. The embedded ones will not get unique keys, file ones have basically unique keys. Well not unique but all will point to the same file so it is fine if another model uses that.

This material cloning will work since the embedded textures wont be cached as their uniqueness can't be vouched for anyway.

oxplay2:
> But if you load the same model multiple times, with embedded texture... you will get multiple (same) textures loaded. So... Kinda could figure a unique name for them, if possible, and then cache them to asset manager as well...

Yes if this would be possible that cache would work for exact GLB file, but not image file or another GLB file with same texture, then i think it would be fine.

tonihele:
To get close to unique name for the embedded textures... I guess they could contain the URI, same as the actual file references. But here the file is the current GLB file, this can be extracted from `AssetInfo.AssetKey`.

oxplay2:
sounds good. GLB file URI + texture URI(or whatever glb store as name/index)?

Ofc still GLTF advantage is they might use same file, while 2 GLB files using same texture 2x times is 2x time more memory usage. But i see no solution for this, because it might break if we cache under same name.

tonihele:
Well... Lol, this doesn't really work. I mean, I totally did it. And wondered why it doesn't trigger... Because if one loads the whole model the second time, it comes from... the... cache. And it is just cloned. So I suspect that this is just all very fine with the material caching.

pspeed42:
The URI proposal is very similar to have Java JAR files work with internal file URIs.

re: "while 2 GLB files using same texture 2x times is 2x time more memory usage."

...yeah, because without computing some kind of hash, there is no way to know that they are the same texture.  Name is not enough.

For lots of reasons, I've switched to using gltf exclusively here... way easier to debug if something goes wrong.  But yeah, it would be nice if glb can work better than it currently does.

tonihele:
I loaded the BistroExterior.glb (960Mb, with 360Mb of textures embedded) 100 times to a scene and it looks to be fine. Textures are shared. FPS is terrible :)

tonihele:
But yeah, it still needs to cache the textures locally. If some textures are shared between the materials. But I think this is better done with the local cache than the AssetManager cache. Since reading the texture data (from memory!) is still somewhat slow. I'll optimize this reading with the other ticket I made.

I'll just cache the textures locally and that I think covers all the cases, right?

These embedded textures are unique to the GLB model, and the model itself is cached anyway in the AssetManager for future use.

oxplay2:
ye, just cache textures for GLB locally (leave loadTexture case of GLTF because it cache anyway) and it should be done.

Ofc i know that GLB enthusiasts would prefer GLB be better than GLTF, but it seems to be opposite.

tonihele:
Should be good to go!

stephengold:
I re-read the entire discussion just now and was impressed by all the thought that went into this pull request. Thank you to all who participated!

Unless there's further substantial discussion, I plan to integrate this PR in about 48 hours.

oxplay2:
Looking at code it seems fine. GLTF or GLB store same textures as indexes, so materials just use same indexes, and code here just cache by indexes, so it should work fine. 

While one line "dataStream.close();" removal should be fine, since based on stackoverflow "DataInputStream holds no system resources of its own, so not closing it will not cause any leaks. You can simply leave it open. Alternatively you can return it from the method so that the caller can close it."

So from my perspective this PR is ready, tho it would be cool if someone else would look too.

tonihele:
The stream is closed. Previously if failed, it was never closed. I fixed
this failing case.

I used Java feature called try-with resource. Which does the try finally if
not null then close thing but cleaner syntax. Im sorry I explained it
poorly the first time.

stephengold:
Unless there's further substantial discussion, I plan to integrate this PR in about 24 hours.

stephengold:
Thank you @tonihele for the solution and @oxplay2 for code review.

Issue #2132, opened 2023-10-26
Texture.java setMinFilter() contains glued code should be also executed by texture.read()


oxplay2:
I still might be wrong, but it seems like an issue.

Texture.java
```
    public void setMinFilter(MinFilter minificationFilter) {
        if (minificationFilter == null) {
            throw new IllegalArgumentException(
                    "minificationFilter can not be null.");
        }
        this.minificationFilter = minificationFilter;
        if (minificationFilter.usesMipMapLevels() && image != null && !image.isGeneratedMipmapsRequired() && !image.hasMipmaps()) {
            image.setNeedGeneratedMipmaps();
        }
    }
```

That is executed always first time by:
```
    public void setImage(Image image) {
        this.image = image;

        // Test if mipmap generation required.
        setMinFilter(getMinFilter());
    }
```

GLTFLoader execute it second time for setting proper filtering by:

```
        if (minFilter != null) {
            texture.setMinFilter(minFilter);
        }
```

BUT J3O importer when using texture.read capsule do not execute it second time, causing not setting proper bollean for generateMipMaps.

Here is texture.read:
```
    public void read(JmeImporter importer) throws IOException {
        InputCapsule capsule = importer.getCapsule(this);
        name = capsule.readString("name", null);
        key = (TextureKey) capsule.readSavable("key", null);

        // load texture from key, if available
        if (key != null) {
            // key is available, so try the texture from there.
            try {
                Texture loadedTex = importer.getAssetManager().loadTexture(key);
                image = loadedTex.getImage();
            } catch (AssetNotFoundException ex){
                Logger.getLogger(Texture.class.getName()).log(Level.SEVERE, "Cannot locate texture {0}", key);
                image = PlaceholderAssets.getPlaceholderImage(importer.getAssetManager());
            }
        }else{
            // no key is set on the texture. Attempt to load an embedded image
            image = (Image) capsule.readSavable("image", null);
            if (image == null){
                // TODO: what to print out here? the texture has no key or data, there's no useful information ..
                // assume texture.name is set even though the key is null
                Logger.getLogger(Texture.class.getName())
                        .log(Level.SEVERE, "Cannot load embedded image {0}", toString());
            }
        }

        anisotropicFilter = capsule.readInt("anisotropicFilter", 1);
        minificationFilter = capsule.readEnum("minificationFilter",
                MinFilter.class,
                MinFilter.BilinearNoMipMaps);
[...]

PR #2134, opened 2023-10-27
Fixes Issue #2132 (Texture.load() doesn't trigger mipmap generation)

JohnLKkk:
Fixes Issue #2132 
see: https://github.com/jMonkeyEngine/jmonkeyengine/issues/2132


oxplay2 (approved):
During Code Check the only concern were exceptions thrown by this methods when value is null.

But BinaryInputCapsule should setup default value if capsule have no field, and if field have somehow null value, then anyway it should throw exception same way because this values are intended to always have value.

So for me code looks fine.

stephengold:
Unless there's further discussion, I plan to integrate this PR in about 48 hours.

stephengold:
Thanks for the fix, @johnlkkk !

PR #2139, opened 2023-11-01
PBRLighting.j3md:  annotate LightMap as being a LINEAR texture

stephengold:
This PR should eliminate the distracting/confusing warnings currently generated when the same texture is specified for both LightMap and another PBR map.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.j3md
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.j3md
@@ -69,7 +69,7 @@ MaterialDef PBR Lighting {
         Boolean HorizonFade
 
         // Set to Use Lightmap
-        Texture2D LightMap
+        Texture2D LightMap -LINEAR
 
         // A scalar multiplier controlling the amount of occlusion applied.
         // A value of `0.0` means no occlusion. A value of `1.0` means full occlusion.


stephengold:
This change eliminates some runtime diagnostics like this one:

```console
Nov 04, 2023 11:44:39 PM com.jme3.material.Material checkTextureParamColorSpace
WARNING: The texture Models/Tank/Tank_Occ_Rough_Metal.png has linear color space, but the material parameter LightMap specifies no color space requirement, this may lead to unexpected behavior.
Check if the image was not set to another material parameter with a linear color space, or that you did not set the ColorSpace to Linear using texture.getImage.setColorSpace().
```

stephengold:
The only evidence I have one way or the other is that it's common to specify the same texture for both "LightMap" and "MetallicRoughnessMap" (with "LightMapAsAOMap" set to `true`). Since "MetallicRoughnessMap" specifies `-LINEAR` it makes sense that "LightMap" should also.

riccardobl:
It seems that when LightMapAsAOMap is true the lightmap is used as ao map that is a grayscale linear texture that can be specified as -LINEAR.
But when it is used as an actual lightmap, it contains colors and so it should be sampled as sRGB.

stephengold:
@riccardobl How inconvenient! 
Perhaps a better solution would be to have distinct material parameters for the light map and the AO map.
What do you think?

riccardobl:
We have so many possible combinations in our PBR shader at this point, but my understanding is that if the AO map is the R component of MetallicRoughnessMap, it should be used by setting AoPackedInMRMap.
While LightMapAsAOMap should be used when the AO map is standalone and provided using the LightMap field.

stephengold:
PBRLighting.j3md does seem needlessly complex.

So if "AoPackedInMRMap" is true then the AO map is linear, but if "LightMapAsAOMap" is true then the AO map is sRGB?

riccardobl:
AO is always linear, but the lightmap is linear if it contains only AO or sRGB if it is a full lightmap

PR #2141, opened 2023-11-06
Adding support for Specular-AA

JohnLKkk:
see:http://www.jp.square-enix.com/tech/library/pdf/ImprovedGeometricSpecularAA(slides).pdf
https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@15.0/manual/Geometric-Specular-Anti-Aliasing.html
https://hub.jmonkeyengine.org/t/geometric-specular-anti-aliasing/47246/2
Usage: Simply set Material.setBoolean("SpecularAA", true);

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -100,10 +100,10 @@ varying vec3 wNormal;
 
 // Specular-AA
 #ifdef SCREEN_SPACE_VARIANCE
-  uniform float m_Sigma;
+  uniform float m_SpecularAASigma;
 #endif
 #ifdef THRESHOLD
-  uniform float m_Kappa;
+  uniform float m_SpecularAAKappa;
 #endif
 
 #ifdef DISCARD_ALPHA
@@ -259,10 +259,10 @@ void main(){
         float sigma = 1.0;
         float kappa = 0.18;
         #ifdef SCREEN_SPACE_VARIANCE
-            sigma = m_Sigma;
+            sigma = m_SpecularAASigma;
         #endif
         #ifdef THRESHOLD
-            kappa = m_Kappa;
+            kappa = m_SpecularAAKappa;
         #endif
     #endif
     float ndotv = max( dot( normal, viewDir ),0.0);
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -256,7 +256,7 @@ void main(){
     #endif
 
     #ifdef SPECULAR_AA
-        float sigma = 0.5f;
+        float sigma = 1.0f;
         float kappa = 0.18f;
         #ifdef SCREEN_SPACE_VARIANCE
             sigma = m_Sigma;
@@ -289,13 +289,17 @@ void main(){
         vec3 directDiffuse;
         vec3 directSpecular;
 
-        float hdotv = PBR_ComputeDirectLight(
         #ifdef SPECULAR_AA
-                            sigma, kappa,
+            float hdotv = PBR_ComputeDirectLightWithSpecularAA(
+                                normal, lightDir.xyz, viewDir,
+                                lightColor.rgb, fZero, Roughness, sigma, kappa, ndotv,
+                                directDiffuse,  directSpecular);
+        #else
+            float hdotv = PBR_ComputeDirectLight(
+                                normal, lightDir.xyz, viewDir,
+                                lightColor.rgb, fZero, Roughness, ndotv,
+                                directDiffuse,  directSpecular);
         #endif
-                            normal, lightDir.xyz, viewDir,
-                            lightColor.rgb, fZero, Roughness, ndotv,
-                            directDiffuse,  directSpecular);
 
         vec3 directLighting = diffuseColor.rgb *directDiffuse + directSpecular;
         
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -99,10 +99,10 @@ varying vec3 wPosition;
 varying vec3 wNormal;
 
 // Specular-AA
-#ifdef SCREEN_SPACE_VARIANCE
+#ifdef SPECULAR_AA_SCREEN_SPACE_VARIANCE
   uniform float m_SpecularAASigma;
 #endif
-#ifdef THRESHOLD
+#ifdef SPECULAR_AA_THRESHOLD
   uniform float m_SpecularAAKappa;
 #endif
 
@@ -258,10 +258,10 @@ void main(){
     #ifdef SPECULAR_AA
         float sigma = 1.0;
         float kappa = 0.18;
-        #ifdef SCREEN_SPACE_VARIANCE
+        #ifdef SPECULAR_AA_SCREEN_SPACE_VARIANCE
             sigma = m_SpecularAASigma;
         #endif
-        #ifdef THRESHOLD
+        #ifdef SPECULAR_AA_THRESHOLD
             kappa = m_SpecularAAKappa;
         #endif
     #endif
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -256,8 +256,8 @@ void main(){
     #endif
 
     #ifdef SPECULAR_AA
-        float sigma = 1.0f;
-        float kappa = 0.18f;
+        float sigma = 1.0;
+        float kappa = 0.18;
         #ifdef SCREEN_SPACE_VARIANCE
             sigma = m_Sigma;
         #endif
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -98,6 +98,14 @@ varying vec3 wPosition;
 #endif
 varying vec3 wNormal;
 
+// Specular-AA
+#ifdef SCREEN_SPACE_VARIANCE
+  uniform float m_Sigma;
+#endif
+#ifdef THRESHOLD
+  uniform float m_Kappa;
+#endif
+
 #ifdef DISCARD_ALPHA
   uniform float m_AlphaDiscardThreshold;
 #endif
@@ -247,6 +255,16 @@ void main(){
        ao = clamp(ao, 0.0, 1.0);
     #endif
 
+    #ifdef SPECULAR_AA
+        float sigma = 0.5f;
+        float kappa = 0.18f;
+        #ifdef SCREEN_SPACE_VARIANCE
+            sigma = m_Sigma;
+        #endif
+        #ifdef THRESHOLD
+            kappa = m_Kappa;
+        #endif
+    #endif
     float ndotv = max( dot( normal, viewDir ),0.0);
     for( int i = 0;i < NB_LIGHTS; i+=3){
         vec4 lightColor = g_LightData[i];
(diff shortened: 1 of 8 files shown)


jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
```diff
@@ -6,6 +6,43 @@
...
+    vec3 dndu = dFdx(normal);
+    vec3 dndv = dFdy(normal);
+    vec2  footprint_bounding_box = fwidth(H); 		//- abs(dfdx(slope_h)) + abs(dfdy(slope_h))
+	//vec2  variance = sigma*sigma*footprint_bounding_box*footprint_bounding_box;
+        float  variance = sigma*sigma*(dot(dndu, dndu) + dot(dndv, dndv));
+    float  kernel_roughness = min(Kappa, variance);
+    alpha_x = sqrt(alpha_x*alpha_x + kernel_roughness);
+}
+void Specular_Anti_Aliasing2(in vec3 normal, in vec3 half_vector,
+                           	inout float alpha_x)
```
codex128, line 28: `Specular_Anti_Aliasing2` is not used.

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
```diff
@@ -6,6 +6,43 @@
...
+{
+
+    float sigma = 0.50f; 							//- screen space variance
+    float Kappa = 0.18f;								//- clamping treshold
+    vec3 h = TBN * half_vector;
+    vec2  H = h.xy;
+    vec3 dndu = dFdx(normal);
+    vec3 dndv = dFdy(normal);
```
codex128, line 36: `dndu` and `dndv` are not used.

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
```diff
@@ -6,6 +6,43 @@
...
+void Specular_Anti_Aliasing(in vec3 normal, in vec3 half_vector,
+                           	inout float alpha_x)
+{
+
+    float sigma = 0.50f; 							//- screen space variance
+    float Kappa = 0.18f;								//- clamping treshold
+    vec2  H = half_vector.xy;
+    vec3 dndu = dFdx(normal);
+    vec3 dndv = dFdy(normal);
+    vec2  footprint_bounding_box = fwidth(H); 		//- abs(dfdx(slope_h)) + abs(dfdy(slope_h))
```
codex128, line 21: `footprint_bounding_box` and `H` are not used.

codex128 (changes requested):
Just some unused variables/functions that should be trimmed out if possible.

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
JohnLKkk: I have removed the related code, thanks for the review.

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
JohnLKkk: I have removed the related code, thanks for the review.

codex128 (approved):
Ok, that looks good! :+1:

riccardobl:
Should we set this on by default?

JohnLKkk:
> Should we set this on by default? Are there any disadvantages in doing so? It doesn't seem computationally intensive

Personally, I'm inclined to set its default value to True, however, others may not want anti-aliasing to be on by default (for some reason it seems most people use the low quality graphics mode in JME), so I'm not sure if the default here should be changed to True...

JohnLKkk:
I forgot one thing, I will expose the two parameters Screen Space Variance and Threshold to material parameters tomorrow as well. For descriptions of them, refer to:
https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@15.0/manual/Geometric-Specular-Anti-Aliasing.html

riccardobl:
I think we should make an exception in this case, since it won't impact performances too much and it is most likely a wanted feature, i doubt anyone likes aliasing in the reflections, but let's wait for a counterargument .

oxplay2:
agree, i think it should be default with just option to disable.

Tho i belive it might be nice to add later to PBRTerrain too?

JohnLKkk:
> agree, i think it should be default with just option to disable. (like its option now)
> 
> Tho i belive it might be nice to add later to PBRTerrain too? (it do not need to be this pr ofc)

Specular-AA has been added to PBRTerrain.

> I think we should make an exception in this case, since it won't impact performances too much and it is most likely a wanted feature, i doubt anyone likes aliasing in the reflections, but let's wait for a counterargument .



> I would say make it default
True has been set as the default value, and two material parameters (Screen Space Variance and Threshold) have been added.

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
```diff
@@ -247,6 +255,16 @@ void main(){
        ao = clamp(ao, 0.0, 1.0);
     #endif
 
+    #ifdef SPECULAR_AA
+        float sigma = 0.5f;
```
riccardobl, line 259: The default values are set twice, here and in the j3md.
Also, can you remove the "f" suffix? Since it is unsupported in older specs, thanks

jme3-terrain/src/main/resources/Common/MatDefs/Terrain/AdvancedPBRTerrain.frag
```diff
@@ -485,6 +493,16 @@ void main(){
     
      
     float ndotv = max( dot( normal, viewDir ),0.0);
+    #ifdef SPECULAR_AA
+        float sigma = 0.5f;
+        float kappa = 0.18f;
```
riccardobl, line 498: The default values are set twice, here and in the j3md.
Also, can you remove the "f" suffix? Since it is unsupported in older specs, thanks

jme3-terrain/src/main/resources/Common/MatDefs/Terrain/PBRTerrain.frag
```diff
@@ -417,6 +425,16 @@ gl_FragColor.rgb = vec3(0.0);
     
      
     float ndotv = max( dot( normal, viewDir ),0.0);
+    #ifdef SPECULAR_AA
+        float sigma = 0.5f;
```
riccardobl, line 429: The default values are set twice, here and in the j3md.
Also, can you remove the "f" suffix? Since it is unsupported in older specs, thanks

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
```diff
@@ -35,7 +51,11 @@ vec3 sphericalHarmonics( const in vec3 normal, const vec3 sph[9] ){
 }
 
 
-float PBR_ComputeDirectLight(vec3 normal, vec3 lightDir, vec3 viewDir,
+float PBR_ComputeDirectLight(
```
riccardobl, line 54: Could you please  refactor  this to a function that accepts a precalculated alpha  eg. `_PBR_ComputeDirectLight(in float alpha,. ..`
and then have two more functions 
- `PBR_ComputeDirectLight` that calls `_PBR_ComputeDirectLight` with `alpha=roughness*roughness` (the current non antialiased bahavior)

and 

- `PBR_ComputeDirectLightAA` that calls `_PBR_ComputeDirectLight` using the antialiasing corrected alpha (new method)

and then call PBR_ComputeDirectLightAA if SPECULAR_AA is defined or PBR_ComputeDirectLight otherwise.

In this way we let the shaders implementing PBR.glsllib deal with their own way to detect if antialiasing is required, instead of forcing them to defined SPECULAR_AA

riccardobl (changes requested):

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
JohnLKkk: Good idea. I think no need to define two PBR_ComputeDirectLight and PBR_ComputeDirectLightAA, just adjust the roughness parameter to alpha as function parameter, then calculate externally via
#SPECULAR_AA
alpha = Specular_Anti_Aliasing();
#else
alpha = roughness * roughness;
#endif
PBR_ComputeDirectLight(，，alpha，...);
What do you think?

jme3-terrain/src/main/resources/Common/MatDefs/Terrain/PBRTerrain.frag
JohnLKkk: I will modify the code later, thanks for pointing it out.

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
riccardobl: > I think the #SPECULAR_AA macro definition belongs to material parameters and should not appear in PBR.glsllib

Indeed, that's what i was proposing, to remove the macro from PBR.glsllib, however the rationale behind keeping the signature of PBR_ComputeDirectLight unchanged and to add another function PBR_ComputeDirectLightAA is to maintain backward compatibility.
PBR_ComputeDirectLightAA doesn't need to check if specular AA is enabled internally, since PBRLighting etc would do the check and call it only if specular AA is enabled.

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
oxplay2: wow, didnt noticed "f" suffix myself. Good catch.

jme3-core/src/main/resources/Common/ShaderLib/PBR.glsllib
JohnLKkk: > > I think the #SPECULAR_AA macro definition belongs to material parameters and should not appear in PBR.glsllib
> 
> Indeed, that's what i was proposing, to remove the macro from PBR.glsllib, however the rationale behind keeping the signature of PBR_ComputeDirectLight unchanged and to add another function PBR_ComputeDirectLightAA is to maintain backward compatibility. PBR_ComputeDirectLightAA doesn't need to check if specular AA is enabled internally, since PBRLighting etc would do the check and call it only if specular AA is enabled.

I roughly understand what you said. Considering backwards compatibility, I will adjust the code here.

JohnLKkk: I have adjusted the related code

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
JohnLKkk: The "f" suffixes have been removed

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.j3md
```diff
@@ -56,6 +56,13 @@ MaterialDef PBR Lighting {
         // Parallax/height map
         Texture2D ParallaxMap -LINEAR
 
+        // Specular-AA
+        Boolean UseSpecularAA : true
+        // screen space variance,Use the slider to set the strength of the geometric specular anti-aliasing effect between 0 and 1. Higher values produce a blurrier result with less aliasing.
+        Float Sigma
+        // clamping threshold,Use the slider to set a maximum value for the offset that HDRP subtracts from the smoothness value to reduce artifacts.
+        Float Kappa
+
```
riccardobl, line 65: Sorry to bother you again with these small details, could you please rename Sigma and Kappa to something like SpecularAASigma SpecularAAKappa ?

riccardobl (changes requested):

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.j3md
JohnLKkk: The related parameter names have been updated

riccardobl: Thank you, can you do the same for the defines SPECULARAA_THRESHOLD SPECULARAA_SCREEN_SPACE_VARIANCE ? After that i think it is ready to merge.

JohnLKkk: > Thank you, can you do the same for the defines SPECULARAA_THRESHOLD SPECULARAA_SCREEN_SPACE_VARIANCE ? After that i think it is ready to merge.

The related macro definitions have been updated

riccardobl (approved):
Thank you! If there are no further objections, I will merge it within 24 hours.

PR #2150, opened 2023-11-25
LightControl:  correct some copy-paste errors in the javadoc

stephengold:


stephengold:
Unless there's something to discuss, I'd like to self-integrate this PR in about 24 hours.

PR #2152, opened 2023-11-28
jme3-core:  add 2 getters for frame interpolators

stephengold:
This PR adds getters for the `FrameInterpolator` of a `MorphTrack` or `TransformTrack`.

The corresponding setters date back to MonkAnim, but Nehon didn't provide getters. I don't know why.

The use case is when I want to deep-clone an `AnimTrack`.  Since these classes don't implement deep cloning using `cloneFields()`, I need access to the private `interpolator` field. Currently such access is possible only by using reflection.


stephengold:
Unless there's something worth discussing here, I plan to self-integrate this pull request in about 72 hours.

Issue #2076, opened 2023-09-30
software skinning shouldn't require vertex normals


stephengold:
Trying to render a model imported from https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/RecursiveSkeletons/glTF using [MonkeyWrench](https://github.com/stephengold/MonkeyWrench):
```txt
Sep 30, 2023 11:58:32 AM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.NullPointerException
	at com.jme3.anim.SkinningControl.resetToBind(SkinningControl.java:332)
	at com.jme3.anim.SkinningControl.controlRenderSoftware(SkinningControl.java:255)
	at com.jme3.anim.SkinningControl.controlRender(SkinningControl.java:304)
	at com.jme3.scene.control.AbstractControl.render(AbstractControl.java:126)
	at com.jme3.scene.Spatial.runControlRender(Spatial.java:763)
	at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:796)
	at com.jme3.renderer.RenderManager.renderSubScene(RenderManager.java:806)
	at com.jme3.renderer.RenderManager.renderScene(RenderManager.java:780)
	at com.jme3.renderer.RenderManager.renderViewPort(RenderManager.java:1205)
	at com.jme3.renderer.RenderManager.render(RenderManager.java:1287)
	at com.jme3.app.SimpleApplication.update(SimpleApplication.java:278)
	at com.jme3.system.lwjgl.LwjglWindow.runLoop(LwjglWindow.java:628)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:717)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

One of the many quirks of the RecursiveSkeletons model is that its mesh doesn't include vertex normals. I believe `SkinningControl` should be able to handle that, just like it handles meshes without tangents. Looking at the code, it's clear this possibility wasn't considered:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/4de10c34f6d078e0cd1e32c16bdf5303ceb61e26/jme3-core/src/main/java/com/jme3/anim/SkinningControl.java#L320-L334

In case you're wondering, the model doesn't include vertex tangents either.

stephengold:
The old animation system has a similar issue:

[...]

PR #2155, opened 2023-12-03
solve issue #2076 (software skinning requires vertex normals)

stephengold:
This adds null checks to the software skinning code, for both the old animation system and the new one.
To catch future regressions, a test case is provided.


stephengold:
Unless there's something worth discussing, I plan to self-integrate this pull request in about 48 hours.

Issue #2096, opened 2023-10-12
NoClassDefFoundError when launching the generated jar test launcher


zzuegg:
All works fine if i start the test chooser trough gradle run. 
When starting the jar created by gradle dist it results in an error:
`Uncaught exception thrown in Thread[#56,jME3 Main,6,main]
NoClassDefFoundError: com/jme3/plugins/json/JsonElement`

same error when launching with java -jar

stephengold:
I suspect this is due to a missing dependency in "jme3-examples/build.gradle", probably "jme3-plugins-json" .

zzuegg:
Would it then, not have the same error when launch with the gradle run? 
I have checked, and jme3-plugins-json is not added to the lib folder, i am suspecting the cause in the task dist at the artifact resolving/copying. But i am not good at this

PR #2157, opened 2023-12-08
solve issue #2096 (error launching examples from the dist jar)

stephengold:
I reproduced issue #2096 using Java 11 on Mint Linux.  Adding 2 library dependencies solved it, as I suspected it might.

However, I'm not very confident this is the right solution.
I don't understand why issue #2096 doesn't occur when running examples in NetBeans.
I don't understand when/why the jme3-plugins-json and jme3-plugins-json-gson libraries are required. Is this documented anywhere?


stephengold:
Even though this PR hasn't been reviewed, I plan to self-integrate it in about 72 hours ... unless there's substantial discussion.

PR #2158, opened 2023-12-09
in J3MD files, always list GLSL versions in descending order

stephengold:
Improve several material definitions used in tests, based on Riccardo's note at the Discourse Forum/Hub: https://hub.jmonkeyengine.org/t/porting-jmonkeyengine-to-the-browser-with-webgl2-and-teavm-sourcecode-released/47020/65

As far as I know, these material definitions aren't causing any issues. This PR is more about setting a good example.


stephengold:
Even though this PR hasn't been reviewed, I plan to self-integrate it in about 48 hours ... unless there's substantial discussion.

Issue #2164, opened 2023-12-20
WaterFilter Material parameter is not defined: causticsMap on method setCausticsTexture()


nickidebruyn:
I found a bug on the WaterFilter class when one wants to setCausticsTexture() it sets the wrong parameter on the material.

java.lang.IllegalArgumentException: Material parameter is not defined: causticsMap
	at com.jme3.material.Material.checkSetParam(Material.java:474)
	at com.jme3.material.Material.setTextureParam(Material.java:543)
	at com.jme3.material.Material.setTexture(Material.java:617)
	at com.jme3.water.WaterFilter.setCausticsTexture(WaterFilter.java:1039)

stephengold:
Good catch, @nickidebruyn !
Do you want to submit a pull request, or shall I?

nickidebruyn:
You are welcome to submit a PR.
I currently don't have the jme src checked out.
Thanks @stephengold .

PR #2165, opened 2023-12-21
solve issue #2164 (setCausticsTexture() sets non-existent parameter)

stephengold:
Material-parameter names are case-sensitive, and the actual name of the parameter is "CausticsMap".
This resolves issue #2164.


stephengold:
Even though this PR hasn't been reviewed, I plan to self-integrate it in about 36 hours ... unless there's substantial discussion.

PR #2193, opened 2024-02-07
Create CompositeAppState.java

scenemax3d:
For complex game state management, JME users can benefit from having a composite app state which handles child app states. This way the developer can easily organize his game states.

This composite app state is a contribution of Paul Speed and was copied as-is from:
[src/main/java/com/simsilica/state/CompositeAppState.java](https://github.com/Simsilica/SiO2/blob/master/src/main/java/com/simsilica/state/CompositeAppState.java)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/state/CompositeAppState.java
+++ b/jme3-core/src/main/java/com/jme3/app/state/CompositeAppState.java
@@ -1,6 +1,7 @@
 /*
  * 
  * Copyright (c) 2014-2024 jMonkeyEngine
+ * Copied with Paul Speed's permission from: https://github.com/Simsilica/SiO2
  * All rights reserved.
  * 
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


scenemax3d:
Hi @pspeed42,
Can you please review the license text and add whatever needed credit / reference to your SiO2 lib

Thanks again :)

pspeed42:
To cover all bases, it might be nice to add a line like:
"Copied with Paul Speed's permission from: https://github.com/Simsilica/SiO2"
...somewhere near the copyright (maybe right before the all rights reserved).
It makes the pedigree clear and the permission explicit.
But it's technically fine as-is.

scenemax3d:
> To cover all bases, it might be nice to add a line like: "Copied with Paul Speed's permission from: https://github.com/Simsilica/SiO2" ...somewhere near the copyright (maybe right before the all rights reserved). It makes the pedigree clear and the permission explicit. But it's technically fine as-is.

Sure! done.

Issue #2194, opened 2024-02-08
Add the ClipAction.getAnimClip() method and remove ClipAction.reset() method


capdevon:
Although the constructor of the `ClipAction` class requires an `AnimClip`, there is no getter method `ClipAction.getAnimClip()`. 
Could this be added please?

Here is a use case:
```java
        StringBuilder sb = new StringBuilder();
        Action action = animComposer.getCurrentAction();

        if (action != null) {
            sb.append("ActionType=").append(action.getClass()).append('\n');
            sb.append("Length=").append(timeLength).append('\n');
            sb.append("Speed=").append(action.getSpeed()).append('\n');

            if (action instanceof ClipAction) {
                ClipAction ca = (ClipAction) action;

                try {
                    // Use reflection here because the ClipAction.getAnimClip() method does not exist. Why ???
                    AnimClip clip = (AnimClip) FieldUtils.readField(ca, "clip", true);
                    sb.append("ClipName=").append(clip.getName()).append('\n');
                    sb.append("Duration (sec)=").append(clip.getLength()).append('\n');
                    sb.append("Tracks=").append(clip.getTracks().length).append('\n');

                } catch (ReflectiveOperationException e) {
                    e.printStackTrace();
                }

                sb.append("TransitionLength=").append(ca.getTransitionLength()).append('\n');
                sb.append("MaxTransitionWeight=").append(ca.getMaxTransitionWeight()).append('\n');
                sb.append("Weight=").append(ca.getWeight()).append('\n');
                sb.append("TargetTransforms=").append(ca.getTargets().size()).append('\n');
            }
        }
```

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/tween/action/ClipAction.java#L13

---

Also could you remove the `ClipAction.reset()` method which does nothing?

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/tween/action/ClipAction.java#L62

capdevon:
completed in https://github.com/jMonkeyEngine/jmonkeyengine/pull/2195

PR #2195, opened 2024-02-09
#2194 add ClipAction.getAnimClip() method, remove ClipAction.reset() method

capdevon:
add jMonkeyEngine license

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/tween/action/ClipAction.java
+++ b/jme3-core/src/main/java/com/jme3/anim/tween/action/ClipAction.java
@@ -97,7 +97,7 @@ public class ClipAction extends BlendableAction {
     }
 
     /**
-     * Gets the animation clip associated with this object.
+     * Gets the animation clip associated with this action.
      * 
      * @return The animation clip
      */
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/anim/tween/action/ClipAction.java
```diff
@@ -59,8 +96,8 @@ private void interpolateMorphTrack(double t, MorphTrack track) {
 //        }
     }
 
-    public void reset() {
-
+    public AnimClip getAnimClip() {
```
Scrappers-glitch, line 99: Could you specify a JavaDoc here, please? 
For example: 
```java
/** 
 * Retrieves the animation clip (containing the animation tracks) 
 * associated with this action.
 * 
 * @return the associated animation clip object
 */
``` 
Thanks.

Scrappers-glitch (approved):

scenemax3d:
Hi @capdevon ,
Can you explain why did you remove the reset() method?

Thanks!

capdevon:
Hi @scenemax3d , the `reset `method is **empty** and does **nothing**. 
It is a method that was left incomplete in the transition to the new animation system. 
Removing it does not cause any problem. In my opinion it only creates confusion and lowers the quality of the code.

Scrappers-glitch:
Yes, either to implement the method or remove it completely, however implementation will require a discussion on forum and some prerequisites, and thus another completely different PR. @scenemax3d If you want to support resetting ClipActions, then you might consider opening an issue for a later implementation.

PR #2197, opened 2024-02-17
update the Gradle wrapper to v7.6.4

stephengold:
We can't easily upgrade to Gradle 8, but at least we can use the latest release of Gradle 7.

Upgrading to 7.6.4 addresses some security vulnerabilities. See [the release notes](https://docs.gradle.org/7.6.4/release-notes.html) for details.

Upgrading is a simple matter of changing the properties file:  the upgrade does not involve any changes to the Gradle JAR. (@tonihele I ran the update command twice, just in case.)

Long-term, we should clean up the project so it can migrate to Gradle 8 ... and perhaps Kotlin-based Gradle as well, since that's where the world appears to be headed.


tonihele:
> @tonihele I ran the update command twice, just in case.

Glad to have caused adulthood trauma :D

PR #2199, opened 2024-02-19
VarType: format code

capdevon:
`VarType.Boolean`: reorder the list of `javaTypes` in accordance with `Float` and `Int`

- format code
- update License year

```java
// before
Float("float", float.class, Float.class), 
Int("int", int.class, Integer.class), 
Boolean("bool", Boolean.class, boolean.class), //<----
```

```java
// after
Float("float", float.class, Float.class), 
Int("int", int.class, Integer.class), 
Boolean("bool", boolean.class, Boolean.class), //<----
```

### DIFF
--- a/jme3-core/src/main/java/com/jme3/shader/VarType.java
+++ b/jme3-core/src/main/java/com/jme3/shader/VarType.java
@@ -46,40 +46,39 @@ import com.jme3.texture.TextureCubeMap;
 
 public enum VarType {
 
-    Float("float", float.class, Float.class), 
-    Vector2("vec2", Vector2f.class), 
+    Float("float", float.class, Float.class),
+    Vector2("vec2", Vector2f.class),
     Vector3("vec3", Vector3f.class),
     Vector4("vec4", Vector4f.class, ColorRGBA.class),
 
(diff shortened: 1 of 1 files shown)


stephengold (commented):
Combining formatting and functional changes in a single PR obscures the functional changes and makes them much harder to find and review. Please don't do this again.

On close inspection, it looks like the `UniformBufferObject` and `ShaderStorageBufferObject` constants have been combined into a new `BufferObject` value. The rationale for this change is unclear.

I'm unsure whether this `enum`'s ordinal value is used anywhere---for instance if it's serialized as part of a `Savable`. If it is, then I worry that change might break existing code.

Similarly, if existing code depends on the names of the changed constants, this might be a breaking change.

capdevon:
Thanks for your review @stephengold , I agree with you. I noticed this discrepancy in the order of `javaTypes` while designing my editor. The correct order would have made my code more elegant, instead of having to write the usual exceptions. Now it is too late to fix past mistakes. I have restored the original code. This PR only formats the class code.

stephengold (approved):
Looks good to me.

PR #2201, opened 2024-02-22
Capdevon mat param

capdevon:


capdevon:
This PR should modify only the MatParam class.

PR #2208, opened 2024-03-01
Fix render states merging when additionalState is not set

riccardobl:
While testing some pbr code, i've noticed that something was off with lighting, after some convoluted debugging i've found that pr https://github.com/jMonkeyEngine/jmonkeyengine/commit/5975b3f791d18fae311a10e209b3ed40b5e5c60f introduced a bug.

The tl;dr is that copyMergedTo changes the passed state when additionalState is set, but returns `this` when it isn't, causing mergedRenderState to not be updated at all.

This PR restores the original behavior of passing the returned state to applyRenderState instead of mergedRenderState.


stephengold:
Note: it seems this issue was introduced by PR #2091.

PR #2220, opened 2024-03-13
Performance improvements

tonihele:
Performance improvements. From static code analysis. All Java 8 compatible (is this the official code level?).

There are few types of improvements:

- Switching manual array copying to a system call (is this ok on Android??)
- String concatenation / string format to use StringBuilder properly
- Creating/copying collections switched to use bulk operation
- Some early breaks to loops
- Use Files to get Input/OutputStreams. At least on Windows this is proven to give faster stream implementation
- Arrays.asList size thing (https://shipilev.net/blog/2016/arrays-wisdom-ancients/)

This should be all pretty safe stuff. I did not benchmark this separately at all. These are based on static code analysis.


jme3-examples/src/main/java/jme3test/texture/TestSkyRotation.java
```diff
@@ -131,7 +131,7 @@ public void onAction(String name, boolean ongoing, float ignored) {
             System.out.print("rotate floor and sky leftward ...");
         } else if (name.equals("right")) {
             angle -= 0.1f; // radians
-            System.out.printf("rotate floor and sky spatials rightward ...");
+            System.out.print("rotate floor and sky spatials rightward ...");
```
pspeed42, line 134: While before/after are equivalent, I wonder if the original writer meant to have these all glom onto one line or if they left out the linefeed.

jme3-networking/src/main/java/com/jme3/network/message/ChannelInfoMessage.java
```diff
@@ -69,6 +70,6 @@ public int[] getPorts() {
     
     @Override
     public String toString() {
-        return "ChannelInfoMessage[" + id + ", " + Arrays.asList(ports) + "]";
+        return "ChannelInfoMessage[" + id + ", " + Collections.singletonList(ports) + "]";
```
pspeed42, line 73: Another case where I think the authors original intent was to output the contents of the array and not just wrap the garbage string in brackets.

pspeed42:
re: your description when it says "Arrays.asList size thing" I think you mean "Collection.toArray() size thing"

Curious: what static code analysis was used?

tonihele:
> Curious: what static code analysis was used?

I used IntelliJ's.


> re: your description when it says "Arrays.asList size thing" I think you mean "Collection.toArray() size thing"

Correct. Fixed to description.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -1398,7 +1398,7 @@ assert isValidNumber(fb) : "Invalid Matrix4f value " + uniform.getValue() + " fo
             case FloatArray:
                 fb = uniform.getMultiData();
                 assert isValidNumber(fb) : "Invalid float array value "
-                        + Arrays.asList((float[]) uniform.getValue()) + " for " + uniform.getBinding();
+                        + Collections.singletonList((float[]) uniform.getValue()) + " for " + uniform.getBinding();
```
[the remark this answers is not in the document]
tonihele: Arrays.toString used instead

jme3-networking/src/main/java/com/jme3/network/message/ChannelInfoMessage.java
tonihele: Arrays.toString used instead

jme3-examples/src/main/java/jme3test/texture/TestSkyRotation.java
tonihele: All system out prints in this class now have a linefeed

jme3-examples/src/main/java/jme3test/texture/TestSkyRotation.java
```diff
@@ -128,17 +128,17 @@ public void onAction(String name, boolean ongoing, float ignored) {
...
             angle -= 0.1f; // radians
-            System.out.print("rotate floor and sky spatials rightward ...");
+            System.out.println("rotate floor and sky spatials rightward ...");
         } else {
             return;
         }
         /*
          * Update the local rotations of both objects based on the angle.
          */
-        System.out.printf(" to %.1f radians left of start%n", angle);
```
stephengold: I don't understand why adding second carriage return to a console message is considered a "performance improvement".

tonihele: I think this is old code? Yeah, better to SQUASH when merge

tonihele: These are related to pspeed's comments. So in few places this goes beyond performance improvements and actually changes the behavior to allegedly better one.

pspeed42: In this case, looking at the other methods... they were meant to be prints to get them all on one line.

Then you added a _second_ carriage return to line 141.

My comments about line feeds were apparently incorrect.

tonihele: I accidentally added the second one yes, but I reverted it quite quickly afterwards (https://github.com/jMonkeyEngine/jmonkeyengine/pull/2220/commits/47a9e9596e4bced184b4b8de731b195f42d193fe). Somehow Stephen managed to review 4h old revision.

I'll revert the comment change then too.

tonihele: > I'll revert the comment change then too.

Done

jme3-examples/src/main/java/jme3test/texture/TestSkyRotation.java
tonihele: Linefeeds reverted

jme3-android/src/main/java/com/jme3/input/android/AndroidJoyInput.java
```diff
@@ -215,7 +215,7 @@ public Joystick[] loadJoysticks(InputManager inputManager) {
         if (!disableSensors) {
             joystickList.add(sensorJoyInput.loadJoystick(joystickList.size(), inputManager));
         }
-        return joystickList.toArray( new Joystick[joystickList.size()] );
```
stephengold, line 218: Why is this considered more performant than the old code? Passing a zero-length array results in 2 heap allocations, only one of which is necessary. The same question also applies to many of the other changes in this PR.

tonihele: This is why https://shipilev.net/blog/2016/arrays-wisdom-ancients/

pspeed42: You can skip to the conclusion if you are not interested in the background.  Bottom line is that for counter-intuitive reasons, toArray(new Foo[0]) is faster than toArray(new Foo[specificSize]).  
I read this lore correction some years back.

stephengold: Facinating.

jme3-core/src/main/java/com/jme3/scene/shape/Surface.java
```diff
@@ -45,10 +45,7 @@
 import com.jme3.util.BufferUtils;
 import java.io.IOException;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
stephengold, line 48: JME's preferred code style is based on [Google style](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports) which specifies "No wildcard imports."

tonihele: Fixed from where I was responsible (3 classes)

stephengold: Thank you!

tonihele:
Is this now ok to merge? I believe I did all the requested changes?

stephengold:
I haven't finished looking at all 107 files.

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryExporter.java
stephengold, line 1: For clarity, I believe `width - (width - bytes.length)` should be replaced by `bytes.length` in 2 places.
Also, `width - bytes.length - bytes.length` should be replaced by `width - 2 * bytes.length`.

tonihele: You are correct. I've made the requested change

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
stephengold, line 1: '0 + offset' should be replaced by 'offset'.

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryOutputCapsule.java
```diff
@@ -744,8 +744,7 @@ else if (possibleMagic == DEFAULT_OBJECT)
 
         byte[] rVal = new byte[1 + size];
         rVal[0] = (byte) size;
-        for (int x = 1; x < rVal.length; x++)
-            rVal[x] = bytes[bytes.length - size - 1 + x];
+        System.arraycopy(bytes, bytes.length - size - 1 + 1, rVal, 1, rVal.length - 1);
```
stephengold, line 747: Delete `- 1 + 1` please.

jme3-core/src/plugins/java/com/jme3/export/binary/ByteUtils.java
```diff
@@ -479,9 +479,8 @@ public static byte[] readData(byte[] store, int bytes, InputStream is) throws IO
     public static byte[] rightAlignBytes(byte[] bytes, int width) {
         if (bytes.length != width) {
             byte[] rVal = new byte[width];
-            for (int x = width - bytes.length; x < width; x++) {
-                rVal[x] = bytes[x - (width - bytes.length)];
-            }
+            if (width - (width - bytes.length) >= 0)
+                System.arraycopy(bytes, width - bytes.length - (width - bytes.length), rVal, width - bytes.length, width - (width - bytes.length));
```
stephengold, line 483: Please simplify `width - bytes.length - (width - bytes.length)` to `0` and `width - (width - bytes.length)` to `bytes.length`.

jme3-core/src/test/java/com/jme3/shader/GLSLPreprocessorTest.java
```diff
@@ -51,16 +51,16 @@
...
     String readAllAsString(InputStream is) throws Exception{
-        String output = "";
+        StringBuilder output = new StringBuilder();
         BufferedReader reader = new BufferedReader(new InputStreamReader(is));
         while (true) {
             String l = reader.readLine();
             if (l == null) break;
-            if (output != "") output += "\n";
-            output += l;
+            if (output.toString() != "") output.append("\n");
```
stephengold, line 59: I think this code might be unreliable because it is using `!=` to compare string.
It was unreliable before this change, for the same reason, but obfuscating it in the name of performance just makes it worse.
Either keep it unchanged, or if you want to improve it, replace `output.toString() != ""` with `output.length() > 0`.

jme3-terrain/src/main/java/com/jme3/terrain/heightmap/FluidSimHeightMap.java
```diff
@@ -211,9 +211,7 @@ public boolean load() {
 
         // put the normalized heightmap into the range [0...255] and into the heightmap
         for (int y = 0; y < size; y++) {
-            for (int x = 0; x < size; x++) {
-                heightData[x + y * size] = tempBuffer[curBuf][x + y * size];
-            }
+            System.arraycopy(tempBuffer[curBuf], 0 + y * size, heightData, 0 + y * size, size);
```
stephengold, line 214: Please delete the `0 +` from 2 expressions.

stephengold (changes requested):
I didn't get very far in my review before losing enthusiasm. Hope my review is helpful.
Many issues I found involve `System.arrayCopy()`. You may want to self-review those changes.

jme3-core/src/test/java/com/jme3/shader/GLSLPreprocessorTest.java
tonihele: Or if we are in the modern Java, then we can use isEmpty(). Can't remember when it came...

tonihele:
> I didn't get very far in my review before losing enthusiasm. Hope my review is helpful. Many issues I found involve `System.arrayCopy()`. You may want to self-review those changes.

Thanks, I'll go through them. These are all changes by automation.

jme3-core/src/test/java/com/jme3/shader/GLSLPreprocessorTest.java
stephengold: As of Java 22, there's no `StringBuilder.isEmpty()`.
But `String.isEmpty()` dates back to Java 1.6, so that would work.
I just assume it would be more efficient to test the length property of the `StringBuilder` than to convert it to a `String` and then test *its* length property, regardless of whether you use `String.isEmpty()` or `String.length`.

tonihele: I think you are right about that.

stephengold:
> > I didn't get very far in my review before losing enthusiasm. Hope my review is helpful. Many issues I found involve `System.arrayCopy()`. You may want to self-review those changes.
> 
> Thanks, I'll go through them. These are all changes by automation.

Submitting a huge PR that's 100% auto generated and then relying on human volunteers review it seems abusive. And perhaps detrimental to the project as a whole.

tonihele:
> > > I didn't get very far in my review before losing enthusiasm. Hope my review is helpful. Many issues I found involve `System.arrayCopy()`. You may want to self-review those changes.
> > 
> > 
> > Thanks, I'll go through them. These are all changes by automation.
> 
> Submitting a huge PR that's 100% auto generated and then relying on human volunteers review it seems abusive. And perhaps detrimental to the project as a whole.

I'm terribly sorry. I did of course go through it myself prior submitting it. And my motivation is to help the project as a whole. And I feel that I am one of the _abused_ human volunteers that you referenced, having been here for a long time too. And to be perfectly clear, I don't feel abused by this PR, any comment or anything, just including myself to this _sorry lot_ :D

My point perhaps was the scope of this PR. I'm more than happy to discuss and make changes you guys request, no problem there. Just kinda a defensive point that the focus point was only on the performance of these scattered lines. Not to make existing problems go away or make actual changes in behavior. Of course there is no intention of making the code less readable either.

stephengold:
@tonihele, I realize of course that you're also a volunteer, and I greatly value and respect you and your past contributions to the project. If this PR had come from someone I didn't know and respect, I would have totally ignored it or closed it as "won't fix." As it is, I've neglected it. I realize now I haven't been forthcoming about my reasons for that neglect, and I apologize for my silence. You deserve better communication, so I'll try to explain myself.

Engine performance is important, but micro improvement efforts should focus on the inner loops of practical workloads. Furthermore, I doubt optimizations auto-generated by static analysis will measurably improve the performance of JME.
I tried to explain this to xtonik last year (https://hub.jmonkeyengine.org/t/newcomer/46680) and didn't do a great job.

A lot of Engine code is semi-obsolete, rarely executed, and/or executed only during testing or initialization or error conditions. Automated static analysis doesn't take such factors into account.

Furthermore, many micro-optimizations reduce readability. For example, I'd rather read a `for` loop than an invocation of `System.arrayCopy()`. I read and write `for` loops daily and find them very readable. OTOH I find `arrayCopy()` very opaque. It has 5 positional arguments (3 of which are integers) that are easily confused. I use it so rarely that I forget its precise semantics: what happens if an argument is negative, what happens if the arrays are too short, and what happens if `src` and `dest` are the same array or are of different types.

IMO, the time and energy of the core developers is better spent on bugs that visibly affect apps. If you enjoy spending your time on micro-optimizations, that's what you should do. But repeatedly and specifically requested my review drags me into work I don't enjoy and have little talent for, all for benefits I don't believe in.

Again, I'm sorry it's taken me so long to articulate my underlying concerns about this PR, and I apologize for that.

PR #2225, opened 2024-03-17
#2224 Fix bug where the clipRect would be left in the openGL state 

### LINKED ISSUE #2224: RenderManager#renderViewPort method leaves a clipRect in the OpenGL state
Within GLRenderer the renderViewPort method leaves a clipRect in the OpenGL state. This means that other calls to GLRenderer #copyFrameBuffer may give surprising (clipped) results

GLRenderer#renderViewPort should clear the clipRect state once it has finished its render

See https://hub.jmonkeyengine.org/t/blitted-to-frame-buffer-only-copies-part-of-source/47564 for an example of the problem caused

(I intend to put in a PR for this myself)

richardTingle:
Fix bug where the clipRect would be left in the openGL state leading to surprising (clipped) results when later using GLRenderer#copyFrameBuffer

closes #2224

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/RenderManager.java
@@ -1246,6 +1246,12 @@ public class RenderManager {
         // clear any remaining spatials that were not rendered.
         clearQueue(vp);
 
+        /*
+         * the call to setCamera will indirectly cause a clipRect to be set, must be cleared to avoid surprising results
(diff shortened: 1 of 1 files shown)


codex128 (approved):

scenemax3d:
Hi @richardTingle ,
Was this tested with v3.7 as well or just with master branch?

richardTingle:
I tested that the bug existed in 3.6 and master, but I tested the fix only in a locally built master

PR #2229, opened 2024-03-19
Added SoftBloomFilter

codex128:
The filter introduced in this PR produces much cleaner and visually natural results than BloomFilter. Only supports the equivalent of GlowMode.Scene (no select object glow).

Basic usage is similar to BloomFilter:
```java
PBRBloomFilter bloom = new PBRBloomFilter();
fpp.addFilter(bloom);
viewPort.addProcessor(fpp);
```

Note: name changed from PBRBloomFilter to SoftBloomFilter.


jme3-effects/src/main/java/com/jme3/post/filters/PBRBloomFilter.java
```diff
@@ -0,0 +1,262 @@
...
+        this.glowFactor = FastMath.clamp(factor, 0, 1);
+        if (material != null) {
+            material.setFloat("GlowFactor", glowFactor);
+        }
+    }
+    
+    /**
+     * Gets the number of downsampling/upsampling passes per step.
+     * 
+     * @return glow factor
```
capdevon, line 245: "`@return glow factor`": error in the javadoc
should be "`@return number of downsampling/upsampling passes`"

jme3-effects/src/main/java/com/jme3/post/filters/PBRBloomFilter.java
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Gets the glow factor.
+     * 
+     * @return glow factor
+     * @see #setGlowFactor(float)
+     */
+    public float getGlowFactor() {
+        return glowFactor;
+    }
+
```
capdevon, line 261: The read/write methods for serializing the filter are missing

eg
```java
    @Override
        super.write(ex);
        oc.write(glowMode, "glowMode", GlowMode.Scene);
        oc.write(blurScale, "blurScale", 1.5f);
        oc.write(exposurePower, "exposurePower", 5.0f);
        oc.write(exposureCutOff, "exposureCutOff", 0.0f);
        oc.write(bloomIntensity, "bloomIntensity", 2.0f);
        oc.write(downSamplingFactor, "downSamplingFactor", 1);
    }

    @Override
        super.read(im);
        InputCapsule ic = im.getCapsule(this);
        glowMode = ic.readEnum("glowMode", GlowMode.class, GlowMode.Scene);
        blurScale = ic.readFloat("blurScale", 1.5f);
        exposurePower = ic.readFloat("exposurePower", 5.0f);
        exposureCutOff = ic.readFloat("exposureCutOff", 0.0f);
        bloomIntensity = ic.readFloat("bloomIntensity", 2.0f);
        downSamplingFactor = ic.readFloat("downSamplingFactor", 1);
    }
```

jme3-examples/src/main/java/jme3test/post/TestPBRBloom.java
```diff
@@ -0,0 +1,148 @@
...
+        fpp=new FilterPostProcessor(assetManager);
+        PBRBloomFilter bloom=new PBRBloomFilter();
+        fpp.addFilter(bloom);
+        //ToneMapFilter toneMap = new ToneMapFilter();
+        //fpp.addFilter(toneMap);
+        viewPort.addProcessor(fpp);
+
+    }
+
+
```
capdevon, line 146: remove extra spaces

jme3-examples/src/main/java/jme3test/post/TestPBRBloom.java
```diff
@@ -0,0 +1,148 @@
...
+        Spatial sky = SkyFactory.createSky(assetManager, 
+                "Textures/Sky/Bright/FullskiesBlueClear03.dds", 
+                EnvMapType.CubeMap);
+        sky.setCullHint(Spatial.CullHint.Never);
+        rootNode.attachChild(sky);
+        EnvironmentProbeControl.tagGlobal(sky);
+        
+        rootNode.addControl(new EnvironmentProbeControl(assetManager, 256));
+        
+        fpp=new FilterPostProcessor(assetManager);
```
capdevon, line 137: Add the spaces before and after =

jme3-examples/src/main/java/jme3test/post/TestPBRBloom.java
```diff
@@ -0,0 +1,148 @@
...
+        PBRBloomFilter bloom=new PBRBloomFilter();
```
capdevon, line 138: Add the spaces before and after =

jme3-examples/src/main/java/jme3test/post/TestPBRBloom.java
```diff
@@ -0,0 +1,148 @@
...
+        fpp.addFilter(bloom);
+        //ToneMapFilter toneMap = new ToneMapFilter();
```
capdevon, line 140: Remove the ToneMapFilter code if unused

jme3-effects/src/main/java/com/jme3/post/filters/PBRBloomFilter.java
```diff
@@ -0,0 +1,262 @@
...
+    public PBRBloomFilter() {}
+    
+    /**
+     * Creates filter with the given number of sampling passes.
+     * 
+     * @param numSamplingPasses 
+     * @see #setNumSamplingPasses(int)
+     */
+    public PBRBloomFilter(int numSamplingPasses) {
+        this.numSamplingPasses = numSamplingPasses;
```
capdevon, line 86: Please add the call to the empty constructor to initialize the filter name.

eg:
```java
public PBRBloomFilter(int numSamplingPasses) {
        this();
        this.numSamplingPasses = numSamplingPasses;
}
```

jme3-effects/src/main/java/com/jme3/post/filters/PBRBloomFilter.java
```diff
@@ -0,0 +1,254 @@
...
+     * <p>
+     * Settings this after the filter has been initialized forces reinitialization.
+     * <p>
+     * default=5
+     * 
+     * @param numSamplingPasses number of passes per donwsampling/upsampling step
+     */
+    public void setNumSamplingPasses(int numSamplingPasses) {
+        if (numSamplingPasses <= 0) {
+            throw new IllegalArgumentException("Expected number of sampling passes to be greater than zero (found: " + numSamplingPasses + ").");
```
capdevon, line 192: Here's a better exception message:
```java
"Number of sampling passes must be greater than zero. Received: " + numSamplingPasses
```

jme3-effects/src/main/java/com/jme3/post/filters/PBRBloomFilter.java
```diff
@@ -0,0 +1,254 @@
...
+     * per fragment). Though, it should be noted that for each downsampling pass the
+     * number of fragments decreases by 75%, and for each upsampling pass, the number
+     * of fragments quadruples (which restores the number of fragments to the original
+     * resolution).
+     * <p>
+     * Settings this after the filter has been initialized forces reinitialization.
+     * <p>
+     * default=5
+     * 
+     * @param numSamplingPasses number of passes per donwsampling/upsampling step
```
capdevon, line 188: better param description:
```java
@param numSamplingPasses The number of sampling passes to use. Must be greater than zero.
```

Add this after param description:
```java
@throws IllegalArgumentException if the provided number of sampling passes is less than or equal to zero.
```

eg:
```java
    /**
     * Sets the number of sampling passes for the filter.
     * <p>
     * This method sets the number of times the filter will sample the input texture
     * when rendering. A higher number of sampling passes typically results in a
     * higher quality filtered image, but also comes at a performance cost.
     * 
     * @param numSamplingPasses The number of sampling passes to use. Must be
     *                          greater than zero.
     * @throws IllegalArgumentException if the provided number of sampling passes is
     *                                  less than or equal to zero.
     */
```

jme3-effects/src/main/java/com/jme3/post/filters/PBRBloomFilter.java
```diff
@@ -0,0 +1,254 @@
...
+ * <a href="https://learnopengl.com/Guest-Articles/2022/Phys.-Based-Bloom">this article</a>.
+ * 
+ * @author codex
+ */
+public class PBRBloomFilter extends Filter {
+    
+    private AssetManager assetManager;
+    private RenderManager renderManager;
+    private ViewPort viewPort;
+    private int width, height;
```
capdevon, line 72: Declare "height" on a separate line.

jme3-effects/src/main/java/com/jme3/post/filters/PBRBloomFilter.java
```diff
@@ -0,0 +1,254 @@
...
+    
+    /**
+     * Creates filter with default settings.
+     */
+    public PBRBloomFilter() {
+        super("PBRBloomFilter");
+    }
+    
+    @Override
+    protected void initFilter(AssetManager am, RenderManager rm, ViewPort vp, int width, int height) {
```
capdevon, line 88: You can simplify the creation of duplicate variables for `w` and `h`. 
Perhaps you can simplify the code in this way:

```java
    @Override
    protected void initFilter(AssetManager am, RenderManager rm, ViewPort vp, int w, int h) {

        this.assetManager = am;
        this.renderManager = rm;
        this.viewPort = vp;
        this.postRenderPasses = new LinkedList<>();
        this.width = w;
        this.height = h;

        Pass[] downsamplingPasses = new Pass[numSamplingPasses];
        Pass[] upsamplingPasses = new Pass[numSamplingPasses];
        
        Renderer renderer = renderManager.getRenderer();

        // the rest of the code is the same...
}
```

capdevon:
* Does the PBRBloomFilter work only with PBR materials ? 
* What effect does it have on Lighting materials (the teapot and the box) ? 
* Can you copy/paste a screenshot in the comments please ? 

The java code looks ok to me, but I am not experienced enough to do a review of the shader code. 
So you need approval from a shader expert and a core member for final integration.

codex128:
> Does the PBRBloomFilter work only with PBR materials?

In general, yes, because this filter relies heavily on HDR.

> What effect does it have on Lighting materials (the teapot and the box in the test class)?

Same effect, technically, though not noticeable because Lighting.j3md doesn't utilize HDR. I copied TestBloomFilter.java, so that is why the teapot is in the test.

> Can you copy/paste a screenshot in the comments please?

riccardobl:
This is a multipass bloom filter that uses hardware filtering to increase the samples, similar to the mipmap bloom filter but it doesn't relies on the opengl mipmap generation function (that is driver dependent and not guaranteed to run on the gpu).

This is much better than the default bloom filter and it has a different optimization (uses hardware filtering instead of two pass blur) so it might be pretty much equivalent or even better from  a performance standpoint. 

My only suggestion is to find a different name, since calling it PBR might be confusing.

I have one question though, what is the default min filter for the filter `getRenderedTexture()` ? This should work best with MinFilter.BilinearNoMipMaps , can you check that? @codex128

codex128:
> My only suggestion is to find a different name, since calling it PBR might be confusing.

Good idea, though PBR Bloom is what the original article calls it. Perhaps SoftBloomFilter?

> what is the default min filter for the filter `getRenderedTexture()` ?

Yup, the default min filter for pass textures is NearestNoMipMaps.

riccardobl:
This is a multipass bloom filter that uses hardware sampling, kinda like the mipmap bloom filter-

> > My only suggestion is to find a different name, since calling it PBR might be confusing.
> 
> Good idea, though PBR Bloom is what the original article calls it. Perhaps SoftBloomFilter?

Yes SoftBloomFilter might be a good option.

> > what is the default min filter for the filter `getRenderedTexture()` ?
> 
> Yup, the default min filter for pass textures is NearestNoMipMaps, so that should be good.

Can you try with **Bilinear**NoMipMaps and show the results? This will enable hardware filtering when downscaling , the result should be smoother.
Might be worth trying also Bilinear MagFilter when upscaling.

codex128:
> Can you try with **Bilinear**NoMipMaps and show the results?

Oh, sorry, my bad. :stuck_out_tongue: Here is the result with min=BilinearNoMipMaps and mag=Bilinear:



It's difficult to tell, but I think the default min/mag filters produce *slightly* rougher glow. For reference:



The differences become more apparent if you zoom in on the glowing sections.

riccardobl:
It should be more visible with complex scenes.
One last thing, can you make it on by default but toggleable with `setBilinearFiltering(false)`?

codex128:
I have implemented that. I also added something to suppress the number of downsampling/upsampling passes made so that textures do not have width or height <= 0.

jme3-effects/src/main/java/com/jme3/post/filters/SoftBloomFilter.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+        }
+    }
+    
+    @Override
+    public void write(JmeExporter ex) throws IOException {
+        super.write(ex);
+        OutputCapsule oc = ex.getCapsule(this);
+        oc.write(numSamplingPasses, "numSamplingPasses", 5);
+        oc.write(glowFactor, "glowFactor", 0.05f);
```
capdevon, line 324: serialize `bilinearFiltering` field

`oc.write(bilinearFiltering, "bilinearFiltering", true);`

jme3-effects/src/main/java/com/jme3/post/filters/SoftBloomFilter.java
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    @Override
+    public void read(JmeImporter im) throws IOException {
+        super.read(im);
+        InputCapsule ic = im.getCapsule(this);
+        numSamplingPasses = ic.readInt("numSamplingPasses", 5);
+        glowFactor = ic.readFloat("glowFactor", 0.05f);
```
capdevon, line 332: deserialize `bilinearFiltering` field

`bilinearFiltering = ic.readBoolean("bilinearFiltering", true);`

jme3-effects/src/main/java/com/jme3/post/filters/SoftBloomFilter.java
codex128: Keep forgetting about that one, thanks for pointing it out. :+1:

riccardobl:
> I have implemented that. I also added something to suppress the number of downsampling/upsampling passes made so that textures do not have width or height <= 0.

Just to be sure to not hurt the feeling of any driver, could you please make it <=2 ?

riccardobl (approved):

riccardobl (commented):
Can you rename PBRBloomFinal.frag and PBRBloomFinal.j3md to SoftBloomFinal.frag/j3md?

riccardobl (approved):
Perfect. Thank you.

capdevon (approved):

PR #2232, opened 2024-03-20
Logs exception as severe on asset locator instantiation failure

### LINKED ISSUE #2230: Unclear exception info on invalid file path in asset locator
If an asset locator is registered with a nonexistent file path and then the system tries to locate an asset with it, a very unclear warning occurs followed by a null pointer exception.

```java
assetManager.registerLocator("NonExistentFolder", FileLocator.class);
assetManager.loadModel("NonExistentModel.j3o");
```
```stacktrace
SEVERE: Cannot create locator of type com.jme3.asset.plugins.FileLocator, does the class have an empty and publicly accessible constructor?
Mar 20, 2024 9:14:45 AM com.jme3.app.LegacyApplication handleError
java.lang.NullPointerException: Cannot invoke "com.jme3.asset.AssetLocator.locate(com.jme3.asset.AssetManager, com.jme3.asset.AssetKey)" because the return value of "com.jme3.asset.ImplHandler$ImplThreadLocal.get()" is null
	at com.jme3.asset.ImplHandler.tryLocate(ImplHandler.java:176)
	at com.jme3.asset.DesktopAssetManager.loadAsset(DesktopAssetManager.java:376)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:441)
	at com.jme3.asset.DesktopAssetManager.loadModel(DesktopAssetManager.java:446)
	at mygame.GeneralTests.simpleInitApp(GeneralTests.java:24)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:610)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:714)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

The warning/exception should instead indicate that a nonexistent path is being used. I am working on a PR to fix this.

codex128:
Fixes #2230 by logging the resulting exception as severe and providing extra info when an asset locator fails to be instantiated.


jme3-core/src/main/java/com/jme3/asset/ImplHandler.java
```diff
@@ -124,10 +124,7 @@ protected T initialValue() {
             } catch (InstantiationException | IllegalAccessException
                     | IllegalArgumentException | InvocationTargetException
                     | NoSuchMethodException | SecurityException ex) {
-                logger.log(Level.SEVERE, "Cannot create locator of type {0}, does"
-                        + " the class have an empty and publicly accessible"
-                        + " constructor?", type.getName());
-                logger.throwing(type.getName(), "<init>", ex);
```
scenemax3d, line 130: Hi @codex128 ,
Why did u remove the logger.throwing? is it OK that no exception will be thrown at that point and a null will be returned instead?

codex128: The null pointer exception does not clearly state what causes it, which is why I changed the log level to severe, since a failure there *will* result in a null pointer later.

scenemax3d: @codex128 ,
Maybe I'm missing something but I see that the code you deleted was also logging a severe exception so what is the significant change you made here?
Also, you deleted the logger.throwing line so I guess you find out that logging the exception object is not clear enough so you removed it right?
Last question - why not using logger.error instead of logger.info? 

10X

codex128: > the code you deleted was also logging a severe exception so what is the significant change you made here?

The message logged previously in the severe category was very unclear. It complains about not being able to instantiate a constructor, regardless of what the actual error is.

> Also, you deleted the logger.throwing line so I guess you find out that logging the exception object is not clear enough so you removed it right?

Logger.throwing is technically clear, but it logs the exception in the finer category. The change logs the exception in the severe category, which is much more visible.

> why not using logger.error instead of logger.info?

I'm not sure what you mean. I am already logging the exception in the severe category.

scenemax3d (approved):

PR #2153, opened 2023-11-29
LwjglCanvas using lwjgl-jawt

JNightRide:
Would this be an acceptable solution for now?

This solves the problem of integrating https://github.com/LWJGLX/lwjgl3-awt (#1868)

There is just one small detail, removing the canvas from its main component and integrating it again breaks it. This issue occurs on the lwjgl3-awt side (https://github.com/LWJGLX/lwjgl3-awt/pull/40)

PR #2235, opened 2024-03-26
Gradle build (zip)

JNightRide:
Hi

This PR resolves dependency issues with **lwjgl** (2 or 3) reported on the [forum](https://hub.jmonkeyengine.org/t/jmonkeyengine-v3-7-feature-freeze/47558/12?u=swiftwolf) when trying to use the zip version of JME3.

The reason for these problems is that jme3-lwjgl3 only loaded the dependencies at runtime and when compiling JME3 it did not export the jar files, this can be solved by changing the way of adding the dependencies:

```Gradle
api "org.lwjgl:lwjgl:${lwjgl3Version}:natives-windows"
```

Should we use jme3-lwjgl3 by default in the examples? (I think it's already quite stable)


jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -497,7 +497,6 @@ protected void createContext(AppSettings settings)  {
         // This will activate the "effective data" scrubber.
         showGLDataEffective.set(settings.getBoolean("GLDataEffectiveDebug"));
         
-        glData.depthSize = settings.getBitsPerPixel();
```
scenemax3d, line 500: Hi @JNightRide ,
How does this line removal relates to solving the gradle zip build issue?

jme3-examples/build.gradle
```diff
@@ -19,8 +19,8 @@ dependencies {
     implementation project(':jme3-effects')
     implementation project(':jme3-jbullet')
     implementation project(':jme3-jogg')
-//    implementation project(':jme3-lwjgl')
-    implementation project(':jme3-lwjgl3')
+    implementation project(':jme3-lwjgl')
```
scenemax3d, line 22: @JNightRide ,
Can you explain why using jme3-lwjgl3 instead of jme3-lwjgl solves the problem?

JNightRide: I didn't understand the question ("using jme3-lwjgl3 instead of jme3-lwjgl"), this part changes from **jme3-lwjgl3 to jme3-lwjgl** to solve the problem.

Currently jme3 uses jme3-lwjgl3 in the examples module.

jme3-lwjgl3 uses runtimeOnly "..." to implement its dependency (which is fine to run with maven or directly), when building the project it fails to export the jar files and is left without binaries.

jme3-lwjgl by default includes jar files (binary):
```
    /*
     * Upgrades the default jinput-2.0.5 to jinput-2.0.9 to fix a bug with gamepads on Linux.
     * See https://hub.jmonkeyengine.org/t/linux-gamepad-input-on-jme3-lwjgl-splits-input-between-two-logical-gamepads
     */
    api 'net.java.jinput:jinput:2.0.9'
    api 'net.java.jinput:jinput:2.0.9:natives-al
```

stephengold (approved):
Looks good to me.

Although I want to encourage developers to use jme3-lwjgl3 instead of jme3-lwjgl, the jme3-examples app chooser *still* works better with jm3-lwjgl. This PR reverts the unintended change to jme3-examples and the dist JAR that was included in PR #2153.

Meanwhile, we should also discourage people from relying the dist JAR (at GitHub) to obtain JME libraries and dependencies, since it doesn't include all of them. People should download JME libraries and dependencies from Maven Central instead, preferably using Gradle build scripts.

PR #2238, opened 2024-03-28
Logic code error (LwjglCanvas - LWJGL3)

JNightRide:
This PR is to remove a '**redundancy'** in the size depth settings in **LwjglCanvas**

```java
...
[Line 500] glData.depthSize = settings.getBitsPerPixel();
...
[Line 504] glData.depthSize = settings.getDepthBits();
```

It doesn't cause any problems but it is better to clean it.


stephengold (approved):
Looks good to me.

PR #2240, opened 2024-04-07
Bright and dark (ColorRGBA - Functions)

JNightRide:
Hello everyone, this PR is about adding two new methods to the **ColorRGBA** class where you can make colors brighter or darker.

The difference between method `brighter()` and `get|setAsSrgb() `is the following:

- It is static (which makes it easier to call)
-  It doesn't saturate the color, just a brighter ("lighter") tone.

Also add an example class (**ColorBrighterTest**) so you can try it out.

[ **Some screenshots** ]

Captura desde 2024-04-06 19-18-57
Captura desde 2024-04-06 19-19-10

What do you think of this idea?

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
+++ b/jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
@@ -42,9 +42,19 @@ import java.io.IOException;
  * @version $Id: ColorRGBA.java,v 1.29 2007/09/09 18:25:14 irrisor Exp $
  */
 public final class ColorRGBA implements Savable, Cloneable, java.io.Serializable {
+    /** Saturation-gamma.*/
     static final float GAMMA = 2.2f;
-
+    /** UID for serialization.*/
     static final long serialVersionUID = 1;
+    
+    /** 
+     * Factor to make colors brighter or darker.
+     * 
+     * @see ColorRGBA#brighter(com.jme3.math.ColorRGBA) 
+     * @see ColorRGBA#darker(com.jme3.math.ColorRGBA) 
+     */
+    private static final float FACTOR = 0.7F;
+    
     /**
      * The color black (0,0,0).
      */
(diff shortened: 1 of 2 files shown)


pspeed42:
darker() seems no different than mult() with a fixed factor... so less flexible.  What not just call mult(0.7)?

brighter() is not that different from mult() either... except again with a fixed factor.  Other than the 0 handling, this is just another mult() call.

The factor seems rather arbitrary.  Why not a parameter?  Using a constant doesn't really make any sense.

Also, neither of these are at all compatible with srgb color because they presume that colors are already in linear space.  (The fact that you compare brighter() to setSrgb() makes me think you might not understand what setAsSrgb() is really doing... hint: it is NOT just making the color brighter.)  I only bring up srgb because you did, though.

These new methods seem unnecessary to me.  They do one thing and they do it rigidly.  A new multRgb(float) that skips alpha and a new minLocal() method would take the place of all of this and have way more uses.

JNightRide:
> darker() seems no different than mult() with a fixed factor... so less flexible. What not just call mult(0.7)?

This is so that the user has a fixed value for the dark color (the intention is not for it to be dynamic in these cases, there are the methods you mentioned).

> Also, neither of these are at all compatible with srgb color because they presume that colors are already in linear space. (The fact that you compare brighter() to setSrgb() makes me think you might not understand what setAsSrgb() is really doing... hint: it is NOT just making the color brighter.) I only bring up srgb because you did, though

Mention method 'setSrgb()' to give an example (not that these new methods are replacing them or anything, not that they are related either)

> brighter() is not that different from mult() either... except again with a fixed factor. Other than the 0 handling, this is just another mult() call.

comparing it to 'mult()' is an error; 'mult()' simply multiplies the values ​​of 2 colors, 'brighter()' increases the clarity of a color (brightness) in a given range so that the user does not have to intervene

> These new methods seem unnecessary to me. They do one thing and they do it rigidly. A new multRgb(float) that skips alpha and a new minLocal() method would take the place of all of this and have way more uses.

As mentioned above, the intention is not to energize it (user intervention); the methods mentioned would not be bad (minLocal() , multRgb(float) -> names)...

JNightRide:
I'll close this PR because it's "redundant".

PR #2241, opened 2024-04-09
Cleanup strange code

TheLivan:
Hi. I simple corrected the oddities in the code.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/texture/plugins/HDRLoader.java
+++ b/jme3-core/src/plugins/java/com/jme3/texture/plugins/HDRLoader.java
@@ -101,18 +101,6 @@ public class HDRLoader implements AssetLoader {
         rgbf[2] = (B / 256.0f) * e;
     }
 
-    public static void convertRGBEtoFloat3(byte[] rgbe, float[] rgbf){
-        int R = rgbe[0] & 0xFF,
-            G = rgbe[1] & 0xFF,
-            B = rgbe[2] & 0xFF,
-            E = rgbe[3] & 0xFF;
-
-        float e = (float) Math.pow(2f, E - (128 + 8) );
-        rgbf[0] = R * e;
-        rgbf[1] = G * e;
-        rgbf[2] = B * e;
-    }
-
     private void writeRGBE(byte[] rgbe){
         if (writeRGBE){
             dataStore.put(rgbe);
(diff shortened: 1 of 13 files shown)


jme3-core/src/plugins/java/com/jme3/texture/plugins/HDRLoader.java
```diff
@@ -101,18 +101,6 @@ public static void convertRGBEtoFloat2(byte[] rgbe, float[] rgbf){
         rgbf[2] = (B / 256.0f) * e;
     }
 
-    public static void convertRGBEtoFloat3(byte[] rgbe, float[] rgbf){
```
codex128, line 104: This method cannot be deleted, in order to maintain backwards compatibility.

jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/objects/FbxMesh.java
```diff
@@ -332,8 +332,8 @@ private List<Geometry> createGeometries() throws IOException {
             indexMap = new ArrayList<>(vCount);
             reverseVertexMap = new ArrayList<>(vCount);
             for(int i = 0; i < vCount; ++i) {
-                vertexMap.set(i, i);
-                indexMap.set(i, i);
+                vertexMap.add(i, i);
+                indexMap.add(i, i);
```
codex128, line 336: I think this should be reverted back. `add` inserts elements, which I don't think is correct here.

codex128 (changes requested):

jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/objects/FbxMesh.java
TheLivan: The set method does replace one element of the collection with another. However, if an element with the given index does not exist, the method will generate an IndexOutOfBoundExeption error. In the above code, the collection is created before the loop, so it will always be empty.

riccardobl: technically add(index,v) is still wrong, the only reason it works is that it is executed in sequence from 0 to n, so i suggest to change it to 
```java
vertexMap.add(i);
indexMap.add(i);
```
without the index

codex128: Also, the lists are not empty at this point. They are initialized with `vCount` null elements. Actually, since these are ArrayLists being used, I think the previous code using `set` was most efficient. `add` causes the internal array to be copied to a new larger array, but `set` allows the list to reuse the internal array. So I suggest reverting these changes.

riccardobl: the vCount initializes the capacity, that is indeed the initial size of the internal array, however set will still fail because size() is not == capacity in arraylist . Kinda weird, but that's how it is, it would need a `for(0..capacity) vertexMap.add(null)` for set to work as expected

PR #2243, opened 2024-04-15
MatParamTexture: duplicate variables, missing javadoc, exceptions

capdevon:
This pull request addresses several issues with the `MatParamTexture` class:

* **Duplicate variable**: The `texture` variable is redundant as it's already stored in the inherited `value` variable. This PR removes the unnecessary texture field.
* **Missing Javadoc**: The class and its methods lacked proper documentation. This PR adds comprehensive Javadoc comments to improve code readability and maintainability.
* **Inconsistent null handling**: The behavior of `setValue` methods regarding null values differed. This PR modifies `setValue` to allow null values, aligning it with the behavior of `setTextureValue` and the superclass's `setValue`.

Key improvements:
* **Code simplification**: By removing the duplicate variable, the code becomes cleaner and easier to understand.
* **Improved documentation**: Clear Javadoc comments explain the class, its purpose, and how to use its methods.
* **Consistent null handling**: All setValue methods now allow null values, preventing unnecessary exceptions.

Overall, this PR enhances the MatParamTexture class by streamlining the code, providing proper documentation, and ensuring consistent behavior.

--Edit:
Instead of overriding `setValue` in `MatParamTexture`, we can leverage the existing behavior in the superclass. The superclass method likely performs type and value checking, making an override unnecessary in this case (see https://github.com/jMonkeyEngine/jmonkeyengine/pull/1797).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/MatParamTexture.java
+++ b/jme3-core/src/main/java/com/jme3/material/MatParamTexture.java
@@ -81,7 +81,7 @@ public class MatParamTexture extends MatParam {
      * Sets the texture associated with this material parameter.
      *
      * @param value the texture object to set
-     * @throws IllegalArgumentException if the provided value is not a {@link Texture}
+     * @throws RuntimeException if the provided value is not a {@link Texture}
      */
     public void setTextureValue(Texture value) {
         setValue(value);
--- a/jme3-core/src/main/java/com/jme3/material/MatParamTexture.java
+++ b/jme3-core/src/main/java/com/jme3/material/MatParamTexture.java
@@ -86,22 +86,6 @@ public class MatParamTexture extends MatParam {
     public void setTextureValue(Texture value) {
         setValue(value);
     }
-    
-    /**
-     * Overrides the base class {@link MatParam#setValue(Object)} to allow null
-     * values.
-     *
-     * @param value the object to set as the value (must be a {@link Texture} or null)
-     * @throws IllegalArgumentException if the provided value is not a {@link Texture} and not null
-     */
-    @Override
-    public void setValue(Object value) {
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/material/MatParamTexture.java
```diff
@@ -40,51 +40,83 @@
...
+     * values.
+     *
+     * @param value the object to set as the value (must be a {@link Texture} or null)
+     * @throws IllegalArgumentException if the provided value is not a {@link Texture} and not null
+     */
     @Override
     public void setValue(Object value) {
-        if (!(value instanceof Texture)) {
-            throw new IllegalArgumentException("value must be a texture object");
+        if ((value == null) || value instanceof Texture) {
```
stephengold, line 99: seems to me this changes the behavior of `setValue(null)`. Was that your intention?

capdevon: yes, currently invoking the MatParamTexture.setValue(null) method raises an exception, while invoking the MatParam.setValue(null) method does not.

It seems reasonable to remove the setValue method from the MatParamTexture class.

The superclass, MatParam, already has a setValue method that performs all the required checks.
By removing the setValue method from MatParamTexture, it would inherit the behavior from the superclass, which aligns with the intended functionality.

stephengold: Go ahead and remove the overriding method. That seems the cleanest solution.

capdevon: Done

stephengold (approved):
looks good to me

PR #2244, opened 2024-04-18
Cleans up warnings in jme3-desktop

MelVimL:
As written in https://hub.jmonkeyengine.org/t/2079-warnings/47608/3 I tried to remove some warnings. 

For the **raw type** warnings, I replaced it with a simple wildcard; maybe I should change some of them to a lower-bound wildcard. Any thoughts on that?

 ```
Class<JmeContext> ctxClazz = (Class<JmeContext>) Class.forName(className);
            return ctxClazz.getDeclaredConstructor().newInstance();
``` 
I think this doesn't work while I am writing this.

Ressource Warning in **/cursors/plugins/CursorLoader.java** was fixed by closing DataInput
the _int flag_ wasn't and what the comment implies shoundn't be used so I removed it and nextInt into "nirvana".

**/system/AWTComponentRenderer.java** and **system/AWTFrameProcessor.java** had a deprecated FrameBuffer creation. I added **AWTUtils** with the function _getFrameBuffer(...)_ and suppressed the warning. Is there a better way to create the buffer?

**jme3tools/converters/ImageToAwt.java** had some unused "constants" I suppressed the warning. I also removed an unused constructor.

### DIFF
--- a/jme3-desktop/src/main/java/com/jme3/cursors/plugins/CursorLoader.java
+++ b/jme3-desktop/src/main/java/com/jme3/cursors/plugins/CursorLoader.java
@@ -95,6 +95,8 @@ public class CursorLoader implements AssetLoader {
         }
     }
 
+
+    // @SuppressWarnings("resource")
     private JmeCursor loadCursor(InputStream inStream) throws IOException {
 
         byte[] icoimages = new byte[0]; // new byte [0] facilitates read()
@@ -107,7 +109,7 @@ public class CursorLoader implements AssetLoader {
             int steps = 0;
             int width = 0;
             int height = 0;
-            int flag = 0; // we don't use that.
+            //int flag = 0; // we don't use that.
             int[] rate = null;
             int[] animSeq = null;
             ArrayList<byte[]> icons;
@@ -135,7 +137,7 @@ public class CursorLoader implements AssetLoader {
                             height = leIn.readInt();
                             leIn.skipBytes(8);
                             jiffy = leIn.readInt();
-                            flag = leIn.readInt();
+                            leIn.readInt(); // flag 
                             nextInt = leIn.readInt();
                         } else if (nextInt == 0x65746172) { // found a 'rate' of animation
 //                            System.out.println("we have 'rate'.");
@@ -207,8 +209,10 @@ public class CursorLoader implements AssetLoader {
                 return setJmeCursor(ciDat);
 
             } else if (riff == 0x58464952) {
+                ((LittleEndien) leIn).close();
                 throw new IllegalArgumentException("Big-Endian RIFX is not supported. Sorry.");
             } else {
+                ((LittleEndien) leIn).close();
                 throw new IllegalArgumentException("Unknown format.");
             }
         } else if (isCur || isIco) {
--- a/jme3-desktop/src/main/java/com/jme3/system/AWTFrameProcessor.java
+++ b/jme3-desktop/src/main/java/com/jme3/system/AWTFrameProcessor.java
@@ -49,13 +49,15 @@ import com.jme3.renderer.RenderManager;
 import com.jme3.renderer.ViewPort;
 import com.jme3.renderer.queue.RenderQueue;
 import com.jme3.texture.FrameBuffer;
-import com.jme3.texture.Image;
 
 /**
  * A frame processor that enables to render JMonkey frame buffer onto an AWT component.
  * <p>
- * This class is based on the <a href="http://www.oracle.com/technetwork/java/javase/overview/javafx-overview-2158620.html">JavaFX</a> original code provided by Alexander Brui (see <a href="https://github.com/JavaSaBr/JME3-JFX">JME3-FX</a>)
+ * This class is based on the
+ * <a href="http://www.oracle.com/technetwork/java/javase/overview/javafx-overview-2158620.html">JavaFX</a>
+ * original code provided by Alexander Brui (see <a href="https://github.com/JavaSaBr/JME3-JFX">JME3-FX</a>)
  * </p>
+ * 
  * @author Julien Seinturier - COMEX SA - <a href="http://www.seinturier.fr">http://www.seinturier.fr</a>
  * @author Alexander Brui (JavaSaBr)
  *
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: jme3-desktop/src/main/java/com/jme3/cursors/plugins/CursorLoader.java [hunk 1]
```diff
@@ -107,7 +109,7 @@ private JmeCursor loadCursor(InputStream inStream) throws IOException {
             int steps = 0;
             int width = 0;
             int height = 0;
-            int flag = 0; // we don't use that.
+            //int flag = 0; // we don't use that.
```
richardTingle, line 112: Why comment out rather than delete?

jme3-desktop/src/main/java/com/jme3/system/AWTFrameProcessor.java
```diff
@@ -417,12 +435,17 @@ public void bind(Component destination, Application application, ViewPort viewPo
...
+     * @param destination
+     *            the destination.
+     * @param application
+     *            the application.
+     * @param viewPort
+     *            the view port.
+     * @param main
+     *            true if this processor is main.
+     */
+    public void bind(final Component destination, final Application application, ViewPort viewPort,
```
richardTingle, line 447: If these redundant finals were removed it might fit on a single line?

jme3-desktop/src/main/java/com/jme3/cursors/plugins/CursorLoader.java
MelVimL: Good Question. I have no strong opinion on keeping the comment :)

jme3-desktop/src/main/java/com/jme3/system/AWTFrameProcessor.java
```diff
@@ -548,12 +576,16 @@ protected void unbindListeners() {
...
-     * @param fixAspect true to fix the aspect ratio.
+     * @param width
+     *            the width.
+     * @param height
+     *            the height.
+     * @param fixAspect
+     *            true to fix the aspect ratio.
      * @return the new frame transfer.
      */
-    protected AWTComponentRenderer reshapeInThread(final int width, final int height, final boolean fixAspect) {
```
richardTingle, line 556: This line was only 112 characters. Surely that can't be too long?

jme3-desktop/src/main/java/com/jme3/system/JmeDesktopSystem.java
```diff
@@ -158,10 +158,10 @@ private JmeContext newContextLwjgl(AppSettings settings, JmeContext.Type type) {
         return null;
     }
 
-    @SuppressWarnings("unchecked")
+    //@SuppressWarnings("unchecked")
```
richardTingle, line 161: Should be removed not commented out

jme3-desktop/src/main/java/jme3tools/converters/ImageToAwt.java
```diff
@@ -87,9 +87,9 @@ public DecodeParams(int bpp, int rm, int rs, int im, int is, boolean alpha){
             this.is = is;
         }
 
-        public DecodeParams(int bpp, int rm, int rs, int im, int is){
-            this(bpp, rm, rs, im, is, false);
-        }
+        /*
+         * public DecodeParams(int bpp, int rm, int rs, int im, int is){ this(bpp, rm, rs, im, is, false); }
+         */
```
richardTingle, line 92: Should be removed rather than commented out

jme3-desktop/src/main/java/jme3tools/converters/ImageToAwt.java
```diff
@@ -104,13 +104,21 @@ public DecodeParams(int bpp, int rm, int rs, int im, int is){
         final int mxxxx = 0xffffffff;
         final int sxxxx = 0;
 
+        @SuppressWarnings("unused")
```
richardTingle, line 107: A comment (just in front of the first one) about why these really are used would be good for future understanding

jme3-desktop/src/main/java/jme3tools/converters/MipMapGenerator.java
```diff
@@ -87,7 +87,7 @@ public static void generateMipMaps(Image image){
         BufferedImage original = ImageToAwt.convert(image, false, true, 0);
         int width = original.getWidth();
         int height = original.getHeight();
-        int level = 0;
+        // int level = 0;
```
richardTingle, line 90: Should be removed

jme3-desktop/src/main/java/com/jme3/system/AWTUtils.java
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * This should be a Temporary solution. FrameBuffer functions and Image-Formats are deprecated.
+     * 
+     * @param width
+     * @param height
+     * @param samples
+     * @return
+     */
+    @SuppressWarnings("deprecation")
+    public static FrameBuffer getFrameBuffer(int width, int height, int samples) {
```
richardTingle, line 16: Should this have @SuppressWarnings("deprecation")? The warnings are valid. More generally. Why extract this out into a seperate method (if we want to discourage using these deprecated things we don't want to make reuse easier)

richardTingle:
Thank you for doing this, a page full of warnings is always a pain

jme3-desktop/src/main/java/com/jme3/cursors/plugins/CursorLoader.java
stephengold: Since everything's under source control, code should be commented out only if there's a plausible reason to un-comment it later. And that rationale only works if it's obvious to someone reading the code what effects un-commenting would have.

jme3-desktop/src/main/java/jme3tools/converters/ImageToAwt.java
```diff
@@ -87,12 +88,19 @@ public DecodeParams(int bpp, int rm, int rs, int im, int is, boolean alpha){
             this.is = is;
         }
 
-        /*
-         * public DecodeParams(int bpp, int rm, int rs, int im, int is){ this(bpp, rm, rs, im, is, false); }
-         */
+        @SuppressWarnings("unused")
```
richardTingle, line 91: Why is this getting a @SuppressWarnings("unused")? There might be legitimate reasons for this suppression but a comment to explain why the warnings are being supressed would be nice.

(I'm surprised a constructor has unused methods. Sometimes it makes sense for a method if it is overridden and the child class needs them)

PR #2248, opened 2024-04-24
New data types supported by class UserData

JNightRider:
Hello dear community.

This PR consists of integrating 3 new types of data.

- enum
- BigInteger
- BigDecimal

I think the 'enum' type is something missing in class `UserData` (essential), the other 2 types are an extra (if the user uses big-numbers).

Since class `UserData` is **public**; The user can use said class to instantiate an object of the same, it is better that the data type be defined internally without the user having direct intervention in it.

What do you think of these new types of data?

[ **NOTE** ]
Modify the `TestUserData` examples class to test the new data types

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2023 jMonkeyEngine
+ * Copyright (c) 2009-2024 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


capdevon:
Hi @JNightRide ,
Thanks for your suggestions! While I appreciate your effort, I believe the current `UserData` class already handles enum properties effectively. Similar to **Gson** and many other libraries, enums are automatically represented by their string values during _serialization/deserialization_.

```java
//e.g.
        Spatial sp = null;
        for (MyEnum val : MyEnum.values()) {
            sp.setUserData(val.name(), val.ordinal());
        }
```
Regarding `BigInteger` and `BigDecimal`, considering the intended use of UserData for storing basic information or game object data, I don't see a need for these data types at this time. The existing primitive types (_int, float, double, etc._) should be sufficient for most scenarios.

JNightRide:
Hi @capdevon.

> I believe the current UserData class already handles enum properties effectively. Similar to Gson and many other libraries, enums are automatically represented by their string values during serialization/deserialization.

It is true that by doing this process you can store data of type 'enum':
```java
//e.g.
        Spatial sp = null;
        for (MyEnum val : MyEnum.values()) {
            sp.setUserData(val.name(), val.ordinal());
        }
```

However, I think we can save it another way, for better encapsulation when the user is using it:

```java
// set
Spatial spatial = null;
spatial.setUserData("1", MyEnum.ENUM_01);
        
// get
MyEnum val = spatial.getUserData("1");
```
When retrieving the data, it will return the enumerated object itself (not a string or integer value), thus preventing the user from searching (converting) its data to an `enum`...

> Regarding BigInteger and BigDecimal, considering the intended use of UserData for storing basic information or game object data, I don't see a need for these data types at this time. The existing primitive types (int, float, double, etc...) should be sufficient for most scenarios.

Regarding the data of type `BigInteger` and `BigDecimal `it may not be necessary now as you mentioned (it was like another option for numbers)...

JNightRider:
I will close this PR due to lack of interest....

Issue #2249, opened 2024-04-24
comparison methods for BoundingBox and BoundingSphere


stephengold:
Currently if you invoke `equals()` on a `BoundingBox` or `BoundingSphere` you get `Object.equals()`, which ignores the object's components. Usually that isn't what's desired.

By analogy with com.jme3.math classes such as `Matrix4f`, `Quaternion`, and `Vector3f`, these 2 classes should:
+ override the `equals()` method to test for precise equality of all components (distinguishing -0 from 0) AND
+ provide an `isSimilar()` method to test for approximate equality of all components.

PR #2252, opened 2024-05-04
implement equals(), hashCode(), and isSimilar() for bounding volumes

stephengold:
This PR addresses the enhancements request in issue #2249.

Since the PR overrides `equals()`, it must also override `hashCode()` because the contract of `hashCode()` specifies that objects that are equal must return the same code.
 
The new methods intentionally ignore the `checkPlane` field, which holds auxiliary data used by the `Camera` class.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/bounding/BoundingBox.java
+++ b/jme3-core/src/main/java/com/jme3/bounding/BoundingBox.java
@@ -613,11 +613,9 @@ public class BoundingBox extends BoundingVolume {
             return false;
         } else if (Float.compare(zExtent, otherBoundingBox.zExtent) != 0) {
             return false;
-        } else if (!center.equals(otherBoundingBox.getCenter())) {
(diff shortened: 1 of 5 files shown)


stephengold:
@pspeed42: If you've got a moment, I'd like your opinion on whether `equals()` and `hashCode()` should ignore the `checkPlane` field or not.

pspeed42:
It probably should ignore... but probably also that's up to the class that contains the field.  It would maybe be more forward-thinking of these subclasses called super.equals() and super.hashCode() as part of their processing rather than hoping to grab the right fields from the super class (which in my opinion should have been private fields in the first place... but that's a different religious hill to die on another day).
Also, was there a reason that Objects.hash() wasn't used vs hand-coding?

stephengold:
Thanks for your insights, Paul.

stephengold:
ready for review

PR #2090, opened 2023-10-10
A framegraph framework that supports 3 render paths

JohnLKkk:
see:[https://hub.jmonkeyengine.org/t/enhancements-to-the-jme3-rendering-engine/47145/27](url)

### Enhance the capabilities of the JME3.7 renderer, mainly including:
1. Introduce a basic framegraph framework for managing renderPasses, while modularizing the existing rendering modules into several core Passes for reuse;
2. Added support for render paths, including Forward (default render path), Deferred (a standard DeferredShading), and TileBasedDeferred (Tile-optimized DeferredShading). They should be compatible with most existing modules.
3. Added some code demos for RenderPath.

Regarding render paths, please refer to: https://docs.unity3d.com/Manual/RenderingPaths.html

### Performance considerations
The rendering overhead of real-time lights in deferred shading is proportional to the number of pixels illuminated by the light and not dependent on Scene complexity. So small Point Lights or Spot Lights are very cheap to render and if they are fully or partially occluded by Scene GameObjects then they are even cheaper.
Of course, lights with shadows are much more expensive than lights without shadows. In deferred shading, shadow-casting GameObjects still need to be rendered once or more for each shadow-casting light. Furthermore, the lighting shader that applies shadows has a higher rendering overhead than the one used when shadows are disabled.

### GBuffer format: 
RT0, RGBA16F: Pack DiffuseColor, Ao, Alpha
RT1, RGBA16F: Pack SpecularColor, FZero, Roughness
RT2, RGBA16F: Pack ShadingModelId, emissiveColor
RT3, RGBA32F: Pack Normal (contains Normal with TBN transform and modelNormal, using cubemap mapping)
RT4, depth: Pack depth

### LightPack:
There are two ways to package lightData, one of which is to use a uniform array, which should be faster, but its quantity is limited, and a large number of light sources will be drawn and accumulated through several passes to complete all light sources;
Another way is to package all light source information through three texture1D, which may have some performance overhead, but supports a large amount of light source data;

### Tile-Based DeferredShading:
TileLightDecode: Packs tile information, x indicates the first light source offset used by the tile (for lookup in TileLightIndex), y indicates how many lights are associated with the tile (currently no limit, i.e. unlimited number of lights), z indicates the uv coordinate offset when sampling TileLightIndex for the tile; 
TileLightIndex: Packs light source ids for each tile

### UseRenderPath:
Enable the corresponding renderPath in the following ways：
[...]

PR #2256, opened 2024-05-07
FrameGraph API

codex128:
This PR adds a FrameGraph rendering pipeline API that is similar in concept to shader nodes. This is based off work from #2090.

For more information:
https://hub.jmonkeyengine.org/t/framegraph-api/47646?u=codex

To Do:
* Clean up several scattered compile errors.
* Implement Forward+ and/or Forward++ (save for another PR?)
* Test FrameGraph under more conditions.
* Test global illumination?
* Javadoc and licenses.

Issues to be investigated further:
* Some unexplained difficulties in combining results from multiple ViewPorts running FrameGraphs in some circumstances.
* Slight artifacts on shadow mappings. This may be a regression of unrelated code on the fork, and not any fault of the FrameGraph.

Any testing is welcome!


codex128:
This PR is ready for review and testing.

To speed things along, I will try to explain how the resource handling works:

Each FrameGraph has a **ResourceList**, which handles **RenderResources**. RenderResources are *not* actual resources  (called **RenderObjects** in the code) , they are simply promises of an actual resource that will exist in the future. **RenderPasses** do not have direct access to the RenderResources, but can perform operations on them through **ResourceTickets**, which are essentially indices pointing to particular RenderResources within an array.

When a RenderPass makes a request for a RenderObject from the ResourceList using a ResourceTicket (called *acquiring*), the ResourceList will check if the RenderResource located by the ResourceTicket is *virtual* (is not associated with a RenderObject). If not, great! The associated RenderObject is simply returned. But if the RenderResource is virtual, the ResourceList asks the **RenderObjectMap** (held by the RenderManager) to either create a new RenderObject to use or find an existing one to reallocate.

The RenderObjectMap considers creating new RenderObjects expensive, so it tries to reallocate existing RenderObjects wherever it can. The **ResourceDef** (which defines the behavior of a RenderResource and the associated RenderObject) attached to the RenderResource is used to determine which RenderObjects are suitable for reallocation. Also the RenderObjectMap doesn't like to keep unused RenderObjects alive, so once a RenderObject has gone a number of frames without being used, it is disposed.

As an example to better understand how render passes and resource handling works, here is a basic render pass that performs downsampling. That is, it takes an input texture and renders to an output texture that is a quarter of the size.

```java
public class MyRenderPass extends RenderPass {
    
    private ResourceTicket<Texture2D> inTex, outTex;
    private TextureDef<Texture2D> texDef;
    
    @Override
    protected void initialize(FrameGraph frameGraph) {
        
        // Initialize is called when the pass is assigned to the FrameGraph.
        
        // Add an input named "Texture". This allows this pass to access a texture
        // from another pass.
        inTex = addInput("Texture");
        
        // Add an output named "Texture". This allows other passes to access the
        // resulting texture.
        outTex = addOutput("Texture");
        
        // Declare the texture definition, defining how the output texture is created
        // and how it behaves.
        texDef = new TextureDef(Texture2D.class, img -> new Texture2D(img));
        
    }
    @Override
    protected void prepare(FGRenderContext context) {
        
        // Prepare is called before execution for every ViewPort rendered, and is used primarily to determine
        // what resources are referenced where, so that unused passes can be culled.
        
        // Note that a prepare call does not necessarily proceed a execute call, due to culling.
        
        // Declare a new resource as the output texture
        declare(texDef, outTex);
        
        // Reserve the output texture. This greatly increases the chances of getting
        // the same texture from frame to frame, and thus minimizes texture binds.
        reserve(outTex);
        
        // Reference the input texture, so this pass can safely access it
        reference(inTex);
        
    }
    @Override
    protected void execute(FGRenderContext context) {
    
        // Execute is called to perform the rendering operations. This is not guaranteed
        // to be called for every ViewPort rendered, because of culling.
    
        // Acquire the input texture provided by another pass
        Texture2D texture = resources.acquire(inTex);
        Image img = texture.getImage();
        
        // Set the size of the output texture as half the input texture on each demension
        int w = img.getWidth() / 2;
        int h = img.getHeight() / 2;
        texDef.setSize(w, h);
        
        // Set the format of the output texture to match the input texture
        texDef.setFormat(img.getFormat());
        
        // Get a framebuffer object matching the width, height, and samples. If no
        // such framebuffer exists, a new one will be created. This is handled only by
        // the RenderPass superclass and not the resource manager.
        FrameBuffer fb = getFrameBuffer(w, h, 1);
        
        // Acquire the output texture, which is immediately attached to the framebuffer.
        // This operation is specifically designed to limit texture binds and framebuffer updates.
        resources.acquireColorTargets(fb, outTex);
        
        // Attach the framebuffer for rendering
        context.getRenderer().setFrameBuffer(fb);
        context.getRenderer().clearBuffers(true, true, true);
        context.getRenderer().setBackgroundColor(ColorRGBA.BlackNoAlpha);
        
        // Render the input texture to the output texture on a fullscreen quad.
        // The depth texture is null, so the quad is rendered at a depth of one.
        context.renderTextures(texture, null);
        
    }
    @Override
    protected void reset(FGRenderContext context) {
        // Reset is called after execution. Passes cannot be culled from this step.
    }
    @Override
    protected void cleanup(FrameGraph frameGraph) {
        // Cleanup is called when this pass is removed from the FrameGraph.
    }
    
}
```

Further details for each function and class can be found in the relavent Javadoc.

codex128:
This branch does conflict with master, but fortunately it only is a small conflict in some Javadoc.

codex128:
After fixing several issues, the branch now builds successfully.

codex128:
Due to some development advances, the entire FrameGraph system is now able to be moved to an [external library](https://github.com/codex128/FrameGraph), which is way better than trying to merge it into core. The external library still depends on some changes on this branch, so I will open a new pull request for that later today.

Issue #2275, opened 2024-06-01
Invalid-matrix4f-for-worldviewprojectionmatrix assertion is hard to diagose


richardTingle:
As seen in https://hub.jmonkeyengine.org/t/solved-error-invalid-matrix4f-for-worldviewprojectionmatrix/47680 and https://hub.jmonkeyengine.org/t/jmonkeyengine-v3-7-0-beta1-release/47653/12 the new assertion in GLRenderer catches incorrect setting of an objects position to infinity or NaN. However the assertion doesn't give much of a clue where the mistake was made. Just that it has.

I'm proposing adding some assertions to the Transform class in order to catch these mistakes "when they go in".

PR #2276, opened 2024-06-01
#2275 Improve assertions for invalid transforms.

richardTingle:
These are already caught by assertions in GLRenderer, but by that time it is unclear what call pushed in the bad values

closes https://github.com/jMonkeyEngine/jmonkeyengine/issues/2275

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Quaternion.java
+++ b/jme3-core/src/main/java/com/jme3/math/Quaternion.java
@@ -1613,7 +1613,7 @@ public final class Quaternion implements Savable, Cloneable, java.io.Serializabl
      * Tests whether the argument is a valid quaternion, returning false if it's
      * null or if any component is NaN or infinite.
      *
-     * @param quaternion the vector to test (unaffected)
+     * @param quaternion the quaternion to test (unaffected)
      * @return true if non-null and finite, otherwise false
      */
     public static boolean isValidQuaternion(Quaternion quaternion) {
--- a/jme3-core/src/main/java/com/jme3/math/Quaternion.java
+++ b/jme3-core/src/main/java/com/jme3/math/Quaternion.java
@@ -1608,4 +1608,27 @@ public final class Quaternion implements Savable, Cloneable, java.io.Serializabl
             throw new AssertionError(); // can not happen
         }
     }
+
+    /**
+     * Tests whether the argument is a valid quaternion, returning false if it's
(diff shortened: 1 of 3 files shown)


richardTingle:
closes #2275

stephengold:
I'm generally supportive of these changes.

I recommend using the existing `Vector3f.isValidVector()` instead of defining a private method for that purpose.
Furthermore, I suggest moving `isValidQuaternion()` to `com.jme3.math.Quaternion` and making it a public method, to encourage re-use.
Finally, for `isValidNumber()` I think we should use `Float.isFinite()`, which was added in Java 1.8 .

richardTingle:
Sure @stephengold I've made those changes

stephengold:
I see you also added an automated test. That's great, but it needs javadoc and a copy of the JME license.

stephengold:
Thanks for your quick responses @richardTingle. Just a few more small things ...

jme3-core/src/main/java/com/jme3/math/Quaternion.java
```diff
@@ -1608,4 +1608,27 @@ public Quaternion clone() {
             throw new AssertionError(); // can not happen
         }
     }
+
+    /**
+     * Tests whether the argument is a valid quaternion, returning false if it's
+     * null or if any component is NaN or infinite.
+     *
+     * @param quaternion the vector to test (unaffected)
```
stephengold, line 1616: The parameter is a quaternion, not a vector.

jme3-core/src/test/java/com/jme3/math/QuaternionTest.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2023 jMonkeyEngine
```
stephengold, line 2: It's 2024, not 2023.

stephengold (approved):

jme3-core/src/test/java/com/jme3/math/QuaternionTest.java
richardTingle: Well that's embarrassing. That isn't a copy-paste fail. I removed the year from the one I got the copyright notice from, thought "what year is it?" and wrote 2023. Anyway I've corrected that and the other issue

stephengold: Thanks for taking care of things.

Issue #2277, opened 2024-06-01
GltfLoader issues


capdevon:
The recent changes to the GLTF loader have introduced several problems:

1. **Duplicate Controllers**: The new loader creates duplicate `AnimComposer` and `SkinningControl` controllers for each geometry in the model. (see [here](https://hub.jmonkeyengine.org/t/jmonkeyengine-v3-7-0-beta1-release/47653/30?u=capdevon))
2. **Nested Geometry**: Each geometry is now encapsulated within two parent nodes, which can complicate scene hierarchy management.
3. **Backward Compatibility**: These changes may cause compatibility issues with existing code that relies on the previous behavior of the GLTF loader. (see [here](https://hub.jmonkeyengine.org/t/jmonkeyengine-v3-7-0-beta1-release/47653/13?u=capdevon))

Other issues reported by adi.barda: (see [here](https://hub.jmonkeyengine.org/t/jmonkeyengine-v3-7-0-beta1-release/47653/5?u=capdevon))
* Blured colors - maybe normal map is missing?
* Sometime missing animations - maybe related to 1

stephengold:
Please provide detailed instructions for reproducing this issue.
Has anyone identified which recent changes caused this issue?

capdevon:
Maybe this: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2103

stephengold:
Would reverting 2103 solve this issue?

capdevon:
> Please provide detailed instructions for reproducing this issue.

Here is the 3D Model for the test [file](https://github.com/capdevon/sharefile/blob/main/YBot.gltf)

jME-3.7.0
![image](https://github.com/jMonkeyEngine/jmonkeyengine/assets/32201701/6e1485a3-3a3a-436b-acf2-7c9e3e423d07)


jME-3.6.1-stable
![image](https://github.com/jMonkeyEngine/jmonkeyengine/assets/32201701/0fe233ce-bd5c-4b5b-9167-dbf98878fd6e)

stephengold:
The issue was introduced by PR #2060.

With "master" branch at f915e56, the test app prints `numComposers = 1`.
With "master" branch at 4c87531, the test app prints `numComposers = 2`.

riccardobl:
> 2. Nested Geometry: Each geometry is now encapsulated within two parent nodes, which can complicate scene hierarchy management.
[...]

PR #2278, opened 2024-06-07
partial revert of PR #2060 (JSON parser abstraction)

stephengold:
The goal of this pull request is to solve the "multiple controls" portion of issue #2277. (It might solve the entire issue; I'm unsure.)

This switches JME's glTF loader back, from using the jme3-plugins-json and jme3-plugins-json-gson libraries to using Google's gson library directly.

This is only a *partial* revert of PR #2060 because it leaves the added jme3-plugins-json and jme3-plugins-json-gson libraries intact. Perhaps they can still be used.


scenemax3d:
@stephengold do you know why the  jme3-plugins JSON implementation caused this issue?

stephengold:
No, I don't.

PR #2286, opened 2024-06-20
TestBrickTower:  delete secondary version of the BSD license

stephengold:
JMonkeyEngine has a policy of including a copyright notice and a copy of its (BSD) license in every Java source file.

TestBrickTower.java is a very old physics test located in the jme3-examples subproject.

Back in October 2012 (at commit 01f1c8f) someone inserted a 2nd copy of the copyright notice and license into the file, probably by accident.

This PR deletes the extra copy. It makes no change to executable code, only comments are affected.


richardTingle:
Looks good to me

PR #2291, opened 2024-07-06
Remove dead comments from gradle files

8Keep:
Deleting some dead code here to simplify the gradle files.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -245,43 +245,6 @@ if (skipPrebuildLibraries != "true" && buildNativeProjects != "true") {
     }
 }
 
-
-//class IncrementalReverseTask extends DefaultTask {
-//    @InputDirectory
-//    def File inputDir
-//
-//    @OutputDirectory
-//    def File outputDir
-//
-//    @Input
-//    def inputProperty
-//
-//    @TaskAction
-//    void execute(IncrementalTaskInputs inputs) {
-//        println inputs.incremental ? "CHANGED inputs considered out of date" : "ALL inputs considered out of date"
-//        inputs.outOfDate { change ->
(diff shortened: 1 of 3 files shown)


jme3-examples/build.gradle
```diff
@@ -20,12 +20,9 @@ dependencies {
     implementation project(':jme3-jbullet')
     implementation project(':jme3-jogg')
     implementation project(':jme3-lwjgl')
-//    implementation project(':jme3-lwjgl3')
```
stephengold, line 23: This commented-out line makes it easier for people to switch between LWJGL v2 and v3, which is something I used to do a lot during beta testing of releases. With a little scripting, that functionality could be moved the gradle.properties file, which might make it easier to find.

jme3-examples/build.gradle
```diff
@@ -20,12 +20,9 @@ dependencies {
...
     implementation project(':jme3-networking')
     implementation project(':jme3-niftygui')
     implementation project(':jme3-plugins')
-//    implementation project(':jme3-plugins-json')
```
stephengold, line 27: I commented these out as part of PR #2278. I'm hoping they can be put back in after somebody figures out why the jme3-plugins JSON implementation wasn't working.

8Keep:
Reverted changes to jme3-examples build file

stephengold:
Thank you, @8Keep.
The rest of the dead comments are mysteries to me.
Have you looked at the commits that added them, to understand why they're there?

8Keep:
The ones I left in this commit were last changed 10+ years ago, I am pretty sure they were just left by the authors

stephengold:
You're probably right! Thanks.

PR #2292, opened 2024-07-06
Add java 21 to the CI matrix

8Keep:
As well, start deploying from java 21, bumped from java 17.

So, now we have all 3 java LTS's building, and deploy from the latest LTS.


stephengold (approved):
Looks good to me.

PR #2293, opened 2024-07-06
Use windows-latest image for windows CI

8Keep:
Use windows-latest image (windows server 2022) instead of windows 2019


stephengold (approved):
Looks good to me.

PR #2294, opened 2024-07-08
Fix deprecated gradle features, prepare for gradle 9

8Keep:
With this change, gradle 8.8 no longer complains about using deprecated gradle features that will be removed in 9, so we will have an easier time upgrading next time.


MeFisto94 (approved):

stephengold (approved):
Looks good. Thank you, @8Keep!

PR #2295, opened 2024-07-08
Upgrade spotbugs to 4.8.6

8Keep:
Upgrading spotbugs versions. 

To run spotbugs, I run:

./gradlew build -PenableSpotBugs=true

This generates the html report in build/reports/spotbugs, for each gradle subproject. I can see that there are a whole bunch of new lints in there for the engine. Because we don't block on spotbugs, this won't start failing any CI or builds, nor will it print any warnings while building. It just creates the report, when you turn on spotbugs with the flag, so this should be a safe change. We can later look at the reports and start fixing some of the issues.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:4.2.0'
         classpath 'me.tatarka:gradle-retrolambda:3.7.1'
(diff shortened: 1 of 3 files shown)


8Keep:
Squash these commits when merging

stephengold:
Thanks for your contribution, @8Keep !

PR #2297, opened 2024-07-17
BatchNode: Fix IndexOutOfBoundsException

### LINKED ISSUE #2296: BatchNode can throw IndexOutOfBoundsException
There are temporary arrays that are reused to work with vertex buffers.
Some conditions cause the index to be used for storing more vertices that they can hold and this throws the exception.

A proposed solution is to validates the arrays' size before accessing them, recreating the arrays if they are too small.

I commented this in more detail in the hub: https://hub.jmonkeyengine.org/t/batchnode-and-indexoutofboundsexception/47766

jcfandino:
- There are temporary arrays that are reused to work with vertex buffers.

- Some conditions cause the arrays to be used for storing more vertices that they can hold.

- This throws the exception.

- This fix validates the arrays size before accessing them, recreating the arrays if they are too small.

Fixes #2296

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/BatchNode.java
+++ b/jme3-core/src/main/java/com/jme3/scene/BatchNode.java
@@ -246,13 +246,7 @@ public class BatchNode extends GeometryGroupNode {
 
         //init the temp arrays if something has been batched only.
         if (matMap.size() > 0) {
-            //TODO these arrays should be allocated by chunk instead to avoid recreating them each time the batch is changed.
-            //init temp float arrays
-            tmpFloat = new float[maxVertCount * 3];
-            tmpFloatN = new float[maxVertCount * 3];
-            if (useTangents) {
-                tmpFloatT = new float[maxVertCount * 4];
-            }
+            initTempFloatArrays();
         }
     }
 
@@ -387,7 +381,6 @@ public class BatchNode extends GeometryGroupNode {
         int maxWeights = -1;
 
         Mesh.Mode mode = null;
-        float lineWidth = 1f;
         for (Geometry geom : geometries) {
             totalVerts += geom.getVertexCount();
             totalTris += geom.getTriangleCount();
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/scene/BatchNode.java
```diff
@@ -387,7 +381,6 @@ private void mergeGeometries(Mesh outMesh, List<Geometry> geometries) {
         int maxWeights = -1;
 
         Mesh.Mode mode = null;
-        float lineWidth = 1f;
```
stephengold, line 390: Why was this line removed?

jcfandino: The variable isn't referenced anywhere else in the function, it's simply dead code.

stephengold:
@jcfandino thanks for your contribution

stephengold:
This hasn't yet been integrated into the v3.7 branch, so it's not in 3.7.0-stable.

pspeed42:
For what it's worth, I'm running an older version of JME that predates this fix and I ran across this during my live stream.  After a little debugging, I know the exact cause of the issue and can guide to a more direct fix to the problem.

The issue stems from the fact that doBatch() will only recalculate the maxVertCount for the geometries that it is rebatching.  A specific scenario where this causes problems is:
1. have a bunch of larger meshes for one material (material A)
2. add a few smaller meshes for another material (material B)
3. batch()
4. move one of the geometries for material A.

The call to batch() would have recalculated maxVertCount as a value that is too small to redo the transforms of any of the objects for material A.  This is because doBatch()'s gatherGeometries() call will only return the smaller geometries... the original material A geometries don't need to be rebatched.

So while this PR is a really nice stop-gap to catch bugs and continue running, it didn't actually fix the bug... just hides it.  (I still think the validateTempFloatArrays() is a good idea.)  As it stands, though, the 'error case' will churn through more temporary buffer allocation than is needed.  Once as an incorrect size when doBatch() is called and then again as a potentially incorrect size every time a larger geometry is moved.

It's possible to get the maxVertCount right in doBatch().

Since gatherGeometries() is going to iterate over all of the geometries, anyway... this is where we should be calculating the maxVertCount.
 
I don't have anything setup to do commits (yet) in modern jme github but I'm attaching a patch with my fix.
[0001-Alternate-fix-for-the-IndexOutOfBounds-exception-whe.patch](https://github.com/user-attachments/files/19522928/0001-Alternate-fix-for-the-IndexOutOfBounds-exception-whe.patch)

Issue #2302, opened 2024-08-18
NativeLibraryLoader fails due to no write permissions


terranprog:
Running Linux.
I am getting an exception when jme tries to extract a native library to a temp folder.

```
Aug 15, 2024 5:40:32 PM com.jme3.system.JmeDesktopSystem initialize
INFO: Running on jMonkeyEngine 3.7.0-beta1.2.2
 * Branch: HEAD
 * Git Hash: 5cadb88
 * Build Date: 2024-07-30
Exception in thread "AWT-EventQueue-0" java.io.UncheckedIOException: Failed to extract native library to: /tmp/jme3/natives_8946616b/libbulletjme-x86_64.so
	at com.jme3.system.NativeLibraryLoader.loadNativeLibrary(NativeLibraryLoader.java:534)
	at com.jme3.system.JmeDesktopSystem.initialize(JmeDesktopSystem.java:290)
	at com.jme3.system.JmeDesktopSystem.newContext(JmeDesktopSystem.java:212)
	at com.jme3.system.JmeSystem.newContext(JmeSystem.java:175)
	at com.jme3.app.LegacyApplication.createCanvas(LegacyApplication.java:560)
	at scamsoft.minecraft.terrain3d.Terrain3D.<init>(Terrain3D.java:112)
	at scamsoft.minecraft.terrain3d.gui.TerrainGui.lambda$main$12(TerrainGui.java:886)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:318)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:773)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:720)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:714)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:86)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:742)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:203)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:124)
[...]

PR #2303, opened 2024-08-19
Fix for issue #2302 - NativeLibraryLoader fails due to no write permissions

terranprog:
This fixes an an issue I encountered while running a jmonkey application in Linux. #2302 - NativeLibraryLoader fails due to no write permissions.
The cause of the bug is a bit of an edge case, but it results in the native library not being extracted and the application not starting.

The main change I made is to extract the native library to a folder directly beneath the user temp folder, instead of using an intermediate tmp/jme subfolder. The reason for this is to avoid the scenario where the /tmp/jme folder is owned by one user and another user tries to write to it.
Note that this change is optional and I could modify the PR to not include it. We can still use /tmp/jme and fall back to calling setExtractionFolderToUserCache() if the /tmp/jme folder is not writable.
 
The other changes are to tighten the logic so the code will call setExtractionFolderToUserCache() instead of crashing.
- Check the result of the call to mkdir. If it fails, call call setExtractionFolderToUserCache()
- Check if the extraction folder is actually writable. If not, call setExtractionFolderToUserCache()
- Updated log message: since a previous change we no longer extract to the working directory.

Some general notes about this PR:
I tried to limit the scope to only fix the immediate issue.
There are other issues with NativeLibraryLoader that this PR does not address
- There are other places in NativeLibraryLoader where the result of mkdir is not checked.
- It should potentially use mkdirs() instead of mkdir()

### DIFF
--- a/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
+++ b/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
@@ -198,14 +198,15 @@ public final class NativeLibraryLoader {
                 setExtractionFolderToUserCache();
             } else {
                 try {
-                    File jmeTempDir = new File(userTempDir, "jme3");
-                    if (!jmeTempDir.exists()) {
-                        jmeTempDir.mkdir();
-                    }
-                    extractionFolder = new File(jmeTempDir, "natives_" + Integer.toHexString(computeNativesHash()));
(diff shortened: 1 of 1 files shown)


MeFisto94 (approved):
You are right with mkdirs and not checking the result utimately, so we may want a tracking issue for that, it's also low hanging fruit for entry level contributions

riccardobl:
I think we can merge this @yaRnMcDonuts

PR #2305, opened 2024-08-20
GSON global version (Gradle)

JNightRider:
It seems that the gson library is used by 2 modules: `jme3-plugins and jme3-plugins-json-gson`, this can be tedious when gson has a new version and needs to be updated, it would have to be changed in 2 places (modules)

This PR mainly changes the way the dependency is implemented, the version has been defined globally the same as with lwjgl3 (version)

### DIFF
--- a/common.gradle
+++ b/common.gradle
@@ -31,7 +31,8 @@ tasks.withType(JavaCompile) { // compile-time options:
 
 ext {
     lwjgl3Version = '3.3.3' // used in both the jme3-lwjgl3 and jme3-vr build scripts
-    niftyVersion = '1.4.3' // used in both the jme3-niftygui and jme3-examples build scripts
+    niftyVersion = '1.4.3'  // used in both the jme3-niftygui and jme3-examples build scripts
+    gsonVersion = '2.9.1'   // used in both the jme3-plugins and jme3-plugins-json-gson build scripts
 }
 
 repositories {
--- a/jme3-plugins-json-gson/build.gradle
+++ b/jme3-plugins-json-gson/build.gradle
@@ -1,16 +1,12 @@
 sourceSets {
     main {
         java {
-           
             srcDir 'src/main/java'
-
         }
     }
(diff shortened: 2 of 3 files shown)


stephengold:
Thanks for your contribution, @JNightRider .

A little history: the situation this PR addresses is a direct consequence of PR #2278, which was a temporary solution for a bug in PR #2060.

Long term, it would be better to figure out why PR #2060 broke glTF importing and solve *that* issue, so we could go back to jme3-plugins not depending on the gson library at all. Such a solution would make this PR unnecessary.

Furthermore, the current best practice for coding library versions in Gradle builds is not to hard-code them into common.gradle or other Gradle scripts, but instead to use [a version catalog](https://docs.gradle.org/current/userguide/platforms.html). If you want to simplify version upgrades, that would be a better way to go.

JNightRider:
Hi @stephengold 

Thanks for reviewing this PR...

> Long term, it would be better to figure out why PR https://github.com/jMonkeyEngine/jmonkeyengine/pull/2060 broke glTF importing and solve that issue

You're right, I think it's the best option. I will do some research on that matter (this may take time)

> Furthermore, the current best practice for coding library versions in Gradle builds is not to hard-code them into common.gradle or other Gradle scripts, but instead to use [a version catalog](https://docs.gradle.org/current/userguide/platforms.html). If you want to simplify version upgrades, that would be a better way to go.

It would be interesting to apply this in regards to how lwjgl3 and nifty are versioned (maybe with gson if the problem is still not resolved - PR #2060). I would like to implement it, this would change the main goal of this PR which is the gson version but it is still the same path.

stephengold:
It would be cleaner to create 2 new PRs: one to create the version catalog and one to properly resolve the issue with PR #2060.
If a version catalog were created, this PR could be closed as obsolete.

JNightRider:
There is a PR (##2311) that replaces this...

PR #2307, opened 2024-08-22
Minor refactor to allow for saving AnimLayers

neph1:
While working on this https://github.com/jMonkeyEngine/sdk/pull/612 I found out that AnimLayers couldn't be saved.

Tested in local SDK.

I'll look it over once more before turning into a proper PR.

The tests don't explicitly test the changes, just general AnimComposer behavior.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -66,7 +66,9 @@ public class AnimComposer extends AbstractControl {
      * Instantiate a composer with a single layer, no actions, and no clips.
      */
     public AnimComposer() {
-        layers.put(DEFAULT_LAYER, new AnimLayer(this, DEFAULT_LAYER, null));
+        if(layers.isEmpty()) {
+            layers.put(DEFAULT_LAYER, new AnimLayer(DEFAULT_LAYER, null));
+        }
     }
 
     /**
@@ -314,7 +316,7 @@ public class AnimComposer extends AbstractControl {
      * @param mask the desired mask for the new layer (alias created)
      */
     public void makeLayer(String name, AnimationMask mask) {
-        AnimLayer l = new AnimLayer(this, name, mask);
+        AnimLayer l = new AnimLayer(name, mask);
         layers.put(name, l);
     }
 
@@ -399,7 +401,7 @@ public class AnimComposer extends AbstractControl {
     @Override
     protected void controlUpdate(float tpf) {
         for (AnimLayer layer : layers.values()) {
-            layer.update(tpf);
+            layer.update(tpf, globalSpeed);
         }
     }
 
@@ -542,6 +544,7 @@ public class AnimComposer extends AbstractControl {
         InputCapsule ic = im.getCapsule(this);
         animClipMap = (Map<String, AnimClip>) ic.readStringSavableMap("animClipMap", new HashMap<String, AnimClip>());
         globalSpeed = ic.readFloat("globalSpeed", 1f);
+        layers = (Map<String, AnimLayer>) ic.readStringSavableMap("layers", new HashMap<String, AnimLayer>());
     }
 
     /**
@@ -557,5 +560,6 @@ public class AnimComposer extends AbstractControl {
         OutputCapsule oc = ex.getCapsule(this);
         oc.writeStringSavableMap(animClipMap, "animClipMap", new HashMap<String, AnimClip>());
         oc.write(globalSpeed, "globalSpeed", 1f);
+        oc.writeStringSavableMap(layers, "layers", new HashMap<String, AnimLayer>());
     }
 }
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -66,9 +66,7 @@ public class AnimComposer extends AbstractControl {
      * Instantiate a composer with a single layer, no actions, and no clips.
      */
     public AnimComposer() {
-        if(layers.isEmpty()) {
-            layers.put(DEFAULT_LAYER, new AnimLayer(DEFAULT_LAYER, null));
-        }
+        layers.put(DEFAULT_LAYER, new AnimLayer(DEFAULT_LAYER, null));
     }
 
     /**
--- a/jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
@@ -85,6 +85,9 @@ public class AnimLayer implements JmeCloneable, Savable {
 
     private boolean loop = true;
     
+    /**
+    * For serialization only. Do not use.
+    */
     protected AnimLayer() {
         
     }
@@ -93,7 +96,6 @@ public class AnimLayer implements JmeCloneable, Savable {
      * Instantiates a layer without a manager or a current Action, owned by the
      * specified composer.
      *
-     * @param composer the owner (not null, alias created)
      * @param name the layer name (not null)
      * @param mask the AnimationMask (alias created) or null to allow this layer
      *     to animate the entire model
@@ -251,14 +253,15 @@ public class AnimLayer implements JmeCloneable, Savable {
      *
      * @param appDeltaTimeInSeconds the amount application time to advance the
      *     current Action, in seconds
+     * @param globalSpeed the global speed applied to all layers.
      */
-    void update(float appDeltaTimeInSeconds, float speed) {
+    void update(float appDeltaTimeInSeconds, float globalSpeed) {
         Action action = currentAction;
         if (action == null) {
             return;
         }
 
-        double speedup = action.getSpeed() * speed;
+        double speedup = action.getSpeed() * globalSpeed;
         double scaledDeltaTime = speedup * appDeltaTimeInSeconds;
         time += scaledDeltaTime;
 
@@ -313,7 +316,7 @@ public class AnimLayer implements JmeCloneable, Savable {
     public void write(JmeExporter ex) throws IOException {
         OutputCapsule oc = ex.getCapsule(this);
         oc.write(name, "name", null);
-        if(mask instanceof Savable) {
+        if (mask instanceof Savable) {
             oc.write((Savable) mask, "mask", null);
         }
     }
--- a/jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
@@ -313,8 +313,8 @@ public class AnimLayer implements JmeCloneable, Savable {
     public void write(JmeExporter ex) throws IOException {
         OutputCapsule oc = ex.getCapsule(this);
         oc.write(name, "name", null);
-        if(mask instanceof ArmatureMask) {
-            oc.write((ArmatureMask) mask, "mask", null);
+        if(mask instanceof Savable) {
+            oc.write((Savable) mask, "mask", null);
         }
     }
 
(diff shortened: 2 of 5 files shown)


jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -66,7 +66,9 @@ public class AnimComposer extends AbstractControl {
      * Instantiate a composer with a single layer, no actions, and no clips.
      */
     public AnimComposer() {
-        layers.put(DEFAULT_LAYER, new AnimLayer(this, DEFAULT_LAYER, null));
+        if(layers.isEmpty()) {
+            layers.put(DEFAULT_LAYER, new AnimLayer(DEFAULT_LAYER, null));
+        }
```
neph1, line 71: Not sure if this is needed, but feels safer in case someone deletes it and saves

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -57,16 +63,12 @@ public class AnimLayer implements JmeCloneable {
      * The name of Action currently running on this layer, or null if none.
      */
     private String currentActionName;
-    /**
-     * The composer that owns this layer. Were it not for cloning, this field
-     * would be final.
-     */
-    private AnimComposer composer;
```
neph1, line 64: Completely unnecessary field

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
neph1, line 1: I decided to only save name and animationMask. The other fields felt like they're only interesting for runtime.

jme3-core/src/main/java/com/jme3/anim/AnimationMask.java
```diff
@@ -31,12 +31,14 @@
...
 
+import com.jme3.export.Savable;
+
 /**
  * Created by Nehon
  * An AnimationMask is defining a subset of elements on which an animation will be applied.
  * Most used implementation is the ArmatureMask that defines a subset of joints in an Armature.
  */
-public interface AnimationMask {
+public interface AnimationMask extends Savable {
```
capdevon, line 41: This change breaks the legacy code. Do not modify the interface. Please remove the Savable extension from the AnimationMask interface and move it to the ArmatureMask class.

neph1: Thanks for the comment. Hmm. Can you explain in what way? It's only used by ArmatureMask it seems, and only created during runtime (since it's not saved)

capdevon: The purpose of the AnimationMask interface is to enable anyone to create their own implementation (I’ve written my own as well). This change, however, makes previous code non-compilable. The ArmatureMask class is just one of many implementations, so don’t focus solely on that one. Serialization of AnimChannels wasn’t provided in the old animation system either, so the absence of AnimLayer/AnimationMask serialization is a deliberate design choice. While the graphics editor in the SDK can help with mixing animations on multiple layers, I don’t think it’s necessary to save it. Multi-level animation is rarely used and complicated to manage in games. Nonetheless, it’s a feature that anyone can implement using a JSON file and a Loader designed according to their needs.

pspeed42: In my own case, I setup blending in my asset pipeline but it won't get stored... so I have to store a completely separate descriptor file to load it into my game so that I can blend animations.  It's a little frustrating... but I've already worked around it so I don't personally have a dog in this fight.

This is only tangentially related to anim layers... but it is frustrating that the only thing that gets saved/loaded is the stuff loaded directly from external asset files.  None of the JME "value add" gets stored.

neph1: How about making ArmatureMask Savable then?

Even if it was a "deliberate design choice", is there a reason it can't be changed? Would it be a problem if this change occurred? Even if it did break backwards compatibility, how many people would be affected? 3? (Which, arguably, is still more than would use this feature)

`Multi-level animation is rarely used`
I'd say any modern fps or 3rd person shooter has multi-level animations. Not to mention IK based on those layers.

`and complicated to manage in games.`
Hence my desire to make tools to simplify them! Being able to view, edit and save them removes several steps a developer has to do.

`it’s a feature that anyone can implement`
Anyone can implement any feature, so why make any tools at all?

pspeed42: I'm against changing the interface because that forces everyone to be savable even if they don't want to be but I'm a fan of changing the JME implementations to be savable.  Unless someone can come up with a good reason why this is a bad idea... JME should drink it's own kool-aid here.

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -313,7 +313,9 @@ public Object jmeClone() {
     public void write(JmeExporter ex) throws IOException {
         OutputCapsule oc = ex.getCapsule(this);
         oc.write(name, "name", null);
-        oc.write(mask, "mask", null);
+        if(mask instanceof ArmatureMask) {
```
neph1, line 316: Maybe Savable here?

pspeed42: Yeah, probably.

pspeed42:
Note that in my case, it's actions that give me heartburn for not being persisted.  But my memory is that there are more demons down that rabbit hole.

But it does mean that for every model rig I load, I have to do a loop like this with my own external fix-up config:
```java
            for( BlendInfo blend : blends ) {
                BlendAction ba = anim.actionBlended(blend.getName(), new LinearBlendSpace(0, 1),
                                                    blend.getBlendedActions());
                ba.getBlendSpace().setValue((float)blend.getDefaultBlend());
            }
```

Actions enter this weird area where I can see defining them one-off in code to do some particular thing... while also wanting to preconfigure some with masks and blending and so on.  But also, every other thing in JME operates this way... if I tweak a "runtime to me" material parameter and then save that j3o then it gets saved.  So I'm not sure I buy arguments against saving them.

neph1:
If I can find a use case from the SDK's perspective, I don't mind making Actions Savable too (I actually started it). But atm I don't see exactly how you would use them. In any case, that's for another PR.

capdevon:
To enhance clarity and facilitate review, please submit the optimizations as a separate pull request (PR). This will prevent confusion and simplify potential rollbacks. By separating these changes, you can maintain the flexibility to revert optimizations without affecting the core functionality.

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -314,7 +314,7 @@ public Action removeAction(String name) {
      * @param mask the desired mask for the new layer (alias created)
      */
     public void makeLayer(String name, AnimationMask mask) {
-        AnimLayer l = new AnimLayer(this, name, mask);
+        AnimLayer l = new AnimLayer(name, mask);
```
capdevon, line 317: Please create a separate PR for optimizations.

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -399,7 +399,7 @@ public Set<String> getAnimClipsNames() {
     @Override
     protected void controlUpdate(float tpf) {
         for (AnimLayer layer : layers.values()) {
-            layer.update(tpf);
+            layer.update(tpf, globalSpeed);
```
capdevon, line 402: Please create a separate PR for optimizations.

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -249,13 +252,13 @@ public void setLooping(boolean loop) {
      * @param appDeltaTimeInSeconds the amount application time to advance the
      *     current Action, in seconds
      */
-    void update(float appDeltaTimeInSeconds) {
+    void update(float appDeltaTimeInSeconds, float speed) {
```
capdevon, line 255: Please create a separate PR for optimizations.

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -249,13 +252,13 @@ public void setLooping(boolean loop) {
...
         Action action = currentAction;
         if (action == null) {
             return;
         }
 
-        double speedup = action.getSpeed() * composer.getGlobalSpeed();
+        double speedup = action.getSpeed() * speed;
```
capdevon, line 261: Please create a separate PR for optimizations.

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -66,7 +66,7 @@ public class AnimComposer extends AbstractControl {
      * Instantiate a composer with a single layer, no actions, and no clips.
      */
     public AnimComposer() {
-        layers.put(DEFAULT_LAYER, new AnimLayer(this, DEFAULT_LAYER, null));
+        layers.put(DEFAULT_LAYER, new AnimLayer(DEFAULT_LAYER, null));
```
capdevon, line 69: Please create a separate PR for optimizations.

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -306,4 +308,20 @@ public Object jmeClone() {
             throw new AssertionError();
         }
     }
+
+    @Override
+    public void write(JmeExporter ex) throws IOException {
+        OutputCapsule oc = ex.getCapsule(this);
+        oc.write(name, "name", null);
+        if(mask instanceof Savable) {
```
capdevon, line 316: Very good!

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
pspeed42: To me, this was not a separate optimization but removing an essentially unnecessary field that would make serialization more difficult.

capdevon: ok, if the change is essential for new functionality, it's good to go

neph1: It's correct. It's redundant, but it also makes serialization difficult (impossible due to SO?)

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -79,9 +81,13 @@ public class AnimLayer implements JmeCloneable {
     /**
      * The name of this layer.
      */
-    final private String name;
+    private String name;
 
     private boolean loop = true;
+    
+    protected AnimLayer() {
```
capdevon, line 88: please add the javadoc to the empty constructor as below:

    /**
     * For serialization only. Do not use.
     */

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
```diff
@@ -92,10 +98,7 @@ public class AnimLayer implements JmeCloneable {
      * @param mask the AnimationMask (alias created) or null to allow this layer
      *     to animate the entire model
      */
-    AnimLayer(AnimComposer composer, String name, AnimationMask mask) {
-        assert composer != null;
-        this.composer = composer;
-
+    AnimLayer(String name, AnimationMask mask) {
```
capdevon, line 101: please remove the `AnimComposer` parameter from the constructor javadoc:

```java
@param composer the owner (not null, alias created)
```

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
capdevon, line 255: please rename the variable _speed_ to `globalSpeed` and add the parameter to the javadoc as below:

```java
@param globalSpeed the global speed applied to all layers.
```

jme3-core/src/main/java/com/jme3/anim/AnimLayer.java
capdevon, line 316: Please add an extra space between `if` and `(` in the code, following the JME style guide.

```java
if (...
```

capdevon:
Ok, everything looks good to me. Thank you.

neph1:
Does anyone have to officially approve it? Or do I just merge it?

stephengold:
> Does anyone have to officially approve it? Or do I just merge it?

The project isn't very formal these days. Merge when ready.

PR #2309, opened 2024-09-02
Fix bug #2277 - GltfLoader

JNightRider:
This PR fixes bug #2277 (GltfLoader issues).

This error originates when comparing 2 objects, the return is always false (result). `jme3-plugins-json-gson` always returns a new object of type `com.jme3.plugins.json.JsonElement*` (when accessing JSON elements), each new object confuses the gltf loader when compared using the `equals(java.lang.Object) ` method.

A clear example where this error occurs is on line 1054 when searching for a JSON object within a list using the `indexOf()` method (class - GltfLoader).

Using the code created by @stephengold, now numComposers is 1
```java
import com.jme3.anim.AnimComposer;
import com.jme3.scene.Node;
import com.jme3.scene.control.Control;
import com.jme3.scene.plugins.gltf.GltfModelKey;
import java.util.List;

public class TestYBot extends SimpleApplication {

        TestYBot app = new TestYBot();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        GltfModelKey modelKey = new GltfModelKey("YBot.gltf");
        Spatial modelRoot = assetManager.loadModel(modelKey);
        int numComposers = countControls(modelRoot, AnimComposer.class);
        System.out.println("numComposers = " + numComposers);
        stop();
    }

    /**
     * Count all controls of the specified type in the specified subtree of a
     * scene graph. Note: recursive!
     *
     * @param <T> subclass of Control
     * @param subtree the subtree to analyze (may be null, unaffected)
     * @param controlType the subclass of Control to search for
     * @return the count (&ge;0)
     */
    private static <T extends Control> int countControls(
            Spatial subtree, Class<T> controlType) {
        int result = 0;

        if (subtree != null) {
            int numControls = subtree.getNumControls();
            for (int controlI = 0; controlI < numControls; ++controlI) {
                Control control = subtree.getControl(controlI);
                if (controlType.isAssignableFrom(control.getClass())) {
                    ++result;
                }
            }
        }

        if (subtree instanceof Node) {
            Node node = (Node) subtree;
            List<Spatial> children = node.getChildren();
            for (Spatial child : children) {
                result += countControls(child, controlType);
            }
        }

        return result;
    }
}
```


stephengold:
Thank you @JNightRider for investigating the issue and proposing a fix.

jme3-plugins-json/src/main/java/com/jme3/plugins/json/JsonElement.java
```diff
@@ -99,4 +99,15 @@ public interface JsonElement {
...
+     * @see java.lang.Object#equals(java.lang.Object) 
+     */
+    @Override
+    public boolean equals(Object obj);
+
+    /*(non-Javadoc)
+     * @see java.lang.Object#hashCode() 
+     */
+    @Override
+    public int hashCode();
```
tonihele, line 112: Hmm, I don't understand the reason declaring equals and hashCode in a interface? Everything that actually implements this interface is inherently a Object that has these methods already.

jme3-plugins-json-gson/src/main/java/com/jme3/plugins/gson/GsonElement.java
```diff
@@ -46,6 +48,22 @@ class GsonElement implements JsonElement {
         this.element = element;
     }
 
+    @Override
+    public int hashCode() {
+        int hash = 7;
```
pspeed42, line 53: All of this weird hash magic seems unnecessary.  Just return element.hashCode().

jme3-plugins-json-gson/src/main/java/com/jme3/plugins/gson/GsonElement.java
```diff
@@ -46,6 +48,22 @@ class GsonElement implements JsonElement {
...
+        hash = 97 * hash + Objects.hashCode(this.element);
+        return hash;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof GsonElement) {
```
pspeed42, line 60: This could lead to bugs if there are subclasses of GsonElement.  Correct .equals pattern goes like:
```java
if(obj == this ) {
    return true;
}
if( obj == null || obj.getClass() != getClass() ) {
    return false;
}
GsonElement other = (GsonElement)obj;
return Objects.equals(element, other.element);
```

jme3-plugins/build.gradle
```diff
@@ -11,6 +11,8 @@ sourceSets {
 
 dependencies {
     api project(':jme3-core')
-    api 'com.google.code.gson:gson:2.9.1'
```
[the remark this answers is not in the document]
stephengold: @pspeed42 I disabled most of Riccardo's changes to work around the bug.

jme3-plugins-json/src/main/java/com/jme3/plugins/json/JsonElement.java
JNightRider: thanks for the observation, fixed...

jme3-plugins-json-gson/src/main/java/com/jme3/plugins/gson/GsonElement.java
JNightRider: fixes applied (magic hash)...

jme3-plugins-json-gson/src/main/java/com/jme3/plugins/gson/GsonElement.java
JNightRider: An error is not generated as such, since if a subclass inherits from the GsonElement class it may be a non-direct instance of it.

Suppose we have the following classes:
```java
public class A { ... }
public class B extends A { ... }
public class C extends A { ... }
public class D extends B { ... }
```

when you check to see if they are an instance of class 'A' they all return true.
```java
    A a = new A();
    B b = new B();
        
    C c = new C();
    D d = new D();
        
        
     System.out.println("A: " +  (a instanceof A));
     System.out.println("B: " +  (b instanceof A));
     System.out.println("C: " +  (c instanceof A));
     System.out.println("D: " +  (d instanceof A));
}
```

[ **NOTE** ]
The changes have already been applied...

JNightRider:
With this 'commit', the changes have already been applied.

jme3-plugins-json-gson/src/main/java/com/jme3/plugins/gson/GsonElement.java
pspeed42: The issue with using instanceof in .equals() methods is that you can end up with cases where a.equals(b) returns true but b.equals(a) does not.  That's why one should always check direct class == comparison.  (Which also saves you from accidentally comparing two different similar looking objects from entirely different classloaders.)

jme3-plugins/build.gradle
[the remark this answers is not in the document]
JNightRider: > or rather, I don't understand how Ric's changes were working at all before without this.... and if it was working

These changes were removed as a temporary fix; Currently modules `jme3-plugins-json` and `jme3-plugins-json-gson` are not used.

> I don't know why suddenly it needs to change just to fix how .equals()/.hashCode() is working.

Ric's changes are no longer present in the loader (Gltf), it would make no sense to fix something that is not being used, that is why the library imports will restore the connections and make sense of the solution.

tonihele (approved):

stephengold:
Thanks for the comments and review.
@pspeed42 have all your concerns been addressed?

pspeed42:
Yeah, I think so.  Mostly I was confused. :)

Issue #2312, opened 2024-09-18
BinaryExporter resets buffer positions to 0 when writing them.


JosiahGoeman:
When a buffer or buffer arraylist is written using BinaryExporter, [the buffer's position gets reset to zero.](https://github.com/jMonkeyEngine/jmonkeyengine/blob/db6353046cc34af7d09211860c37e4c4a722516f/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryOutputCapsule.java#L955)  XMLExporter handles this by [restoring the old position after writing.](https://github.com/jMonkeyEngine/jmonkeyengine/blob/db6353046cc34af7d09211860c37e4c4a722516f/jme3-plugins/src/xml/java/com/jme3/export/xml/DOMOutputCapsule.java#L856C9-L856C29)  This seems easy enough to fix.  It does however bring up another issue currently masked by the position clobbering: shouldn't the position also be serialized?  If a buffer with a non-zero position is serialized and the position is preserved, the deserialized buffer will not equals() the original because it will have a zero position.  I can easily fix this for XMLExporter with an additional attribute, but from my brief look at BinaryOutputCapsule, it looks like this can't be changed because only one int (length) is expected before the start of the contents.

<details><summary>BinaryExportBufferPositionMRE.java</summary>

```java
package com.mygame;

import com.jme3.export.JmeExporter;
import com.jme3.export.JmeImporter;
import com.jme3.export.OutputCapsule;
import com.jme3.export.Savable;
import com.jme3.export.binary.BinaryExporter;
import java.io.IOException;
import java.io.OutputStream;
import java.nio.ByteBuffer;
import org.lwjgl.BufferUtils;

public class BinaryExportBufferPositionMRE
{
	static class SavableWithBuffer implements Savable
	{
		@Override
		public void write(JmeExporter je) throws IOException
		{
			ByteBuffer bb = BufferUtils.createByteBuffer(16);
			bb.position(4);
			
			OutputCapsule capsule = je.getCapsule(this);
			capsule.write(bb, "buffer", null);
			
			if(bb.position() != 4)
			{
				throw new IOException("Expected buffer position 4, got position: " + bb.position());
			}
		}

		@Override
		public void read(JmeImporter ji) throws IOException
		{
			
		}
	}

	public static void main(String[] args)
	{
		BinaryExporter be = BinaryExporter.getInstance();

		try
		{
[...]

Issue #2310, opened 2024-09-12
XMLExporter has several bugs/inconsistencies with BinaryExporter


JosiahGoeman:
I've been struggling a bit with the XMLExporter recently in my game and this prompted me to write some unit tests for the JmeExporter/JmeImporter implementations.  You can see the tests on my fork here:
https://github.com/JosiahGoeman/jmonkeyengine/blob/master/jme3-plugins/src/test/java/com/jme3/export/InputOutputCapsuleTest.java.
All write* and read* methods in OutputCapsule and InputCapsule respectively are tested here.  I put in all the edge cases I could think of.  BinaryExporter/BinaryImporter pass all tests no problemo, but XMLExporter/XMLImporter do not.

Here's the problems I've discovered so far:

1. ☑~~Write an empty string -> Reads defVal as if attribute was not present~~
2. ☑~~Write a string containing an apostrophe/single quote -> throws IOException when reading~~
3. ☑~~Write a string containing tab, newline, or carriage return -> Reads string with these characters replaced with spaces~~
4. ☑ ~~Write any BitSet object -> Reads BitSet containing a single zero~~
5. ☑~~Write String[] containing a null string -> Reads array with null string having been replaced with an empty string~~
6. ☑~~Write a 2d array of any type except int[][] containing a null element -> Throws NullPointerException when writing~~
7. ☑~~Write an ArrayList containing a null element -> Throws IOException when reading~~
8. ☑~~Write an ArrayList\<ByteBuffer> -> Reads ArrayList with same number of entries, but all are null~~

<details><summary>XMLExporterMREs.java</summary>

```java
package com.mygame;

import com.jme3.asset.AssetInfo;
import com.jme3.export.InputCapsule;
import com.jme3.export.JmeExporter;
import com.jme3.export.JmeImporter;
import com.jme3.export.OutputCapsule;
import com.jme3.export.Savable;
import com.jme3.export.binary.BinaryExporter;
import com.jme3.export.binary.BinaryImporter;
import com.jme3.export.xml.XMLExporter;
import com.jme3.export.xml.XMLImporter;
import com.jme3.math.Vector3f;
import java.io.ByteArrayInputStream;
import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.InputStream;
import java.nio.ByteBuffer;
import java.util.ArrayList;
import java.util.Arrays;
import java.util.BitSet;
import java.util.Objects;
[...]

PR #2313, opened 2024-09-18
Fix for XMLExporter issues in #2310

JosiahGoeman:
Includes fixes for all issues listed in https://github.com/jMonkeyEngine/jmonkeyengine/issues/2310 and refactors related classes for readability, adherence to style guide, and elimination of duplicate code.  Also adds unit tests for implementations of the OutputCapsule and InputCapsule interfaces.  Currently one [test is commented out](https://github.com/JosiahGoeman/jmonkeyengine/blob/fdaa7b7b1013e4158f6be8d0ceade7c36830aca6/jme3-plugins/src/test/java/com/jme3/export/InputOutputCapsuleTest.java#L641) because BinaryExporter fails it, see https://github.com/jMonkeyEngine/jmonkeyengine/issues/2312.


JosiahGoeman:
Wait, why did GitHub include the commit from my previous pr?  It doesn't change any files anymore.  Does it just always do that if you don't make a separate branch for each issue?

stephengold:
I'm unsure what happened. It might have something to do with creating a pull request in your "master" branch. Does the "files changed" tab look correct to you?

JosiahGoeman:
Yes, the files changed are correct.

JosiahGoeman:
Alright, I think this should be ready for review.

stephengold:
If there's no substantive review or discussion, I'll run some simple tests of my own and then integrate this PR.

stephengold:
My tests turned up an issue that's probably related to this change. I'll troubleshoot and post my findings here.

stephengold:
The issue I'm seeing is an infinite recursion that occurs while writing a Minie `PhysicsVehicle` to an `XMLExporter`. However, I don't fully understand what's going wrong...

The test app looks like this:
```java
        NativeLibraryLoader.loadNativeLibrary("bulletjme", true);
        CollisionShape shape = new SphereCollisionShape(1f);
        PhysicsVehicle body = new PhysicsVehicle(shape, 1f);

        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
        XMLExporter exporter = new XMLExporter();
        try {
            exporter.save(body, outputStream);
        } catch (IOException exception) {
            throw new RuntimeException(exception);
        }
```

The infinite recursion is as follows:
`PhysicsVehicle.write()` line 846 invokes
`PhysicsRigidBody.write()`. At line 1442, that invokes
`DOMOutputCapsule.write()`. At line 406, that invokes
`RigidBodyMotionState.write()`. At line 366, that invokes 
`DOMOutputCapsule.write()`. And at line 406, that invokes
`PhysicsVehicle.write()` again.

There's only one `PhysicsVehicle` object in the test app, but it appears `DOMOutputCapsule` has forgotten it's already in the process of writing that `PhysicsVehicle`. The cycle repeats many times, until eventually the test app crashes with a `StackOverflowError`.

The same test app works fine if I replace `XMLExporter` with `BinaryExporter`.

codex128:
I took a quick peek at it, and it looks like PhysicsVehicle contains a PhysicsRigidBody, which it attempts to save. The PhysicsRigidBody contains a RigidBodyMotionState, which it also attempts to save. Finally, the RigidBodyMotionState contains the original PhysicsVehicle, which it also attempts to save, and the process starts all over again.

I don't know why it only fails with XMLExporter, but I'd guess that BinaryExporter maybe has some sort of check for that sort of dependency issue.

stephengold:
> I'd guess that BinaryExporter maybe has some sort of check for that sort of dependency issue.

The test succeeds with `XMLExporter` in v3.7.0-stable, which means there was *also* a check in `XMLExporter`, and this PR disabled that check. I'm hoping @JosiahGoeman will solve this for us.

stephengold:
Thank you for your quick response.
How about adding a reference loop to jme3-plugins/src/test ?

JosiahGoeman:
Looks like it was indeed that simple.  And here I was thinking my test suite was so thorough. -_-  I added a test for the reference-loop case.

stephengold:
Thanks, @JosiahGoeman .
7f522dc passes my tests as well, so I think this PR is ready for integration---unless someone cares to review it.

PR #2325, opened 2024-10-28
Stencil buffer fix

zzuegg:
This fixes the use of the stencil buffers. 

As reference:
https://hub.jmonkeyengine.org/t/how-to-do-outline-with-stencil-buffer/47954/5

Also added the TestStencilOutline.java to the examples. 
Note: Somehow the TestChooser app is overwriting the appsettings in the main method. 
As a fix i added a postprocessor to the example application. This is not required when running
the test directly.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/RenderState.java
+++ b/jme3-core/src/main/java/com/jme3/material/RenderState.java
@@ -1511,6 +1511,10 @@ public class RenderState implements Cloneable, Savable {
             hash = 79 * hash + (this.backStencilDepthPassOperation != null ? this.backStencilDepthPassOperation.hashCode() : 0);
             hash = 79 * hash + (this.frontStencilFunction != null ? this.frontStencilFunction.hashCode() : 0);
             hash = 79 * hash + (this.backStencilFunction != null ? this.backStencilFunction.hashCode() : 0);
+            hash = 79 * hash + (this.frontStencilMask);
+            hash = 79 * hash + (this.frontStencilReference);
+            hash = 79 * hash + (this.backStencilMask);
+            hash = 79 * hash + (this.backStencilReference);
             hash = 79 * hash + Float.floatToIntBits(this.lineWidth);
             
             hash = 79 * hash + this.sfactorRGB.hashCode();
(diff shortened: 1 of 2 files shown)


jme3-core/src/main/java/com/jme3/material/RenderState.java
```diff
@@ -1665,6 +1679,10 @@ public void set(RenderState state) {
         backStencilDepthPassOperation = state.backStencilDepthPassOperation;
         frontStencilFunction = state.frontStencilFunction;
         backStencilFunction = state.backStencilFunction;
+        frontStencilMask=state.frontStencilMask;
+        frontStencilReference=state.frontStencilReference;
+        backStencilMask=state.backStencilMask;
+        backStencilReference=state.backStencilReference;
```
codex128, line 1685: Add spacing between around '='.

jme3-core/src/main/java/com/jme3/material/RenderState.java
```diff
@@ -1636,6 +1645,11 @@ public RenderState copyMergedTo(RenderState additionalState, RenderState state)
 
             state.frontStencilFunction = frontStencilFunction;
             state.backStencilFunction = backStencilFunction;
+
+            state.frontStencilMask=frontStencilMask;
+            state.frontStencilReference=frontStencilMask;
+            state.backStencilMask=backStencilMask;
+            state.backStencilReference=backStencilMask;
```
codex128, line 1652: Add spacing between around '='.

jme3-core/src/main/java/com/jme3/material/RenderState.java
```diff
@@ -1623,6 +1627,11 @@ public RenderState copyMergedTo(RenderState additionalState, RenderState state)
 
             state.frontStencilFunction = additionalState.frontStencilFunction;
             state.backStencilFunction = additionalState.backStencilFunction;
+
+            state.frontStencilMask=additionalState.frontStencilMask;
+            state.frontStencilReference=additionalState.frontStencilMask;
+            state.backStencilMask=additionalState.backStencilMask;
+            state.backStencilReference=additionalState.backStencilMask;
```
codex128, line 1634: Add spacing between around '='.

codex128 (commented):
These changes look generally fine to me. I'd just suggest that some spacing be added. I'm a bit surprised that RenderState didn't already copy those parameters.

Would you mind explaining what the bug turned out being exactly?

zzuegg:
Thanks for point out the spacing error. 
Due the fact that the Mask/Reference was not copied/included in the hash, the actual values have not been send to opengl, you were stuck with the reference set to 0 and the mask to 255. 

Since i added those values back in 2022, and the fix was already present in my local fork, it seems i forgot to commit them back then. As far i remember the original usecase used only Increment and NotEqual to 0 so it slipped trough the testing.

codex128 (approved):

Issue #2327, opened 2024-11-12
Mouse pointer moves to centre of canvas on click in OpenGL3 with Swing


terranprog:
With JME3.7 we can now embed OpenGL3 canvas in a Swing window. I have noticed an issue with the mouse with this configuration.

**Steps to reproduce:**

1. Use JME 3.7.0-stable
2. Be running OpenGL3
3. Have a jme canvas embedded in a swing window
4. Listen for jme mouse events using InputManager.addListener(ActionListener)
5. Left click on the canvas
6. In the ActionListener, get the position of the mouse using InputManager.getCursorPosition()

**Observed**
The mouseDown (i.e. keyPressed=true) will have the correct mouse position.
By the time the mouseUp event is processed (i.e. keyPressed=false), the position of the mouse has been moved to the centre of the canvas.
- The mouse is physically displayed in the centre of the canvas
- InputManager.getCursorPosition() will return a mouse position in the centre of the canvas

**Notes:**
- This does not happen if you are not using Swing
- This does not happen in OpenGL2 with Swing
- Seen on both Windows 10 and current Debian linux stable (12.7)

A workaround (more a hack) is to
- Store the location of the mousedown event
- In the mouseup event, use java.awt.Robot to move the mouse back to where the mousedown event happened.

terranprog:
I should have added that I am using FlyByCamera with dragToRotate set to true. This turns out to be relevant.

A few things I have learnt so far:

The reason for the difference in behaviour between jme3-lwjgl and jme3-lwjgl3 is that they use different instances of com.jme3.input.MouseInput

jme3-lwjgl uses an instance of com.jme3.input.lwjgl.LwjglMouseInput (even if running in a Swing window)
jme3-lwjgl3 uses an instance of com.jme3.input.awt.AwtMouseInput

The unwanted recentering of the mouse cursor that I am seeing happens in com.jme3.input.awt.AwtMouseInput.setCursorVisible.

```
component.setCursor(newVisible ? null : getTransparentCursor());
    if (!newVisible) {
        recenterMouse(component);
    }
```

So every time the mouse cursor is set to not visible the cursor is moved to the centre of the canvas.

com.jme3.input.lwjgl.LwjglMouseInput has different code in setCursorVisible:
`Mouse.setGrabbed(!visible);`

com.jme3.input.awt.AwtMouseInput.recenterMouse has this comment:
[...]

PR #2328, opened 2024-11-15
Fix for #2327 Mouse pointer moves to centre of canvas on click in OpenGL3 with Swing

terranprog:
Modify FlyByCamera to only hide the mouse cursor when the user actually rotates the camera, rather than when the user first presses the left mouse button. When the mouse is clicked, instead of immediately hiding the cursor, a flag (hideCursorOnNextRotate) is set. If and when a rotate is detected, the flag is checked and if true the cursor is hidden and the flag is cleared.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
+++ b/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
@@ -107,6 +107,7 @@ public class FlyByCamera implements AnalogListener, ActionListener {
     protected boolean dragToRotate = false;
     protected boolean canRotate = false;
     protected boolean invertY = false;
+    private boolean hideCursorOnNextRotate = false;
     protected InputManager inputManager;
 
     /**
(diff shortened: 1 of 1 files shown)


capdevon:
I tested the behavior of `FlyByCamera` with **LWJGL2** and **LWJGL3** in combination with Swing. In both cases, setting `flyCam.setDragToRotate(true)` repositions the mouse pointer to the center of the window after the left mouse button is released to enable camera rotation. This behavior does not occur in typical applications that do not use Swing, where the mouse pointer returns to its last position at the time of the click.

While this isn’t overly inconvenient and you can get used to it, it’s worth noting that `FlyByCamera` is built into the `SimpleApplication` class by default and is likely used by thousands of applications. Changing its behavior could introduce more problems than benefits without extensive testing. FlyByCamera was probably not designed specifically for use with Swing, but rather for simple JME applications.

If your goal is to develop an editor with Java Swing + JME, as I am, I recommend creating a custom `EditorCamera` by duplicating the `FlyByCamera` code and adapting it to your specific needs. JME is designed to be customizable in almost all of its components.

codex128 (commented):
If these changes introduce functionally different behaviors (particularly on glfw applications), I'd suggest adding an option to disable or enable the changes, with it disabled by default.

Also, I don't foresee there being any problems, but it would be good to make sure nothing unexpected happens when using glfw with any opengl versions along with these changes.

terranprog:
While it seems incorrect that a mouse press is treated as a drag, I agree that FlyByCamera is widely used and changing it would be high risk to fix a relatively niche use case.
I suspect the real fix for this issue is to change AwtMouseInput.

PR #2332, opened 2024-11-29
Fix: Incorrect mouse position on canvas LWJGL3 (Canvas)

JNightRider:
This PR resolves bad mouse position when using canvas with lwjgl3, this fixes issue #2327 (Cursor centered on click -> flying camera with `setDragToRotate(true)` option enabled).

Now when the cursor is grabbed, if there is a point (coordinates) coming from the user, the cursor will stay at that position; otherwise it will focus on the component (canvas).


jme3-desktop/src/main/java/com/jme3/input/awt/AwtMouseInput.java
```diff
@@ -283,7 +285,7 @@ public void mouseDragged(MouseEvent awtEvt) {
 
     @Override
     public void mouseMoved(MouseEvent awtEvt) {
-        if (isRecentering) {
+        if (isRecentering) {
```
MeFisto94, line 288: Accidental whitespace?

jme3-desktop/src/main/java/com/jme3/input/awt/AwtMouseInput.java
```diff
@@ -304,7 +306,7 @@ public void mouseMoved(MouseEvent awtEvt) {
             }
             lastKnownLocation.x = awtEvt.getX();
             lastKnownLocation.y = awtEvt.getY();
-
+
```
MeFisto94, line 309: same accidental whitespace here?

jme3-desktop/src/main/java/com/jme3/input/awt/AwtMouseInput.java
```diff
@@ -341,4 +347,4 @@ private int getJMEButtonIndex(MouseEvent awtEvt) {
     @Override
     public void setNativeCursor(JmeCursor cursor) {
     }
-}
+}
```
MeFisto94, line 350: One more linebreak here

MeFisto94 (commented):
Can't comment on the functionality, probably needs testing, but code wise it looks good

jme3-desktop/src/main/java/com/jme3/input/awt/AwtMouseInput.java
JNightRider: It was probably a couple of accidental keystrokes (I'm working on it).

jme3-desktop/src/main/java/com/jme3/input/awt/AwtMouseInput.java
JNightRider: It was probably a couple of accidental keystrokes (I'm working on it).

MeFisto94: Yeah, it's the opposite, you are missing an empty line here.
That's an arguably odd convention required for linux, because otherwise when inspecting a file in a terminal, the next terminal prompt is written right after the `}` from the source code, so you want to have an empty line, causing linux to start the prompt on the next line.

JNightRider: It's strange that this jump doesn't appear in my IDE, but it does appear in a Linux terminal (strange).

yaRnMcDonuts:
I'll leave this open for another 2-3 day, then I plan to merge this so it can be tested in 3.8 beta.

Issue #2330, opened 2024-11-18
PBR Lighting issue


zzuegg:
To continue the discussion from: 
https://hub.jmonkeyengine.org/t/zombiegirl-gltfloader-vs-monkeywrench/48007/15

Summary:

According to gltf specs, and on the reference viewers, the default value for metallic factor is 1. Jme's importers, when using are setting the metallic factor to 0 to pass the test cases. The problematic line in my opinion is:
`
//223 of PBRLighting
vec4 diffuseColor = albedo - albedo * Metallic;
`

I have checked, a few other repositories, and i have not yet found that piece of math again. Additionally, the whole block:
`
        float specular = 0.5;
        float nonMetalSpec = 0.08 * specular;
        vec4 specularColor = (nonMetalSpec - nonMetalSpec * Metallic) + albedo * Metallic;
        vec4 diffuseColor = albedo - albedo * Metallic;
        vec3 fZero = vec3(specular);
`
is quite unique. F0 is in all other shaders defined as: 

`    vec3 F0 = vec3(0.04);
    F0 = mix(F0, albedo, metallic);
`

zzuegg:
When testing the MetalRoughSpheres jme shows also different results than the khronos sample viewer. 
![Screenshot 2024-11-18 211130](https://github.com/user-attachments/assets/a52a0ab0-5a97-4267-99ac-2e17c4e07f39)

stephengold:
Thanks for investigating.

When it comes time to solve this, instead of modifying the PBRLighting shaders (and potentially breaking many apps) I suggest defining a new PBR material with a new name.

yaRnMcDonuts:
This looks like it is quite a big bug in our PBR implementation.

I've always speculated that something looks slightly off with some of my PBR textures in jme compared to the original source, especially under certain lighting conditions... and I think this might have been it it. 

I just tested the results with the old (broken) fZero calculation compared to the new code you suggested, and it appears to look much better with your code.

The difference is even more noticeable first-hand, but you can still see a difference in these 2 screenshots (both taken with identical lighting in same place and the same time of day in-game) : 

Here is with broken fZero calculation (you can notice a very slight annoying glare to parts of the scene)
![image](https://github.com/user-attachments/assets/cce2500d-6a0e-4c16-b271-4dbf2e10cd05)


And here is the same time/place with the fixed fZero calculation. It looks much better in my opinion, especially the grass:
[...]

PR #2191, opened 2024-02-02
Modularize PBRLighting.frag

yaRnMcDonuts:
Addresses this issue: https://github.com/jMonkeyEngine/jmonkeyengine/issues/2122

To re-summarize:

This PR extracts all of the texture reads and base PBR matParam assignment into a .glslib file (currently named PBRLightingParamReads.glsllib)

And then it also extracts all of the lighting calculation into a .glslib file (currently named PBRLighting.glsllib)

I also fixed alot of formatting and indentation mistakes, and overall reorganized the shader.

This should now make it as easy as possible for jme devs to fork PBRLighting to make changes, and they will no longer have to update the lighting or texReads to stay on par with master, as all future changes will all be in the glsllibs. This also reduces the chances that a lesser-skilled shader dev (or even skilled shader devs on a bad day) mistakenly mess up the lighting calculations when forking the shader.

Any feedback and review is greatly appreciated.

PR #2334, opened 2024-12-01
Fix incorrect metallic calculations in PBRLighting.frag

yaRnMcDonuts:
addresses this issue: https://github.com/jMonkeyEngine/jmonkeyengine/issues/2330

There were 2 apparent issues with our metallic calculation that were causing our PBR implementation to render different visual results compared to the khronos PBR standards:

1) fZero was being calculated incorrectly, as mentioned in the linked issue 
2) the nonMetalSpec float constant was being set incorrectly. According to what I've read, it should always be 0.08

I also checked tested changes in my own app and it appears the rendering difference is more accurate now. But I'm not an expert in lighting math like this, so let me know if anyone sees anything else that looks incorrect with the PBR shader's code.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -217,11 +217,9 @@ void main(){
         Roughness = 1.0 - glossiness;
         vec3 fZero = specularColor.xyz;
     #else
-        float nonMetalSpec = 0.08; //nonMetalSpec is always constant value of 0.8 in metallic workflow
-        vec4 specularColor = (nonMetalSpec - nonMetalSpec * Metallic) + albedo * Metallic;
+        vec4 specularColor = (0.08 - 0.08 * Metallic) + albedo * Metallic;  // 0.08 represents the base specular reflectance for non-metallic surfaces. This can be slightly adjusted to values like 0.04, however 0.08 is considered the standard for PBR.
         vec4 diffuseColor = albedo - albedo * Metallic;
-        vec3 fZero = vec3(0.04); 
-        fZero = mix(fZero, albedo.rgb, Metallic);          
+        vec3 fZero = mix(vec3(0.04), albedo.rgb, Metallic);
     #endif
 
     gl_FragColor.rgb = vec3(0.0);
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -217,7 +217,7 @@ void main(){
         Roughness = 1.0 - glossiness;
         vec3 fZero = specularColor.xyz;
     #else
-        vec4 specularColor = (0.08 - 0.08 * Metallic) + albedo * Metallic;  // 0.08 represents the base specular reflectance for non-metallic surfaces. This can be slightly adjusted to values like 0.04, however 0.08 is considered the standard for PBR.
+        vec4 specularColor = (0.04 - 0.04 * Metallic) + albedo * Metallic;  // 0.04 is the standard base specular reflectance for non-metallic surfaces in PBR. While values like 0.08 can be used for different implementations, 0.04 aligns with Khronos' PBR specification.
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
```diff
@@ -217,11 +217,11 @@ void main(){
         Roughness = 1.0 - glossiness;
         vec3 fZero = specularColor.xyz;
     #else
-        float specular = 0.5;
-        float nonMetalSpec = 0.08 * specular;
+        float nonMetalSpec = 0.08; //nonMetalSpec is always constant value of 0.8 in metallic workflow
```
MeFisto94, line 220: Comment: "0.08 in metallic workflow"! Classic example of how a comment describing existing code causes confusion ;)
Could probably reword it as that the specularity is always constant (or independent of (whatever it is usually)).

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
```diff
@@ -217,11 +217,11 @@ void main(){
...
         vec4 specularColor = (nonMetalSpec - nonMetalSpec * Metallic) + albedo * Metallic;
         vec4 diffuseColor = albedo - albedo * Metallic;
-        vec3 fZero = vec3(specular);
+        vec3 fZero = vec3(0.04); 
+        fZero = mix(fZero, albedo.rgb, Metallic);
```
MeFisto94, line 224: Another nit but couldn't you combine this in a one liner? `vec3 fZero = mix(vec3(0.04), albedo.rgb, Metallic)`?
Or is there a specific meaning?

MeFisto94 (commented):
Again not really an expert too, but I guess it's a case of "try and see"

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
yaRnMcDonuts: Yeah, I think I should probably just remove that comment as well as the unnecessary nonMetalSpec variable declaration so I can condense the code. nonMetalSpec is just an internal constant variable that isn't used anywhere else anyways, so I'm not sure that it ever needed declared as a variable in the first place.

jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
yaRnMcDonuts: I think there's a few segments of code like this that could probably be condensed in the PBR shader.

It looks like the shader was originally written fairly verbosely, and I assume that was because the original author was creating this shader based on a reference doc full of PBR math and equations, so he was likely focused more on accurately following the pbr standards than creating condensed code. Which is understandable very early in the development process, since its often times easier to debug verbose code than it is to debug a big 1 liner.

But I think it should be okay to condense some of this code now that our PBR shader is in a stable state. So I'll go ahead and combine it into one line.

yaRnMcDonuts:
I updated my comment about the constant value in the metallic equation to be more understandable.

I also am finding conflicting information about whether or not that constant value should be 0.04 or 0.08 in this line.:

`        vec4 specularColor = (0.08 - 0.08 * Metallic) + albedo * Metallic; .`

From my research, it sounds like either could work for PBR and it is often tweaked slightly for different implementation depending on how shiney they want their non-metallic substance to be by default. So that would explain why I previously was finding conflicting values. But now I believe 0.04 is the value that would match Khronos standards.

Either way, a small change to this value ends up having very miniscule bearing on the rendering results, and changing this doesn't cause nearly the same issue as was being caused by the fZero calculation being incorrect. The fZero fix is definitely the more important aspect of this PR.

yaRnMcDonuts:
I've merged this PR so it is ready for 3.8.0-alpha1. The only real change ended up being the straightforward fix for the broken fZero calculation, and I left nonMetalSpec as is, since it was already correct.

And there will also be some other major changes (refactoring and feature-wise) happening to the PRB shader in another soon to be merged PR anyways, so if any further refactoring or code condensing changes are to be discussed, it is better off here: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2191

PR #2338, opened 2024-12-29
Fix NullPointer in MikktspaceTangentGenerator

yaRnMcDonuts:
Fixes a null pointer exception that occurred when trying to generate tangents on a model without a normal buffer: https://hub.jmonkeyengine.org/t/mikktspacetangentgenerator-fails-with-npe/48114/4

This PR makes it so that MikktspaceTangentGenerator handles a non existant normal (or texCoord) buffer in the same way that the old TangentBinormalGenerator handled it.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/util/mikktspace/MikktspaceTangentGenerator.java
+++ b/jme3-core/src/main/java/com/jme3/util/mikktspace/MikktspaceTangentGenerator.java
@@ -117,33 +117,35 @@ public class MikktspaceTangentGenerator {
             Geometry g = (Geometry) s;
             Mesh mesh = g.getMesh();
 
-            Mesh.Mode mode = mesh.getMode();
-            boolean hasTriangles;
-            switch (mode) {
-                case Points:
-                case Lines:
-                case LineStrip:
-                case LineLoop:
-                    hasTriangles = false; // skip this mesh
-                    break;
+            if (mesh.getBuffer(Type.TexCoord) != null && mesh.getBuffer(Type.Normal) != null) {
+                Mesh.Mode mode = mesh.getMode();
+                boolean hasTriangles;
+                switch (mode) {
+                    case Points:
+                    case Lines:
+                    case LineStrip:
+                    case LineLoop:
+                        hasTriangles = false; // skip this mesh
+                        break;
 
-                case Triangles:
-                case TriangleFan:
-                case TriangleStrip:
-                    hasTriangles = true;
-                    break;
+                    case Triangles:
+                    case TriangleFan:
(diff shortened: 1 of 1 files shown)


MeFisto94:
Did you guys discuss how error-feedbacky you want this method to be? Because for me, there's opportunity in changing behavior, albeit implicitly, which is usually a hard no for APIs, but:

On one hand, there's the Exception for not supported mesh types, but things like Lines are silently ignored, too.
Personally, the requirement of generating tangents for meshes that don't have a normal map feels illegal/unsupported, so throwing would be fine. It's not what we *expect* to do. But then we should do the same for Lines.

This could probably be shifted to user-code if we had the main Method take a Geometry and only keep the Spatial method as convenience/deprecate it and then having the user be responsible to not try to generate nonsense.

The risk with being fault tolerant is always that faults fly under the radar, so we could/should at least emit a warning in these cases

yaRnMcDonuts:
> there's opportunity in changing behavior, albeit implicitly, which is usually a hard no for APIs

I personally was aiming to avoid this, since I have very little understanding of this class, especially the math behind it. So I wanted to make the PR as minimal as possible, and to follow the same approach that TangentBinormalGenerator took (which also just silently ignores meshes without normals, and generates tangents for meshes that have them)

The main goal of this PR is just to fix the crashing NPE that occurred when another jme user tried generating tangents on a model that had some meshes with normals and some without. So ideally we should fix it so they don't have to go through manually and determine which meshes are valid and then have to generate tangents one mesh a time. It is important that generating tangents on a parentNode handles the generation of all children safely and error free, even in cases where a child mesh has no normals.

But I'd say it would be okay to add some code this PR to make MikktSpaceTangentGenerator output a simple warning or message of some sort whenever a mesh without normals/triangles/texCoords is detected. However there is also the argument that adding more (potentially unecssary) warning clutters up the console output.

yaRnMcDonuts:
There was another PR related to some fixes to MikktSpaceTangentGenerator that required me to look at the code some more, so I went ahead and added the suggested loggign output in case of no triangle or no buffers, and removed some other redundant logging that would've been there twice after having merged that other PR. 

Unless anyone has issues with the changes, I'll merge this in the next day or two.

fba-rio:
Okay, I'm just not sure if I agree with the log level. Should it really be SEVERE or would a WARNING be enough? Or would even a FINE be okay? I mean, if it's okay not to have normals (no tangents in this case), then why log a SEVERE?

I may just be seeing my side of the fence here or just being nonsensical, but the thing is that my log file is getting flooded with logs from MikktspaceTangentGenerator class.

PR #2339, opened 2024-12-29
Add RawInputListenerAdapter for easier listener creation

capdevon:
Implement `RawInputListenerAdapter` to provide default empty implementations for all listener methods.


MeFisto94:
LGTM, the only gripe I have is with the constructor: Do we even explicitly need one? I guess for abstract classes it may implicitly be protected? Also the javadoc on it is kind of "old school" java, not really providing any additional value. (compare "javadoc: getter function for foo, that gets the variable foo" 😄 )

MeFisto94 (approved):

capdevon:
You're correct that abstract classes have an implicit protected constructor by default. 
We can remove the explicit constructor declaration to avoid redundancy.

I utilized Gemini AI to assist in the rapid generation of the initial Javadoc, which helped me quickly establish a baseline and avoid common grammatical errors :)

PR #2342, opened 2025-01-01
Proposed fix for Issue#2341 - Missing default layer in AnimComposer

### LINKED ISSUE #2341: Exception playing Animation without ArmatureMask
I receive the following Exception when attempting to play an animation on an AnimComposer that was created without custom armature masks 

`Caused by: java.lang.IllegalArgumentException: Unknown layer Default
	at com.jme3.anim.AnimComposer.getLayer(AnimComposer.java:447)
	at com.jme3.anim.AnimComposer.setCurrentAction(AnimComposer.java:147)
	at com.jme3.anim.AnimComposer.setCurrentAction(AnimComposer.java:135)
	at com.jme3.anim.AnimComposer.setCurrentAction(AnimComposer.java:124)
	at com.theafflictedforests.InteractiveArchitecture.SpiritStatue.Animal.StagStatue.<init>(StagStatue.java:29)
	... 17 more
`

The majority of my models that have ArmatureMasks setup work fine, so it looks like a recent PR must've made AnimComposer not work without armatureMasks.

It looks like the issue is being caused because the DEFAULT_LAYER (which gets created when you do not provide custom layers in the Armature constructor) is returning null.

There's 3 recent PRs to AnimCompose I see that could've caused this:

https://github.com/jMonkeyEngine/jmonkeyengine/commits/36aac2579b852094df29ec81a3c2074dbb276480/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java

Although none of those 3 stand out to me as code that would cause this to break, except maybe the change to check if layers.isEmpty() before adding the default layer... but I'd assume it is empty at that point and that shouldn't be blocking it being added. 

IDK though, I'd probably have to look through more code in the Armature class to find the issue.

If anyone has ideas, let me know.

neph1:
may fix #2341

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -52,6 +52,7 @@ import java.util.*;
  * @author Nehon
  */
 public class AnimComposer extends AbstractControl {
+
     /**
      * The name of the default layer.
      */
@@ -121,7 +122,7 @@ public class AnimComposer extends AbstractControl {
      * @return The action corresponding to the given name.
      */
     public Action setCurrentAction(String name) {
-        return setCurrentAction(name, DEFAULT_LAYER);
+        return setCurrentAction(name, DEFAULT_LAYER, true);
     }
 
     /**
@@ -144,9 +145,9 @@ public class AnimComposer extends AbstractControl {
      * @return The action corresponding to the given name.
      */
     public Action setCurrentAction(String actionName, String layerName, boolean loop) {
-        AnimLayer l = getLayer(layerName);
+        AnimLayer layer = getLayer(layerName);
         Action currentAction = action(actionName);
-        l.setCurrentAction(actionName, currentAction, loop);
+        layer.setCurrentAction(actionName, currentAction, loop);
 
         return currentAction;
     }
@@ -239,7 +240,8 @@ public class AnimComposer extends AbstractControl {
     /**
      *
      * @param name The name of the action to return.
-     * @return The action registered with specified name. It will make a new action if there isn't any.
+     * @return The action registered with specified name. It will make a new
+     * action if there isn't any.
      * @see #makeAction(java.lang.String)
      */
     public Action action(String name) {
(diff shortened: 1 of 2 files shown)


jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -526,7 +530,9 @@ public void cloneFields(Cloner cloner, Object original) {
         for (String key : layers.keySet()) {
             newLayers.put(key, cloner.clone(layers.get(key)));
         }
-
+        if (!newLayers.containsKey(DEFAULT_LAYER)) {
```
capdevon, line 533: Please use `map.putIfAbsent(key, value)` instead. This is generally considered more concise and potentially more efficient than the manual check.

eg:
```java
map.putIfAbsent(DEFAULT_LAYER, new AnimLayer(DEFAULT_LAYER, null);
```

capdevon (changes requested):

jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
```diff
@@ -546,6 +552,9 @@ public void read(JmeImporter im) throws IOException {
         animClipMap = (Map<String, AnimClip>) ic.readStringSavableMap("animClipMap", new HashMap<String, AnimClip>());
         globalSpeed = ic.readFloat("globalSpeed", 1f);
         layers = (Map<String, AnimLayer>) ic.readStringSavableMap("layers", new HashMap<String, AnimLayer>());
+        if (!layers.containsKey(DEFAULT_LAYER)) {
```
capdevon, line 555: Please use `map.putIfAbsent(key, value)` instead. This is generally considered more concise and potentially more efficient than the manual check.

eg:
```java
map.putIfAbsent(DEFAULT_LAYER, new AnimLayer(DEFAULT_LAYER, null);
```

capdevon (changes requested):

yaRnMcDonuts:
Thanks for the quick PR to fix this!

Let me know if you are still working on anything with this PR, otherwise I'll merge it in the next few days and have it included in the next 3.8.0alpha2 release.

neph1:
I've nothing to add.

PR #2360, opened 2025-02-03
Update tank material pointing to deleted copy of PBRLighting

yaRnMcDonuts:
While preparing to cut the next alpha release, I did a final engine build and ran some important pbr engine tests/examples, and I ran into a crash with the SimplePBRTest example.

I had forgotten that riccardo had changed this tank material to test the new modular pbr shaders when the shaders were in a separate directory for preliminary testing last year.

But now the duplicated modular shaders were deleted (and have replaced the main PBR shader), so this test crashes with an assetNotFoundException, and just needed updated to point back to Common/MatDefs/Light/PBRLighting.j3md.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.frag
@@ -20,7 +20,7 @@ uniform vec4 g_LightData[NB_LIGHTS];
 uniform vec3 g_CameraPosition;
 
 #ifdef USE_FOG
-    #import "MatDefs/ShaderLib/MaterialFog.glsllib"
+    #import "Common/ShaderLib/MaterialFog.glsllib"
 #endif
 
 void main(){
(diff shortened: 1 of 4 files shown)


yaRnMcDonuts:
I had to update a few more asset paths to point to the correct place, I moved things around alot after i started testing on my larger projects, and forgot to revert them back to point to the common/ directories.

But I've now tested PBRTerrainTest, TestPBRSimple, TestPBRLighting, and PBRTerrainAdvancedTest.

All 4 of these PBR tests/examples now run with the new modular PBR shader. 

However, the recent changes to modularize the terrain shaders were only applied to AdvancedPBRTerrain.j3md, so this has temporarily caused PBRTerrainTest and the non-advanced pbr terrain shader (PBRTerrain.j3md) to crash for the time being, and it is not advised to be used PBRTerrain.j3md until it is modularized as well, which will come is hortly after AdvancedPBRTerrain.j3md is done being tested and finalized in the next alpha release that will be out in a few hours.

PR #2314, opened 2024-10-22
update LWJGL3 to v3.3.4

stephengold:
to address issue #2298

PR #2362, opened 2025-02-05
LwjglWindow:  solve window centering incorrect in JME 3.8.0-alpha3

stephengold:
Due to a logic error, PR #2314 broke window centering with X windows and LWJGL3.
The breakage first appeared in 3.8.0-alpha3 .
The positioning calculations had a couple errors.
This PR corrects the errors.


stephengold:
I verified the fix in v3.8.0-alpha4.

PR #2365, opened 2025-02-06
Fix Rendering Issue in PBRTerrain

yaRnMcDonuts:
This PR attempts to fix a rendering bug that a few jme users have experienced on certain devices, as discussed in this thread: https://hub.jmonkeyengine.org/t/requesting-help-troubleshooting-pbrterrain-bug/47895

I believe the issue was due to a combination of things that were done incorrectly in the process of reading normal maps, calculating tangents, and then blending the normal values.

I believe it is also important to normalize the normal value prior to blending each layer, which I previously was not doing.

There's also some code that is necissary for putting a normal map in the proper range, and this was being done improperly for tri-planar normal mapping. It should be done 3 times (once per read) and then normalized, and this was incorrect previously.

Once this PR is done I'll post back to that thread and request testing from the users who were experiencing the bug on their device.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.vert
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/PBRLighting.vert
@@ -30,7 +30,7 @@ attribute vec3 inNormal;
 varying vec3 wNormal;
 varying vec3 wPosition;
 
-varying vec4 lPosition;
+varying vec3 lPosition;
 
 attribute vec4 inTangent;
 varying vec4 wTangent;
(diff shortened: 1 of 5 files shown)


yaRnMcDonuts:
I plan to merge this in the next 24 - 48 hours

stephengold:
@yaRnMcDonuts: perhaps it's time to delete the branch

yaRnMcDonuts:
I deleted the branch, and also went back through and deleted the branches in a couple other PRs where I forgot to do so.

I also realized I've been forgetting to flag some PRs to the 3.8.0 milestone recently, but I'll make sure I am remembering to do this going forward. Thanks for flagging the PRs I've missed in the meantime!

stephengold:
@yaRnMcDonuts: You're welcome.

I think it's particularly important to make the milestones accurate before creating a release branch, since that's when the meaning of "future release" suddenly changes!

PR #2367, opened 2025-02-10
Fix missing wViewDir variable and undeclared/wrongly declared vars

yaRnMcDonuts:
The wViewDir varying was changed to be stored in the surface.viewDir value, so the references to that variable needed to be updated to match this.

I also changed some floats to be in decimal format.


yaRnMcDonuts:
I plan to merge this PR in 24 hours.

Issue #2366, opened 2025-02-10
Spline's type can only be deserialized as CatmullRom


xiaojiao-zsw:
When Spline is serialized, type will be serialized to SplineType.CatmullRom by default, regardless of the current type value.
https://github.com/jMonkeyEngine/jmonkeyengine/blob/1f31c14af577e9d5b6b8f6d153823cbd7b30cb04/jme3-core/src/main/java/com/jme3/math/Spline.java#L465-L468

Also, when deserializing, type reads the key name as “pathSplineType” instead of “type”.
https://github.com/jMonkeyEngine/jmonkeyengine/blob/1f31c14af577e9d5b6b8f6d153823cbd7b30cb04/jme3-core/src/main/java/com/jme3/math/Spline.java#L497-L509

stephengold:
Good catch!
Thank you, @xiaojiao-zsw, for documenting this issue.
I'll write some tests and try to get this corrected in the next release.

stephengold:
On closer inspection, the `Spline.write()` code might be correct:
The value written by `OutputCapsule.write()` is the 1st argument. The 3rd argument to `OutputCapsule.write()` is a default value that is never written.

However, the mismatch of names ("pathSplineType" versus "type") is clearly a bug.

stephengold:
Testing uncovered another issue with spline serialization: in the case of an NURBS spline, the knot list is a non-null `List<Float>`. Line 483

https://github.com/jMonkeyEngine/jmonkeyengine/blob/98c0f9d20a038eb058a01b46bf684904b75ef679/jme3-core/src/main/java/com/jme3/math/Spline.java#L483

causes a `ClassCastException` to be thrown in `BinaryOutputCapsule.writeSavableArrayList()` because `Float` cannot be cast to `Savable`.

stephengold:
I submitted PR #2368 to solve the bugs, which apparently date back to at least 2014:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/a16857c8f4e7aadf3ca5fcb79bff5c8f16eb5481/jme3-core/src/main/java/com/jme3/math/Spline.java

In other words, they not recent regressions. It appears they were missed due to inadequate test coverage.

xiaojiao-zsw:
> On closer inspection, the `Spline.write()` code might be correct: The value written by `OutputCapsule.write()` is the 1st argument. The 3rd argument to `OutputCapsule.write()` is a default value that is never written.
> 
[...]

PR #2368, opened 2025-02-10
solve issue #2366 (bugs related to saving/loading a Spline)

stephengold:
This pull request solves issue #2366 and adds an automated test to prevent regressions.

The issue of spines being loaded as the wrong type is solved by passing the same tag ("type") to `readEnum()` as was passed to `oc.write()`.

The `ClassCastException` issue is solved by converting `knots` to/from a `float[]` array and saving/loading those arrays. Compatibility with serialized data from past releases is maintained in the case where `knots` is `null`---in other words for non-NURB splines. However, past releases couldn't serialize NURB splines without crashing, so this change shouldn't cause any issues in practice.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Spline.java
+++ b/jme3-core/src/main/java/com/jme3/math/Spline.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


yaRnMcDonuts:
I plan to merge this PR in 24 hours.

yaRnMcDonuts:
Oops, I said i'd merge in 24 hours... 2 weeks ago. Sorry for the delay!

stephengold:
> Sorry for the delay!

As a former release manager, I understand that life doesn't always go according to plan. No harm done.

PR #2369, opened 2025-02-11
Finish modularizing PBRTerrain.j3md

yaRnMcDonuts:
Previously I only modularized AdvnacedPBRTerrain.j3md

But now this PR applies the same concept to PBRTerrain.j3md

I ended up deleting the PBRTerrain.frag file, so now both of the terrain's .j3md files reference the same fragment shader, but the advanced version has a boolean called UseTextureArrays set to true to tell the fragment shader to read from texture arrays instead of normal textures.

I also added suppor for metallicRoughnessAoEi maps in the regular PRRTerrain shader (Ao being ambient occlusion that is packed in blue channel, and Ei being emissiv intensity which gets packed in the alpha channel). You will still be limited to 16 textures without using texture arrays, and as always I'd suggest most people juts use the advanced version of the shader primarily, but atleast the option is there.


stephengold:
To revert a commit, I'd use the "gitk" tool, not to be confused with "GitKraken".

https://git-scm.com/docs/gitk for more info about "gitk".

Gitk makes reverting a simple procedure. Make sure the relevant branch ("master" in this case) is checked out. Select the commit using the Left Mouse Button, then pull down with the Right Mouse Button and select "Revert This Commit" from the pulldown menu. Then push the branch to "origin".

Or I can do this to https://github.com/jMonkeyEngine/jmonkeyengine/commit/508967205d8d439a64d61483256daa3f201589e5 for you if you'd like.

yaRnMcDonuts:
> Or I can do this to [5089672](https://github.com/jMonkeyEngine/jmonkeyengine/commit/508967205d8d439a64d61483256daa3f201589e5) for you if you'd like.

Yes if its not too much of a hassle, that would be helpful.

I looked through the doc you linked but I may be another few days until I'm feeling better from being sick before I get back to working on the next release.

stephengold:
Sorry to hear you've been ill. I hope you're back to full health very soon.

I reverted 508967205d8d439a64d61483256daa3f201589e5 for you, at 2de6a7834d67f0db49136bb294693d18f9ee8e79 .

yaRnMcDonuts:
For some reason, my changes that I made to PBRTerrainUtils.glsllib are not showing up... the changes are there in the branch I'm using for this PR, but its not detecting changes when I look at that file compared to jme's master branch. I think I glitched something out wit this PR when I made that one commit directly, so I just went ahead and remade this whole PR: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2377

PR #2371, opened 2025-02-11
update LWJGL from v3.3.4 to v3.3.6

stephengold:
Last week, in [jMonkeyEngine 3.8.0-alpha3](https://github.com/jMonkeyEngine/jmonkeyengine/releases/tag/v3.8.0-alpha3), we finally upgraded jme3-lwjgl3 and jme3-vr to use [LWJGL v3.3.4](https://github.com/LWJGL/lwjgl3/releases/tag/3.3.4).

V3.3.4 is over 6 months old. Already there are 2 newer releases out: v3.3.5 and v3.3.6. The differences between them are so slight that I propose we upgrade directly from v3.3.4 to v3.3.6. This pull request implements that upgrade.

I'm unsure whether this change should go public in JME v3.8.0-stable or in a later release. I'll leave that decision to the release manager .

Either way, we'll need someone to test on VR hardware. Currently, I don't have access to such hardware. @jseinturier and @phr00t can you help?

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,7 +3,7 @@
 [versions]
 
 checkstyle = "9.3"
-lwjgl3 = "3.3.4"
+lwjgl3 = "3.3.6"
 nifty = "1.4.3"
 
 [libraries]
--- a/jme3-vr/src/main/java/com/jme3/input/vr/osvr/OSVR.java
+++ b/jme3-vr/src/main/java/com/jme3/input/vr/osvr/OSVR.java
@@ -33,6 +33,7 @@ import com.ochafik.lang.jnaerator.runtime.NativeSizeByReference;
 import com.sun.jna.Pointer;
 import com.sun.jna.ptr.PointerByReference;
 import java.nio.FloatBuffer;
+import java.nio.IntBuffer;
 import java.util.logging.Logger;
 
 /**
@@ -78,6 +79,7 @@ public class OSVR implements VRAPI {
      */
     public static byte[] OpenGLString = { 'O', 'p', 'e', 'n', 'G', 'L', (byte)0 };
 
+    private final IntBuffer lastError = IntBuffer.allocate(2);
     private final Matrix4f[] eyeMatrix = new Matrix4f[2];
 
     private PointerByReference grabRM;
@@ -180,7 +182,7 @@ public class OSVR implements VRAPI {
      */
     public void grabGLFWContext() {
         // get current context
-        wglGLFW = org.lwjgl.opengl.WGL.wglGetCurrentContext();
+        wglGLFW = org.lwjgl.opengl.WGL.wglGetCurrentContext(lastError);
         glfwContext = org.lwjgl.glfw.GLFW.glfwGetCurrentContext();
     }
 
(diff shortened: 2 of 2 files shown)


yaRnMcDonuts:
> I'm unsure whether this change should go public in JME v3.8.0-stable or in a later release. I'll leave that decision to the release manager .

I'd say it is a good decision to update to the newest LWJGL3 version, and to try to keep up to date anytime they release a new one.

If you can submit another PR upgrading the version, then I'll add it to the list of PRs ready for review in the 3.8 thread on the jme hub, and we can hopefully get some users with VR to test it in time.

stephengold:
> If you can submit another PR upgrading the version

This is that PR! If I was unclear, I apologize.

Based on your comments, I've gone ahead and added this to v3.8.0 milestone.

yaRnMcDonuts:
> This is that PR! If I was unclear, I apologize.

Oops that's my bad for misunderstanding, I am still sick and not thinking straight at times, so I mistook this as an issue and not a PR at first glance when I responded.

richardTingle:
I tried this with Tamarin and it crashes with

    Exception in thread "main" java.lang.NoSuchFieldError: Class org.lwjgl.openxr.XrVector3f does not have member field 
    'sun.misc.Unsafe UNSAFE'

But that is pretty much always the case with a change in LWJGL version. When jme 3.8.0 comes out I'll create a new Tamarin version and update the [compatibility table](https://github.com/oneMillionWorlds/Tamarin/wiki/jMonkeyEngine-Compatibility-Table). 

This problem is caused because Tamarin has to ship lwjgl-openxr separately (and so has to keep the JME LWJGL and Tamarin LWJGL versions in sync)

More worryingly when I do align the LWJGL versions I get this

    com.onemillionworlds.tamarin.openxr.OpenXrSessionManager$OpenXrException: Open XR returned an error code -4
    
I think this is the problem described at https://stackoverflow.com/questions/79048790/how-to-query-the-openxr-runtimes-api-version. I'll have an investigate. The stack overflow post does suggest an (ugly) solution.

p.s. I notice that this PR changes OSVR.java. Are you specifically looking for someone to test exercising that bit of code? OSVR is horrifically deprecated (like even more deprecated than the other VR support in JME). E.g. see this post on the OSVR reddit 6 years ago saying "don't go with OSVR https://www.reddit.com/r/OSVR/comments/bgby9y/what_happened_to_osvr_read_me_if_youre_new/)

stephengold:
> OSVR is horrifically deprecated

Perhaps v3.8.0 presents a good opportunity to remove OSVR from JMonkeyEngine. Worth discussing at the Forum, anyway.

richardTingle:
I've started a discussion at https://hub.jmonkeyengine.org/t/should-osvr-very-old-vr-standard-be-removed-from-jme/48234

richardTingle:
I've put in a PR to remove OSVR support at https://github.com/jMonkeyEngine/jmonkeyengine/pull/2380.

I tested it by exercising OpenVR (LWJGL version) on a Quest 3 just to make sure I hadn't inadvertently broken OpenVR support when I deleted OSVR 

Interestingly my virus checker (bitdefender) is convinced that LWJGL V3.3.4 has trojans in it (but not v3.3.6). A false alarm I'm sure but another reason to update. As a result I tested my changes with your changes merged into it (which is probably a good thing anyway - confirms 3.3.6 is fine with OpenVR).

It will merge conflict with this PR so whoever goes second will have to deal with that but it is a very mild merge conflict (we just have to accept the side that deletes OSVR.java

Issue #2317, opened 2024-10-23
Remove SingleLayerInfluenceMask due to Serialization Issues and Dependency Conflicts. A Custom Solution Not Suited for Core


capdevon:
The recent change to `ArmatureMask` implementing `Savable` has caused an issue with `SingleLayerInfluenceMask` (which extends `ArmatureMask`) serialization. 
Due to its dependencies on `AnimComposer` and `SkinningControl`, serialization is now impossible and the class exhibits inconsistent behavior.

Given its reliance on these external controls and potential for conflicts, I believe `SingleLayerInfluenceMask` is a custom solution best suited for specific needs. Its presence in the jME engine core can lead to maintenance issues and hinder the future evolution of animation base classes.

We recommend its removal from the jME engine core.
NOTE: this class was introduced in the latest 3.7.0-stable release.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/SingleLayerInfluenceMask.java#L42

codex128:
Hmm, that's a shame. I think it is a particularly useful class. Perhaps this would work, along with some documentation?

```java
@Override
public void write(JmeExporter ex) throws IOException {
    throw new UnsupportedOperationException("Read/write not supported.");
}
@Override
public void read(JmeImporter im) throws IOException {
    throw new UnsupportedOperationException("Read/write not supported.");
}
```

If not, I can move SingleLayerInfluenceMask to a utility library.

capdevon:
Personally, I do not think this is an acceptable solution. This class is not necessary for the proper functioning of the engine. It is a utility class and should be in an external library.

To maintain flexibility and avoid potential conflicts, I propose that `SingleLayerInfluenceMask` be moved to an external library. This would allow for more focused development and maintenance, while also preserving its accessibility for those who find it useful.

capdevon:
Alternatively, I propose moving this class to the `jme3-examples` module, where it can serve as a basis for users who wish to optimize or adapt it for their projects.

codex128:
Not moving to examples, certainly, but moving to `jme3-plugins` might be a good idea. I've personally been very busy lately, so I do not have the time to do this myself, but you're welcome to open a PR for this.

What I am most concerned about is that SingleLayerInfluenceMask was already released with 3.7-stable, so it can't really be moved anymore without breaking backwards compatibility.

capdevon:
[...]

PR #2372, opened 2025-02-12
SingleLayerInfluenceMask:  fix serialization issues

capdevon:
see https://github.com/jMonkeyEngine/jmonkeyengine/issues/2317

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/SingleLayerInfluenceMask.java
+++ b/jme3-core/src/main/java/com/jme3/anim/SingleLayerInfluenceMask.java
@@ -66,6 +66,17 @@ public class SingleLayerInfluenceMask extends ArmatureMask {
         this.targetLayer = targetLayer;
         this.animComposer = animComposer;
     }
+    
+    /**
+     * Instantiate a mask that affects no joints.
+     * 
+     * @param targetLayer  The layer this mask is targeted for.
+     * @param animComposer The animation composer associated with this mask.
+     */
+    public SingleLayerInfluenceMask(String targetLayer, AnimComposer animComposer) {
+        this.targetLayer = targetLayer;
+        this.animComposer = animComposer;
+    }
 
     /**
      * Get the layer this mask is targeted for.
--- a/jme3-core/src/main/java/com/jme3/anim/SingleLayerInfluenceMask.java
+++ b/jme3-core/src/main/java/com/jme3/anim/SingleLayerInfluenceMask.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 jMonkeyEngine
+ * Copyright (c) 2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


yaRnMcDonuts:
Is there an inherit reason that the serialization issue cannot be fixed?

I am not an expert in serialization, so I apologize if I'm overlooking something obvious here.

codex128:
The issue would indeed be resolved if AnimComposer and SkinningControl were removed. The problem is that AnimComposer is being used to check if any higher layers are using a joint. Without it, SingleLayerInfluenceMask is useless.

SkinningControl I think could be removed. A lot of neat utility methods would have to go with it, but it can probably be removed anyway.

yaRnMcDonuts:
> A lot of neat utility methods would have to go with it, but it can probably be removed anyway.

What if you change those methods to take the AnimComposer or SkinningControl as a param through the constructor? This would allow that method to use AnimComposer without needing AnimComposer as a class variable:


`    public void makeLayer(AnimComposer animComposer) {
        animComposer.makeLayer(layer, this);
    }
`

You would then obviously have to change some of your own code that calls these methods to pass in the new AnimComposer param. But this way these non-serializable variables would no longer need stored in this class that needs serialized

yaRnMcDonuts:
Hi @codex128 , I'm curious if you have considered my suggestion? I'd be willing to keep this class in the engine if you can make the changes I suggested to resolve the serialization issues. Or is there something more to this issue I am possibly missing that makes my suggestion invalid? 

Removing it would be the last resort, and there's a good case for why this may not have been a good addition in the first place. But removing an existing class altogether isn't typically a great decision once users may be relying on it, so I think the optimal solution would be to fix the serialization issues so it doesn't cause any future issues.

codex128:
Ah, yes, thanks for reminding me. It slipped my mind and I've been rather busy lately. I've submitted #2376 that incorporates your suggestion.

capdevon:
Hi guys, 
I've edited the PR, you can consider it a revision of @codex's PR to fix the serialization issues already discussed.

I rewrote the `SingleLayerInfleunceMask` class removing all problematic methods.
I have optimized the variable names, the code and the javadoc to make it clearer and simpler. I removed the static methods because they only add complexity for future maintenance and the methods of the `ArmatureMask` superclass are already sufficient to achieve the same result. I rewrote the test class more clearly, adding the missing mask serialization test.

I believe this is the only alternative to not deprecate the class.

Please let me know your thoughts.

@yaRnMcDonuts @codex128

codex128:
It looks fine to me.

> I removed all methods for adding Joints to the mask because they add complexity to future maintenance and are already present in the ArmatureMask superclass.

I would only suggest keeping `addAll` or moving it to ArmatureMask, because it is new and useful functionality. If I remember correctly the alternative is to use `addFromJoint` on the root joint, which is little bit of a hassle. I'm ok with not keeping it, though.

capdevon:
> I would only suggest considering keeping `addAll` or moving it to ArmatureMask, because it is new and useful functionality. If I remember correctly the alternative is to use `addFromJoint` on the root joint, which is little bit of a hassle. I'm ok with not keeping it, though.

The `addAll` method is not necessary. The constructor of the `ArmatureMask` superclass with the `Armature` parameter does exactly the same thing. I've updated the test class to highlight this detail.

@yaRnMcDonuts  if you also agree, I propose to move forward with this solution.

yaRnMcDonuts:
That sounds good to me!

yaRnMcDonuts:
If everything looks good to both @capdevon and @codex128 and if there is nothing else you plan to change, then I will merge this in about 48 hours. Thanks to both of you for collaborating on this to find a good solution!

PR #2373, opened 2025-02-17
update LWJGLX (AWT) from 0.1.8 to 0.2.3

JNightRider:
This PR updates the lwjglx-awt library to the latest version ([V0.2.3](https://central.sonatype.com/artifact/org.lwjglx/lwjgl3-awt)). In this new version, an error that was reported in the [forum](https://hub.jmonkeyengine.org/t/failed-to-pass-test-case-jme3test-awt-testcanvas-on-jme-3-7-0-with-lwjgl3/47990?u=swiftwolf)  has been solved (correction in [lwjgl-awt](https://github.com/LWJGLX/lwjgl3-awt/pull/81))

I would like to inform you that these changes are related to PR #2371, that is; that this PR depends on PR #2371 so that AWT does not break when updating LWJGL3.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,7 +3,7 @@
 [versions]
 
 checkstyle = "9.3"
-lwjgl3 = "3.3.4"
+lwjgl3 = "3.3.6"
 nifty = "1.4.3"
 
 [libraries]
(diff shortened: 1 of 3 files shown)


yaRnMcDonuts:
It looks like this is the first PR that has failed from the new screenshot tests. However I don't think there's anything inherently wrong with this PR, so I will look into it and see how to resolve it as soon as I can.

yaRnMcDonuts:
Hi @richardTingle

If you get a chance, could you take a look at the failed screenshot test and let me know if it looks like it is something to be worried about?

It appears there is nothing wrong and the rest of the tests passed, so the PR should be good to merge.

However this is the first instance I have seen where the new test fails, so I am curious to hear your analysis of the situation. Thanks!

JNightRider:
For this PR to pass the test, I think it is better to merge #2371 first and then update this PR.

JNightRider:
Everything indicates that this PR has passed all the tests (it is finally ready).

yaRnMcDonuts:
Thanks for the quick follow up on resolving the merge conflicts! I will go ahead and merge this in the next 24-48 hours

Issue #2374, opened 2025-02-17
NoSuchMethodError AnimComposer.makeLayer()


stephengold:
The MavDemo1 application in my jme-vehicles project uses v1.1.1 of the JmePower library, which was compiled for JME v3.6.0-stable.

When I build MavDemo1 with JME v3.8.0-alpha3, initialization of the app fails with the following diagnostic messages on the console:
```console
> Task :MavDemo1:run
Feb 17, 2025 2:28:59 PM com.jayfella.jme.vehicle.lemurdemo.MavDemo1 main
WARNING: Assertions are enabled.
Feb 17, 2025 2:29:03 PM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[jME3 Main,5,main]
java.lang.NoSuchMethodError: 'void com.jme3.anim.AnimComposer.makeLayer(java.lang.String, com.jme3.anim.AnimationMask)'
	at com.github.stephengold.jmepower.JmeLoadingState.setupCinematic(JmeLoadingState.java:409)
	at com.github.stephengold.jmepower.JmeLoadingState.startCinematic(JmeLoadingState.java:557)
	at com.github.stephengold.jmepower.JmeLoadingState.update(JmeLoadingState.java:319)
	at com.jme3.app.state.AppStateManager.update(AppStateManager.java:371)
	at com.jme3.app.SimpleApplication.update(SimpleApplication.java:260)
	at com.jme3.system.lwjgl.LwjglWindow.runLoop(LwjglWindow.java:707)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:797)
	at java.base/java.lang.Thread.run(Thread.java:840)

Exception in thread "Thread-3" java.lang.NullPointerException: Cannot invoke "com.jme3.input.InputManager.addRawInputListener(com.jme3.input.RawInputListener)" because the return value of "com.jme3.app.Application.getInputManager()" is null
	at com.simsilica.lemur.event.KeyInterceptState.<init>(KeyInterceptState.java:65)
	at com.simsilica.lemur.GuiGlobals.<init>(GuiGlobals.java:157)
	at com.simsilica.lemur.GuiGlobals.initialize(GuiGlobals.java:125)
	at com.jayfella.jme.vehicle.lemurdemo.LemurLoadingState.initializeLemur(LemurLoadingState.java:100)
[...]

PR #2231, opened 2024-03-20
AnimComposer: return AnimLayer on the makeLayer and removeLayer methods.

capdevon:
Standardize the `AnimComposer.makeLayer()` and `AnimComposer.removeLayer()` methods similar to `AnimComposer.makeAction()` and `AnimComposer.removeAction()` to simplify interaction with created and deleted objects.

PR #2375, opened 2025-02-19
solve issue #2374 (NoSuchMethodError)

stephengold:
This restores the pre-PR #2231 API of `AnimComposer`, in order to improve compatibility with JME 3.7 .

It then adds the 2 new methods (which return objects) under different names, to provide the functionality desired by @capdevon.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
+++ b/jme3-core/src/main/java/com/jme3/anim/AnimComposer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2024 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


capdevon:
Hi everyone,
Thank you for identifying the backwards compatibility issue and for all your hard work. While Stephen's PR offers a great solution to preserve my proposal, adding two more methods might lead to confusion and increased maintenance in the future.
Given that my PR isn't crucial and may cause more issues than benefits, I believe the best solution in this case is to discard my PR and revert the API to version 3.7
I appreciate your understanding and time. Let's continue working together to make the jme engine even better!

stephengold:
@capdevon thanks for your perspective and insight

yaRnMcDonuts:
I'll go ahead and close this PR, and I've created a separate PR that reverts the changes in the original PR that caused the regression

PR #2377, opened 2025-02-20
Finish Modularizing PBRTerrainUtils.glsllb

yaRnMcDonuts:
Previously I only modularized AdvnacedPBRTerrain.j3md

But now this PR applies the same concept to PBRTerrain.j3md

I ended up deleting the PBRTerrain.frag file, so now both of the terrain's .j3md files reference the same fragment shader, but the advanced version has a define that gets set to true to tell the fragment shader to read from texture arrays instead of normal textures.

I also added suppor for metallicRoughnessAoEi maps in the regular PRRTerrain shader (Ao being ambient occlusion that is packed in blue channel, and Ei being emissiv intensity which gets packed in the alpha channel). You will still be limited to 16 textures without using texture arrays, and as always I'd suggest most people juts use the advanced version of the shader primarily, but atleast the option is there.


yaRnMcDonuts:
I plan to merge this within the next 24-48 hours, then I'll cut the next alpha release. 

I might make some minor final changes  if I run into errors testing things one last time tomorrow, but everything is solidified now and all redundant code is condensed into AdvancedPBRTerrain.frag and the PBRTerrainUtils glsllib. 

The only places that still have duplicate code are the 2 .j3md files, however there is no realistically clean way to combine both pbr terrain shaders into one .j3md file and I'd say it is best they still each have their own .j3md file anyways so that it is obvious that one is intended to use texture arrays and the other basic textures.

yaRnMcDonuts:
Making a few more changes to add support for NORMAL_TYPE so that you can flip terrain normals in shader the same way its done in PBRLighting.j3md.

yaRnMcDonuts:
Everything is fixed and thoroughly tested now. 

Both terrain examples run with no exceptions, and I ensured that the final render is identical in both triplanar and non-triplanar mode for both examples. 

I also tested everything in the Afflicted Forests (my own game) as well as my custom scene editor app, and I ensured that there were no exceptions or rendering issues there. 

Despite some fixes that improve the final render and improve the effects from normal maps, I'm still not satisfied with the final results of how things look in the terrain examples, but I think that has more to do with the free textures I used, and also the lighting settings. The examples could probably use a brighter light probe and a different direction set for the DirectionalLight ( help emphasize the poor normal maps that are used by those examples). But at least the examples still work to show how to use the shader, I suppose that's the important part.

PR #2378, opened 2025-02-20
Revert "AnimComposer: return AnimLayer on the makeLayer and removeLayer methods."

yaRnMcDonuts:
Reverts jMonkeyEngine/jmonkeyengine#2231

As discussed here: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2375


capdevon:
thanks @yaRnMcDonuts

Issue #2329, opened 2024-11-17
GLSLCompat.glsllib contains ';' at the end of the first few line and Android is not happe with it


nickidebruyn:
While test jme3.7.0 with my android projects I found that the jme3-core/src/main/resources/Common/ShaderLib/GLSLCompat.glsllib
code contains ';' semicolons at the end of the first few lines and android is not happy with that.

Error I am experiencing:
```
ERROR: 0:11: ';' : Syntax error:  syntax error
INTERNAL ERROR: no main() function!
ERROR: 1 compilation errors.  No code generated.                                                                                                    
                                                                                                    	at com.jme3.renderer.opengl.GLRenderer.updateShaderSourceData(GLRenderer.java:1658)
                                                                                                    	at com.jme3.renderer.opengl.GLRenderer.updateShaderData(GLRenderer.java:1685)
                                                                                                    	at com.jme3.renderer.opengl.GLRenderer.setShader(GLRenderer.java:1750)
```

stephengold:
Thanks for documenting the issue, @nickidebruyn!

stephengold:
According to "git blame", the failing code was last modified by PR #2217, which was integrated between v3.7.0-alpha2 and v3.7.0-beta1.

@nickidebruyn Do your Android projects work with JME v3.7.0-alpha2 ?

nickidebruyn:
I tested a few scenarios for this. I can confirm that none of the new version (v3.7.0-alpha2, v3.7.0-beta1, v3.7.0-stable) works.
I then tested again with v3.6.1 and it worked again.
After that I changed back to v3.7.0-stable and created my own local GLSLCompat.glsllib and changed the frag shader to point to that and it worked again. This is the code I removed from my local: GLSLCompat.glsllib

```
#ifdef GL_ES
  #ifdef FRAGMENT_SHADER
    precision highp float;
    precision highp int;
    precision highp sampler2DArray;
    precision highp sampler2DShadow;
    precision highp samplerCube;
    precision highp sampler3D;
    precision highp sampler2D;
    #if __VERSION__ >= 310
      precision highp sampler2DMS;
    #endif
  #endif
#endif
```

stephengold:
Thank you for the investigation.
Based on your findings, I'm convinced PR #2061 caused the regression.
At the time (August 2023), Riccardo indicated 2061 would need to be tested on Android.
Apparently it never got such testing, and neither did JME 3.7!
[...]

PR #2381, opened 2025-02-25
Fix Android crash from precision qualifiers in GLSLCompat.glsllib

yaRnMcDonuts:
Addresses this issue: https://github.com/jMonkeyEngine/jmonkeyengine/issues/2329#issuecomment-2676145431

Feedback and review is appreciated, as I only just delved into learning about these precision qualifiers and I also don't do any jme dev on android to test this myself.


pavly-gerges:
Thanks for working on it, I am willing to take the initiative to test this. However, first I want to understand what the issue was (Because I am not very familiar with GLSL). It seems to me to be a compilation issue due to using semi-colons? But that was not addressed in this PR?

Issue #2386, opened 2025-03-05
implicit cast from "vec4" to "vec3" in PBRLighting.frag


stephengold:
Shortly after loading the "SpecGlossVsMetalRough" model from glTF-Sample-Assets, my CompareLoaders app crashes with a bad shader compile.

This crash is reproducible with JME v3.8.0-alpha4 and -alpha3 but not -alpha2, so I think it's a recent regression.

The test model can be found online:
https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/SpecGlossVsMetalRough

```console
Successfully loaded "SpecGlossVsMetalRough" from the gltf-sample-assets group using the Default loader(s).
======
Mar 05, 2025 11:19:00 AM com.jme3.renderer.opengl.GLRenderer updateShaderSourceData
WARNING: Bad compile of:
1	#version 150 core
2	#define SRGB 1
3	#define FRAGMENT_SHADER 1
4	#define BOUND_DRAW_BUFFER 0
5	#define BASECOLORMAP 1
6	#define EMISSIVE 1
7	#define SPECGLOSSPIPELINE 1
8	#define SPECULAR_AA 1
9	#define NORMAL_TYPE -1.0
10	#define SINGLE_PASS_LIGHTING 1
11	#define NB_LIGHTS 3
12	#define NB_PROBES 1
13	#define USE_AMBIENT_LIGHT 1
14	#extension GL_ARB_shader_texture_lod : enable
15	// -- begin import Common/ShaderLib/GLSLCompat.glsllib --
16	#ifdef GL_ES
17	  #ifdef FRAGMENT_SHADER
18	    precision highp float;
19	    precision highp int;
20	    #if __VERSION__ >= 130
21	      precision highp sampler2DArray;        
22	    #endif
23	    precision highp sampler2DArray;
24	    precision highp sampler2DShadow;
25	    precision highp samplerCube;
26	    precision highp sampler3D;
27	    precision highp sampler2D;
28	    #if __VERSION__ >= 310
29	      precision highp sampler2DMS;
30	    #endif
31	  #endif
32	#endif
33	
34	#if defined GL_ES
35	#  define hfloat highp float
36	#  define hvec2  highp vec2
37	#  define hvec3  highp vec3
38	#  define hvec4  highp vec4
39	#  define lfloat lowp float
40	#  define lvec2 lowp vec2
41	#  define lvec3 lowp vec3
42	#  define lvec4 lowp vec4
[...]

PR #2387, opened 2025-03-06
Fix logic errors and exception in SpecGloss pipeline

yaRnMcDonuts:
This PR addresses this issue: 
#2386 
And then I also found another critical logic error in the spec gloss pipeline that was likely there even before riccardo adapted the shader to use structs:

It looks like the specular and glossiness value taken from specularMaps was being totally ignored and the glossiness value was not being accounted for in the final roughness calculation. The code further down in the shader was calculating the roughness solely based on m_Glossiness with no regard for the glossiness scalar read from texture maps, which is incorrect.

The logic error and exception should all be fixed in this PR now, but it will need testing from a spec gloss model with packed specularGlossinesMap as well as a model with seperate specular and glossiness maps in order to ensure everything works properly now.

### DIFF
--- a/jme3-core/src/main/resources/Common/ShaderLib/module/pbrlighting/PBRLightingUtils.glsllib
+++ b/jme3-core/src/main/resources/Common/ShaderLib/module/pbrlighting/PBRLightingUtils.glsllib
@@ -384,7 +384,7 @@
             //spec gloss tex reads:
 
             #ifdef SPECGLOSSPIPELINE
-                flota glossiness = m_Glossiness;
+                float glossiness = m_Glossiness;
                 #ifdef USE_PACKED_SG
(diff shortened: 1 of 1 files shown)


yaRnMcDonuts:
I also condensed and cleaned up the code. 

Some lines of code were repeating in the packed and non packed specular and glossiness texture reads, and as a result past editors ended up applying changes to one spot and not the other, leading to inconsistent functionality when using packed vs non packed maps.

stephengold:
I built a local snapshot of JME at Git hash 154022e and re-ran the test that uncovered issue #2386.
There was no crash, and the rendered model matched my expectations, so I believe this PR will solve the issue.

yaRnMcDonuts:
I also tested a model with a packed specGloss map and it renders properly now too. So I will merge this in the next 24 hours

Issue #2370, opened 2025-02-11
MotionEvent.cloneFields() should deeply clone the path field


stephengold:
The `MotionEvent` class implements `Control` and could therefore be included in a model asset. It also contains a mutable `MotionPath` field, which gets serialized and de-serialized. And `MotionPath` contains a mutable `Spline` field, which gets serialized and de-serialized.

The usual use case for deserialization is to load scene-graph assets. Scene graphs are "smart assets" that get cached on first load. Subsequent loads are supposed to yield deep clones of the cached assets.

`MotionPath` and `Spline` are mutable objects; they ought to be replicated in any deep clone. However,  `path` field isn't cloned in `MotionEvent.cloneFields()`, probably because `MotionPath` isn't cloneable. And `MotionPath` can't be cloned because `Spline` isn't cloneable.

`MotionPath` and `Spline` should implement the `JmeCloneable` interface, and `MotionEvent.cloneFields()` should clone the `path` field.

Issue #879, opened 2018-08-20
Cloner fails on enums and wrapped primitives


stephengold:
It would be convenient if com.jme3.util.clone.Cloner worked on enums and wrapped primitives (such as Float and Boolean).

Currently, the test code (below) fails with an IllegalArgumentException because f is a Float, which is not an array, nor Cloneable, nor JmeCloneable.

```
package mygame;

import com.jme3.app.SimpleApplication;
import com.jme3.font.LineWrapMode;
import com.jme3.util.clone.Cloner;

public class Main extends SimpleApplication {

    public static void main(String[] args) {
        Main app = new Main();
        app.start();
    }

    @Override
    public void simpleInitApp() {
        Boolean bo = false;
        Byte by = 0;
        Double d = 0.0;
        LineWrapMode e = LineWrapMode.Clip;
        Float f = 0f;
        Integer i = 0;
        Long l = 0L;

        Object clone;
        clone = Cloner.deepClone(f);
        clone = Cloner.deepClone(bo);
        clone = Cloner.deepClone(by);
        clone = Cloner.deepClone(d);
        clone = Cloner.deepClone(e);
        clone = Cloner.deepClone(i);
        clone = Cloner.deepClone(l);
    }
}
```

pspeed42:
There is no reason to deep clone Float or enums... in fact, if cloning an enum does anything but just copying the reference then the enum would be broken.

So, here is how you "clone" an enum above:
clone = e;

stephengold:
For situations where the type of Object is unknown at compile time, I work around this issue like so:
```
        if (object instanceof Boolean
                || object instanceof Byte
                || object instanceof Character
                || object instanceof Enum) {
            // a unique instance for every possible value, so alias
            clone = object;

        } else if (object instanceof String) {
            // instances are immutable, so alias
            clone = object;

        } else if (object instanceof Double) {
            clone = (double) object;
        } else if (object instanceof Float) {
            clone = (float) object;
        } else if (object instanceof Integer) {
            clone = (int) object;
        } else if (object instanceof Long) {
            clone = (long) object;
        } else if (object instanceof Short) {
            clone = (short) object;

        } else {
            clone = Cloner.deepClone(object);
        }
```

pspeed42:
In what cases does this come up?

[...]

PR #2390, opened 2025-03-06
solve issue #2370 (deeply clone MotionEvent path)

stephengold:
Currently a work in progress.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/cinematic/MotionPath.java
+++ b/jme3-core/src/main/java/com/jme3/cinematic/MotionPath.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


stephengold:
This PR was created to address issue #2370. Here's what it does:

+ It adds code to `MotionEvent.cloneFields()` to clone the `path` field. While implementing this change, I discovered that the `lookAt` and `rotation` fields also needed to be cloned. (The `direction` and `upVector` fields are cloned in the `jmeClone()` method, which is irregular but suffices for most use cases.)

+ It makes `MotionPath` and `Spline` deeply cloneable, adding `jmeClone()` and `cloneFields()` methods. Unfortunately, `Cloner.clone()` does not work on `ArrayList<Float>` because `Float` is not cloneable---see issue #879. In `Spline.cloneFields()` I used copy constructors to work around this limitation.

+ It adds automated cloning tests for all 3 classes to jme3-core/src/test. The cloning tests for `Spline` use the same test splines as the existing serialization tests. To avoid duplicating code, I refactored the generation of test splines to private methods.

+ The copyright notices in the affected source files were updated to include the year 2025.

yaRnMcDonuts:
Since no one else has reviewed this yet, I did my best to look through the code and I don't see any apparent issues. 

So I will leave this open for another day or two and then I plan to merge this in ~48 hours.

PR #2391, opened 2025-03-07
Fix exception when using Instancing with Fog in PBR

yaRnMcDonuts:
The fog code in PBRLighting.j3md is encountering the same issue that was reported before with Lighting.j3md : https://github.com/jMonkeyEngine/jmonkeyengine/pull/1458

This PR should fix the issue.

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Light/Lighting.vert
+++ b/jme3-core/src/main/resources/Common/MatDefs/Light/Lighting.vert
@@ -10,8 +10,8 @@
 
 // fog - jayfella
 #ifdef USE_FOG
-varying float fog_distance;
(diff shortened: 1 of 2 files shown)


yaRnMcDonuts:
I just realized that the variable name for the varying distance variable is different in PBRLighting.vert compared to Lighting.vert. And this means fog is broken in lighting

one is fog_distance and the other is fogDistance.

I'm fixing both varyings to be named fogDistance. I typically don't like using the _ character in regular variable names unless im working with a define (since defines are in all caps so _ character becomes useful since you can't use upper and lowercase to differentiate words in the variable name when its already all uppercase. For example I'd name the corresponding define FOG_DISTANCE but a uniform should just be named fogDistance)

PR #2395, opened 2025-03-10
Fix TextureArray not exporting WrapMode

### LINKED ISSUE #2394: TextureArray not exporting WrapMode
The TextureArray class does not implement write() and read() and so when it is exported to .j3o and read back it does no longer wrap in the same way as originally set.
I compared with Texture2D and that one does implement these methods.
I will submit a PR with this change.

jcfandino:
Fixes #2394

### DIFF
--- a/jme3-core/src/main/java/com/jme3/texture/TextureArray.java
+++ b/jme3-core/src/main/java/com/jme3/texture/TextureArray.java
@@ -175,6 +175,9 @@ public class TextureArray extends Texture {
 
     @Override
     public boolean equals(Object other) {
+        if (this == obj) {
+            return true; 
+        }
         if (!(other instanceof TextureArray)) {
             return false;
         }
--- a/jme3-core/src/main/java/com/jme3/texture/TextureArray.java
+++ b/jme3-core/src/main/java/com/jme3/texture/TextureArray.java
@@ -175,7 +175,7 @@ public class TextureArray extends Texture {
 
     @Override
     public boolean equals(Object other) {
-        if (this == obj) {
+        if (this == other) {
             return true; 
         }
         if (!(other instanceof TextureArray)) {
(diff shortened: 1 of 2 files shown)


jme3-core/src/test/java/com/jme3/texture/TextureArrayTest.java
```diff
@@ -0,0 +1,42 @@
+package com.jme3.texture;
```
stephengold, line 1: Please add a copy of the JME license.

jcfandino: Done!

jme3-core/src/main/java/com/jme3/texture/TextureArray.java
```diff
@@ -151,4 +156,41 @@ public void setWrap(WrapMode mode) {
...
+    @Override
+    public void read(JmeImporter importer) throws IOException {
+        super.read(importer);
+        InputCapsule capsule = importer.getCapsule(this);
+        wrapS = capsule.readEnum("wrapS", WrapMode.class, WrapMode.EdgeClamp);
+        wrapT = capsule.readEnum("wrapT", WrapMode.class, WrapMode.EdgeClamp);
+    }
+
+    @Override
+    public boolean equals(Object other) {
```
yaRnMcDonuts, line 177: The equals() method just needs updated to include the == check first:

```
       if (this == obj) {
            return true; 
        }
```

Otherwise everything looks good with the PR and it should be ready to merge after this change.

jcfandino: Good catch!
I'll try to fix it as soon as I can.
I might not find the time in the next few days so if you want to make the change quick please go ahead.
thx

yaRnMcDonuts: Sounds good, I've went ahead and made the change for you, so everything should be good and ready to merge now. Thanks for your contribution! 

And I also owe you a second thanks, since reporting this issue also helped me figure out how to properly save my project's Texture Arrays

yaRnMcDonuts:
I made the changes that I requested in my review (with the permission of the PR author) and everything looks good. So I will plan to merge this in ~48 hours.

PR #2396, opened 2025-03-11
update the jinput library to v2.0.10

stephengold:
Currently JME is using version 2.0.9 of the jinput library, which was released in 2018. This library is used only with LWJGL 2.x (jme3-lwjgl) never with LWJGL 3.x (jme3-lwjgl3).

This PR upgrades the library to v2.0.10, published in 2024. The release notes are here: https://github.com/jinput/jinput/releases/tag/2.0.10

I'll run a few tests in the hope of catching any major regressions. Until then, this PR should be considered a draft.


stephengold:
I just realized this PR requires JDK 17. It'll have to be deferred until JME stops supporting JDK 11.

PR #2403, opened 2025-03-30
Improved fix for IndexOutOfBoundsException

tonihele:
Ok, this is @pspeed42 's patch applied on top of master. I did not revert https://github.com/jMonkeyEngine/jmonkeyengine/pull/2297, I kept the good parts from the original patch. But I did not include the _fix for the symptom_ just to be sure that everything is always ok and we don't accidentally always just fallback to costly bandaid.


yaRnMcDonuts:
Looks good to me. I'll merge this within the next day or two.

PR #2406, opened 2025-04-17
Add defensive copy to EmitterShape's Vector3f

capdevon:
* Update licence year
* Update javadoc
A defensive copy is created to ensure that any external modifications to the original Vector3f objects will not affect the EmitterShape's internal state.


capdevon:
Closing PR: Based on Feedback:

This PR has the potential to break existing code that relies on a shared shape position across multiple emitter shapes. Specifically, any code designed to have changes propagate through external modifications would be affected

PR #2409, opened 2025-04-23
Update SeparateJointModelTransform.java

utsavsavani01:
Avoid mutating input

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/SeparateJointModelTransform.java
+++ b/jme3-core/src/main/java/com/jme3/anim/SeparateJointModelTransform.java
@@ -30,9 +30,12 @@ public class SeparateJointModelTransform implements JointModelTransform {
 
     @Override
     public void applyBindPose(Transform localTransform, Matrix4f inverseModelBindMatrix, Joint parent) {
-        localTransform.fromTransformMatrix(inverseModelBindMatrix.invert());
+        Matrix4f safeInverseBind = inverseModelBindMatrix.clone().invert(); // avoid mutating input
+        localTransform.fromTransformMatrix(safeInverseBind);
+
         if (parent != null) {
-            localTransform.combineWithParent(parent.getModelTransform().invert());
+            Transform safeParentTransform = parent.getModelTransform().clone().invert(); // same here
+            localTransform.combineWithParent(safeParentTransform);
         }
     }
 


utsavsavani01:
Avoid mutating input

jme3-core/src/main/java/com/jme3/anim/SeparateJointModelTransform.java
```diff
@@ -30,9 +30,12 @@ public void getOffsetTransform(Matrix4f outTransform, Matrix4f inverseModelBindM
 
     @Override
     public void applyBindPose(Transform localTransform, Matrix4f inverseModelBindMatrix, Joint parent) {
-        localTransform.fromTransformMatrix(inverseModelBindMatrix.invert());
+        Matrix4f safeInverseBind = inverseModelBindMatrix.clone().invert(); // avoid mutating input
```
capdevon, line 33: The `Matrix4f.invert()` method already creates a new object, so there is no need to clone the input matrix before inverting it.

jme3-core/src/main/java/com/jme3/anim/SeparateJointModelTransform.java
```diff
@@ -30,9 +30,12 @@ public void getOffsetTransform(Matrix4f outTransform, Matrix4f inverseModelBindM
...
+        localTransform.fromTransformMatrix(safeInverseBind);
+
         if (parent != null) {
-            localTransform.combineWithParent(parent.getModelTransform().invert());
+            Transform safeParentTransform = parent.getModelTransform().clone().invert(); // same here
```
capdevon, line 37: The `Transform.invert()` method already creates a new object, so there is no need to clone the parent model Transform of before inverting it.

capdevon:
The `Matrix4f.invert()` method already creates a new object, so there is no need to clone the input matrix before inverting it.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/math/Matrix4f.java#L1484

The `Transform.invert()` method already creates a new object, so there is no need to clone the parent model Transform of before inverting it.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/math/Transform.java#L424

Given the changes you have introduced, this PR can be closed.

capdevon:
Hi @utsavsavani01 ,
You don't need to clone a `Matrix4f` before inverting it because the `invert()` method already produces a new, independent result.
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/math/Matrix4f.java#L1484

You don't need to clone a `Transform` before inverting it because the `invert()` method already produces a new, independent result.
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/math/Transform.java#L424

Edit:
To properly evaluate this change, please explain the problem it solves and provide details on how the results were tested. Otherwise, I suggest closing this PR.

yaRnMcDonuts:
> You don't need to clone a `Matrix4f` before inverting it because the `invert()` method already produces a new, independent result.

> To properly evaluate this change, please explain the problem you are trying to solve and provide details on how the results were tested. Otherwise, I suggest closing this PR.

I agree. If you still are having issues, I would recommend posting on the jme hub for troubleshooting help if you are still having any issues, but the code is already safely avoiding mutating input and no changes are needed, so I'm going to close this PR for now.

PR #2411, opened 2025-04-25
iOS support

joliver82:
- Added jme3-ios-native library mostly based on old native iOS code prepared to work with both MOE and RoboVM (robovm is still WIP)
- Modified jme3-ios adding buffer allocator and proper getPlatform method

Pending to add jme3-ios-native library to be uploaded to maven. Any help is appreciated

### DIFF
--- a/jme3-ios-native/src/JmeAppHarness.java
+++ b/jme3-ios-native/src/JmeAppHarness.java
@@ -16,7 +16,7 @@ import java.util.logging.Logger;
 public class JmeAppHarness extends IosHarness{
 
     private static final Logger logger = Logger.getLogger(JmeAppHarness.class.getName());
-	protected GLRenderer renderer;
+	protected Renderer renderer;
 	protected IosInputHandler input;
 	protected boolean autoFlush = true;
 	protected Vector2f resizePending = null;
@@ -66,7 +66,7 @@ public class JmeAppHarness extends IosHarness{
         logger.log(Level.FINE, "JmeAppHarness appDraw");
         if (renderer == null) {
             JmeContext iosContext = app.getContext();
-            renderer = (GLRenderer)iosContext.getRenderer();
+            renderer = iosContext.getRenderer();
             renderer.initialize();
             input = (IosInputHandler)iosContext.getTouchInput();
             input.initialize();


codex128:
You should probably write at least one test for this.

stephengold:
> Pending to add jme3-ios-native library to be uploaded to maven. Any help is appreciated

If `gradlew install` installs the library to your Maven Local repository, then I expect `gradlew release` will upload it to the correct staging repo.

joliver82:
Thanks for the comments

@codex128 , OK, I would do some, but... Which kind of tests do you want me to do? Also, how could I make them run either when compiling locally or in github actions if this code is intended to run only on iOS?

@stephengold , checked and the artifact is saved properly to mavenLocal. I had to make a small change so the jar is properly created when running "gradlew install". Thanks for the tip

jme3-ios-native/src/jme-ios.m
```diff
@@ -0,0 +1,192 @@
...
+
+    int bpp = (*e)->CallIntMethod(e, imageFormat, bitsPerPixel);
+    int comps = 4; // Components (Bytes) per Pixel
+    
+    if ((bpp % 8) == 0)
+    {
+        comps = bpp / 8;
+    } else {
+        jclass assetExClazz = (*e)->FindClass(e, "com/jme3/asset/AssetLoadException");
+        (*e)->ThrowNew(e, assetExClazz, "Unsupported ImageFormat: Bits per Pixel is no multiple of 8");
```
codex128: "Bits per Pixel is no multiple of 8" -> "Bits per Pixel is **not** multiple of 8"

jme3-ios-native/src/JmeAppHarness.java
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * You can extend this class to perform iOS-only operations from java,
+ * native methods can reside either in .c/.m files in this directory
+ * or in the XCode project itself.
+ * @author normenhansen
+ */
+public class JmeAppHarness extends IosHarness{
+
+    private static final Logger logger = Logger.getLogger(JmeAppHarness.class.getName());
+	protected GLRenderer renderer;
```
codex128: I recommend changing this type to Renderer if possible. It's difficult to tell on github, but this file at least should be fine with that change.

jme3-ios/src/main/java/com/jme3/util/IosBufferAllocator.java
```diff
@@ -0,0 +1,118 @@
...
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Jesus Oliver
+ * Quick test to solve instancing issue in 3.6.1 ios. As did in Android, should change to a native way
+ * @deprecated implemented {@link IosNativeBufferAllocator} instead.
+ */
+@Deprecated
```
codex128: This file looks like it should be deleted, given that it is added by this PR as deprecated.

codex128 (changes requested):
Huge thanks for this PR, and I apologize that it didn't get better attention sooner. Just a few things I noticed that need changing.

jme3-ios-native/src/jme-ios.m
joliver82: Updated ;)

jme3-ios/src/main/java/com/jme3/util/IosBufferAllocator.java
joliver82: Yes, I left both when trying the native buffer allocator but forgot to remove this one. Already removed. Thanks

jme3-ios-native/src/JmeAppHarness.java
joliver82: I agree it can be safely changed to Renderer. I just did it ;)

codex128 (approved):
Looks good, thanks @joliver82 

@yaRnMcDonuts this is ready whenever you are

PR #2418, opened 2025-05-03
jme3-examples: TestPostWater code cleanup

capdevon:
Refactored the `TestPostWater `example code for improved clarity. This commit restructures the demo to explicitly highlight its individual components, making it easier to understand the flow and functionality.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/water/TestPostWater.java
+++ b/jme3-examples/src/main/java/jme3test/water/TestPostWater.java
@@ -56,7 +56,6 @@ import com.jme3.post.filters.LightScatteringFilter;
 import com.jme3.renderer.queue.RenderQueue.ShadowMode;
 import com.jme3.scene.Node;
 import com.jme3.scene.Spatial;
-import com.jme3.system.AppSettings;
 import com.jme3.terrain.geomipmap.TerrainLodControl;
 import com.jme3.terrain.geomipmap.TerrainQuad;
 import com.jme3.terrain.geomipmap.lodcalc.DistanceLodCalculator;
@@ -287,7 +286,7 @@ public class TestPostWater extends SimpleApplication {
     private float waterHeight = 0.0f;
     private final float initialWaterHeight = 90f;
     private boolean underWater = false;
-    
+
     private AudioNode waves;
     private final LowPassFilter aboveWaterAudioFilter = new LowPassFilter(1, 1);
     private final LowPassFilter underWaterAudioFilter = new LowPassFilter(0.5f, 0.1f);
--- a/jme3-examples/src/main/java/jme3test/water/TestPostWater.java
+++ b/jme3-examples/src/main/java/jme3test/water/TestPostWater.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2022 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -40,6 +40,7 @@ import com.jme3.font.BitmapText;
 import com.jme3.input.KeyInput;
 import com.jme3.input.controls.ActionListener;
 import com.jme3.input.controls.KeyTrigger;
+import com.jme3.input.controls.Trigger;
 import com.jme3.light.AmbientLight;
 import com.jme3.light.DirectionalLight;
 import com.jme3.material.Material;
@@ -55,7 +56,10 @@ import com.jme3.post.filters.LightScatteringFilter;
 import com.jme3.renderer.queue.RenderQueue.ShadowMode;
 import com.jme3.scene.Node;
 import com.jme3.scene.Spatial;
+import com.jme3.system.AppSettings;
+import com.jme3.terrain.geomipmap.TerrainLodControl;
 import com.jme3.terrain.geomipmap.TerrainQuad;
+import com.jme3.terrain.geomipmap.lodcalc.DistanceLodCalculator;
 import com.jme3.terrain.heightmap.AbstractHeightMap;
 import com.jme3.terrain.heightmap.ImageBasedHeightMap;
 import com.jme3.texture.Texture;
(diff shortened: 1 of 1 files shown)


jme3-examples/src/main/java/jme3test/water/TestPostWater.java
```diff
@@ -87,225 +85,241 @@ public static void main(String[] args) {
...
 
+    private final ActionListener actionListener = new ActionListener() {
+        @Override
+        public void onAction(String name, boolean isPressed, float tpf) {
+            if (!isPressed) return;
 
-        Spatial sky = SkyFactory.createSky(assetManager, 
-                "Scenes/Beach/FullskiesSunset0068.dds", EnvMapType.CubeMap);
-        sky.setLocalScale(350);
+            if (name.equals("foam1")) {
```
codex128, line 162: This should probably be converted to a switch.

jme3-examples/src/main/java/jme3test/water/TestPostWater.java
```diff
@@ -87,225 +85,241 @@ public static void main(String[] args) {
...
     private float waterHeight = 0.0f;
-    final private float initialWaterHeight = 90f;//0.8f;
-    private boolean uw = false;
+    private final float initialWaterHeight = 90f;
+    private boolean underWater = false;
+    
+    private AudioNode waves;
+    private final LowPassFilter aboveWaterAudioFilter = new LowPassFilter(1, 1);
+    private final LowPassFilter underWaterAudioFilter = new LowPassFilter(0.5f, 0.1f);
+    private boolean useDryFilter = true;
```
codex128, line 294: I would suggest moving these fields to the top, or at least have all the fields together.

jme3-examples/src/main/java/jme3test/water/TestPostWater.java
```diff
@@ -66,18 +70,12 @@
...
 
 /**
- * test
- *
  * @author normenhansen
  */
 public class TestPostWater extends SimpleApplication {
 
-    final private Vector3f lightDir = new Vector3f(-4.9236743f, -1.27054665f, 5.896916f);
+    private static final Vector3f lightDir = new Vector3f(-4.9236743f, -1.27054665f, 5.896916f);
```
codex128, line 77: Maybe I missed something, but I don't think this field should be static.

codex128 (commented):
These changes look fine to me. There are just a couple things I would change, but that is up to you.

jme3-examples/src/main/java/jme3test/water/TestPostWater.java
capdevon: Here's why keeping those class variables nearby enhances the code for `addAudioClip()` and `simpleUpdate()`:

**Improved Flow and Understandability**: When reading addAudioClip(), you can easily see the waves variable being created and the initial state of the audio filters being set based on the nearby underWater and useDryFilter variables. Similarly, in simpleUpdate(), the logic for adjusting the audio based on the changing underWater state and the useDryFilter flag becomes clearer when these related variables are in close proximity. This avoids having to jump to the top of the class to understand the context of these variables.

**Easier Reasoning and Modification**: If you need to modify the audio system's behavior in addAudioClip() or simpleUpdate(), having the relevant variables declared nearby simplifies the reasoning process. You can see at a glance which variables are involved in controlling the audio and how they are being used. For instance, if you wanted to add a new audio effect based on the underWater state, having waves, underWater, and the audio filters declared together makes it easier to implement this change consistently.

**Reduced Risk of Contextual Errors**: When the declaration of a variable is far from its usage, there's a higher risk of losing track of its purpose and potential side effects. Keeping these audio-related variables close to the methods that directly manipulate them reduces this risk and makes the code more self-contained and easier to understand in isolation.

In essence, for the audio management aspects of the _TestPostWater_ class, keeping the `waves`, `underWater`, `aboveWaterAudioFilter`, `underWaterAudioFilter`, and `useDryFilter `class variables close to the `addAudioClip()` and `simpleUpdate()` methods significantly improves the code's clarity, makes it easier to follow the audio logic, and simplifies future maintenance and modifications related to the underwater sound effects.

jme3-examples/src/main/java/jme3test/water/TestPostWater.java
capdevon: A switch statement based on strings would be a less ideal choice for this code snippet for several reasons:

1. **Readability and Verbosity**: A series of `if-else if` conditions, as you have, is already quite readable for a small number of distinct string comparisons. A `switch `statement with string cases can become more verbose due to the `case `keyword and the need for `break `statements (though in Java 17+, break is implicit for non-fallthrough behavior). For this specific number of cases, the current `if-else if` structure is arguably clearer.

2. **Performance (Minor in this case but a general consideration)**: While modern Java implementations have optimized string `switch `statements, they often involve calculating hash codes and performing equality checks, which can be slightly less efficient than direct string equality comparisons in a short `if-else if` chain, especially if the order of checks is optimized for likely inputs.

3. **No Significant Advantage**: A string-based `switch `statement doesn't offer a significant advantage in terms of conciseness or functionality over the existing `if-else if` structure for this particular code. The logic within each case is distinct and doesn't inherently benefit from the `switch `construct.

In summary, while a string-based `switch `is a valid feature in modern Java, for this specific snippet with a small number of distinct string comparisons and actions, the existing `if-else if` structure is already clear and doesn't suffer from significant drawbacks that would necessitate a change to a `switch `statement. Other design patterns would likely offer more substantial benefits if the number of actions were to grow significantly.

jme3-examples/src/main/java/jme3test/water/TestPostWater.java
capdevon: variable modified to be `non-static`

codex128: Excellent. Thank you.

codex128 (approved):

jme3-examples/src/main/java/jme3test/water/TestPostWater.java
codex128: In that case, this can be resolved.

jme3-examples/src/main/java/jme3test/water/TestPostWater.java
codex128: That makes sense. This can be resolved then.

PR #2439, opened 2025-05-10
 PBR Fixes: Apply band factor to spherical harmonic coefficients in accelerated baker light and prevent fireflies artifacts

riccardobl:
Band factors were not applies to spherical harmonic coefficients in the accelerated baker, causing bad lighting on high roughness materials.
This PR fixes that for both IBLHybridEnvBakerLight (gpu+cpu) and IBLGLEnvBaker**Light** (gpu) bakers.

This PR also clamps the luminance in the prefiltered envmap kernel, to avoid fireflies artifacts when sampling over bright areas (that could be due to artistic choice or bad tonemap).

### DIFF
--- a/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
+++ b/jme3-core/src/main/java/com/jme3/environment/EnvironmentProbeControl.java
@@ -287,8 +287,9 @@ public class EnvironmentProbeControl extends LightProbe implements Control {
     }
 
     void rebakeNow(RenderManager renderManager) {
-        IBLHybridEnvBakerLight baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F, Format.Depth,
-                envMapSize, envMapSize);
+        IBLHybridEnvBakerLight baker = new IBLGLEnvBakerLight(renderManager, assetManager, Format.RGB16F,
+                Format.Depth,
+                        envMapSize, envMapSize);
                     
         baker.setTexturePulling(isRequiredSavableResults());
         baker.bakeEnvironment(spatial, getPosition(), frustumNear, frustumFar, filter);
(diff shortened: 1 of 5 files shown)


riccardobl:
The screenshot test fails because it needs to be re-tuned to the correct lighting after the fix.
However now both the accelerated and non-accelerated baker outputs look mostly the same:

image
image

yaRnMcDonuts:
Looks good to me. Let me know if this all done and ready to merge, and then I'll work on a 3.8.1 release that includes this PR as well as the other PR related to instancing breaking.

Also if you're able to, it would be best to upload the new screenshot for the screenshot tests to fix that in this PR before merging. But if you're busy and unable to do that, then I can fix the screenshot test in another PR once this one is merged.

riccardobl:
Thank you, it should be good to go, but it is untested on android, also i would want to test it on a real pbr scene.
Currently i am working on a fork that is a bit of a mess, so i am unable to do the tests atm, but i should be able to upload the screenshot in a while. I just need to include the one generated by the test, right?

yaRnMcDonuts:
> I just need to include the one generated by the test, right?

Yes, you should be able to just use the newly generated image to replace the old one in the jme-screenshottests directory, and then that should make the test pass again.

I also won't be of much help testing this currently. Even though I do have some larger PBR scenes using light probes in my project, I am still not at the point of generating my own probes yet, and I just use the DefaultProbe.j3o (from jme-examples) in most of my scenes. But I do plan to generate custom probes for my scenes eventually, so when that time comes, I'll be able to offer much more help testing generated probes.

yaRnMcDonuts:
> I'll work on a 3.8.1 release that includes this PR as well as the other PR related to instancing breaking.

I ended up just putting the other PR in 3.8.1 and held off on this since you said it still needs tested.

And since its not standard to do a beta/alpha for a patch release I figured its best probably to hold off on this PR so that this be more thoroughly tested during the 3.9 alpha release that should be coming soon.

yaRnMcDonuts:
Merging this so it can be tested in the soon-to-come 3.9.0-alpha1 release. I'll also submit another PR to update the screenshot test very soon.

PR #2447, opened 2025-05-21
Reduce object allocations in MotionEvent's direction computation

capdevon:
This PR introduces a minor but impactful performance optimization within the `MotionEvent` class.

Previously, the `computeTargetDirection()` method repeatedly created new `Quaternion` objects (`q` and `q2`) on each call when handling `Path` and `PathAndRotation` direction types. This constant object allocation can lead to unnecessary garbage collection overhead, particularly in performance-critical game loops or animations.

To address this, I've refactored the method to utilize a single, reusable `Quaternion` class member variable (`tempRotation`). By reusing this existing object, we significantly reduce memory allocations and subsequent garbage collection cycles, resulting in a modest but noticeable performance improvement, especially when many MotionEvents are active or when the method is called frequently.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/cinematic/events/MotionEvent.java
+++ b/jme3-core/src/main/java/com/jme3/cinematic/events/MotionEvent.java
@@ -70,7 +70,7 @@ public class MotionEvent extends AbstractCinematicEvent implements Control, JmeC
     protected Direction directionType = Direction.None;
     protected MotionPath path;
     private boolean isControl = true;
-    private final Quaternion tempQuaternion = new Quaternion();
+    private final Quaternion tempRotation = new Quaternion();
     /**
      * the distance traveled by the spatial on the path
      */
@@ -239,10 +239,8 @@ public class MotionEvent extends AbstractCinematicEvent implements Control, JmeC
     }
 
     /**
-     * Determines if the motion path needs direction information. This method
-     * is intended to be called by the motion path only.
-     *
-     * @return True if direction information is needed, otherwise false.
+     * This method is meant to be called by the motion path only.
+     * @return true if needed, otherwise false
      */
     public boolean needsDirection() {
         return directionType == Direction.Path || directionType == Direction.PathAndRotation;
@@ -251,8 +249,8 @@ public class MotionEvent extends AbstractCinematicEvent implements Control, JmeC
     private void computeTargetDirection() {
         switch (directionType) {
             case Path:
-                tempQuaternion.lookAt(direction, upVector);
-                spatial.setLocalRotation(tempQuaternion);
+                tempRotation.lookAt(direction, upVector);
+                spatial.setLocalRotation(tempRotation);
                 break;
             case LookAt:
                 if (lookAt != null) {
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/cinematic/events/MotionEvent.java
```diff
@@ -249,9 +249,8 @@ public boolean needsDirection() {
     private void computeTargetDirection() {
         switch (directionType) {
             case Path:
-                Quaternion q = new Quaternion();
-                q.lookAt(direction, upVector);
-                spatial.setLocalRotation(q);
+                tempRotation.lookAt(direction, upVector);
```
terranprog, line 252: This approach will cause issues if computeTargetDirection() is called from different threads. 

Also this approach forces a Quaternion instance to be created and retained upfront every time a MotionEvent object is created. Which will change the performance profile of this class. It will mean less creation and destruction of temporary objects (so less Garbage Collection) but more memory required per MotionEvent instance.

capdevon: * **Irrelevant Multi-threading Concerns**: Given the single-threaded design assumption for `MotionEvent` and the private nature of `computeTargetDirection()`, raising concerns about this specific method being called from "different threads" is irrelevant and misplaced. It fundamentally misunderstands the class's intended usage and the principles of encapsulation. The internal implementation of a private method does not introduce new thread safety risks if the containing class itself is not designed for concurrent use.

* **Reduced GC Overhead**: The core benefit, as outlined in the PR description, is significantly reducing memory allocations and subsequent garbage collection cycles. In performance-critical loops, repeated small object allocations can lead to noticeable GC pauses, impacting frame rates and responsiveness. Reusing an object is a well-established pattern to mitigate this.
* **Negligible Memory Impac**t: While it's true there's a minor increase in memory required per `MotionEvent` instance (the size of one Quaternion), this is a negligible cost compared to the benefits of eliminating repeated allocations and the associated GC load. This trade-off is highly favorable for overall performance, especially when `MotionEvent` instances are frequently processed.
The aim of this PR is to improve the net performance by minimizing CPU cycles spent on memory management, leading to a smoother experience when MotionEvent instances are active or frequently processed.

I hope this clarifies the rationale behind the changes.

yaRnMcDonuts: I would lean towards @capdevon 's solution that prioritizes reduced GC load even if it uses a tiny amount of extra memory. Freezes from GC cycles are certainly irritating and probably one of the few valid criticisms that java haters can make about a java based game engine, so I'd say that's one of the more important things to focus on optimizing for the engine.

But ram usage is nearly 99% from your assets for games anyways, so I personally find memory reductions from minor code optimizations to be like removing a cup of water in the ocean, essentially useless most of the time, especially when working with something like a 16 byte Quaternion. I'd typically only worry about memory in cases where close to 1mb or atleast a few 100 kb can be saved at minimum (which is extremely rare to be able to optimize by that much unless working with textures or meshes since that's where all the ram usage comes from in games).

So I will go ahead and mark this resolved, and I will plan to merge this PR in the next 24-48 hours unless anymore review is brought up before then.

yaRnMcDonuts (approved):

PR #2451, opened 2025-05-23
Improve LightControl Javadoc & Add Light Type Validation

capdevon:
This PR refactors the Javadoc for the `LightControl` class to enhance clarity and completeness. It also introduces the `validateSupportedLightType()` method for robust light type validation.

1. **Class-level Javadoc**: Expanded description outlining the control's purpose and typical use cases, particularly regarding dynamic light behavior.
2. **Constructor Javadoc**: Clarified parameters and added @throws for invalid light types.

3. **Robust Light Type Validation**: Implemented the new `validateSupportedLightType()` method. This ensures that the `LightControl` only accepts `PointLight`, `DirectionalLight`, and `SpotLight` instances, providing earlier error detection and preventing unexpected behavior with unsupported light types. This method is called in relevant constructors and setters.

4. **New Feature**: Added `axisRotation` property to `LightControl`, enabling precise control over which of the spatial's world axes (X, Y, or Z) defines the direction of the light when using `SpatialToLight` synchronization. This allows for more flexible alignment of lights with complex models.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
@@ -158,6 +158,10 @@ public class LightControl extends AbstractControl {
     }
 
     private void validateSupportedLightType(Light light) {
+        if (light == null) {
+            return;
+        }
+
         switch (light.getType()) {
             case Point:
             case Directional:
@@ -172,6 +176,10 @@ public class LightControl extends AbstractControl {
 
     @Override
     protected void controlUpdate(float tpf) {
+        if (light == null) {
+            return;
+        }
+
         switch (controlDir) {
             case SpatialToLight:
                 spatialToLight(light);
--- a/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
@@ -87,20 +87,10 @@ public class LightControl extends AbstractControl {
         X, Y, Z
     }
 
-    /**
-     * Represents the direction (positive or negative) along the chosen axis.
-     * This influences how the light's direction is set when `SpatialToLight`
-     * and how the spatial's rotation is derived from the light's direction
-     * when `LightToSpatial`.
-     */
-    public enum Direction {
-        Positive, Negative
-    }
-
     private Light light;
     private ControlDirection controlDir = ControlDirection.SpatialToLight;
     private Axis axisRotation = Axis.Z;
-    private Direction axisDirection = Direction.Positive;
+    private boolean invertAxisDirection = false;
 
     /**
      * For serialization only. Do not use.
@@ -159,12 +149,12 @@ public class LightControl extends AbstractControl {
         this.axisRotation = axisRotation;
     }
 
-    public Direction getAxisDirection() {
-        return axisDirection;
+    public boolean isInvertAxisDirection() {
+        return invertAxisDirection;
     }
 
-    public void setAxisDirection(Direction axisDirection) {
-        this.axisDirection = axisDirection;
+    public void setInvertAxisDirection(boolean invertAxisDirection) {
+        this.invertAxisDirection = invertAxisDirection;
     }
 
     private void validateSupportedLightType(Light light) {
@@ -206,7 +196,7 @@ public class LightControl extends AbstractControl {
 
         final Vector3f lightDirection = vars.vect2;
         spatial.getWorldRotation().getRotationColumn(axisRotation.ordinal(), lightDirection);
-        if (axisDirection == Direction.Negative) {
+        if (invertAxisDirection) {
             lightDirection.negateLocal();
         }
 
@@ -246,7 +236,7 @@ public class LightControl extends AbstractControl {
         } else if (light instanceof DirectionalLight) {
             DirectionalLight dl = (DirectionalLight) light;
             lightDirection.set(dl.getDirection());
-            if (axisDirection == Direction.Negative) {
+            if (invertAxisDirection) {
                 lightDirection.negateLocal();
             }
             rotateSpatial = true;
@@ -255,7 +245,7 @@ public class LightControl extends AbstractControl {
             SpotLight sl = (SpotLight) light;
             lightPosition.set(sl.getPosition());
             lightDirection.set(sl.getDirection());
-            if (axisDirection == Direction.Negative) {
+            if (invertAxisDirection) {
                 lightDirection.negateLocal();
             }
             translateSpatial = true;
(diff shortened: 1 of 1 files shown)


capdevon:
Hi @yaRnMcDonuts ,
Please hold off on merging this class for now. I'd like to do some further verification on its functionality. 
I'll let you know as soon as it's ready ;)

capdevon:
I think the PR is ready now!

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
```diff
@@ -115,86 +141,133 @@ public void setControlDir(ControlDirection controlDir) {
...
+
+    public boolean isInvertAxisDirection() {
+        return invertAxisDirection;
+    }
+
+    public void setInvertAxisDirection(boolean invertAxisDirection) {
+        this.invertAxisDirection = invertAxisDirection;
+    }
+
+    private void validateSupportedLightType(Light light) {
```
yaRnMcDonuts, line 160: I see this no longer allows a null liight and would throw an exception in that case. This could break things for users that are currently setting the light null for some reason. 

There could be situation where someones code intentionally sets the light to null at cleanup, or set null temporarily before being used with a new light. I don't know how probable that is, but its certainly possible and would be an acceptable way to do things that we should support. 

And I suspect that, considering a setLight() method exists, that it is intended that this class should allow the light to be changed at any time. In which case it should also be allowed to be null in cases where there's a delay between 2 lights being used.

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
```diff
@@ -115,86 +141,133 @@ public void setControlDir(ControlDirection controlDir) {
...
+                break;
+            default:
+                throw new IllegalArgumentException(
+                        "Unsupported Light type: " + light.getType());
+        }
+    }
+
     @Override
     protected void controlUpdate(float tpf) {
-        if (spatial != null && light != null) {
```
yaRnMcDonuts, line 121: I disagree with the removal of this null check here.

Even if the class is written to be as null-proof as possible with the new validate method you added (which I also disagree with, as mentioned in my other comment), I still think its best to leave a null check in a case like this. Even if you are 99.99% sure that you null-proofed everything and that spatial or light will never be null, I would still leave that null check here just to be safe. It's already there, possibly being relied upon by some existing code, and the benefit from removing it is extremely negligible. 

In my experience, tying to write classes to always make a publicly modifiable field non-null and to never allow null values like this almost always backfires and makes it harder to maintain in the future.  Allowing null values gives the class much more flexibility, and I nearly always find it to be safer to just do null checks and accept null values whenever possible.

jme3-core/src/main/java/com/jme3/scene/control/LightControl.java
```diff
@@ -172,6 +176,10 @@ private void validateSupportedLightType(Light light) {
 
     @Override
     protected void controlUpdate(float tpf) {
+        if (light == null) {
```
yaRnMcDonuts, line 179: I would also add back the spatial == null check.

Even though it wouldn't be proper usage, Spatial could still be null here is if someone forcibly sets the spatial null in the control class while the control is still attached to the spatial

I looked at other controls' update loops, and some check spatial != null while others don't... so I don't tihnk its a big deal but I still think it's best to just leave the spatial null check here since that's how it originally was. Best to change as little as possible unless absolutely necessary.

capdevon: When working with jMonkeyEngine's `AbstractControl`, checking if `spatial` is null within the `controlUpdate()` method is redundant and unnecessary. Here's why:

1.  **Correct Control Lifecycle:** The `AbstractControl`'s `setSpatial()` method (which is called by `Spatial.addControl()`) includes a check:
    ```java
    @Override
        if (this.spatial != null && spatial != null && spatial != this.spatial) {
            throw new IllegalStateException("This control has already been added to a Spatial");
        }
        this.spatial = spatial;
    }
    ```
    This ensures that a control is only ever associated with one spatial at a time. More importantly, when a control is added to a spatial, `setSpatial(this)` is called. When it's removed via `spatial.removeControl(control)`, `setSpatial(null)` is called. The `update()` method of the `Control` interface (which `AbstractControl` implements) is only invoked *if* the control is enabled and *if* it's attached to a spatial that is being updated by the scene graph.

2.  **`Spatial.updateLogicalState()`:** The `Spatial` class explicitly calls `runControlUpdate(tpf)` within its `updateLogicalState()` method. Before calling `runControlUpdate`, `Spatial` checks `if (controls.isEmpty()) return;`. This means if a control has been properly removed (and its `spatial` field set to `null` through `removeControl`), it won't be in the `controls` list, and its `update` method (and thus `controlUpdate`) won't be called.

3.  **Encapsulation and API Usage:** The jME API provides the proper way to manage controls: `spatial.addControl(control)` and `spatial.removeControl(control)`. Directly calling `control.setSpatial(null)` bypasses the intended lifecycle management that `Spatial.removeControl()` provides, which includes updating the parent node's update list if needed. Adhering to the public API is crucial for correct behavior and maintainability.

it's impossible and a waste of time to write "code barriers" for every conceivable misuse of an API. Good API design aims for clarity and provides the correct pathways for interaction. If a developer deviates from the documented and intended usage, it's generally accepted that they assume responsibility for the consequences. Focusing on robust core logic and clear API contracts is far more productive than trying to bullet-proof against every possible improper manual manipulation.

yaRnMcDonuts: > 1. **Correct Control Lifecycle:** The `AbstractControl`'s `setSpatial()` method (which is called by `Spatial.addControl()`) includes a check:
>    ```java
>    @Override
>    public void setSpatial(Spatial spatial) {
>        if (this.spatial != null && spatial != null && spatial != this.spatial) {
>            throw new IllegalStateException("This control has already been added to a Spatial");
>        }
>        this.spatial = spatial;
>    }
>    ```

It looks like this check doesn't stop a null value from being set, so that's part of why I was being cautious

But I do agree that the way the controls life cycle is written and update loop is controlled by the spatial should protect the spatial variable from ever being null while the control update loop is being run, as long as the controls are being setup and and cleaned up properly. So it should be okay to leave out this spatial null check here then.

yaRnMcDonuts: > it's impossible and a waste of time to write "code barriers" for every conceivable misuse of an API.

And I do agree with this point you raised. However in this case we are talking about removing existing null checks, rather than writing new ones. And it is a very small null check that is not complex or convoluted. 

I completely agree that it is not a good idea to add new unnecissary null checks or barriers in existing classes. If the class has been running fine for users this long then there's no need to add overly cautious new null checks.

But it's a much different story when removing an existing null check like this, even it seems almost entirely useless, especially considering neither of us were the original author to know if there was a strange reason they did something that seems otherwise incorrect. 

Caution should always be taken when removing existing code even if that code may seem unnecessary or incorrect. Once something like a null check has been written and used for so long, the ramifications of removing it do need to be fully considered before it is decided to be removed.

yaRnMcDonuts (approved):

Issue #2460, opened 2025-05-24
getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT) causes error 1280 (invalid enum) on Windows


capdevon:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java#L3492

```java
    @Override
    public void setMainFrameBufferSrgb(boolean enableSrgb) {
        // Gamma correction
        if (!caps.contains(Caps.Srgb) && enableSrgb) {
            // Not supported, sorry.
            logger.warning("sRGB framebuffer is not supported " +
                    "by video hardware, but was requested.");

            return;
        }

        setFrameBuffer(null);

        if (enableSrgb) {
            if (
                // Workaround: getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT) causes error 1280 (invalid enum) on macos
                JmeSystem.getPlatform().getOs() != Platform.Os.MacOS
                && !getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT)
            ) {
                logger.warning("Driver claims that default framebuffer " + "is not sRGB capable. Enabling anyway.");
            }

            gl.glEnable(GLExt.GL_FRAMEBUFFER_SRGB_EXT);

            logger.log(Level.FINER, "SRGB FrameBuffer enabled (Gamma Correction)");
        } else {
            gl.glDisable(GLExt.GL_FRAMEBUFFER_SRGB_EXT);
        }
    }
```

```
INFO: OpenGL Renderer Information
 * Vendor: Intel
 * Renderer: Intel(R) Iris(R) Xe Graphics
 * OpenGL Version: 4.0.0 - Build 32.0.101.5763
 * GLSL Version: 4.00 - Build 32.0.101.5763
 * Profile: Core
[JME3] OpenGL debug message
        ID: 1280
        Source: API
        Type: ERROR
        Severity: HIGH
        Message: Error has been generated. GL error GL_INVALID_ENUM in GetBooleanv: (ID: 659224037) Generic error
java.lang.Exception: Stack trace
	at java.base/java.lang.Thread.dumpStack(Thread.java:2209)
	at com.jme3.system.lwjgl.LwjglGLDebugOutputHandler.handleMessage(LwjglGLDebugOutputHandler.java:76)
[...]

PR #2461, opened 2025-05-27
#2460: Fix getBoolean(GLExt.GL_FRAMEBUFFER_SRGB_CAPABLE_EXT) causes error 1280 (invalid enum)

capdevon:
`GL_FRAMEBUFFER_SRGB_CAPABLE_EXT` is likely not a queryable state via `glGetBooleanv`. It's a capability flag for *configuring* framebuffers, not for *querying* the global sRGB state. It's good that it's commented out.

This PR should fix https://github.com/jMonkeyEngine/jmonkeyengine/issues/1996

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2025 jMonkeyEngine
+ * Copyright (c) 2009-2026 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -129,7 +129,7 @@ public final class GLRenderer implements Renderer {
         this.glext = glext;
         this.texUtil = new TextureUtil(gl, gl2, glext);
     }
-
+    
     /**
      * Enable/Disable default automatic generation of mipmaps for framebuffers
      * @param v  Default is true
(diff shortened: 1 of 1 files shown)


riccardobl:
GL_FRAMEBUFFER_SRGB_CAPABLE_EXT is actually queryable via glGetBooleanv, but it is provided by the EXT_framebuffer_sRGB extension.

I suspect that some drivers that provided opengl3+ capabilities, do not implement a fallback for this extension.

jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -3490,19 +3490,11 @@ public void setMainFrameBufferSrgb(boolean enableSrgb) {
...
-                logger.warning("Driver claims that default framebuffer " + "is not sRGB capable. Enabling anyway.");
-            }
-
             gl.glEnable(GLExt.GL_FRAMEBUFFER_SRGB_EXT);
-
-            logger.log(Level.FINER, "SRGB FrameBuffer enabled (Gamma Correction)");
+            logger.log(Level.INFO, "SRGB FrameBuffer enabled (Gamma Correction)");
         } else {
             gl.glDisable(GLExt.GL_FRAMEBUFFER_SRGB_EXT);
+            logger.log(Level.INFO, "SRGB FrameBuffer disabled (Gamma Correction)");
```
riccardobl: These two logs should be kept at Level FINER

capdevon: Done

capdevon:
> Since this check is only used for debugging purposes and, as noted, it is not reliable across different platforms, i am ok with removing it. But this PR needs to be cleaned from the unrelated formatting and logging changes, can you do that?
> 
> Thanks

Done

PR #2466, opened 2025-05-30
Refactor: Torus Mesh Optimization and Javadoc

capdevon:
`This PR significantly improves the `Torus` mesh by:

* Refining vertex and index generation logic.
* Removing the unused `vertexIndex` variable in `generateGeometryData`.
* Adding thorough Javadoc comments for all methods and fields.
* Introducing validation for constructor parameters to prevent invalid mesh creation. These changes enhance code readability, maintainability, and robustness.`

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/shape/Torus.java
+++ b/jme3-core/src/main/java/com/jme3/scene/shape/Torus.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


capdevon:
@yaRnMcDonuts @richardTingle 
Hi guys, can you help me understand why this PR doesn't pass the test?

richardTingle:
The tests that fail are one that uses the Torus so something has changed (although the difference is pretty subtle). Are the meshes you generate after your change identical to the ones before you change?

richardTingle:
You have subtle differences in your position and normal data. I used this within the torus class to print them out

        System.out.println("Position: " + Arrays.toString(BufferUtils.getFloatArray(fpb)));
        System.out.println();
        System.out.println("normals: " + Arrays.toString(BufferUtils.getFloatArray(fnb)));
        
For example position index 14 is 0.0 in master but 2.1855694E-8 in your branch

image

capdevon:
The number $2.1855694 \times 10^{-8}$ is a very small number, meaning it tends to zero. Therefore, the difference between $0.0f$ and $2.1855694 \times 10^{-8}$ is also very small, and not considered a large difference.

richardTingle:
Well yes, it is a very small difference. Why has it changed at all? Isn't this supposed to just be a refactor?

(If it used to be 2.1855694E-8 and was now 0 I might say that was an improvement, but it's the other way round)

If you really really wanted to do this you'd need to update the reference images (as the bot explained) but unexpected numerical changes worries me a bit. What's the reason for introducing this change?

capdevon:
The last 85 PRs aim to enhance the `jme3-core` module through optimization, feature additions, duplication removal, stabilization, and comprehensive Javadoc implementation. My work isn't a critique of the original authors; I respect their contributions and am confident they did their best. However, some areas have become complex and challenging to read, often lacking comments or containing 'magic numbers.' I've used various analysis tools to identify and address critical issues.

Regarding this PR, it significantly simplifies the original code. While there might be minor differences in decimal values compared to the old algorithm—which duplicated initial vertices for loop closure and texture smoothing—I believe the overall result is a substantial improvement. So I revert the `setGeometryData()` method. I'm satisfied with the current implementation and suggest we proceed with merging this PR to allow me to focus on testing the remaining PRs.

PR #2467, opened 2025-05-30
Refactor: LightsDebugState optimization, javadoc, performance

capdevon:
This PR significantly enhances the `LightsDebugState` class by:

* **Optimizing performance**: Reduced overhead for more efficient debugging.
* **Fixing bugs**: Addressed several issues for improved stability.
* **Improving resource management**: Ensured proper cleanup and allocation.
* **Adding Javadoc**: Provided comprehensive documentation for better understanding and maintainability.
* **Unified Light Visualization**: `LightProbe`, `DirectionalLight`, `PointLight` & `SpotLight`.
* **Dynamic Scene Integration**: The state automatically detects and creates/updates gizmos for lights present in the configured scene graph. Includes logic to clean up gizmos for lights that have been removed from the scene, ensuring the debug overlay remains accurate and clutter-free.

Edit:
This PR is still work in progress, and is not yet ready to be merged... 

Edit:
Ready!

This comprehensive `LightsDebugState` streamlines the process of debugging lighting setups, making it easier to visualize light behavior and troubleshoot common lighting issues in JME applications.

image
image
image

### DIFF
--- a/jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
+++ b/jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
@@ -220,7 +220,6 @@ public class LightsDebugState extends BaseAppState {
     @Override
     public void render(RenderManager rm) {
         debugNode.updateGeometricState();
-        rm.renderScene(debugNode, viewPort);
     }
 
     /**


jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
```diff
@@ -33,147 +33,374 @@
...
+
+    private AssetManager assetManager;
     private Material debugMaterial;
-    private float probeScale = 1.0f;
-    private Spatial scene = null;
-    private final List<LightProbe> probes = new ArrayList<>();
+
+    private float lightProbeScale = 1.0f;
+    private final Quaternion tempRotation = new Quaternion();
+    private final ColorRGBA GIZMO_COLOR = ColorRGBA.DarkGray;
```
codex128: Make `GIZMO_COLOR` static?

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
```diff
@@ -33,147 +33,374 @@
...
+
+    /**
+     * Recursively traverses the scene graph to find and update light gizmos.
+     * New gizmos are created for new lights, and existing gizmos are updated.
+     *
+     * @param spatial The current spatial to process for lights.
+     */
+    private void updateLightGizmos(Spatial spatial) {
+        // Add or update gizmos for lights attached to the current spatial
+        for (Light light : spatial.getWorldLightList()) {
```
codex128: I would've expected this to use `getLocalLightList`. As you move down the spatial tree, it seems you might add the same lights to `lightDeque` multiple times.

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
```diff
@@ -33,147 +33,374 @@
...
 
     /**
-     * returns the scale of the probe's debug sphere
-     * @return the scale factor
+     * Sets the scene for which to render light gizmos.
+     * If no scene is set, it defaults to the first scene in the viewport.
+     *
+     * @param scene The root of the desired scene.
+     */
+    public void setScene(Spatial scene) {
```
codex128, line 376: Do you think it'd be worthwhile to be able to use a list of scene spatials instead of just one?

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
```diff
@@ -33,147 +33,374 @@
...
+     */
+    private void cleanUpRemovedLights() {
+
+        Iterator<Map.Entry<Light, Spatial>> iterator = lightGizmoMap.entrySet().iterator();
+
+        while (iterator.hasNext()) {
+            Map.Entry<Light, Spatial> entry = iterator.next();
+            Light light = entry.getKey();
+
+            if (!lightDeque.contains(light)) {
```
codex128, line 360: This is the only place `lightDeque` is read from. If this can be replaced by something else, then `lightDeque` can be deleted.

jme3-core/src/main/resources/Common/MatDefs/Misc/Dashed.j3md
codex128, line 1: Is there any other existing matdef that could be used instead? Not knocking your shader, it's just a good idea not to create unnecessary files.

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
```diff
@@ -33,147 +33,374 @@
...
-                    probeGeom.setLocalScale(probeScale);
-                    if (probe.isReady()) {
-                        m.setTexture("CubeMap", probe.getPrefilteredEnvMap());
-                    }
-                    n.setLocalTranslation(probe.getPosition());
-                    n.getChild(1).setLocalScale(probe.getArea().getRadius());
-                    break;
-                default:
-                    break;
+        rm.renderScene(debugNode, getApplication().getViewPort());
```
codex128: This seems a little suspicious to me. It might be better practice to create a dedicated viewport for the debug scene, or else attach the debug scene directly the the main viewport.

jme3-core/src/main/java/com/jme3/environment/util/Circle.java
```diff
@@ -0,0 +1,182 @@
...
+            cos[i] = FastMath.cos(angle);
+            sin[i] = FastMath.sin(angle);
+        }
+
+        // Define the color for the entire circle.
+        ColorRGBA color = ColorRGBA.Orange;
+
+        // Populate the position, color, and texture coordinate buffers.
+        for (int i = 0; i < numVertices; i++) {
+            posBuf.put(cos[i]).put(sin[i]).put(0);
```
codex128: Since this is the only place where the `sin` and `cos` arrays are read from, I think the calculations should be moved here, and then the `sin` and `cos` arrays can be eliminated.

jme3-core/src/main/java/com/jme3/environment/util/Circle.java
```diff
@@ -0,0 +1,182 @@
...
+
+        int numVertices = radialSamples + 1;
+
+        FloatBuffer posBuf = BufferUtils.createVector3Buffer(numVertices);
+        FloatBuffer colBuf = BufferUtils.createFloatBuffer(numVertices * 4); // 4 floats per color (RGBA)
+        FloatBuffer texBuf = BufferUtils.createVector2Buffer(numVertices);
+
+        setBuffer(Type.Position, 3, posBuf);
+        setBuffer(Type.Color, 4, colBuf);
+        setBuffer(Type.TexCoord, 2, texBuf);
```
codex128: Personally, I think it'd be a good idea to move these `setBuffer` calls until after you finish writing to the buffers.

codex128 (changes requested):

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
capdevon: correct, I am considering whether to make it an editable parameter or not....

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
capdevon: correct, I will make the suggested change from `getWorldLightList()` to `getLocalLightList()`

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
capdevon: I don't think it's necessary. Allowing a list of scene spatials would needlessly complicate the code and internal data structures. It could also lead to duplicate lights if we accept scattered fragments of the scene tree as input.

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
capdevon: The `lightDeque` is a crucial data structure. It's responsible for tracks the lights present in the scene and is also used to efficiently clear out the `lightGizmoMap` of any lights that have been removed from the scene graph. **Due to these specific and essential functions, the lightDeque cannot be removed.**

see `updateLightGizmos(Spatial spatial)`

jme3-core/src/main/resources/Common/MatDefs/Misc/Dashed.j3md
capdevon: No, I did not find an equivalent shader in the `jme3-core` module. The shaders 
(`Dashed.j3md`, `DashedPattern.j3sn`, `DashedPattern100.frag`) that I added I got from the **SDK**.

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
capdevon: I will create a dedicated `viewPort` to which I will attach the `debugNode`

jme3-core/src/main/java/com/jme3/environment/util/LightsDebugState.java
codex128: Ah, ok. I see what it's doing now.

codex128 (approved):
Looks good to me. Thanks!

capdevon:
Feat: Add light filter to customize debug light display

Introduced a `lightFilter` parameter and its corresponding setter method (`setLightFilter`) to allow for selective display of lights in the debug state. This enables users to define a `Predicate<Light>` to control which lights are visualized, improving debugging flexibility.

yaRnMcDonuts:
Looking forward to testing this out in my own custom editor soon! My editor's light tool is pretty bare bones and lacks any visualization (my lights are just represented by empty invisible nodes that can be moved around) so this will be very useful in my editor for improving my light tool, and should be very useful for other jme users' custom editors. Thanks for working on this class!

I may have more feedback once I've used the class, but so far everything looks good to me,. So just let me know whenever you're all done and ready for it to be merged.

capdevon:
I am undecided about whether or not to add this instruction:
`viewPort.setClearFlags(false, true, true);`

Otherwise, I think the class is ready to be tested.

codex128:
> viewPort.setClearFlags(false, true, true);

I think you should add this, but also add an option to enable/disable depth clearing.

capdevon:
Configuring the `clearDepth` parameter has turned out to be more complex than anticipated, and the visual results aren't consistently clear. I explored two options: 
1. add `LightsDebugState.getViewPort()` method or
2. add `LightsDebugState.setClearDepthViewPort(boolean)` method.

The main hurdle is that the `viewport` is only initialized within `LightsDebugState.initialize()`. Introducing these methods would create scenarios where they're invoked on a `null` viewport, leading to confusion and the need for additional error handling.

Therefore, my conclusion is that the most sensible approach is to forego adding any new methods and maintain the default `viewPort` flags. The reasoning is that enabling `clearDepth=true` by default would cause all debug geometries and light icons to always render in the foreground, which would likely be more disorienting than helpful.

For those who need to customize the viewport flags, the flexibility remains: 
you can always retrieve the `viewport` directly from the `renderManager` like this:

```
ViewPort vp = renderManager.getMainView("EnvDebugView");
vp.setClearDepth(true);
```

Edit:
*****************************************************
That is why I would rename the **viewPort** to either `LightsDebugView` or `LightGizmosView`.
*****************************************************

e.g. clearDepth=true
image

e.g. clearDepth=false
image

yaRnMcDonuts:
> The reasoning is that enabling `clearDepth=true` by default would cause all debug geometries and light icons to always render in the foreground, which would likely be more disorienting than helpful.

On one hand I agree that setting true makes it disorienting and clutters the screen in small scenes with many lights. like the test scene you are using in your screenshots.

But on the other hand, setting it false will also cause the debug circles and light gizmos to be easily lost and hard to find when viewing very large scenes, especially at a distance and especially if the light is under tree canopies or inside a building.

I know that I will have an option to toggle this in my editor's light tool when I use this class, since an editor can get cluttered when enabling extra detail, but you also still want to be able to have the option to toggle that extra detail when the task requires it.

> For those who need to customize the viewport flags, the flexibility remains: you can always retrieve the `viewport` directly from the `renderManager` like this:
> 
> ```
> ViewPort vp = renderManager.getMainView("EnvDebugView");
> vp.setClearDepth(true);
> ```

I'd say that putting this code in a simple convenience method in the LightsDebugState is the best option, because otherwise many users will not find this information to know how to change this, especially newer users who don't fully understand how this class works using view ports and depth writing. Ideally the class should be as convenient and easy to use to its full potential without the typical user having to delve too far into the source code or understand too much of the class' deeper functionality. 

So I'd agree with @codex128 that making a convenience method with that code in it is best. Something straightforward like setShowLightsDebugOverScene(boolean showOverScene) or something similar would be a good name in my opinion.

codex128:
> The main hurdle is that the viewport is only initialized within LightsDebugState.initialize().

You could add a boolean that sets the depth flag during initialization. I realize it's not very convenient, but it'll work.

capdevon:
Ok, I will make the requested changes. Should the default value of the `clearDepth` flag be true or false?

Edit:
It's currently set to true. Let me know if I need to change it to false.

PR #2470, opened 2025-06-01
Make initial velocity influence newtornian particles

riccardobl:
I've changed the newtonian influence to **not** ignore the initial velocity since it is a default property.
I need a second opinion on whether this is a good idea or now, but it might help making it more versatile.


yaRnMcDonuts:
> I need a second opinion on whether this is a good idea or now,

As long as initial velocity is 0,0,0 by default, then this shouldn't break anything and sounds like a good idea.

richardTingle:
This change seems reasonable.

I think the screenshot pipeline failed (rather than the tests themselves). Worth just rerunning them and they may pass (not sure if the github UI allows that. Pushing an empty commit would do it)

PR #2473, opened 2025-06-02
Title: Refine Light toString() Methods

capdevon:
Adds missing `toString()` implementations for `PointLight` and `SpotLight`, and refines existing ones for `DirectionalLight` and `ProbeLight` for better debug output.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/DirectionalLight.java
+++ b/jme3-core/src/main/java/com/jme3/light/DirectionalLight.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2012, 2015-2016 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


yaRnMcDonuts:
I notice this PR includes changed return values for some existing toString() methods, which I had also mentioned being worried about in another PR.

The chance of this breaking something for someone is definitely low, although still possible. 

But I still honestly am not sure if this is something important to worry about or if I'm being over cautious. In my own personal code I'd change it without worrying much, but from what I'm googling it sounds like a toString() method's return values should avoid being changed unless necessary for public APIs like jme. But maybe it would be useful to get a core dev to comment and give their opinion on this since I'm just speculating and being overly cautious.

yaRnMcDonuts:
So I think this PR should be okay now after taking another look and considering it more.

Adding the missing toString() methods for the the light classes that don't have it is definitely useful.

The only thing I was hesitant about was the change in format for the string that was returned by the existing method in LightProbe.

I'd typically say to keep with the existing format for all lights, just to minimize changes. But in this case it already looks like DirectionalLight and LightProbe had inconsistent formats in their existing toString() methods . And regardless of which formatting is better, the important part is probably just to be consistent with all lights, which looks to be the case now. 

So I'd say this PR looks good to me now.

yaRnMcDonuts (approved):

PR #2474, opened 2025-06-02
Add Missing Joystick Button Definitions

capdevon:
Extends the `JoystickButton` interface with definitions for buttons `12` through `15`, ensuring full coverage for common joystick layouts. These buttons are needed with a Logitec F310 and `lwjgl3`.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/JoystickButton.java
+++ b/jme3-core/src/main/java/com/jme3/input/JoystickButton.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


yaRnMcDonuts (approved):

yaRnMcDonuts:
Oops, right after merging this I saw something that I had somehow overlooked both other time I looked at this PR.

It looks like you removed the static keyword from the JoyStick button String declarations and I did not realize that. Was this a mistake, or is there an intentional reason for doing so?

I don't use Joysticks or controllers in my app, so maybe I'm mistaken. But I'm fairly sure that these fields are intended to be static so they can be referenced and used similar to static KeyTrigger values for normal key and mouse input. In which case setting them non static will break every app where a user is referencing them this way.

Edit: this engine example cofirms that the JoyStickButton static vars are indeed intended to be left static, and examples like this would break if not static: https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-examples/src/main/java/jme3test/input/TestJoystick.java#L258

So I would suggest submitting another PR to add the static keywords back, or if you are busy and don't get aroudn to it soon then I could do so for you.

Regardless of this mistake, your contribution is still appreciated and the added support for buttons 12-15 is good to have now. And I apologize for merging it before noticing this issue.

PR #2479, opened 2025-06-06
Camera: Add optimized screenPointToRay(Vector2f) for efficient raycasting

capdevon:
This PR introduces a new public method, `screenPointToRay(Vector2f click2d)`, to the `Camera` class. This method provides a highly efficient way to generate a picking `Ray` in world coordinates from a given 2D screen coordinate.

**Motivation**:

The ability to accurately and efficiently convert a 2D screen point (e.g., from a mouse click) into a 3D ray in world space is fundamental for various interactive features, such as:

* **Object Picking**: Determining which 3D object a user has clicked on.
* **Raycasting**: Performing intersections with geometry for game logic (e.g., shooting, selection).
* **Gizmo Manipulation**: Interacting with 3D manipulation tools.

See https://docs.unity3d.com/6000.1/Documentation/ScriptReference/Camera.ScreenPointToRay.html

usage example:

```java
Ray pickingRay = cam.screenPointToRay(inputManager.getCursorPosition());

// Now 'pickingRay' can be used for intersection tests with 3D objects
// e.g., pickingRay.intersects(someSpatial.getWorldBound());
```

Edit:
add `TestMousePick`
image

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/Camera.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/Camera.java
@@ -1583,9 +1583,6 @@ public class Camera implements Savable, Cloneable {
      * <pre>{@code
      * // Usage Example:
      * Ray pickingRay = cam.screenPointToRay(inputManager.getCursorPosition());
-     *
-     * // Now 'pickingRay' can be used for intersection tests with 3D objects
-     * // e.g., pickingRay.intersects(someSpatial.getWorldBound());
      * }</pre>
      */
     public Ray screenPointToRay(Vector2f pos) {


jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -1590,9 +1625,14 @@ public int getHeight() {
 
     @Override
     public String toString() {
-        return "Camera[location=" + location + "\n, direction=" + getDirection() + "\n"
```
yaRnMcDonuts, line 1593: It looks like toString() will now return a different value, and I'm not entirely certain that this is something we want to change especially in a core class like Camera. The improved code formatting is okay, but if it now returns a different String then I worry that may be problematic.

So I am curious what is the reasoning behind changing the return value here and if it is a necessary change.

capdevon: Hi @yaRnMcDonuts 

Thanks for your feedback and for raising the question about the `toString()` method modification in the `Camera` class. I understand your concern about a change in such a core class and appreciate the opportunity to clarify the reasoning behind this modification.

The fundamental point is that **no Java programmer should ever write code that relies on an object's string representation (i.e., the output of `toString()`) for its application logic**. The `toString()` method is primarily intended for debugging and logging purposes, providing a human-readable representation of an object's state. It's not a contractual API, and its output can (and often should) change to improve clarity or diagnostics.

In this specific case, I modified the `toString()` method of the `Camera` class to adhere to a cleaner, more consistent programming standard. Adopting a simpler, linear format, we make the representation more robust and less prone to creating implicit dependencies.

So, to answer your question:

### Reasoning Behind the Return Value Change

The reasoning is twofold:

1.  **Adherence to programming best practices:** To prevent `toString()` from being interpreted as a source of structured data.
2.  **Improved readability and robustness:** To remove special characters that could be problematic in other contexts, making the `toString()` a clearer and more standard representation.

### Is This a Necessary Change?

From a strictly functional perspective, the program will continue to run. However, it's a **highly recommended** change to promote better programming practices and prevent future bugs or unexpected behavior stemming from improper use of `toString()`. If any code were to break because of this change, it would indicate a pre-existing, significant problem in that code, relying improperly on `toString()`'s output instead of accessing data via appropriate getter methods.

I hope this explanation clarifies my point of view. I believe this modification improves the overall code quality and promotes a more correct use of Java APIs.

Please let me know if you have any further questions.

capdevon: For now, I've decided not to proceed with the `toString()` method modification. I want to keep the focus of this PR on the addition of the new method. We can always revisit the `toString()` change in a separate PR in the future.

This approach should help us streamline the review process for this specific feature.

Thanks

yaRnMcDonuts: I do agree with your original analysis that, in best practice, the toString() method probably _shouldn't_ be used in a way that would allow a changed return value to break things.

But there will inevitably be users and small teams with their own approaches that don't follow best practices to a T for whatever reasons they may have, and there may be times where a dev needs to implement a hacky solution that goes entirely against best practices but may still be necessary. 

So I agree with your final decision to ear on the side of caution and leave the toString() return value unchanged since there's no crucial reason or significant benefit to change it.

jme3-core/src/main/java/com/jme3/renderer/Camera.java
```diff
@@ -1570,6 +1571,37 @@ public Vector3f getScreenCoordinates(Vector3f worldPosition, Vector3f store) {
...
+     * @param pos A {@link Vector2f} representing the 2D screen coordinates (in pixels)
+     * @return A {@link Ray} object representing the picking ray in world coordinates.
+     *
+     * <pre>{@code
+     * // Usage Example:
+     * Ray pickingRay = cam.screenPointToRay(inputManager.getCursorPosition());
+     * }</pre>
+     */
+    public Ray screenPointToRay(Vector2f pos) {
+        TempVars vars = TempVars.get();
```
codex128, line 1589: You can't use TempVars on a public utility method because only so many TempVar instances can be checked out at once.

capdevon: I don't understand the point of your comment. According to your comment, would the `Quaternion.lookAt()` method also be wrong?

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/math/Quaternion.java#L1514

```java
    public Quaternion lookAt(Vector3f direction, Vector3f up) {
        TempVars vars = TempVars.get();
        vars.vect3.set(direction).normalizeLocal();
        vars.vect1.set(up).crossLocal(direction).normalizeLocal();
        vars.vect2.set(direction).crossLocal(vars.vect1).normalizeLocal();
        fromAxes(vars.vect1, vars.vect2, vars.vect3);
        vars.release();
        return this;
    }
```

codex128: You're right. I was thinking that if someone happens to use it in the wrong place they would be unintentionally introducing a bug, but that situation would produce an uncaught exception which they should see when they test it.

In that case, everything looks fine to me.

PR #2485, opened 2025-06-11
Materials: new material definitions for common shaders

capdevon:
Provides a material definition that visualizes the mesh normals, useful for debugging geometry and lighting issues.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/material/Materials.java
+++ b/jme3-core/src/main/java/com/jme3/material/Materials.java
@@ -43,6 +43,7 @@ public class Materials {
     public static final String LIGHTING     = "Common/MatDefs/Light/Lighting.j3md";
     public static final String PBR          = "Common/MatDefs/Light/PBRLighting.j3md";
     public static final String PARTICLE     = "Common/MatDefs/Misc/Particle.j3md";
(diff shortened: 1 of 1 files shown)


yaRnMcDonuts:
I'm not sure if the original author intended to only include only the primary stock 3 shaders here (pbr, lighting, and unshaded) but I don't see any reason that more shaders shouldn't be added for convenience. It also serves as a good list of available shaders for anyone who looks at this class and may see one they didn't know existed.

So unless anyone else has issue with adding to this list, then I'd say it looks good to me.

I'd also even say it would be beneficial to add some more (like the particle shader, thats one that I constantly find myself retyping or copy/pasting everytime I make a new ParticleEmitter), but that can always be done later or by someone else if you prefer to keep the scope of this PR limited.

yaRnMcDonuts (approved):

capdevon:
Hi @yaRnMcDonuts ,
Adding the `ParticleEmitter` shader to the list also seems like a good idea! 
I'll add it to the PR and edit the title.

yaRnMcDonuts:
I was going to merge this right now

But I'll leave it open for a little while just in case there's any more you want to add to the list. Otherwise let me know when you think its ready and then I'll merge.

capdevon:
Hi @yaRnMcDonuts , I think the PR is now complete and ready to be merged.

Issue #2472, opened 2025-06-01
2D screen coordinates broken after updating to 3.8.0-stable


Ivan1pl:
When using fullscreen mode, coordinates in 3.8.0 are shifted. These are the settings I am using for the application:

```java
        AppSettings settings = new AppSettings(true);
        settings.setHeight(1080);
        settings.setWidth(1920);
        settings.setFullscreen(true);
```

I'm adding a blue quad that should be fully outside of the screen:

```java
        Quad rectangleMesh = new Quad(100, 100);
        Spatial geometry = new Geometry("Rectangle", rectangleMesh).move(0, 1080, 0);
        Material material = new Material(assetManager,"Common/MatDefs/Misc/Unshaded.j3md");
        material.setColor("Color", ColorRGBA.fromRGBA255(0, 64, 128, 128));
        geometry.setMaterial(material);
        getGuiNode().attachChild(geometry);
```

The first screenshot is from 3.7.0:

![Image](https://github.com/user-attachments/assets/c8734909-3c95-487c-bdc7-dc6d1e40c344)

The second screenshot is from 3.8.0:

![Image](https://github.com/user-attachments/assets/04163dab-bc1a-4bd9-bc2d-7ac3a84866cf)

Notice that part of the quad is visible in the top left corner of the screen, and the framerate is cut off at the bottom.

I am using Linux with Wayland as display server protocol.

Everything is fine when using windowed mode. However, I noticed that 3.8.0 uses a custom tile bar. Past experiences tell me that this might be the culprit.

yaRnMcDonuts:
Thanks for reporting this.

This may require some further investigation, but so far I do see that there were a few minor PRs related to screen resolution in 3.8 from @stephengold. However, I don't think any of his PRs did anything related to the custom title bar you mentioned, so I suspect that your issue is more likely be related to a recent PR from @bob0bob that added support for multiple monitors, it was a fairly large PR so there could have been some changes in that PR inadvertently caused this issue.

JNightRider:
I think I'm a little late.

> I am using Linux with Wayland as display server protocol.

After seeing this thread, I started testing and yes, there is indeed a bug with Wayland (I don't know if it is present on x11).

[...]

PR #2492, opened 2025-06-12
Wayland fullscreen issue fixed

JNightRider:
This PR fixes issue #2472, which relates to fullscreen on Wayland. Interestingly, the custom bar isn't automatically disabled when launching glwf in fullscreen during a Wayland session.


yaRnMcDonuts:
Thanks for fixing this! Everything looks good to me, so I'll merge this soon.

It also looks like something may have bugged out with the checks (3 different PRS that were submitted on the same day all failed, so I assume it was a glitch on GitHub's end), but you should be able to just re-run the checks for this PR and it should pass the checks this time.

JNightRider:
> It also looks like something may have bugged out with the checks (3 different PRS that were submitted on the same day all failed, so I assume it was a glitch on GitHub's end), but you should be able to just re-run the checks for this PR and it should pass the checks this time.

You're right! Now this PR passed the tests without any problems...

PR #2495, opened 2025-06-13
feat: illboardControl optimization + javadoc + bugfix + new Constructor

capdevon:
This PR introduces significant performance optimizations, a new constructor for easier initialization, and comprehensive documentation for the `BillboardControl` class.

**Key Changes & Improvements:**

* **New Constructor for Convenience:**
    * Added a constructor `BillboardControl(Alignment alignment)` that allows specifying the initial billboard alignment directly upon instantiation. This simplifies common usage patterns.
* **Performance Optimization:**
    * Replaced `new Quaternion()` with a reusable `tempQuat` instance in `rotateScreenAligned()` to reduce per-frame object allocations, thereby minimizing garbage collection pauses and improving overall frame rate stability.
* **Code Quality & Maintainability:**
    * Added comprehensive Javadoc comments to the class, its methods, and member variables. This improves code readability, makes the control easier to understand, and facilitates future maintenance and development.
* **Demon:**
    * Updated the `TestBillboard` class to demonstrate the functionality of the optimized `BillboardControl` and its various alignment options at runtime.

These changes ensure the `BillboardControl` remains functionally identical while significantly improving its runtime efficiency and developer experience.
```

![image](https://github.com/user-attachments/assets/917a428c-ed00-4cc4-bb69-c5cd5929f7e8)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/control/BillboardControl.java
+++ b/jme3-core/src/main/java/com/jme3/scene/control/BillboardControl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: jme3-core/src/main/java/com/jme3/scene/control/BillboardControl.java [hunk 1]
```diff
@@ -43,15 +43,34 @@
...
+ * <li>AxialZ: The billboard faces the camera but keeps its local Z-axis fixed.</li>
+ * </ul>
+ */
 public class BillboardControl extends AbstractControl {
 
-    private Matrix3f orient;
-    private Vector3f look;
-    private Vector3f left;
-    private Alignment alignment;
+    // Member variables for calculations, reused to avoid constant object allocation.
```
richardTingle, line 66: orient, look & left looked to already be reused temp variables. They were effectively final other than when they were set by the read method. Why not keep the understandable names?

(You're right about the new Quaternion() though, might as well introduce a new temp variable for that)

richardTingle:
In the example, how is the yellow quad supposed to behave. Even in JME today it behaves weirdly (so not your change) but I want to understand what it is supposed to do (it doesn't seem to billboard, just flip between two different orientations)

capdevon:
> In the example, how is the yellow quad supposed to behave. Even in JME today it behaves weirdly (so not your change) but I want to understand what it is supposed to do (it doesn't seem to billboard, just flip between two different orientations)

That's a good question about the yellow quad's intended behavior. Honestly, I'm not entirely sure what its original purpose was or how it's supposed to behave. I've encountered a fair amount of code in the jME core module that seems redundant, incomplete, or a bit strange, and this might be another instance of that.

From my perspective, the most useful and commonly utilized behaviors are `Alignment.Screen` (for true billboarding) and `Alignment.AxialY` (for rotation around the Y-axis).

PR #2500, opened 2025-06-16
Support for switching platforms (from Wayland EGL to X11 GLX)

JNightRider:
PR where support is implemented to change the default [GLFW](https://www.glfw.org/docs/latest/intro_guide.html#platform) platform on Linux distributions or derivatives that support Wayland to use X11 (enable GLX).

This will open the ability to reposition windows and set icons for them, as Wayland currently does not have support for this, this works if you have XWayland enabled to have a bridge.

Wayland uses EGL by default to create GL contexts, which is great, but sometimes we need to use GLX for our applications, which is very useful. XWaylanda for GLX context.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/system/AppSettings.java
+++ b/jme3-core/src/main/java/com/jme3/system/AppSettings.java
@@ -1518,7 +1518,7 @@ public final class AppSettings extends HashMap<String, Object> {
      * @param preferred true to prefer GLX (native X11) for the GL context, false to prefer EGL (native Wayland).
      */
     public void setX11PlatformPreferred(boolean preferred) {
-        put("X11PlatformPreferred", preferred);
+        putBoolean("X11PlatformPreferred", preferred);
     }
     
     /**
--- a/jme3-core/src/main/java/com/jme3/system/AppSettings.java
+++ b/jme3-core/src/main/java/com/jme3/system/AppSettings.java
@@ -1510,32 +1510,24 @@ public final class AppSettings extends HashMap<String, Object> {
     }
 
     /**
-     * Sets the native platform to be used to create the GL context.
-     * 
+     * Sets the preferred native platform for creating the GL context on Linux distributions.
      * <p>
-     * This only affects Linux distributions or derivatives that use a Wayland session in conjunction 
-     * with X11 via the XWayland bridge, which enables or disables GLX for window positioning and/or 
-     * icon configuration.
-     * </p>
+     * This setting is relevant for Linux distributions or derivatives that utilize a Wayland session alongside an X11 via the XWayland bridge.
+     * Enabling this option allows the use of GLX for window positioning and/or icon configuration.
      *
-     * <p>
-     * <strong>NOTE:</strong> Note that disabling this option uses GLX (native X11) instead of EGL (native WL).
-     * </p>
-     * 
-     * @param nplaf true if you want to enable GLX, otherwise false when using EGL (native)
+     * @param preferred true to prefer GLX (native X11) for the GL context, false to prefer EGL (native Wayland).
      */
-    public void setX11PlatformPreferred(boolean nplaf) {
-        put("X11PlatformPreferred", nplaf);
+    public void setX11PlatformPreferred(boolean preferred) {
+        put("X11PlatformPreferred", preferred);
     }
     
     /**
-     * Gets what type of platform is being used.
-     * 
+     * Determines which native platform is preferred for GL context creation on Linux distributions.
      * <p>
-     * Only valid on Linux distributions or derivatives that support Wayland, where it indicates whether GLX or EGL is enabled.
-     * </p>
-     * 
-     * @return returns true if GLX is enabled, otherwise false if used in EGL (native)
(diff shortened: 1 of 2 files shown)


riccardobl:
nice, i suggest to change the setting to 

- `setPreferX11Session(boolean)` : prefer x11 session when possible

i think "native platform" is not very self explanatory.
This also makes clear that jme will "try" but not guarantee an x11 session if x11 is not available.

capdevon:
```java
        if (glfwPlatformSupported(GLFW_PLATFORM_WAYLAND)) {
            /*
             * Change the platform GLFW uses to enable GLX on Wayland as long as you 
             * have XWayland (X11 compatibility)
             */
            if (!settings.isNativePlatform() && glfwPlatformSupported(GLFW_PLATFORM_X11)) {
                glfwInitHint(GLFW_PLATFORM, GLFW_PLATFORM_X11);
            }
        }
```

The code snippet explicitly checks for **X11** compatibility within a **Wayland** platform. 
The method name should reflect this specific check.

So,  `settings.isX11PlatformPreferred()` and `settings.setX11PlatformPreferred(boolean)` would be a significant improvement.

JNightRider:
@riccardobl , @capdevon . Thank you for your suggestions and for reviewing this PR.

In the end I think `setX11PlatformPreferred()` would be a better name for this option when enabling GLX.

jme3-core/src/main/java/com/jme3/system/AppSettings.java
```diff
@@ -1507,4 +1508,36 @@ public int getDisplay() {
...
+     * icon configuration.
+     * </p>
+     *
+     * <p>
+     * <strong>NOTE:</strong> Note that disabling this option uses GLX (native X11) instead of EGL (native WL).
+     * </p>
+     * 
+     * @param nplaf true if you want to enable GLX, otherwise false when using EGL (native)
+     */
+    public void setX11PlatformPreferred(boolean nplaf) {
```
capdevon: Please rename `nplaf` -> `preferred` and update the javadoc as follow:

```java
    /**
     * Sets the preferred native platform for creating the GL context on Linux distributions.
     * <p>
     * This setting is relevant for Linux distributions or derivatives that utilize a Wayland session alongside an X11 via the XWayland bridge.
     * Enabling this option allows the use of GLX for window positioning and/or icon configuration.
     *
     * @param preferred True to prefer GLX (native X11) for the GL context, false to prefer EGL (native Wayland).
     */
```

jme3-core/src/main/java/com/jme3/system/AppSettings.java
```diff
@@ -1507,4 +1508,36 @@ public int getDisplay() {
...
+     * Gets what type of platform is being used.
+     * 
+     * <p>
+     * Only valid on Linux distributions or derivatives that support Wayland, where it indicates whether GLX or EGL is enabled.
+     * </p>
+     * 
+     * @return returns true if GLX is enabled, otherwise false if used in EGL (native)
+     */
+    public boolean isX11PlatformPreferred() {
+        return getBoolean("X11PlatformPreferred");
```
capdevon, line 1533: would you consider a slightly refined version to enhance javadoc readability and precision? Something along these lines might be beneficial:

```java
    /**
     * Determines which native platform is preferred for GL context creation on Linux distributions.
     * <p>
     * This setting is only valid on Linux distributions or derivatives that support Wayland,
     * and it indicates whether GLX (native X11) or EGL (native Wayland) is enabled for the GL context.
     *
     * @return True if GLX is preferred, otherwise false if EGL is preferred (native Wayland).
     */
```

jme3-core/src/main/java/com/jme3/system/AppSettings.java
```diff
@@ -1507,4 +1508,28 @@ public int getDisplay() {
...
+    /**
+     * Sets the preferred native platform for creating the GL context on Linux distributions.
+     * <p>
+     * This setting is relevant for Linux distributions or derivatives that utilize a Wayland session alongside an X11 via the XWayland bridge.
+     * Enabling this option allows the use of GLX for window positioning and/or icon configuration.
+     *
+     * @param preferred true to prefer GLX (native X11) for the GL context, false to prefer EGL (native Wayland).
+     */
+    public void setX11PlatformPreferred(boolean preferred) {
+        put("X11PlatformPreferred", preferred);
```
capdevon: Please change `put("X11PlatformPreferred", preferred);` to `putBoolean("X11PlatformPreferred", preferred);`

We should use the type-specific `putBoolean` method here for clarity and correctness.

yaRnMcDonuts:
Hi @JNightRider thanks for this PR and for quickly addressing reviews! If everything is ready, then I'll merge this PR in the next couple of days. (I meant to merge it for alhpa2 a few days ago, but it got lost amongst many others and I've had a busy week, so I apologize for the delay)

JNightRider:
> If everything is ready, then I'll merge this PR in the next couple of days

With this final confirmation (to add a message to users indicating whether the changes have been applied to the platform), you should be all set!

Thank you for your patience.

PR #2501, opened 2025-06-17
Reduce object allocations in AudioListener

capdevon:
Reduce object allocations when `ALAudioRenderer` update listener each time the `Camera` is moved:

This pull request refactors the Listener class in the jMonkeyEngine audio system. The main changes include initializing the `location`, `velocity`, and `rotation` fields at declaration instead of in the constructor, removing redundant constructor assignments, and introducing reusable `Vector3f` instances (`left`, `up`, `direction`) for direction calculations. 

Methods that return direction vectors now store results in these fields rather than creating new objects, improving memory efficiency and reducing unnecessary allocations. Overall, the changes streamline object management and optimize performance in the Listener class.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/audio/openal/ALAudioRenderer.java#L768

```java
    private void applyListenerRotation(Listener listener) {
        Vector3f dir = listener.getDirection(); // <--
        Vector3f up = listener.getUp(); // <--
        // Use the shared FloatBuffer fb
        fb.rewind();
        fb.put(dir.x).put(dir.y).put(dir.z);
        fb.put(up.x).put(up.y).put(up.z);
        fb.flip();
        al.alListener(AL_ORIENTATION, fb);
    }
```

### DIFF
--- a/jme3-core/src/main/java/com/jme3/audio/Listener.java
+++ b/jme3-core/src/main/java/com/jme3/audio/Listener.java
@@ -166,9 +166,8 @@ public class Listener {
      * @return The listener's left direction as a {@link Vector3f}.
      */
     public Vector3f getLeft() {
-        return rotation.getRotationColumn(0);
+        return getLeft(null);
     }
-
 
     /**
      * Gets the left direction vector of the listener. This vector is derived from the listener's rotation.
@@ -177,7 +176,6 @@ public class Listener {
      * @return The listener's left direction as a {@link Vector3f}.
      */
     public Vector3f getLeft(Vector3f store) {
-        if (store == null) store = new Vector3f();
         return rotation.getRotationColumn(0, store);
     }
 
@@ -188,7 +186,7 @@ public class Listener {
      * @return The listener's up direction as a {@link Vector3f}.
      */
     public Vector3f getUp() {
-        return rotation.getRotationColumn(1);
+        return getUp(null);
     }
 
     /**
@@ -199,7 +197,6 @@ public class Listener {
      * @return The listener's up direction as a {@link Vector3f}.
      */
     public Vector3f getUp(Vector3f store) {
-        if (store == null) store = new Vector3f();
         return rotation.getRotationColumn(1, store);
     }
 
@@ -210,7 +207,7 @@ public class Listener {
      * @return The listener's forward direction.
      */
     public Vector3f getDirection() {
-        return rotation.getRotationColumn(2);
+        return getDirection(null);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/audio/Listener.java
```diff
@@ -130,32 +131,29 @@ public Vector3f getVelocity() {
 
     /**
      * Gets the left direction vector of the listener.
-     * This vector is derived from the listener's rotation.
      *
      * @return The listener's left direction as a {@link Vector3f}.
      */
     public Vector3f getLeft() {
-        return rotation.getRotationColumn(0);
+        return rotation.getRotationColumn(0, left);
```
richardTingle: This changes the behaviour. Previously this returned a fresh Vector3f that you had exclusive ownership of. Now you are returned a vector3f that may randomly change at some future point.

Unless this is giving a major boost in performance (from profiling) in typical flows I don't think we should do this

capdevon: This solution maintains consistency with the existing `getLocation()`, `getRotation()`, and `getVelocity()` methods, which already expose internal variables. This approach is not only consistent but also **crucial for performance optimization** and efficient reuse of internal variables, preventing new vector allocations every time.

The primary reason for this approach is that `Listener` is an internally managed class by `ALAudioRenderer` and `AudioListenerState`. It's designed as a unique instance created by `LegacyApplication` and is not intended for external direct use. Therefore, the concerns you've raised regarding its internal exposure are mitigated by its strictly internal scope and established usage patterns.

Edit:
Furthermore, specifically within `ALAudioRenderer.applyListenerRotation()`, having redundant copies of `up` and `direction` variables is unnecessary. The underlying audio buffer only needs to read the `x`, `y`, and `z` coordinates directly from the existing variables. Creating and passing extra copies would just add needless overhead.

riccardobl:
I've reverted the changes to the existing getters. Since the listener isn't only an internal API, changing their behavior could break existing code, and because the defect isn't severe, it isn't worth risking hard-to-debug regressions for developers, imo.

Instead, I added new getters that populate the passed store object (a pattern used frequently elsewhere in jME), and updated the Audio Renderer to use these new getters.

I think this is a good resolution.

capdevon:
Refactor `Listener.getLeft()`, `Listener.getUp()`, and `Listener.getDirection()` to rely on `Quaternion.getRotationColumn(int, Vector3f)`, which already handles null store vectors.
This removes redundant null checks and eliminates duplicated code.

PR #2507, opened 2025-06-19
Refactor: optimize ColorRGBA class + reduce code duplication

capdevon:
This pull request focuses on a significant code cleanup and optimization of the `ColorRGBA` class. The primary goal is to reduce code duplication, improve readability, and lay a stronger foundation for future development.

**Key improvements:**

* **DRY Principle Adherence:** Extensively applied the "Don't Repeat Yourself" principle by refactoring constructors and various `set` methods to delegate to a single, authoritative `set(float r, float g, float b, float a)`. This ensures consistent behavior and simplifies updates.
* **Improved Bitwise Operations Readability:** Complex conversions between `float` color components and `int`/`byte` representations (`asIntARGB`, `fromIntARGB`, etc.) have been abstracted into clear, concise private helper methods (`toByte`, `fromByte`, `toInt`). This makes the underlying logic much easier to follow and debug.
* **Consolidated Array and Interpolation Logic:** Methods like `getColorArray()` and `toArray()` now share a common internal implementation, as do the `interpolateLocal` overloads, leading to less redundant code.

These changes result in a more efficient, less error-prone, and more maintainable `ColorRGBA` class.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
+++ b/jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2021 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -30,18 +30,23 @@
  */
 package com.jme3.math;
 
-import com.jme3.export.*;
+import com.jme3.export.InputCapsule;
+import com.jme3.export.JmeExporter;
+import com.jme3.export.JmeImporter;
+import com.jme3.export.OutputCapsule;
+import com.jme3.export.Savable;
+
 import java.io.IOException;
 
 /**
  * <code>ColorRGBA</code> defines a color made from a collection of red, green
- * and blue values stored in Linear color space. An alpha value determines is
+ * and blue values stored in Linear color space. An alpha value determines its
  * transparency.
  *
  * @author Mark Powell
- * @version $Id: ColorRGBA.java,v 1.29 2007/09/09 18:25:14 irrisor Exp $
  */
 public final class ColorRGBA implements Savable, Cloneable, java.io.Serializable {
+
     static final float GAMMA = 2.2f;
 
     static final long serialVersionUID = 1;
@@ -153,15 +158,15 @@ public final class ColorRGBA implements Savable, Cloneable, java.io.Serializable
      * Copy constructor creates a new <code>ColorRGBA</code> object, based on
      * a provided color.
      *
-     * @param rgba The <code>ColorRGBA</code> object to copy.
+     * @param color The <code>ColorRGBA</code> object to copy.
      */
-    public ColorRGBA(ColorRGBA rgba) {
-        this.a = rgba.a;
-        this.r = rgba.r;
-        this.g = rgba.g;
-        this.b = rgba.b;
+    public ColorRGBA(ColorRGBA color) {
+        this.a = color.a;
+        this.r = color.r;
+        this.g = color.g;
+        this.b = color.b;
     }
-    
+
     /**
      * Constructor creates a new <code>ColorRGBA</code> object, based on
      * a provided Vector4f.
@@ -170,12 +175,9 @@ public final class ColorRGBA implements Savable, Cloneable, java.io.Serializable
      * values copied to this color's r, g, b, and a values respectively.
      */
     public ColorRGBA(Vector4f vec4) {
-        this.a = vec4.w;
-        this.r = vec4.x;
-        this.g = vec4.y;
-        this.b = vec4.z;
-    }    
-    
+        set(vec4);
+    }
+
     /**
      * Constructor creates a new <code>ColorRGBA</code> object, based on
      * a provided Vector3f, at full opacity with a 1.0 alpha value by default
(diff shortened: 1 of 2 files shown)


JNightRider (commented):
Hi @capdevon 

I was testing the differences these changes can make and noticed some strange things with the following methods:

1. asIntRGBA()
2. asIntABGR()
3. asIntARGB()

The new changes produce erroneous results. I'm not entirely sure why, but it could be a serious problem, as it doesn't produce the correct output.

This is my test class that I generated, to test the changes I created a new class **_ColorRGBAX_** that literally copies what is in this PR (the new changes), **_ColorRGBA_** is the one in 'master'.

```java
public class TestColor {

        float r = 1,
              g = 0.2f,
              b = 0.6f,
              a = 0.8f;

        ColorRGBA original  = new ColorRGBA(r, g, b, a);
        ColorRGBAX newColor = new ColorRGBAX(r, g, b, a);

        int rgba1 = original.asIntRGBA(),
            rgba2 = newColor.asIntRGBA();

        System.out.println("asIntRGBA()");
        System.out.println("ORIGINAL: " + rgba1);
        System.out.println("NEW     : " + rgba2);

        int abgr1 = original.asIntABGR(),
            abgr2 = newColor.asIntABGR();

        System.out.println("\nasIntABGR()");
        System.out.println("ORIGINAL: " + abgr1);
        System.out.println("NEW     : " + abgr2);

        int argb1 = original.asIntARGB(),
            argb2 = newColor.asIntARGB();

        System.out.println("\nasIntARGB()");
        System.out.println("ORIGINAL: " + argb1);
        System.out.println("NEW     : " + argb2);

        System.out.println("\n");
        System.out.println("COLOR RGBA : " + new ColorRGBA().fromIntRGBA(rgba1));
        System.out.println("COLORX RGBA: "+ new ColorRGBAX().fromIntRGBA(rgba2));

        System.out.println("\nCOLOR ABGR : " + new ColorRGBA().fromIntABGR(abgr1));
        System.out.println("COLORX ABGR: "  + new ColorRGBAX().fromIntABGR(abgr2));

        System.out.println("\nCOLOR ABGR : " + new ColorRGBA().fromIntARGB(argb1));
        System.out.println("COLORX ABGR: "  + new ColorRGBAX().fromIntARGB(argb2));
    }
}
```

With which I obtain the following result:

```shell
asIntRGBA()
ORIGINAL: -13395508
NEW     : -52

asIntABGR()
ORIGINAL: -862374913
NEW     : -1

asIntARGB()
ORIGINAL: -855690343
NEW     : -103


COLOR RGBA : Color[1.0, 0.2, 0.6, 0.8]
COLORX RGBA: Color[1.0, 1.0, 1.0, 0.8]

COLOR ABGR : Color[1.0, 0.2, 0.6, 0.8]
COLORX ABGR: Color[1.0, 1.0, 1.0, 1.0]

COLOR ABGR : Color[1.0, 0.2, 0.6, 0.8]
COLORX ABGR: Color[1.0, 1.0, 0.6, 1.0]
```

jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
```diff
@@ -825,4 +805,25 @@ public ColorRGBA getAsSrgb() {
...
+     */
+    private byte toByte(float channel) {
+        return (byte) ((int) (channel * 255) & 0xFF);
+    }
+
+    /**
+     * Helper method to convert an int (shifted byte) to a float (0-1).
+     */
+    private float fromByte(int channelByte) {
+        return ((byte) (channelByte) & 0xFF) / 255f;
```
yaRnMcDonuts, line 820: I think that this method would need to take in an extra param for the function to do the casting and shifting in the proper order so that it returns the correct origianl value as mentioned by JNightRider (and I think this would also make it more understandable by keeping all of the shifting operation in one line of code within the new method):

    private float fromByte(int color, int shift) {
        return ((byte) (color >> shift) & 0xFF) / 255f;
    }

And if that works to make the new value match the old, then the screenshot tests should also pass. 

(Although i'm also no expert in bit shifting and usually only stumble into bitshifting code every few years when I run into trouble with an internal jme class that uses it, so I inevitably always have to refigure things out with bit shifting anytime I encounter it)

yaRnMcDonuts (changes requested):

jme3-core/src/main/java/com/jme3/math/ColorRGBA.java
capdevon: You cannot add a `shift` parameter to the `fromByte()` method and use it universally because the last color component, is *not* bitshifted before being passed to `fromByte()`.

Consider the `fromIntABGR` method:

```java
public ColorRGBA fromIntABGR(int color) {
    a = fromByte(color >> 24); // Alpha is shifted
    b = fromByte(color >> 16); // Blue is shifted
    g = fromByte(color >> 8);  // Green is shifted
    r = fromByte(color);       // Red is NOT shifted
    return this;
}
```

If you were to change `fromByte` to `private float fromByte(int color, int shift)` and try to apply a shift within it, for the call `r = fromByte(color);`, there is no shift value that would make sense. The `color` value, when passed for `r`, already has the red component in its least significant 8 bits. Any `shift` value other than 0 (or a shift that aligns the red bits properly if it were not already aligned) would incorrectly manipulate the red component's value.

Therefore, the `fromByte` method needs to operate on a value that *already has the desired channel in its lowest 8 bits*, making a `shift` parameter redundant and problematic for the non-shifted components.

capdevon:
@JNightRider @yaRnMcDonuts  thanks for your feedback. I fixed the bug on the `toInt()` method, now all tests passed successfully.

yaRnMcDonuts (approved):

capdevon:
I also added the `ColorRGBATest` class in the test package ;)

PR #2513, opened 2025-06-23
Refactor: DirectionalLightShadowRenderer reduce object allocations

capdevon:


yaRnMcDonuts (approved):

jme3-core/src/main/java/com/jme3/shadow/DirectionalLightShadowRenderer.java
```diff
@@ -51,56 +51,67 @@
...
+     * For serialization only. Do not use.
      */
     protected DirectionalLightShadowRenderer() {
         super();
     }
 
     /**
-     * Creates a DirectionalLight shadow renderer. More info on the technique at <a
-     * href="https://developer.nvidia.com/gpugems/GPUGems3/gpugems3_ch10.html">https://developer.nvidia.com/gpugems/GPUGems3/gpugems3_ch10.html</a>
+     * Creates a DirectionalLight shadow renderer. This renderer implements the
```
richardTingle: Why remove the link? Still seems to work

jme3-core/src/main/java/com/jme3/shadow/DirectionalLightShadowRenderer.java
```diff
@@ -151,16 +162,16 @@ protected void updateShadowCams(Camera viewCam) {
         ShadowUtil.updateFrustumPoints(viewCam, frustumNear, zFar, 1.0f, points);
 
         shadowCam.setFrustumFar(zFar);
-        shadowCam.getRotation().lookAt(light.getDirection(), shadowCam.getUp());
+        shadowCam.getRotation().lookAt(light.getDirection(), shadowCam.getUp(tempVec));
```
richardTingle, line 157: This is only called once per frame, is it really worth it to create a reusable vector? (Did this come from profiling and finding lots of memory allocations?)

jme3-core/src/main/java/com/jme3/shadow/DirectionalLightShadowRenderer.java
richardTingle: Ah, I see, you've moved it to the top. Makes sense

jme3-core/src/main/java/com/jme3/shadow/DirectionalLightShadowRenderer.java
capdevon: While it's just one call per frame, creating a new Vector each time means allocating a new object in memory every single frame. Over a typical game's thousands of frames per second, this quickly accumulates into significant garbage collection pressure. Utilizing a reusable vector prevents these continuous allocations, which is a fundamental practice for maintaining performance in game loops, even for 'small' allocations.

More importantly, this isn't an isolated case. If multiple classes or methods each perform 'small' per-frame allocations like this, the cumulative effect on memory churn and garbage collector pressure becomes significant, leading to noticeable framerate drops and stuttering. Proactive object pooling or reuse, as done here, is a standard and crucial optimization to avoid these continuous per-frame allocations and ensure smoother gameplay.

richardTingle: Micro optimisations like this should always be led by profiling. If you don't profile you waste loads of time optimising stuff that doesn't matter. The cumulative effect of loads of micro optimisations is to reduce readability; that can be worth it if it really does improve performance but without profiling you can't know if that's the case. Usually the vast vast majority of memory allocations (and CPU time) will be within tight loops that are called tens of thousands of times per frame

> Over a typical game's thousands of frames per second

A game will typically have 60 frame per second

riccardobl (approved):
Thanks, this looks good to me.

I undid some formatting and documentation changes to keep this PR focused on the fix.

PR #2514, opened 2025-06-23
Refactor: PointLightShadowRenderer reduce object allocations

capdevon:
This PR refactors the `PointLightShadowRenderer` class to improve readability and maintainability. Key changes include:

- Replaces repeated inline vector negations (e.g., `Vector3f.UNIT_X.mult(-1f)`) with clearly named constants (`X_NEG`, `Y_NEG`, `Z_NEG`).
- Updates camera axis assignments in `updateShadowCams()` to use these new constants, reducing code duplication and potential errors.
- Minor improvements to JavaDoc for clarity and consistency.

These changes make the codebase cleaner and easier to understand without altering any functionality.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/shadow/PointLightShadowRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/shadow/PointLightShadowRenderer.java
@@ -51,19 +51,25 @@ import com.jme3.util.clone.Cloner;
 import java.io.IOException;
 
 /**
- * PointLightShadowRenderer renders shadows for a point light
+ * Renders shadows for a {@link PointLight}. This renderer uses six cameras,
+ * one for each face of a cube map, to capture shadows from the point light's
+ * perspective.
  *
  * @author Nehon
  */
 public class PointLightShadowRenderer extends AbstractShadowRenderer {
 
+    /**
+     * The fixed number of cameras used for rendering point light shadows (6 for a cube map).
+     */
     public static final int CAM_NUMBER = 6;
+
     protected PointLight light;
     protected Camera[] shadowCams;
-    private Geometry[] frustums = null;
-    private final Vector3f X_NEG = Vector3f.UNIT_X.mult(-1f);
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/shadow/PointLightShadowRenderer.java
```diff
@@ -51,33 +51,39 @@
 import java.io.IOException;
 
 /**
- * PointLightShadowRenderer renders shadows for a point light
+ * Renders shadows for a {@link PointLight}. This renderer uses six cameras,
+ * one for each face of a cube map, to capture shadows from the point light's
+ * perspective.
  *
- * @author Rémy Bouquet aka Nehon
+ * @author Nehon
```
richardTingle: Why the author change?

richardTingle:
(Apart from the author change) looks good. Makes sense to avoid pointless negative vector allocations

jme3-core/src/main/java/com/jme3/shadow/PointLightShadowRenderer.java
capdevon: The original author's alias 'Nehon' is already the standard signature for most of their contributions, consistent with their GitHub nickname. To maintain uniformity with the other 500 classes in the jme3-core module and to align with this established practice, I've adjusted the signatures in these four Shadow classes. My intention is never to remove an original author's alias from any jME class, only to ensure consistency where a clear alias already exists and is widely used. This change also implicitly supports privacy by standardizing the alias used.

capdevon:
Hi guys,
Could you tell me why there is an error in the Screenshot Test? I just restored the original name of the class author after a previous change that passed the test.

riccardobl:
> Hi guys, Could you tell me why there is an error in the Screenshot Test? I just restored the original name of the class author after a previous change that passed the test.

looks like a timeout, sometime github action is a bit flaky, i've rerunned it

PR #2515, opened 2025-06-23
Reduce object allocations in SpotLightShadowRenderer

capdevon:
This PR refactors the `SpotLightShadowRenderer` class to improve code readability, consistency, and performance. Key changes include:

- Replaces hardcoded array allocation of `Camera[] cameras` in methods with a class-level field to avoid redundant allocations.
- Cleans up and clarifies Javadocs and comments.
- Minor improvements to method parameter usage (e.g., use of temporary vectors).
- Removes unnecessary lines and improves code formatting and spacing.

These changes do not modify the core logic or functionality but streamline the implementation and improve maintainability.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/shadow/SpotLightShadowRenderer.java
+++ b/jme3-core/src/main/java/com/jme3/shadow/SpotLightShadowRenderer.java
@@ -142,7 +142,7 @@ public class SpotLightShadowRenderer extends AbstractShadowRenderer {
         //shadowCam.setDirection(direction);
 
         shadowCam.setFrustumPerspective(light.getSpotOuterAngle() * FastMath.RAD_TO_DEG * 2.0f, 1, 1f, light.getSpotRange());
-        shadowCam.getRotation().lookAt(light.getDirection(), shadowCam.getUp());
+        shadowCam.getRotation().lookAt(light.getDirection(), shadowCam.getUp(tempVec));
         shadowCam.setLocation(light.getPosition());
 
         shadowCam.update();


jme3-core/src/main/java/com/jme3/shadow/SpotLightShadowRenderer.java
```diff
@@ -60,38 +60,37 @@
...
         super();
     }
     
     /**
-     * Create a SpotLightShadowRenderer This use standard shadow mapping
+     * Creates a SpotLightShadowRenderer.
      *
-     * @param assetManager the application asset manager
-     * @param shadowMapSize the size of the rendered shadowmaps (512,1024,2048,
-     * etc...) The more quality, the fewer fps.
```
yaRnMcDonuts, line 85: I would add back the comment mentioning "The more quality, the fewer fps." however that could certainly be worded better.

Maybe somthing like "Larger shadowmaps not only use more memory but also will result in a lower framerate"

Most users would expect a larger image to use more memory (so that may not be important to add) but it is still important to mention that a larger texture also correlate with a lower framerate (since that is not always the case with textures in other use cases, and is useful to know in this situation)

capdevon: I also update the javadoc of the other filters with this detail

jme3-core/src/main/java/com/jme3/shadow/SpotLightShadowRenderer.java
```diff
@@ -159,8 +164,6 @@ protected GeometryList getOccludersToRender(int shadowMapIndex, GeometryList sha
     @Override
     protected void getReceivers(GeometryList lightReceivers) {
         lightReceivers.clear();
-        Camera[] cameras = new Camera[1];
-        cameras[0] = shadowCam;
```
codex128, line 163: I suggest keeping `cameras[0] = shadowCam` in, rather than only assigning during initialization. It makes this section much clearer.

jme3-core/src/main/java/com/jme3/shadow/SpotLightShadowRenderer.java
```diff
@@ -52,46 +52,51 @@
...
  */
 public class SpotLightShadowRenderer extends AbstractShadowRenderer {
 
-    protected Camera shadowCam;    
+    protected Camera shadowCam;
     protected SpotLight light;
-    protected Vector3f[] points = new Vector3f[8];
-    //Holding the info for fading shadows in the far distance 
-   
+    protected final Camera[] cameras = new Camera[1];
```
codex128: Perhaps rename to `tempCams` since it is used as temporary storage of shadowCam.

PR #2516, opened 2025-06-23
Refactor: AbstractShadowFilter reduce object allocations

capdevon:
**Refactor AbstractShadowFilter: Remove Unused Serialization and Legacy Fields**

- Removed unused imports and legacy serialization methods (`write`/`read`), simplifying the class.
- Cleaned up and clarified Javadoc comments for constructors and author attribution.
- Replaced legacy temporary variables with final temp fields for matrix and vector operations, improving consistency.
- Minor improvements in naming and parameter descriptions for clarity and maintainability.

This refactor streamlines the code, reduces clutter, and adheres to modern coding practices without affecting functionality.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
+++ b/jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
@@ -88,6 +88,15 @@ public abstract class AbstractShadowFilter<T extends AbstractShadowRenderer> ext
         return true;
     }
 
+    /**
+     * @deprecated Use {@link #getMaterial()} instead.
+     * @return The Material used by this filter.
+     */
+    @Deprecated
+    public Material getShadowMaterial() {       
+        return material;
+    }
+
     @Override
     protected void preFrame(float tpf) {
         shadowRenderer.preFrame(tpf);
--- a/jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
+++ b/jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
@@ -56,6 +56,9 @@ public abstract class AbstractShadowFilter<T extends AbstractShadowRenderer> ext
     protected T shadowRenderer;
     protected ViewPort viewPort;
 
+    private final Vector4f tempVec4 = new Vector4f();
+    private final Matrix4f tempMat4 = new Matrix4f();
+
     /**
      * For serialization only. Do not use.
      */
@@ -99,13 +102,8 @@ public abstract class AbstractShadowFilter<T extends AbstractShadowRenderer> ext
     protected void preFrame(float tpf) {
         shadowRenderer.preFrame(tpf);
         Matrix4f m = viewPort.getCamera().getViewProjectionMatrix();
-        TempVars vars = TempVars.get();
-        try {
-            material.setMatrix4("ViewProjectionMatrixInverse", vars.tempMat4.set(m).invertLocal());
-            material.setVector4("ViewProjectionMatrixRow2", vars.vect4f1.set(m.m20, m.m21, m.m22, m.m23));
-        } finally {
-            vars.release();
-        }
+        material.setMatrix4("ViewProjectionMatrixInverse", tempMat4.set(m).invertLocal());
+        material.setVector4("ViewProjectionMatrixRow2", tempVec4.set(m.m20, m.m21, m.m22, m.m23));
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
```diff
@@ -48,35 +44,34 @@
...
 import com.jme3.util.clone.JmeCloneable;
 
-import java.io.IOException;
-
 /**
  * Generic abstract filter that holds common implementations for the different
  * shadow filters
  *
- * @author Rémy Bouquet aka Nehon
+ * @author Nehon
```
richardTingle: Why have you changed the author?

jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
```diff
@@ -48,35 +44,34 @@
...
  */
-public abstract class AbstractShadowFilter<T extends AbstractShadowRenderer> extends Filter implements Cloneable, JmeCloneable {
+public abstract class AbstractShadowFilter<T extends AbstractShadowRenderer> extends Filter implements JmeCloneable {
```
richardTingle, line 54: Why remove Cloneable? (Why was it here in the first place?)

capdevon: The `JmeCloneable` interface already extends` java.lang.Cloneable`. Therefore, the explicit extension is redundant and its removal simplifies the code without altering its behavior. Therefore, what checks have you already performed that led you to raise this question?

jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
capdevon: The original author's alias 'Nehon' is already the standard signature for most of their contributions, consistent with their GitHub nickname. To maintain uniformity with the other 500 classes in the jme3-core module and to align with this established practice, I've adjusted the signatures in these four Shadow classes. My intention is never to remove an original author's alias from any jME class, only to ensure consistency where a clear alias already exists and is widely used. This change also implicitly supports privacy by standardizing the alias used.

jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
richardTingle: Ah yes good point. No objection to removing it then

jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
capdevon: I have re-added the original author's full name to speed up review process

richardTingle:
Looks good to me

jme3-core/src/main/java/com/jme3/shadow/AbstractShadowFilter.java
```diff
@@ -93,18 +88,12 @@ protected boolean isRequiresDepthTexture() {
         return true;
     }
 
-    public Material getShadowMaterial() {
```
yaRnMcDonuts, line 96: I see you removed this redundant method (I assume because there is already a getMaterial() method that does the same thing)

However, it may be best to deprecate this method first, and then maybe remove it at a future date (if at all).

I know that having a duplicate method may feel sloppy, but now that it has been there for so long, it can be better to leave it so that existing apps that are using this method don't break with the new version.

capdevon: I agree. Done.

riccardobl:
Thanks for this PR and good find.

I have only one question:
The engine usually uses [TempVars](https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/util/TempVars.java#L50)  in this case, is there any big downside in using that to be consistent with the rest of the code?

capdevon:
> I have only one question:
> The engine usually uses [TempVars](https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/util/TempVars.java#L50) in this case, is there any big downside in using that to be consistent with the rest of the code?

Done! Thank you for your feedback guys.

richardTingle:
Weird, I can't see anything in your commit that would have broken shadows but the shadow has disappeared from the screenshot 

image

capdevon:
I think I understand the problem. The temp variables passed to the material must be permanent, not volatile. In this case, `TempVars` cannot be used, and persistent class variables must be used as before.

riccardobl (approved):

PR #2521, opened 2025-06-24
MatParam: code cleanup + javadoc

capdevon:


jme3-core/src/main/java/com/jme3/material/MatParam.java
```diff
@@ -391,11 +391,10 @@ public void read(JmeImporter im) throws IOException {
 
     @Override
     public boolean equals(Object obj) {
-        // Optimization for same instance
-        if (this == obj) {
```
yaRnMcDonuts: What is the reasoning for changing the equals() method?

I suppose that splitting the obj == null check into its own if statment is okay since it technically doesn't change functionality, although that's debatably a negligible change since it could be preferred either way by different devs. 

However I'm more worried about the removal of the this == obj check. Typically that is standard to check first in an overridden equals() method; is there a reason it was removed that I may be overlooking?

capdevon: The check was re-added after being removed due to an oversight. Unfortunately, not all classes perform a standard, uniform check of the possible cases of the equals() method. When analyzing classes, it is easy to become confused or to use the wrong example as a model. Anyway, the problem is solved. Thanks for pointing it out. After weeks of software checks and reviews, I am feeling a bit tired, especially in this summer heat ;)

PR #2525, opened 2025-06-24
Refactor: BinaryImporter optimization + javadoc

capdevon:
This PR improves the `BinaryImporter` class by modernizing its code and enhancing maintainability:

- Updates copyright years.
- Replaces wildcard imports with explicit imports.
- Adds comprehensive Javadoc comments to public methods for better documentation and usability.
- Refactors resource management by using try-with-resources for streams to prevent resource leaks.
- Improves logging, especially for debug statistics.
- Changes some method visibilities and names for better encapsulation and clarity.
- Performs minor formatting and cleanup for code consistency.

These changes make the codebase easier to understand, safer, and more maintainable, while preserving existing functionality.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
+++ b/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
@@ -382,15 +382,15 @@ public final class BinaryImporter implements JmeImporter {
         return capsuleTable.get(id);
     }
 
-    private String readString(InputStream is, int length) throws IOException {
+    protected String readString(InputStream f, int length) throws IOException {
         byte[] data = new byte[length];
-        for (int j = 0; j < length; j++) {
-            data[j] = (byte) is.read();
+        for(int j = 0; j < length; j++) {
+            data[j] = (byte)f.read();
         }
         return new String(data);
     }
 
-    private String readString(int length, int offset) {
+    protected String readString(int length, int offset) throws IOException {
         byte[] data = new byte[length];
         for (int j = 0; j < length; j++) {
             data[j] = dataArray[j + offset];


jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
```diff
@@ -44,158 +56,203 @@
...
     private byte[] dataArray;
     private int aliasWidth;
     private int formatVersion;
 
     private static final boolean fastRead = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN;
-    
+
+    /**
+     * Creates a new {@code BinaryImporter}.
+     */
```
richardTingle, line 93: Does this Javadoc provide any non obvious information?

capdevon: An empty default constructor clearly just creates a new instance. However, Javadoc for every constructor is a standard and necessary practice for the Java API. It ensures generated documentation is complete and maintains consistency throughout the project's style. Even if the content seems minimal, its presence is **expected**. 

Raising questions on these well-established conventions slows down the review without adding significant value, and it takes away valuable time that could be dedicated to more complex and relevant code issues.

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
```diff
@@ -289,50 +382,57 @@ public InputCapsule getCapsule(Savable id) {
         return capsuleTable.get(id);
     }
 
-    protected String readString(InputStream f, int length) throws IOException {
+    private String readString(InputStream is, int length) throws IOException {
```
codex128: This changes the API. Please revert back to protected.

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
```diff
@@ -289,50 +382,57 @@ public InputCapsule getCapsule(Savable id) {
...
-            data[j] = (byte)f.read();
+        for (int j = 0; j < length; j++) {
+            data[j] = (byte) is.read();
         }
-
         return new String(data);
     }
 
-    protected String readString(int length, int offset) throws IOException {
+    private String readString(int length, int offset) {
```
codex128: Revert back to protected. Removing the throw clause also breaks API, and try blocks catching an IOException won't compile if the nothing in the block has that throw clause.

capdevon: This class is declared "final" and cannot be extended. Therefore, the "protected" method cannot be overridden. For this reason and to reduce code confusion, using the private modifier is more appropriate.

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
capdevon: This class is declared "final" and cannot be extended. Therefore, the "protected" method cannot be overridden. For this reason and to reduce code confusion, using the private modifier is more appropriate.

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryImporter.java
pspeed42: private loses friendly access.  Whether that's ok or not is up to debate but it technically IS a change in the API from a certain perspective.

codex128:
Closing because requested changes were ignored.

PR #2527, opened 2025-06-25
Matrix3f: making mXX fields public 

capdevon:
**Description:**  
This PR updates the `Matrix3f` class by changing the visibility of its matrix element fields (`m00` to `m22`) from `protected` to `public`. This modification allows direct access to these fields from outside the class, making it easier for users to manipulate matrix values without relying on getter or setter methods. 
The copyright notice has also been updated to 2025.

In this way we unify the Matrix3f class with the [Matrix4f](https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/math/Matrix4f.java#L63) class whose matrix elements have public access modifiers.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/math/Matrix3f.java
+++ b/jme3-core/src/main/java/com/jme3/math/Matrix3f.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2022 jMonkeyEngine
+ * Copyright (c) 2009-2025 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -63,39 +63,39 @@ public final class Matrix3f implements Savable, Cloneable, java.io.Serializable 
     /**
      * The element in row 0, column 0.
      */
-    protected float m00;
+    public float m00;
     /**
      * The element in row 0, column 1.
      */
-    protected float m01;
+    public float m01;
     /**
      * The element in row 0, column 2.
      */
-    protected float m02;
+    public float m02;
     /**
      * The element in row 1, column 0.
      */
-    protected float m10;
+    public float m10;
     /**
      * The element in row 1, column 1.
      */
-    protected float m11;
+    public float m11;
     /**
      * The element in row 1, column 2.
      */
-    protected float m12;
+    public float m12;
     /**
      * The element in row 2, column 0.
      */
-    protected float m20;
+    public float m20;
     /**
      * The element in row 2, column 1.
      */
-    protected float m21;
+    public float m21;
     /**
      * The element in row 2, column 2.
      */
-    protected float m22;
+    public float m22;
     /**
      * Shared instance of the all-zero matrix. Do not modify!
      */


richardTingle:
Similar to the quaternion class I dont think we should be exposing internal implementation details unless we have a really good use case

capdevon:
Regarding the proposal to directly expose fields in `Quaternion` and `Matrix3f`, I believe this is a **fundamental step** toward optimizing crucial aspects of our engine and aligning with an already **established design pattern**.

Core math classes like `Vector2f`, `Vector3f`, `Vector4f`, `ColorRGBA`, `Ray`, and `Matrix4f` **already expose their variables via public modifiers**. This wasn't an arbitrary design choice; it has proven **extremely effective** and is widely used throughout `jme3-core` (especially by `Material` parsers and `ShadowRenderers`), as well as by our entire developer community.

All these classes are declared as **`final`**. This means they cannot be extended, which makes typical arguments related to inheritance and encapsulation via getters/setters less pertinent in this context. Altering this structure now would mean breaking compatibility with all current and past jME projects—a risk none of us want to take.

Bringing `Quaternion` and `Matrix3f` into this consistent design is a **logical progression**. It's not just about aesthetics; it's about enabling the **performance and direct usage advantages** I've already detailed, such as reduced allocations and more efficient access. Even without precise measurements, any native optimization can positively contribute to overall performance, which is crucial for achieving high frame rates on less powerful hardware like smartphones or mid-range devices.

The primary benefit of this adjustment is the ability to have **direct variable access** in high-computation areas such as `Material` parsers, advanced development tools, and demanding `shadow` computation functions, where direct field access in our existing math classes is already instrumental and widely exploited.

I understand that strict adherence to getters/setters is generally considered good programming practice. However, in the context of `final` math classes within a game engine like jME, where the **priority is efficiency and the consistency of an already established API**, opposing this change purely for formal reasons becomes counterproductive. The consistency with the other six `final` classes already following this pattern, combined with the tangible benefits in performance and usability, makes this proposal not just valid, but **necessary**.

yaRnMcDonuts:
My biggest concern is what @riccardobl mentioned about thinking the engine should go in the opposite direction, and that even the vector fields should not be public in the first place. 

I already question if there is any benefit of these changes, but his comment also raises the question that there could be downfalls as well. 

This type of design choice seems to ultimately be negligible and can vary from dev to dev based on coding preferences. Someone else may just come along and say they want to change this back again in the future if they have different preferences. 

For me, this falls int of the category of refactoring that seems very theoretical and has no clear benefit that I can see from my own real experiences using jme.  (I'd have to see a real example of an app that could substantially benefit from these changes to be convinced)

So I'd say this PR should be closed, especially due to the high volume of PRs right now that offer more objective benefits. Even if the debate continues and you prove that there is indeed a proven benefit, I still worry that it would be so minimal that is not worth the time debating over.

If you still really disagree then you can keep this PR (or any similar ones that get closed/denied)  open and it can be revisited in the future. But right now, my opinion would be to reject this PR.

capdevon:
The `protected` class variables have been restored. If you notice any technical errors or potential bugs in the proposed changes, please let me know.

jme3-core/src/main/java/com/jme3/math/Matrix3f.java
```diff
@@ -1253,7 +1245,7 @@ public int hashCode() {
      */
     @Override
     public boolean equals(Object o) {
-        if (o == null || o.getClass() != getClass()) {
+        if (!(o instanceof Matrix3f)) {
```
codex128: This change technically allows subclasses to be equal. The previous method is generally considered to be best practice.

https://dev.to/prabhatkjena/java-equals-best-practice-instanceof-vs-getclass-466l

jme3-core/src/main/java/com/jme3/math/Matrix3f.java
```diff
@@ -307,7 +312,10 @@ public Matrix3f normalize(Matrix3f store) {
             store = new Matrix3f();
         }
 
-        float mag = 1.0f / FastMath.sqrt(
+        float mag;
```
codex128, line 315: Why split declaration and assignment here?

codex128 (changes requested):

pspeed42:
Coming late to the party because I only recently got access to Slack again and so I now get to see all of the commits again.  And I know this has already been adjudicated...

re: the public fields on some math classes... as others said, this is for performance.

In the case of Quaternion, 99% of the time if a callers wants to access the individual values directly then they are misunderstanding quaternions and trying to do something that doesn't make sense.  That is why they are not public.  Changing the 'x' value without changing the other values simultaneously pretty much never makes sense for a Quaternion.

A matrix doesn't really have that same problem... though it is true that almost always when one value is updated at least two more should also be updated.  I'm on the fence about matrix.

jme3-core/src/main/java/com/jme3/math/Matrix3f.java
capdevon: So the problem is widespread across almost all math `Vector` and `Quaternion` classes. The engine is so full of errors that it is no longer clear which standard to use. I have restored the previous code for this class.

pspeed42: The correct standard is the one to use. :)  "if (!(o instanceof Matrix3f)) {" has been super-bad since "Thinking in Java" was first written... just taking a while for the code base to catch up.  :)

PR #2529, opened 2025-06-26
Armature: optimization + javadoc

capdevon:
This PR refactors the `Armature` class to enhance clarity, maintainability, and reliability. Key changes include:

- Improved and expanded Javadoc documentation for public methods and class members.
- Renamed `skinningMatrixes` to `skinningMatrices` for correct and consistent naming.
- Replaced wildcard and unused imports with explicit imports for better readability.
- Enhanced type safety and code clarity in serialization/deserialization logic.
- Updated copyright year.

These changes are non-breaking and aimed at improving code quality and maintainability.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/Armature.java
+++ b/jme3-core/src/main/java/com/jme3/anim/Armature.java
@@ -108,7 +108,7 @@ public class Armature implements JmeCloneable, Savable {
                 rootJointList.add(joint);
             }
         }
-        rootJoints = rootJointList.toArray(new Joint[rootJointList.size()]);
+        rootJoints = rootJointList.toArray(new Joint[0]);
 
         createSkinningMatrices();
 
@@ -215,9 +215,9 @@ public class Armature implements JmeCloneable, Savable {
      * @return the pre-existing instance or null if not found
      */
     public Joint getJoint(String name) {
-        for (Joint joint : jointList) {
-            if (joint.getName().equals(name)) {
-                return joint;
+        for (int i = 0; i < jointList.length; i++) {
+            if (jointList[i].getName().equals(name)) {
+                return jointList[i];
             }
         }
         return null;
@@ -353,13 +353,13 @@ public class Armature implements JmeCloneable, Savable {
     public void read(JmeImporter im) throws IOException {
         InputCapsule ic = im.getCapsule(this);
 
-        Savable[] rootJointsSavable = ic.readSavableArray("rootJoints", null);
-        rootJoints = new Joint[rootJointsSavable.length];
-        System.arraycopy(rootJointsSavable, 0, rootJoints, 0, rootJointsSavable.length);
+        Savable[] jointRootsAsSavable = ic.readSavableArray("rootJoints", null);
+        rootJoints = new Joint[jointRootsAsSavable.length];
+        System.arraycopy(jointRootsAsSavable, 0, rootJoints, 0, jointRootsAsSavable.length);
 
-        Savable[] jointListSavable = ic.readSavableArray("jointList", null);
-        jointList = new Joint[jointListSavable.length];
-        System.arraycopy(jointListSavable, 0, jointList, 0, jointListSavable.length);
+        Savable[] jointListAsSavable = ic.readSavableArray("jointList", null);
+        jointList = new Joint[jointListAsSavable.length];
+        System.arraycopy(jointListAsSavable, 0, jointList, 0, jointListAsSavable.length);
 
         String className = ic.readString("modelTransformClass", MatrixJointModelTransform.class.getCanonicalName());
         try {


jme3-core/src/main/java/com/jme3/anim/Armature.java
```diff
@@ -83,7 +108,7 @@ public Armature(Joint[] jointList) {
                 rootJointList.add(joint);
             }
         }
-        rootJoints = rootJointList.toArray(new Joint[rootJointList.size()]);
+        rootJoints = rootJointList.toArray(new Joint[0]);
```
yaRnMcDonuts, line 111: Why was this changed?

Does it solve a bug? Is there any proven benefit?

I also see this change was not mentioned in the description; however, this appears to be part of a significant refactoring and changes to code that may affect functionality and should be at the forefront of this PR's discussion. 

Without diving deeper into the code (because I don't have time and would need to know what benefit this change brings to determine if its worth my time investigating further) it appears that the size could have been >0 in cases where jointList had a size over 0. So unless there's something more going on here then this may indeed break things.

If there is a good reason for changing this, it should be explicitly detailed in the PR description so that a reviewer like myself doesn't have to spend more time studying the code to see if this is a mistake or intentional. 

Additionally, a seemingly minor change like this is 10000x more important than the low risk changes you mention in the PR's description. Thesee refactoring changes you've made are fairly major, and I think it is incorrect to label this as a javadoc and optimization PR. Those may be aspects of this PR, but code changes that may effect functionality should always be put first. Javadoc and minor code cleanup is nice but it is also extremely unimportant comparatively, and it is risky to have a PR with such heavy changes to the codebase to be labeled as javadoc and optimizations. 

Changes like this that effect functionality (or even have a chance to) should also preferably be in minimal PRs by themselves so that the focus can be solely on the thing that might break. And then changes to javadoc should preferably be in PRs where no code is changed. Combining things into a big multi-purpose PR like this slows me down reviewing the javadoc code that otherwise would've been a quick approval if it were a purely javadoc PR, and it also opens the door for errors slipping through.

capdevon: This modification is a standard and widely accepted best practice in modern Java (specifically since Java 8).

This change does not introduce any functional difference or risk of breaking existing functionality. It is purely an optimization and a move towards more idiomatic and robust Java code.

The PR was indeed labeled as "Javadoc and Optimization" because the core intent was to improve code readability, documentation, and apply minor, safe optimizations like the `toArray` idiom. This particular change falls squarely within the "optimization" or "code cleanup" category, as it improves performance characteristics and code style without altering logic. It is not a "major refactoring" in the sense of changing core algorithms or system architecture, nor does it carry the high risk you've attributed to it.

This specific `toArray` modification, being a widely adopted and safe Java idiom, does not warrant its own dedicated PR. Combining such minor, non-functional improvements with Javadoc updates helps streamline the review process for general code hygiene. Separating every single idiomatic correction would lead to an excessive number of PRs, potentially increasing review overhead rather than reducing it.

Finally, I want to emphasize the importance of a respectful and constructive approach in code reviews. Criticism should be based on accurate verification and an up-to-date understanding of language best practices. As a Release Manager, a deep understanding of core language idioms and modern best practices is crucial for efficient and accurate code review, enabling a quicker distinction between genuine functional changes/risks and standard code improvements. My goal with this PR, and all my contributions, is to ensure the game engine's code quality and maintainability evolve positively.

I hope this clarifies the rationale behind the change.

yaRnMcDonuts: My main concern is that the `new Joint[rootJointList.size()]);` was changed to be hardcode with a 0 with  `(new Joint[0]);` which looks risky at firs glance.

It appears that it would be possible that rootJoinList.size() is greater than 0 and hardcoding it as 0 may change things.

What is the reason for changing that, and is it really worth it? At first glance, this type of change does indeed infer some possibility of risk. And if I'm incorrect, it should be explained why this seemingly notable change to functionality is there. It shouldn't be the job of a reviewer to learn the entirity of the new code from scratch, and it greatly helps if you explain why these changes happened in the description. Reviewers do not have the same sense of reasoning that you have as the conitributor who saw this PR through from the get-go, so it is always important to explain a change like this so reviewers' jobs are easier.

yaRnMcDonuts: Also I think that you need to explain more the benefit in regards to "maintainability and consistency"

In cases like this, I cotinuously ask for a proven benefit of this.

All that this has achieved is making me look up more ways to code the same line of code to realize that the way it was before is no different than the way it is now. And actually my findings say that the old way was beter, but again thats preference so i take it with a grain of salt and ultimately say either way is fine and it is negligible.

From my experience, these minor negligible changes reuqire more maintenance work to apply and review, and I don't see any benefit.

And while I do value maintanability and organized code, you can't just keep using buzzwords like "maintainability and consistency" without explaining where the benefit actually is. Ther certainly are places where changes can make the engine easier to maintain, but I do not agree that this change is one of them. If anything, focusing on these tiny changes increase the immediate work for maintainers. I do not forsee this changed line of code making my future as a maintainer any easier, and I am finding it frustrating that you repeatedly talk about making the engine more maintainable without proving how in each individual specific situation. I feel that I've repeated myself many times and just get told I am wrong because of the importance of maintainability and consistency despite my adamant disagreeance with the level of importance you place on both of those abstract concepts.

capdevon: It's clear that we have different points of view. However, I don't want to waste any more time on this. I'm going to restore the original code snippet and move on.

yaRnMcDonuts: It’s completely fine if you want to move on from this specific change — I respect your time.

That said, I do want to express that I find this an unsatisfying conclusion to a valid concern I’ve raised multiple times. When “maintainability” is cited as justification, I think it’s fair to ask for specific, demonstrable benefits. In this case, I’ve repeatedly asked for that and haven’t gotten an answer. As the primary maintainer currently, it is frustrating when someone is telling me that something will make my job easier when I disagree, and then when I ask for a proven example in this specific scenario, you choose to disengage instead of providing an example or acknowledging that one doesn't exist and that there may be some merit in my perspective. 

I do not see how this one changed line of code makes maintenance easier, and I (quite honestly) find it alarming that you still believe otherwise and are only conceding out of apparent frustration with me. Additionally, getting PRs passed is a time-consuming process and shouldn't be taken lightly, so I apologize if you feel your time is wasted, but this is just how the process goes, and I'm trying to give feedback that should be considered in future PRs and reviewals.

To be clear: I’m not interested in a debate over who's point of view is right or wrong; I’m asking for clear, objective reasoning behind proposed changes that affect the engine’s codebase. If a change cannot be clearly justified in terms of benefit, then it should be reconsidered or openly acknowledged as a matter of personal preference rather than an improvement. And coming to a resolution on this issue related to what is and isn't considered easier to maintain (and not just disengaging from it) is important as to avoid the same issue arising again and again in some of your other PRs.

jme3-core/src/main/java/com/jme3/anim/Armature.java
```diff
@@ -173,9 +215,9 @@ public Joint getJoint(int index) {
      * @return the pre-existing instance or null if not found
      */
     public Joint getJoint(String name) {
-        for (int i = 0; i < jointList.length; i++) {
-            if (jointList[i].getName().equals(name)) {
-                return jointList[i];
+        for (Joint joint : jointList) {
```
codex128: Revert back to iteration by index (discussion at #2530).

jme3-core/src/main/java/com/jme3/anim/Armature.java
```diff
@@ -307,21 +351,21 @@ public void cloneFields(Cloner cloner, Object original) {
     @Override
     @SuppressWarnings("unchecked")
     public void read(JmeImporter im) throws IOException {
-        InputCapsule input = im.getCapsule(this);
+        InputCapsule ic = im.getCapsule(this);
 
-        Savable[] jointRootsAsSavable = input.readSavableArray("rootJoints", null);
-        rootJoints = new Joint[jointRootsAsSavable.length];
-        System.arraycopy(jointRootsAsSavable, 0, rootJoints, 0, jointRootsAsSavable.length);
+        Savable[] rootJointsSavable = ic.readSavableArray("rootJoints", null);
```
codex128: These two renames are redundant. The previous names were fine.

codex128: In my opinion, it's not worth the extra effort to revert all renames like this across all pending PRs. When they're private/local it shouldn't make much difference anyway. I'm going to stop pointing these out unless people are **really** unhappy with them.

jme3-core/src/main/java/com/jme3/anim/Armature.java
```diff
@@ -33,27 +33,49 @@
...
+     */
+    private transient Matrix4f[] skinningMatrices;
+    /**
+     * The class used to instantiate {@link JointModelTransform} instances for each joint.
+     * Defaults to {@link SeparateJointModelTransform}. This allows for customization
+     * of how joint model transformations are handled (e.g., separate scale/rotation/translation
+     * or combined in a single matrix).
      */
-    private transient Matrix4f[] skinningMatrixes;
     private Class<? extends JointModelTransform> modelTransformClass = SeparateJointModelTransform.class;
```
codex128, line 79: I am not in favor of documenting private fields in general. Since all these private fields have corresponding public methods available, all javadoc on these fields should be moved to those methods.

codex128 (changes requested):

jme3-core/src/main/java/com/jme3/anim/Armature.java
yaRnMcDonuts: > it's not worth the extra effort to revert all renames like this across all pending PRs

In this case I will agree, since capdevon has so many open PRs that make negligible variable name changes, and at this point it would be asking a lot to make him go through and revert variable names. 

So in this situation, we can make an exception and allow the variable renaming changes.

Going forward, I still would advise that changes to variable names be minimized unless the original name is exceptionally poor. 

Or if someone wants to change many variable names again in the future, it should all be put together in one focused PR that doesn't change anything else. Simply because that helps keep PRs targeted and easy to review.

codex128:
Closing because the requested changes were ignored. This PR also doesn't offer enough to justify the amount of work we're putting into it.

PR #2531, opened 2025-06-30
ArmatureDebugAppState: minor fix and improvements

capdevon:
This PR introduces the following improvements to the `ArmatureDebugAppState` class:

- **Input Handling:** Removes the right mouse button trigger from the joint picking input mapping, leaving only the left mouse button for selecting joints.
- **Action Listener:** Adds an early return in the action listener if the app state is disabled, preventing unintended processing.
- **Logging Enhancement:** Refactors joint info logging to use the logger at `INFO` level instead of printing to `System.err`, and formats the output as a single multi-line string for improved readability.

These changes improve usability, maintainability, and adhere to best practices for logging and input management.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/debug/custom/ArmatureDebugAppState.java
+++ b/jme3-core/src/main/java/com/jme3/scene/debug/custom/ArmatureDebugAppState.java
@@ -294,22 +294,22 @@ public class ArmatureDebugAppState extends BaseAppState {
             }
         }
 
-        private void printJointInfo(Joint selectedjoint, ArmatureDebugger ad) {
+        private void printJointInfo(Joint selectedJoint, ArmatureDebugger ad) {
             if (enableJointInfoLogging) {
                 String info = "\n-----------------------\n" +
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/scene/debug/custom/ArmatureDebugAppState.java
```diff
@@ -121,7 +121,7 @@ protected void initialize(Application app) {
     }
 
     private void registerInput() {
-        inputManager.addMapping(PICK_JOINT, new MouseButtonTrigger(MouseInput.BUTTON_LEFT), new MouseButtonTrigger(MouseInput.BUTTON_RIGHT));
```
codex128, line 124: Why was the right mouse button trigger removed?

capdevon: Assigning the left and right mouse buttons to the same function is redundant and not very useful. This way, the right mouse button (RMB) can be assigned to another action without interfering with this debug class.

PR #2532, opened 2025-06-30
Spatial: code cleanup & javadoc enhancement

capdevon:


jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -275,57 +291,57 @@ protected void setRequiresUpdates(boolean f) {
...
     }
 
     /**
      * Indicate that the bounding of this spatial has changed and that
      * a refresh is required.
      */
     protected void setBoundRefresh() {
-        refreshFlags |= RF_BOUND;
+        this.refreshFlags |= RF_BOUND;
+        propagateRefreshFlag(RF_BOUND);
```
yaRnMcDonuts, line 325: What is the reason for this call being added here?

This also appears to go beyond the scope of this PR that was labeled as code cleanup and javadoc.

While it is understandable that "code cleanup" can vary in meaning, this is veering into the territory of a heavy refactoring of a core stable class which carries a ton of risk, so every change to this class should be made with intense scrutiny and consideration.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -1608,22 +1646,21 @@ public boolean matches(Class<? extends Spatial> spatialSubclass,
     }
 
     @Override
-    @SuppressWarnings("unchecked")
     public void write(JmeExporter ex) throws IOException {
-        OutputCapsule capsule = ex.getCapsule(this);
```
yaRnMcDonuts, line 1613: As mentioned in 2 previous reviews, changing variable names like this is not warranted. 

While I understand your goal of making every write and read method consistent throughout the engine, that is a goal with no benefit and will only add to the maintenance and reviewing burden if we set such a high and non practical standard. 

Even if you achieve this level of consistency now, the effort of trying to maintain that consistency and enforcing it upon all future contributors is not worth it. As an open source project, we simply cannot afford to spend time on aiming for such a level of perfection, nor can we afford to hold future contributors to the same standards when they are inevitably contributing for free. 

If JME had an infinite budget and infinite manpower, then I'd be glad to say lets go ahead and make code consistent and perfect everywhere and try to maintain that. But resources are limited, and things like using consistent variable naming accross every class' read/write method is a futile effort that is only adding extra lines of code changed to the review load. Aiming for such a level of perfection also adds constraints for future contributors who are now expected to follow our rigorous formatting and variable naming standards. Consistency just isn't important in this context. And while it may be a low risk thing to change variable names like this, it inevitably opens the door to adding onto maintainers workload if we start allowing users to impose their own naming conventions anytime they make changes to a class.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
capdevon: I understand your concerns regarding the scope of this PR and the risks associated with modifying stable, core classes. However, I'd like to clarify the reason for the call you referenced, as it's directly linked to the DRY principle, which is fundamental for code maintainability and quality.

The modification in question aimed to eliminate code duplication by consolidating logic that was repeated in multiple places. While the PR was labeled as "code cleanup and javadoc," I believe that eliminating duplication fully falls under code cleanup, as it reduces complexity, simplifies future changes, and lowers the risk of introducing bugs. The DRY principle, in fact, is a cornerstone of good Java programming practices and significantly contributes to the long-term stability of the software by reducing the attack surface for errors and making the code easier to understand.

Every change to critical classes was made with the utmost care and scrutiny, including thorough testing to ensure no regressions or negative impacts on stability. My intention was never to perform an unjustified "heavy refactoring" but rather to improve the intrinsic quality of the code where duplication was evident.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
capdevon: I consider the renaming of variables, especially within methods like `write` and `read`, to be an integral part of the **optimization process** outlined in the PR title. Consistent and clear variable naming significantly enhances **code readability** and **maintainability**. When variable names like `ic` and `oc`, are consistently used across related methods throughout the codebase, it reduces cognitive load for developers, makes the code easier to scan, and minimizes potential confusion, particularly in an engine as complex as JME.

---

### Importance of Consistency

While I appreciate your perspective on resource constraints in an open-source project, I respectfully disagree that consistency in this context offers no benefit or that maintaining it is a futile effort. **Code consistency**, especially in fundamental aspects like variable naming within standard methods, is not about "perfection" but about **reducing friction and errors** for present and future contributors.

-   **Reduced Learning Curve:** New contributors can onboard faster when they encounter familiar patterns and naming conventions across different parts of the codebase.
-   **Easier Debugging and Maintenance:** When variable names are uniform, it's quicker to trace logic, understand data flow, and identify issues. Inconsistent naming can lead to subtle bugs or misinterpretations.
-   **Improved Code Quality:** Establishing and adhering to clear standards, even for naming, elevates the overall quality and professionalism of the project. It sets a positive precedent for all contributions.

The argument that this adds "extra lines of code changed to the review load" overlooks the long-term benefits in terms of reduced future maintenance burden and increased clarity. The initial investment in consistency pays dividends over time by making the codebase more predictable and easier to navigate for everyone.

---

### Clarifying Expectations for Contributions

Regarding future contributors, setting clear **coding standards** (including naming conventions) is a common and beneficial practice in almost all well-maintained open-source projects. It doesn't impose undue burden; rather, it provides a clear framework that helps contributors integrate their work seamlessly and maintain the project's integrity. It ensures that the project evolves in a structured and coherent manner, rather than becoming a disparate collection of individual styles.

I believe that these changes, including variable renaming, are a valuable part of improving the codebase's overall health and are well within the scope of code optimization.


Edit:
@JNightRider should know, too, since I reviewed his PR as well.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
yaRnMcDonuts: But (unless I'm missing something) the code in the  propagateRefreshFlag() method was previously not called there.

I'm mostly concerned by the addition of that method call.

This code block used to just change the value of refreshFlags , but now it calls a whole extra method and I am wondering what is the reason for calling propagateRefreshFlag() here now.

capdevon: It appears this block of code was indeed invoked before. Let's take a look at the history to confirm:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/scene/Spatial.java#L318

image

jme3-core/src/main/java/com/jme3/scene/Spatial.java
yaRnMcDonuts: Ultimately, I'm just trying to make the decision that I believe past core devs and release managers would've made in regards to a mass variable renaming effort, which would be to deny such a change and consider it unnecessary.

Unfortunately there isn't much documentation about engine policy. So I am just doing my best to follow suit with the engine culture and inferred policies I've observed over my past decade working with jme.

There is one contirbution read-me that sgold recently pointed out to me, wnich outline the process for making major changes starting from the planning phase and through the PR and reviewal phase. In hindsight I should have linked you to that before submitting all these PRs, because it does have some good recommendations about what steps to take in order to receive pre-approval before making big changes. And in this case, I think a thread may have been a better place to start to determine if the community as a whole agrees with each aspect of your refactoring efforts. 

So if you want to debate this, then I'd suggest starting a thread titled something like "renaming variables consistently accross multiple jme classes" so that this can be discussed and potentially approved. My suspicion is that certain core devs will disagree (hence my position thus far) but I also am willing to change my policy on this if other core devs or contributors turn out to feel otherwise.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -117,14 +132,14 @@ public enum BatchHint {
         Never;
     }
     /**
-     * Refresh flag types
+     * Refresh flag types. These flags indicate what data of the spatial
+     * needs to be updated to reflect the correct state.
```
codex128, line 136: This comment will describe only `RF_TRANSFORM` in javadoc, which would probably be confusing. Change this comment so it is not javadoc.

codex128 (changes requested):

jme3-core/src/main/java/com/jme3/scene/Spatial.java
codex128: What is the status of this? Should @capdevon revert variable renames or no?

Personally, I don't think these sorts of renames are worth the effort, but I'm fine with @capdevon doing them, as long as the new name is at least a little more descriptive than the previous one.

yaRnMcDonuts: I agree and still believe that variable renaming changes like this should be avoided as much as possible. The only exception would be in cases where the original variable names are exceptionally poor., and I don't think that was the case in this situation. I think that the original code that names the OutputCapsule object as "capsule" is perfectly fine. And it isn't an issue if other classes have a different name for OutputCapsule in their write method. Consistency has no benefit here in my opinion.

Enforcing strict naming conventions just adds more barriers for future contributors and reviewers who now have to follow these rules.

And if we start allowing name changes like this just for the sake of consistency, then that quickly opens the door to more needless name changing and more work reviewing things that could be spent working on more exciting and useful aspects of the engine.

capdevon:
Fix javadoc + Restore variable names:
* `capsule` for `OutputCapsule`
* `ic` for `InputCapsule`

codex128:
Closing because requested changes haven't been addressed in a reasonable amount of time. The PR also doesn't offer much benefit for the work we're putting in.

PR #2540, opened 2025-08-06
Fixes addControlAt

codex128:
This fixes `addControlAt` in Spatial awkwardly repositioning the control to add after calling `addControl` to do "bookkeeping". `addControl` now delegates everything to `addControlAt`.


codex128:
Screenshots are failing mostly on PBR tests for some reason (and not with any noticeable differences). This is a change I would not expect screenshots to fail on. @richardTingle would you mind taking a look?

richardTingle:
Hmm, this does seem really weird. I'm at work currently but I'll take a look at it after that

richardTingle:
Ok, I understand the problem. I use mesa, libgl1 and Xvfb to do CPU openGL rendering. And the pipeline just grabs the most recent one, and the whole thing is very sensitive to exact implementations to be pixel identical. I think one of these has released a new version.

What I need to do is create a docker container to effectively pin this so we can have a stable version. I'm going to do that now but are you blocked by this (there would be quicker ways to unblock you if needs be)

riccardobl:
I think this is potentially  a breaking change.
addControl is not final, so it can be overriden. 
In the current code addControlAt calls addControl so if you override addControl this affects also addControlAt. 
This change flips the flow, so that you need to override addControlAt to get the same behavior.

imo if not broken, it is better to not touch this, or if you really want to do this, we need to look into the engine to be sure there are no side effects

codex128:
> In the current code addControlAt calls addControl so if you override addControl this affects also addControlAt. This change flips the flow, so that you need to override addControlAt to get the same behavior.

Good point. That's a bummer because addControlAt is inefficient and the validation checks are different between the two methods, but I'm not willing to put in the extra effort for it.

I'm going to close this.

richardTingle:
Probably no longer relevant but for the purposes of linking everything together: this is the screenshot test fix: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2542

PR #2542, opened 2025-08-09
#2541 move screenshots to docker

### LINKED ISSUE #2541: Move screenshot tests over to use a docker container
The current screenshot tests pull whatever versions of mesa etc are the most recent. This can mean the underlying machine can change leading to test failures due to pixel level variation. Use a docker container to pin the versions

richardTingle:
This PR achieves the following

- Moves the screenshot tests over to a docker container. This is the main purpose of this PR, this means if Mesa, xvfb etc get new versions releases we shouldn't end up accepting those new versions by default (and so getting pixel level differences in the rendering triggering test failures)
- Rebake the images for this pinned set of libraries
- Include the full set of logs in the extent report (I really should have done that before, makes life a lot easier)
- Changed from a wait object to a CountDownLatch, hopefully avoiding random wakings and resultant flakeyness

Fixes #2541

The docker image being used is created with https://github.com/oneMillionWorlds/opengl-docker-image (if we wanted that could be moved into jMonkeyEngine world and a jMonkeyEngine branded docker container created)


richardTingle:
@yaRnMcDonuts This should fix the screenshot tests (and make us no longer vulnerable to version changes). Other pipelines will fail until this is merged

yaRnMcDonuts:
Thanks for the fix! 

If everything is ready here, then I'll go ahead and merge this in a few hours so that the tests are fixed for other PRs asap (however let me know if you want me to wait any longer)

richardTingle:
Yes please, as soon as possible would be great

PR #2545, opened 2025-08-09
Vulkan

codex128:
This is an experiment I've been working on for the past month or two. I'm opening this PR now to get feedback on whether Vulkan support should be pursued further, and if so, whether OpenGL support should be dropped. Let me know what you guys think. 

The primary goal is to determine how difficult it would be to port the engine over to Vulkan. Currently, I've made a set of tools that match 1:1 with Vulkan's base elements and have managed to render a spinning textured quad with those tools running from a JME application. I'm finding it to be not as bad as expected in terms of complexity.

Going forward, I'm trying to be as faithful as possible to JME's original high-level design. That means Mesh, Material, Geometry, etc. will hopefully be publically identical for their most common functions.

* Mesh should be relatively simple to port. Vulkan's vertex buffers are fairly similar.
* Geometry shouldn't need to change much since its primary job is to associate a material with a mesh.
* Material will probably be the hardest. Vulkan is completely different from OpenGL in this area.

Other considerations:

* Vulkan is manually synchronized. The engine components will have to be a lot more context-aware than they currently are.
* Vulkan benefits from having multiple frames-in-flight. That may require that some data structures have duplicates for each possible concurrent frame.
* We will have to drop lwjgl2 and jogl support. Most lwjgl-related advice is to move to lwjgl3 anyway, so I doubt anyone will be missing lwjgl2.

Because Vulkan uses structs for storing object creation settings that can often hold a lot of settings, I decided to use make use of builders and try-with-resources blocks. This way, it is obvious what methods are for initialization only, and everything is initialized and closed properly with the try-with-resources block. It also works very nicely with MemoryStack.

```java
VulkanInstance instance = new VulkanInstance(VK_API_VERSION_1_3);
try (VulkanInstance.Builder i = instance.build()) {
    i.addGlfwExtensions();
    i.addDebugExtension();
    i.addLunarGLayer();
    i.setApplicationName(VulkanHelperTest.class.getSimpleName());
    i.setApplicationVersion(1, 0, 0);
}
```
```java
LogicalDevice<GeneralPhysicalDevice> device = new LogicalDevice<>(instance);
try (LogicalDevice.Builder d = device.build(id -> new GeneralPhysicalDevice(instance, surface, id))) {
    d.addFilter(surface);
    d.addFilter(DeviceFilter.swapchain(surface));
    d.addCriticalExtension(KHRSwapchain.VK_KHR_SWAPCHAIN_EXTENSION_NAME);
    d.addFeature(DeviceFeature.anisotropy(1f, true));
}
```


riccardobl:
wow you've done a ton of work!

I don't think we should have the goal to drop opengl entirely, because it is still the most widely supported api, even on platforms where vulkan is not available yet (ie. web with webgl).

But i think we can decide to drop everything that is below GL ES 3 and GL ~3.2.
And if it helps harmonizing  vulkan and opengl renderers (actually i don't know, so i am asking you), we could switch to UBOs for material parameters.

codex128:
Thanks, your response is very encouraging! I will take that as support to continue working on this project.

> I don't think we should have the goal to drop opengl entirely, because it is still the most widely supported api, even on platforms where vulkan is not available yet (ie. web with webgl).

That is the position I expected, and I agree. I ask only because not supporting OpenGL would probably have made things a lot easier. :wink:

> if it helps harmonizing vulkan and opengl renderers, we could switch to UBOs for material parameters.

That's a good idea! I hadn't thought about doing it that way. I don't know for certain about the CPU-side stuff just yet, but it will help tremendously with having shaders work with either platform. This is definitely worth looking into more.

I'm planning on tackling Material first, as it looks to be the most difficult. I will have more details to provide then.

richardTingle:
Please ignore the screenshot test failure. This is caused by a version change. I have a PR out for it

codex128:
The major part of the material system has been implemented, finally. It turned out to be even more difficult than I anticipated, mostly due to Vulkan supporting frames-in-flight, and I wanted to get that part right.

A little explanation for those unfamiliar with Vulkan. Vulkan uses DescriptorSets to bind resources to certain uniform bindings. A set can contain multiple uniform bindings, which Material class reflects. DescriptorSets must be allocated from DescriptorPools using DescriptorSetLayouts to describe them. Additionally, DescriptorSets may only be bound to a pipeline containing the DescriptorSetLayout it was created with.

The Material class's job is to store uniform values and associate them with a specific binding in a specific set. To this end, it also manages DescriptorSets, and attempts to not create more of them than necessary. More could possibly be done in this area, but it is surprisingly difficult to manage DescriptorSets in relation to uniform values.

With that said, here is how a custom material would be created programmatically:

```java
public class TestMaterial extends Material {

    private final BufferUniform matrices = new BufferUniform(
            "Matrices", Descriptor.UniformBuffer, 0, ShaderStage.Vertex);
    private final TextureUniform baseColorMap = new TextureUniform(
            "BaseColorMap", Image.Layout.ShaderReadOnlyOptimal, 1, ShaderStage.Fragment);

    public TestMaterial(DescriptorPool pool) {
        super(pool);
        addSet(matrices, baseColorMap);
    }

    public BufferUniform getMatrices() {
        return matrices;
    }

    public TextureUniform getBaseColorMap() {
        return baseColorMap;
    }

}
```

For materials created from matdef files, I've made sure to include a way to access uniforms by name. This is how to create a material and assign a buffer value to it.

```java
UpdateFrameManager frames = ...
DescriptorPool pool = ...
TestMaterial material = new TestMaterial(pool);
material.getMatrices().setValue(frames.wrap(n -> new PersistentBuffer(device, MemorySize.floats(16), BufferUsage.Uniform, false)));
```

Material uniforms actually store VersionedResources, which handle resource versions for frames-in-flight. In a nutshell, a Vulkan application can have multiple frames-in-flight so that the CPU is building the instructions for the GPU while the GPU is working on the previous frame's instructions. This is good for performance, but it means that some resources will need a copy (what I call a version) for each possible frame. `frames.wrap` is creating a VersionedResource to manage a PersistentBuffer per frame-in-flight.

Rendering with the material is done by calling `bind`. `bind` will automatically select and bind DescriptorSets, which contain the values to be used by the shader.

```java
CommandBuffer graphicsCommands = ...
Pipeline currentPipeline = ...
material.bind(graphicsCommands, currentPipeline);
```

To update uniform values (especially buffers), simply access the buffer *for the current frame*, map, write, and unmap it as normal.

```java
try (MemoryStack stack = MemoryStack.stackPush()) {
    GpuBuffer matrixBuffer = material.getMatrices().getValue().getVersion();
    FloatBuffer matrixFloats = matrixBuffer.mapFloats(stack, 0, matrixBuffer.size().getElements(), 0);
    // write to matrixFloats
    matrixBuffer.unmap();
}
```

I haven't managed to do much with respect to matdefs. I don't see much risk in that area, and I'm tired of working on materials, so I'll probably move on to meshes next. Also, I would appreciate a quick lookover of the code; if something is obviously bad I want to know about it.

All things considered, this is going very well, and I'm seeing ways to abstract that would enable some overlap between OpenGL and Vulkan (specifically on buffers, which would be half the battle anyway). I hope to get it so that as many engine utilities as possible work for both OpenGL and Vulkan elements.

codex128:
Update:

Mesh is working now. Vulkan mesh vertex buffers are required to be laid out in a particular way as defined by the pipeline they're running under. Since I think users will often want to customize the layout, I've made AdaptiveMesh which will automatically format and fill the vertex buffers according to a given MeshDescription. Eventually all the pre-built meshes (Box, Quad, etc.) will extend AdaptiveMesh so they can be used in OpenGL and Vulkan.

I had a lot of trouble getting buffers to update properly. Eventually I landed on submitting buffers to particular CommandBatches depending on how each buffer should be treated. For example, buffers shared across frames (i.e. StaticBuffers) are submitted to a particular CommandBatch which will explicitly wait for all previous frames to complete. I have plans to improve this system using secondary command buffers.

I'm currently working on making the rest of the engine compatible with the larger parts of Vulkan to have as many aspects of the engine able to operate with both OpenGL and Vulkan components. This has involved abstracting Mesh, Material, VertexBuffer, Image, and Texture into interfaces, which has absolutely killed backwards compatibility.

codex128:
I've made progress for JME4, but not as much as I'd hoped. Most of the time was spent fleshing out or redesigning stuff I'd already built. I want the engine to not get in your way and not be confusing, and I'm actively looking at ways to improve it further. This is a general overview of the changes I've made so far. Please give me feedback on what I should change.

Here's an example of the "hello world" of jme:

```java
Engine engine = new SimpleVulkanEngine();
//Engine engine = new OpenGLEngine();
Geometry g = new Geometry("geom_jme4", new Box(1f, 1f, 1f));
Material mat = engine.createMaterial("Common/MatDefs/Misc/Unshaded.j3md");
try (StructMapping<UnshadedParams> m = mat.mapStruct("Parameters")) {
    UnshadedParams p = m.get();
    p.color.set(ColorRGBA.Blue);
    p.glowColor.set(ColorRGBA.Blue.mult(0.2f));
    p.vertexColor.set(true);
}
g.setMaterial(mat);
rootNode.attachChild(g);
```

The struct `p` is writing directly into the uniform buffer, which I do like, but having the try-with-resources in order to interface with the buffer seems a bit on the nose for something like this.

Viewport area is handled by ViewPort instead of Camera. Idk why it was in Camera in the first place, but I think it's more at home in ViewPort.

Render queues are handled by ViewPort instead of RenderManager. You can change what queues are present and how they behave per ViewPort.

```java
viewPort.addGeometryBucket(new GeometryBucket(new OpaqueComparator())  {
    @Override
    public void setupRender(ViewPort vp, StandardRenderSettings settings) {
        // getViewPort() returns a ViewPortArea, not a ViewPort!
        settings.pushViewPort(settings.getViewPort().clone().toMaxDepth());
    }
    @Override
    public void cleanupRender(ViewPort vp, StandardRenderSettings settings) {
        settings.popViewPort(); // undo the most recent push
    }
});
```

Different camera modes have been moved to subclasses: PerspectiveCamera for perspective mode, ParallelCamera for parallel projection, GuiCamera for gui rendering, etc. For engine internals, you no longer need to make the explicit distinction between gui and regular rendering for culling since that is handled by the camera implementations themselves. If you need to change the mode of a camera, encapsulate that camera inside the camera mode you need.

```java
Camera camToChange = ...
Camera newCamMode = new ParallelCamera(camToChange);
```

`newCamMode` delegates to `camToChange` for each interface call (with some tweaking here and there), so changes to `camToChange` are immediately reflected in `newCamMode`.

The primary way to interact with structured native memory now is through Struct, and Struct has undergone some fairly major changes. Fields are to be manually registered in the constructor rather than automatically through reflection (to improve performance).

```java
public class Transforms extends Struct {

    public final Field<Matrix4f> worldViewProjectionMatrix = new Field<>(new Matrix4f());
    public final Field<Matrix4f> viewProjectionMatrix = new Field<>(new Matrix4f());

    public Transforms() {
        addFields(worldViewProjectionMatrix, viewProjectionMatrix);
    }

}
```

If you have a buffer you'd like to interact with using `Transforms`:

```java
MappableBuffer buffer = ...
try (StructMapping<Transforms> m = buffer.mapAllStructs(new Transforms().bind(StructLayout.std140))) {
    Transforms t = m.get();
    m.sample(0); // bind t to offset 0 in the buffer
    t.worldViewProjectionMatrix.set(Matrix4f.IDENTITY);
    m.increment(); // bind t to the current offset plus sizeof(t)
    t.worldViewProjectionMatrix.set(Matrix4f.IDENTITY);
    // binds t to each multiple of sizeof(t) from 0 to the buffer's end in order.
    // preserves the current offset
    for (int i : m) {
        t.worldViewProjectionMatrix.set(t.worldViewProjectionMatrix.get());
    }
}
```

Struct does not store any data itself; it only determines where to read and write from the buffer, so the StructMapping is able to move the same Struct instance around to interact with multiple memory locations. It helps avoid having to recalculate the struct's layout each time and we don't create a lot of garbage.

Meshes use string names instead of enums to identify vertex attributes. I also thought it'd be neat to use structs to determine how vertex buffers are laid out and to interact with them. Structs for vertex buffers are forced to use `StructLayout.packed`.

```java
public static class VertexData extends Struct<VertexAttr> {
    public final VertexAttr<Vector3f> position = new VertexAttr<>("Position", new Vector3f());
    public final VertexAttr<Vector2f> texCoord = new VertexAttr<>("TexCoord", new Vector2f());
    public VertexData() { addFields(position, texCoord; }
}
// name "AdaptiveMesh" is to be changed
Mesh mesh = new AdaptiveMesh(4, 1); // vertices, instances
VertexBuffer data = new VertexBuffer(InputRate.Vertex, new VertexData(), JmePlatform.allocateStandardBuffer(1, BufferUsage.Vertex, UpdateHint.Static));
try (StructMapping<VertexData> m = data.map()) {
    VertexData v = m.get();
    for (int i : m) {
        v.position.alias().set(0f, i, i * i); // alias acts as a temporary Vector3f, but is attached to the struct field
        v.position.set(); // set from alias
        v.texCoord.alias().set(i, i * i);
        v.texCoord.set();
    }
}
mesh.addVertexBuffer(data);
```

If you need to interact with certain attributes by name but don't know in which buffer or by which struct they are stored:

```java
try (AttributeMapping m = mesh.mapAttributes(InputRate.Vertex, "Position", "TexCoord")) {
    StructField<Vector3f> pos = m.poll(); // fetches the first attribute named
    StructField<Vector2f> tex = m.poll(); // fetches the second attribute named
    m.sample(0);
    pos.set(Vector3f.ZERO);
    tex.set(Vector2f.ZERO);
}
```

Anytime a struct writes to a buffer, it also registers the written area as needing to be updated in some fashion (i.e. uploaded to the gpu).

Interacting with MappableBuffer directly is similar to that of structs. A key difference to watch out for is that you'll have to manually register the buffer areas to update yourself.

```java
MappableBuffer buffer = ...
try (BufferMapping m = buffer.map()) {
    m.getFloats().position(24).put(5f);
    m.stage(24 * Float.Bytes, Float.Bytes); // register the changed region for update
}
```

The try-with-resources pattern for mapping buffers could possibly be removed. I don't know if it's worth the effort of ripping all that infrastructure out or how performant it would be.

I'm still working out exactly how applications should go about creating buffers. Vulkan requires a lot of extra stuff with managing buffers and the different ways they can behave, while OpenGL barely cares. I'm thinking about something like this:

```java
MappableBuffer b = BufferUtils.allocateGraphicsBuffer(sizeInBytes, BufferUsage.Uniform, UpdateHint.Dynamic);
```

`allocateGraphicsBuffer` will defer to the engine object (or something of that nature) to create the specific buffer implementation.

To assist with type safety on backend flags and enums, I've created Flag and IntEnum interfaces. The intention was that it would enable the ease-of-use of java enums, but also allow users to use flags/enums not available by those java enums. For example:

```java
public enum BufferUsage implements Flag<BufferUsage> { ... }
public void doSomething(Flag<BufferUsage> usageFlag);
doSomething(BufferUsage.Vertex); // is accepted
doSomething(Flag.of(VK_BUFFER_USAGE_VERTEX)); // is accepted
```

Type "safety" is done by telling the user that `userFlag` should accept some form of `VK_BUFFER_USAGE...` and not some other flag type. Flag also includes a bunch of super useful helper methods and is immutable. IntEnum acts similarly for integer enums, but doesn't have the useful helper methods, so I'm considering removing it.

On the internal rendering side, I haven't landed on an overall system I'm happy enough with, but here's what I'm thinking at the moment.

```java
// lambda executed for each visible spatial
for (GeometryBucket b : vp.gatherGeometry(s -> s.runControlRender(engine, vp))) {
    // apply render settings
    b.setupRender(vp, settings);
    settings.applySettings();
    // encapsulate geometries in BucketElements, then sort the elements.
    // ExampleBucketElement is not real
    for (ExampleBucketElement e : b.sort(g -> new ExampleBucketElement(vp.getCamera(), g)) {
        // bind element resources (i.e. pipelines, materials)
        // update material parameters (lighting, etc)
        // update/upload buffers
        // render mesh
    }
    // cleanup render settings that were applied
    b.cleanupRender(vp, settings);
}
```

BucketElement encapsulates a Geometry and all information needed to properly sort that element in the bucket (camera, pipeline, material, textures, etc), so that information does not need to be stored in the geometry itself or in other less convenient places. The exact BucketElement implementation used will depend on the renderer, hence why the BucketElement type is not tied to GeometryBucket.

I'm not sure how things like lighting will be handled because on one hand we may want to give shaders flexibility in how they get light data, but I also want light processing to be as efficient as possible and that will likely require a more monolithic approach to fetching and packing light data.

A little bonus utility I thought would be nice to have is being able to iterate over a spatial's descendents with an enhanced for-loop. SceneGraphIterator already does this, but I decided it'd be more efficient to have it integrated into Spatial. I'm only bringing this up because I'm also toying around with accumulating inherited spatial properties outside `updateGeometricState`, which could help clean up Spatial.

```java
Deque<Spatial.CullHint> cullHint = new ArrayDeque();
// cullHint is automatically popped each time the iterator goes up in the graph hierarchy
for (Spatial s : scene.iterator(cullHint)) {
    if (s.getLocalCullHint() != Spatial.CullHint.Inherit) {
        cullHint.push(s.getLocalCullHint());
    } else if (cullHint.peek() != null) {
        cullHint.push(cullHint.peek());
    } else {
        cullHint.push(Spatial.CullHint.Dynamic);
    }
    if (cullHint.peek() == Spatial.CullHint.Dynamic) {
       // do frustum culling
    }
}
```

I don't have any strong opinions about this. I just thought it'd be an interesting alternative to explore.

codex128:
From the feedback previously, the two main complaints were both about the material system, so I have plans to completely redesign it from the ground up. Here are my goals for the new system from your feedback.

1. We want setters for setting material parameters. String names are too difficult to remember. A good option for a jme3-like system is richtea's build-time generator, but I think I have a better idea for jme4.
2. We don't want shader implementation to leak into the public, user-facing api. In other words, direct access to the struct of a uniform buffer is discouraged.
3. I want the system to be more optimize-able from the renderer's perspective. This means a more monolithic approach to rendering so that the renderer can micro-manage everything.
4. I don't want materials to be permanently locked into a particular rendering technique. If you want to have all PBR materials use your custom PBR fork, the engine should allow you to do that easily.

### General Solutions

Material will only be a container for material parameters. ShadingInterfaces can be attached to a material that allow the material to be rendered a certain way (i.e. "PBRInterface" would allow the material to be rendered as PBRLighting.j3md) and provide setters for interacting with related parameters. Each parameter's update state will be tracked using a `long` version number, similar to how lemur tracks updates.

To render materials containing a particular ShadingInterface, a "ShadingTechnique" object must be submitted at render time that maps to that ShadingInterface. In other words, a "PBRTechnique" object should be submitted to render all materials containing "PBRInterface". ShadingTechnique is responsible for choosing the shaders to render with and how parameters are mapped to those shader's inputs. If you want to use a different set of shaders to render all PBR materials, all you need to do is submit your own custom ShadingTechnique. ShadingTechnique parallels Technique and TechniqueDefLogic from jme3, but is not local to any one material or type of material per se.

I'm also planning on adding a user-controllable layer that manages global parameters/state/logic for all ShadingTechniques. This would allow things like lights to be computed and packed once per frame instead of once per material rendered as jme3 does. I currently have no concrete plans for how to do this.

Buffers that store material parameters expected to be updated less than once per frame (dynamic/static) are stored in and unique to its assigned material. Buffers with either parameters that are updated at least once per frame or global data are shared among all materials. I have yet to decide exactly how this will work out.

Unless we end up going solely with slang, shader metadata will be parsed either at build-time or run-time using lwjgl3-spvc (SPIRV-Cross). I hope there will be enough information from that to do away with matdefs all together, and it looks very promising. All parameters will automatically be mapped to a shader `#define` with an auto-generated name, so there will be no need to manually map them.

I haven't made a ton of progress on this yet because there seems to be a million tiny details that I need to account for, and some of my existing architecture is getting in the way. I'd like to get an opinion on this before I go to much further.

richardTingle:
@codex128 How would it look from the end users perspective? Say to instantiate an UnshadedMaterial and set the color?

codex128:
I think it would look something like this:

```java
Material m = new Material();
UnshadedInterface u = m.addInterface(new UnshadedInterface());
u.setColor(ColorRGBA.Blue);
```

richardTingle:
Interesting, thats definitely better than today, but why not:

```
UnshadedMaterial m = new UnshadedMaterial();
m.setColor(ColorRGBA.Blue);
```

Where UnshadedMaterial extends Material

codex128:
Sure, either way should work just fine in tandem.

codex128:
Buffer mapping was tending to get a bit unwieldy and it required that whoever is interacting with a structure backed by a buffer know about that buffer. It would be much better to simply be able to interact with a structure without needing to know what is backing it.

To fix this, I've introduced a few new concepts:

1. MappingArena: auto-closeable to which any number of memory mappings can be registered to get unmapped when the arena is closed.
2. Memory: represents a structure that can be mapped with a MappingArena to read/write data.
3. MemoryPointer: extends Memory; binds to another memory object which memory operations are then deferred to.

Buffers will implement Memory, and classes that are backed by a buffer will implement MemoryPointer. Since MemoryPointer extends Memory, you can have a MemoryPointer chain that eventually leads back to a raw Memory object.

Struct will implement MemoryPointer, so it can be bound to another memory object. This neatly solves the problem of needing to pass around both the struct and its backing buffer separately; you only need to pass around the struct. StructField will also implement Memory and defer to its struct for memory operations. So you might have a memory pointer chain that goes field -> struct -> field -> struct -> struct array -> buffer.

```java
static void example() {

    StructArray<MyStruct> array = new StructArray<>(new MyStruct(), 100); // struct automatically bound to array of length 100
    GpuBuffer data = ...; // backing buffer, will only be needed directly when binding new memory pointers

    // bind array to data so that mappings/reads/writes will draw from data
    array.bind(data);

    // begin an arena with array mapped
    try (MappingArena arena = new MappingArena(array)) {

        // iterate over struct array
        for (MyStruct s : array) {
            s.value.set(1f);
            s.sub.vec.set(Vector4f.UNIT_X);
            for (Sub e : s.array) {
                e.vec.set(Vector4f.UNIT_Y);
            }
        }

        // set by index
        array.index(10).value.set(2f);

        // iterate over float field
        FieldArray<StructField<Float>> fArray = array.field(s -> s.value);
        fArray.map(arena); // ensure fArray is mapped (redundent since array is mapped)
        for (StructField<Float> f : fArray) {
            f.set(1f);
        }

        // set by index
        fArray.index(5).set(5f);

        // bind new struct so that value from 10 can be transferred to value from 15
        array.index(15, new MyStruct()).value.set(array.index(10).value.get()); // warning: read operation can be slow depending on buffer memory properties!

        // iterate over struct array field
        for (SubStructArrayField<Sub> e : array.field(s -> s.arrayField)) {
            for (Sub sub : e.get()) {
                sub.vec.set(Vector4f.UNIT_XYZW);
            }
            e.index(7).vec.set(Vector4f.UNIT_W);
        }
    }

    MyStruct struct = new MyStruct();
    struct.bind(data); // bind struct to data

    try (MappingArena arena = new MappingArena(struct)) {
        struct.value.set(1f);
        struct.sub.get().vec.set(Vector4f.ZERO);
        struct.array.index(6).vec.set(Vector4f.UNIT_Z);
    }

    // destroy the buffer backing array
    NativeResource.destroy(array);

}

class MyStruct extends Struct {

    public final StructField<Float> value = new Field<>(0f);
    public final Sub sub = new Sub();
    public final StructArray<Sub> array = new StructArray<>(new Sub(), 10);
    
    // make distinction between struct and fields for more convenient access
    public final SubStructField<Sub> subField = new SubStructField<>(sub);
    public final SubStructArrayField<Sub> arrayField = new SubStructArrayField<>(array);

    public MyStruct() {
        addFields(value, subField, arrayField);
    }

}

class Sub extends Struct {

    public final Field<Vector4f> vec = new Field<>(new Vector4f());

    public Sub() {
        addFields(vec);
    }

}
```

codex128:
The FieldArray structure will also be particularly useful for things like mesh attributes where the user isn't expected to know how attributes are laid out in memory.

PR #2547, opened 2025-08-10
KHR pbr neutral tonemap

riccardobl:
[Khronos PBR Neutral Tone Map](https://www.khronos.org/news/press/khronos-pbr-neutral-tone-mapper-released-for-true-to-life-color-rendering-of-3d-products) for jme. It is basically just [this](https://github.com/KhronosGroup/ToneMapping/blob/main/PBR_Neutral/pbrNeutral.glsl) shader put in a filter with some adaptations.

It is a another standard algorithm for tone mapping, that is available also in blender, making it  a better choice to have renders that are consistent with what you see  in blender.

### DIFF
--- a/jme3-core/src/main/resources/Common/ShaderLib/Hdr.glsllib
+++ b/jme3-core/src/main/resources/Common/ShaderLib/Hdr.glsllib
@@ -63,3 +63,29 @@ vec3 HDR_ToneMap2(in vec3 color, in float lumAvg, in float a, in float white){
     float scale = a / (lumAvg + 0.001);
     return (vec3(scale) * color) / (color + vec3(1.0));
 }
+
+// Based on https://github.com/KhronosGroup/ToneMapping/blob/main/PBR_Neutral/pbrNeutral.glsl
+// Input color is non-negative and resides in the Linear Rec. 709 color space.
+// Output color is also Linear Rec. 709, but in the [0, 1] range.
+vec3 HDR_KHRToneMap(in vec3 color, in vec3 exposure, in vec3 gamma) {
+    color *= pow(vec3(2.0), exposure); 
+
+    const float startCompression = 0.8 - 0.04;
+    const float desaturation = 0.15;
+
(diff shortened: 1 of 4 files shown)


jme3-effects/src/main/java/com/jme3/post/filters/KHRToneMapFilter.java
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Tone-mapping filter that uses khronos neutral pbr tone mapping curve.
+ */
+public class KHRToneMapFilter extends Filter {
+
+    private static final float DEFAULT_EXPOSURE = 0.0f;
+    private static final float DEFAULT_GAMMA = 1.0f;
+    
+    private final Vector3f exposure = new Vector3f(DEFAULT_EXPOSURE, DEFAULT_EXPOSURE, DEFAULT_EXPOSURE);
+    private final Vector3f gamma = new Vector3f(DEFAULT_GAMMA, DEFAULT_GAMMA, DEFAULT_GAMMA);
```
riccardobl, line 55: i've used vectors here so it gives more artistic control, kinda doubling as color correction

jme3-core/src/main/resources/Common/ShaderLib/Hdr.glsllib
```diff
@@ -62,4 +62,30 @@ vec3 HDR_ToneMap(in vec3 color, in float lumAvg, in float a, in float white){
...
+    float peak = max(color.r, max(color.g, color.b));
+    if (peak < startCompression) return color;
+
+    const float d = 1. - startCompression;
+    float newPeak = 1. - d * d / (peak + d - startCompression);
+    color *= newPeak / peak;
+
+    float g = 1. - 1. / (desaturation * (peak - newPeak) + 1.);
+    color = mix(color, newPeak * vec3(1, 1, 1), g);
+    color = pow(color, gamma);
```
riccardobl, line 89: afaik this is the order used in blender to apply gamma and exposure, but i am not 100% sure on this

PR #2548, opened 2025-08-10
Add support for global lights

riccardobl:
JME lights only affect children of the nodes to which they are attached.
This makes sense from a performance/optimization standpoint, but it does not reflect the behavior of blender and other renderers.

This PR adds support for global lights ie. lights that affect all the spatials up to the rootNode.

Lights can be marked as global by setting to true the  "global" boolean flag on the lights' constructors . The flag is false by by default, so existing scene won't be affected.

This implementation exploits the jme scenegraph update logic to make global lights possible without any change to the renderer, this is an overview of how it works:

- A new refresh flag `RF_GLOBAL_LIGHTS`  is added.
- This flag is always propagated upstream up to the rootNode
- The rootNode seeing this flag will do a deep traversal of its children to find all the global lights to add to its own world light list.
- As a side effect, when the children are be updated, immediately after, they will pull the world lights list of their parent (first one being the rootNode with the global lights) propagating the global lights down the scene graph.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/DirectionalLight.java
+++ b/jme3-core/src/main/java/com/jme3/light/DirectionalLight.java
@@ -59,6 +59,17 @@ public class DirectionalLight extends Light {
      * Creates a DirectionalLight
      */
     public DirectionalLight() {
+        super();
+    }
+
+    /**
+     * Creates a DirectionalLight
+     * @param global if true, the light affects the entire tree from the root node,
+     * otherwise it only affects the children of the node in which it is attached.     
+     */
+    public DirectionalLight(boolean global) {
+        this();
+        this.global = global;
     }
 
     /**
@@ -66,7 +77,19 @@ public class DirectionalLight extends Light {
      * @param direction the light's direction
      */
     public DirectionalLight(Vector3f direction) {
+        super();
         setDirection(direction);
+    }
+
+    /**
+     * Creates a DirectionalLight with the given direction
+     * @param direction the light's direction
+     * @param global if true, the light affects the entire tree from the root node,
+     * otherwise it only affects the children of the node in which it is attached.
+     */
+    public DirectionalLight(Vector3f direction, boolean global) {
+        this(direction);
+        this.global = global;
     }
 
     /**
@@ -78,6 +101,19 @@ public class DirectionalLight extends Light {
         super(color);
         setDirection(direction);
     }
+
+    /**
+     * Creates a DirectionalLight with the given direction and the given color
+     * @param direction the light's direction
+     * @param color the light's color
+     * @param global if true, the light affects the entire tree from the root node,
+     * otherwise it only affects the children of the node in which it is attached.
+     */
+    public DirectionalLight(Vector3f direction, ColorRGBA color, boolean global) {
+        this(direction, color);
+        this.global = global;
+    }
+
 
     @Override
     public void computeLastDistance(Spatial owner) {
--- a/jme3-core/src/main/java/com/jme3/light/Light.java
+++ b/jme3-core/src/main/java/com/jme3/light/Light.java
@@ -116,6 +116,7 @@ public abstract class Light implements Savable, Cloneable {
      * The light name. 
      */
     protected String name;
+    protected boolean global = false;
     
     boolean frustumCheckNeeded = true;
     boolean intersectsFrustum  = false;
@@ -123,8 +124,28 @@ public abstract class Light implements Savable, Cloneable {
     protected Light() {
     }
 
+
+    protected Light(boolean global) {
+        this.global = global;
+    }
+    
+    protected Light(ColorRGBA color, boolean global) {
+        setColor(color);
+    }
+
     protected Light(ColorRGBA color) {
         setColor(color);
+    }
+
+
+    /**
+     * Returns true if this light affects the entire tree from the root node,
+     * otherwise returns false, meaning it only affects the children of the node
+     * in which it is attached.
+     * @return true if the light is global, otherwise false.
+     */
+    public boolean isGlobal() {
+        return global;
     }
 
     /**
@@ -265,6 +286,7 @@ public abstract class Light implements Savable, Cloneable {
         oc.write(color, "color", null);
         oc.write(enabled, "enabled", true);
         oc.write(name, "name", null);
+        oc.write(global, "global", false);
     }
 
     @Override
@@ -273,6 +295,7 @@ public abstract class Light implements Savable, Cloneable {
         color = (ColorRGBA) ic.readSavable("color", null);
         enabled = ic.readBoolean("enabled", true);
         name = ic.readString("name", null);
+        global = ic.readBoolean("global", false);
     }
 
     /**
--- a/jme3-core/src/main/java/com/jme3/light/LightList.java
+++ b/jme3-core/src/main/java/com/jme3/light/LightList.java
@@ -38,6 +38,7 @@ import com.jme3.util.clone.JmeCloneable;
 import com.jme3.util.SortUtil;
 import java.io.IOException;
 import java.util.*;
+import java.util.function.Predicate;
 
 /**
  * <code>LightList</code> is used internally by {@link Spatial}s to manage
@@ -230,6 +231,20 @@ public final class LightList implements Iterable<Light>, Savable, Cloneable, Jme
      * @param parent the parent's world-space LightList
      */
     public void update(LightList local, LightList parent) {
+        update(local, parent, null);
+    }
+
+
+    /**
+     * Updates a "world-space" light list, using the spatial's local-space
+     * light list, its parent's world-space light list and an optional filter.
+     *
+     * @param local the local-space LightList (not null)
+     * @param parent the parent's world-space LightList
+     * @param filter an optional filter to apply to the lights
+     *               (null means no filtering)
+     */
+    public void update(LightList local, LightList parent, Predicate<Light> filter) {
         // clear the list as it will be reconstructed
         // using the arguments
         clear();
@@ -237,30 +252,32 @@ public final class LightList implements Iterable<Light>, Savable, Cloneable, Jme
         while (list.length <= local.listSize) {
             doubleSize();
         }
-
-        // add the lights from the local list
-        System.arraycopy(local.list, 0, list, 0, local.listSize);
-        for (int i = 0; i < local.listSize; i++) {
-//            list[i] = local.list[i];
-            distToOwner[i] = Float.NEGATIVE_INFINITY;
+        
+        int localListSize = 0;// local.listSize;
+        for(int i=0;i<local.listSize;i++){
+            Light l = local.list[i];
+            if (filter != null && !filter.test(l))  continue; 
+            list[localListSize] = l;
+            distToOwner[localListSize] = Float.NEGATIVE_INFINITY;
+            localListSize++;
         }
 
         // if the spatial has a parent node, add the lights
         // from the parent list as well
         if (parent != null) {
-            int sz = local.listSize + parent.listSize;
+            int sz = localListSize + parent.listSize;
             while (list.length <= sz)
                 doubleSize();
 
             for (int i = 0; i < parent.listSize; i++) {
-                int p = i + local.listSize;
+                int p = i + localListSize;
                 list[p] = parent.list[i];
                 distToOwner[p] = Float.NEGATIVE_INFINITY;
             }
 
-            listSize = local.listSize + parent.listSize;
+            listSize = localListSize + parent.listSize;
         } else {
-            listSize = local.listSize;
+            listSize = localListSize;
         }
     }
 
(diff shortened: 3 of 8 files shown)


jme3-core/src/main/java/com/jme3/light/DirectionalLight.java
```diff
@@ -59,16 +59,39 @@ public class DirectionalLight extends Light {
...
     public DirectionalLight() {
+        super();
+    }
+
+    /**
+     * Creates a DirectionalLight
+     * @param global if true, the light affects the entire tree from the root node,
+     * otherwise it only affects the children of the node in which it is attached.     
+     */
+    public DirectionalLight(boolean global) {
```
riccardobl, line 70: i didn't make a setter for the global param, to make it defacto immutable.
This makes implementation much easier and performant, since we never need to check if the flag changed.

jme3-core/src/main/java/com/jme3/light/LightList.java
```diff
@@ -230,37 +231,53 @@ public void sort(boolean transformChanged) {
...
+    /**
+     * Updates a "world-space" light list, using the spatial's local-space
+     * light list, its parent's world-space light list and an optional filter.
+     *
+     * @param local the local-space LightList (not null)
+     * @param parent the parent's world-space LightList
+     * @param filter an optional filter to apply to the lights
+     *               (null means no filtering)
+     */
+    public void update(LightList local, LightList parent, Predicate<Light> filter) {
```
riccardobl, line 247: added a filter here so that we can exclude global lights from being added the children's internal world lights list.

That's because the rootNode propagates down all the global lights, so when a global light is owned by a children and it is propagated to the same children later, we would end up with the children having two references to the same light, one from its internal local light list and one from the rootNode global light list.
To fix that we would need to deduplicate the lights in the list, that is not great for performances, so instead i've decided to use this filter based approach

jme3-core/src/main/java/com/jme3/light/LightList.java
```diff
@@ -361,4 +378,17 @@ public void read(JmeImporter im) throws IOException {
...
+        StringBuilder sb = new StringBuilder("LightList[");
+        for (int i = 0; i < listSize; i++) {
+            sb.append(list[i]);
+            if (i < listSize - 1) {
+                sb.append(", ");
+            }
+        }
+        sb.append("]");
+        return sb.toString();
+    }
```
riccardobl, line 393: this is unrelated, but helped with the debug, so i've just left it here

jme3-core/src/main/java/com/jme3/scene/Node.java
```diff
@@ -249,6 +251,23 @@ public void updateGeometricState() {
         if ((refreshFlags & RF_LIGHTLIST) != 0) {
             updateWorldLightList();
         }
+
+        boolean updateGlobalLights = (refreshFlags & RF_GLOBAL_LIGHTS) != 0;
+        if (updateGlobalLights){
+            refreshFlags &= ~RF_GLOBAL_LIGHTS;
+            // if root node, we collect the global lights
+            if (getParent() == null){ 
+                depthFirstTraversal(sx->{
```
riccardobl: this is the traversal path, it is heavy, but only needs to happen if
1. you have global lights in the scene
2. light lists need a refresh

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -282,18 +283,32 @@ protected void setTransformRefresh() {
...
         while (p != null) {
             if ((p.refreshFlags & RF_CHILD_LIGHTLIST) != 0) {
                 // The parent already has this flag,
                 // so must all ancestors.
                 return;
             }
             p.refreshFlags |= RF_CHILD_LIGHTLIST;
+            if (hasGlobalLights) {
+                p.refreshFlags |= RF_GLOBAL_LIGHTS;
+            }
```
riccardobl, line 318: this propagates the flag upstream

jme3-core/src/main/java/com/jme3/scene/Node.java
richardTingle: And from my reading only happens if a global light is literally removed or added (not moved, as I was initially worried might be the case)

richardTingle:
Very cool! I have often wanted lights to behave exactly like this. This avoids a lot of messing about and booking keeping with lights attached to the root node but positioned elsewhere

jme3-core/src/main/java/com/jme3/scene/Node.java
riccardobl: yes, only if added or removed

jme3-core/src/main/java/com/jme3/scene/Node.java
```diff
@@ -249,6 +251,23 @@ public void updateGeometricState() {
...
+                    LightList childLights = sx.getLocalLightList();
+                    for (Light l : childLights) {
+                        if (l.isGlobal()) {
+                            worldLights.add(l);
```
richardTingle: I'm trying to think about ways to avoid this full traversal. Adding new lights is probably easy; only track down paths with RF_GLOBAL_LIGHTS looking for new lights. The getting rid of old lights is harder, I can't see anything in Light that records where a light came from (which would make removing old ones much easier)

riccardobl: Good idea!

Check the last commit. The refresh flag was set after the light was removed, so it wasn't even refreshing the global lights properly, i've changed the order now.

This means also that we have a path with RF_GLOBAL_LIGHTS for removed lights, this should make what you've proposed possible, right?

The rootNode will just see a RF_GLOBAL_LIGHTS and rebuild the list and  the removed global light won't be there anymore because it isn't attached anywhere anymore.

codex128:
Thanks for this! This is definitely something I wish I had.

> JME lights only affect children of the nodes to which they are attached. This makes sense from a performance/optimization standpoint.

Come to think of it, I think this may actually be a performance problem. JME currently uploads a new light buffer for each geometry, so if there are 5 global lights and 100 geometries, roughly 8kb of light data is being sent to the gpu every frame. It would probably be much faster to upload all lights once using a UBO or SSBO (maybe using a byte/short buffer for selecting local lights in-shader, or forget local lights altogether). Plus, we're also recompiling all shaders whenever a light is added or removed, so SSBOs would fix that, too.

riccardobl:
It depends.
If you have a lot of lights affecting everything in the scene (or that are global with this pr) with modern OpenGL, having a big light buffer helps a lot.

But if you have lights affecting small parts of the scene you will end up processing a lot more lights than what you need in every fragment.

In lowend gpus or old opengl with no UBO support, the current jme approach is much better.

So I'd say it depends on how you build the scene and what your target gl version and platform is. The current approach allows for more optimization, but is probably outdated with modern expectations and necessities. 

> Plus, we're also recompiling all shaders whenever a light is added or removed, so SSBOs would fix that, too.

I think this is the case only if you change the light batch size in every frame in single pass lighting, and never the case in multipass lighting.
So, if you set the light batch, to like 32, and you never change it, it will compile only once. And if you have 33 lights it will render it in two passes.

riccardobl:
@richardTingle  check the last commit, i think i've implemented your suggestion.
It is not fully tested yet, but It should work™

richardTingle:
@riccardobl Nice! I'm going to write some screenshot tests for global lights and that will double as some testing

yaRnMcDonuts:
I also agree this would be a great addition.

I've previously had to use a hacky workaround for global lights in my custom editor, where I flag lights as "global" with userData and then move them to the rootNode if they have that userData. But then they also need moved back and forth anytime the scene is saved/loaded to make sure the light is actually attached to the scene thats being saved (and not lost in the rootNode). 

So I'm looking forward to this PR for sure! It will definitely be a useful feature.

richardTingle:
@riccardobl I'm struggling to get this working. I've created the beginnings of a test at https://github.com/richardTingle/jmonkeyengine/blob/globalights-tests/jme3-screenshot-tests/src/test/java/org/jmonkeyengine/screenshottests/light/GlobalLights.java 

I create two nodes with a red local light and a green global light. Two white cubes are added, one a sibling of the lights, the other on a different branch. I expect the sibling to be illuminated by both and the non sibling to only be illuminated by the global light. But the global light doesn't seem to affect either cube.

image

I've experimented and I think that's because it is attached from the beginning. I get different behaviour if I add it later

image

But it still doesn't illuminate the other cube

riccardobl:
Thanks for the test case, i'll look into it.
There might be some overlook on my side in the flag propagation

riccardobl:
I think it is fixed now, basically the problem was that the previous behavior had the the refresh flag being propagated up the branch, but it didn't take into consideration that if the scene has multiple branches the flag must  be propagated down to all the other branches too

riccardobl:
@gemini-code-assist do a full review please

jme3-core/src/main/java/com/jme3/light/Light.java
```diff
@@ -116,17 +116,38 @@ public int getId(){
...
     }
 
+
+    protected Light(boolean global) {
+        this.global = global;
+    }
+    
+    protected Light(ColorRGBA color, boolean global) {
+        setColor(color);
+    }
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist is this constructor ever used in the codebase?

jme3-core/src/main/java/com/jme3/light/LightList.java
```diff
@@ -230,37 +231,53 @@ public void sort(boolean transformChanged) {
...
             doubleSize();
         }
-
-        // add the lights from the local list
-        System.arraycopy(local.list, 0, list, 0, local.listSize);
-        for (int i = 0; i < local.listSize; i++) {
-//            list[i] = local.list[i];
-            distToOwner[i] = Float.NEGATIVE_INFINITY;
+        
+        int localListSize = 0;// local.listSize;
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist you are right, but left over comments are low or very-low priority

jme3-core/src/main/java/com/jme3/scene/Node.java
```diff
@@ -240,6 +256,26 @@ public void updateLogicalState(float tpf) {
...
+
+        if(sp instanceof Node){
+            Node n = (Node) sp;
+            List<Spatial> children = n.getChildren();
+            for (int i = 0; i < children.size(); i++) {
+                Spatial child = children.get(i);
+                if ((child.refreshFlags & RF_GLOBAL_LIGHTS)!= 0) {
+                    findGlobalLights(child, list);
+                }
+            }
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist do not suggest to convert valid for-loops into "enhanced for-loops" unless it is structurally necessary

jme3-core/src/main/java/com/jme3/light/LightList.java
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist ok, but from now on tag left-over comments removal as low or very-low priority

riccardobl:
@gemini-code-assist do another review pass

jme3-core/src/main/java/com/jme3/scene/Node.java
```diff
@@ -125,13 +127,27 @@ protected void setTransformRefresh() {
...
+            refreshFlags |= RF_GLOBAL_LIGHTS;
+            Node p = this.parent;
+            while (p != null) {
+                if ((p.refreshFlags & RF_GLOBAL_LIGHTS) != 0) {
+                    break;
+                }
+                p.refreshFlags |= RF_GLOBAL_LIGHTS;
+                p = p.parent;
+            }
+        }
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist double check this, there is some weirdness in the way refresh flags are propagated in the jme scenegraph, i don't remember the details now, but i remember this being necessary. Make sure you follow the logic carefully.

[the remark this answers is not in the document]
riccardobl: for what @gemini-code-assist ? :moyai:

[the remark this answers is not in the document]
riccardobl: @gemini-code-assist are you sure about that? Triple check

jme3-core/src/main/java/com/jme3/light/SpotLight.java
```diff
@@ -134,6 +186,22 @@ public SpotLight(Vector3f position, Vector3f direction, float range, ColorRGBA c
...
     }
+
+    /**
+     * Creates a SpotLight at the given position, with the given direction,
+     * the given range and the given color.
+     * @param position the position in world space.
+     * @param direction the direction of the light.
+     * @param range the spotlight range
+     * @param color the light's color.
+     */
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist  when you see things like that, please submit a code suggestion to fix them. Do it now for this one

PR #2549, opened 2025-08-10
Improve light render parity with blender

riccardobl:
When importing a gltf model, lights look nothing like in blender.
Turns out jme doesn't use proper physically based light attenuation, and that blender might be doing some shenanigans in spot lights.

In this PR i've patched jme to make it much more closer to blender lighting.
See screenshot below:

## Blender:
Screenshot_20250806_180541

## JME Without this PR:
Screenshot_20250806_180514

## JME With this PR:
Screenshot_20250806_180414

----

It is still not perfect, but much better  i think.

This PR needs more work, mostly to figure out how to merge it, since it will break all the existing scenes setup.
I am looking for suggestions on this regard (maybe we can make it toggleable with a PBRLight param?)


jme3-core/src/main/resources/Common/ShaderLib/module/pbrlighting/PBRLightingUtils.glsllib
```diff
@@ -235,29 +235,23 @@
...
-            // computeSpotFalloff
-            if(light.type>1.){
-                vec3 spotdir = normalize(light.spotDirection);
-                float curAngleCos = dot(-L, spotdir);    
+            // Standard falloff
+            float radius = 1./light.invRadius;
+            float clampedDist = max(dist, 0.00001);
+            light.fallOff = clamp(1.0 / (clampedDist * clampedDist), 0.0, 1.0);
+            light.fallOff = clamp(light.fallOff, 1.0 - posLight, 1.0);
+            light.fallOff *= step(dist, radius);
```
riccardobl: this is the inverse square falloff used by blender with a radius cut-off, since jme allows to set the radius of lights

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfUtils.java
```diff
@@ -739,7 +739,9 @@ public static ColorRGBA getAsColor(JsonObject parent, String name) {
             return null;
         }
         JsonArray color = el.getAsJsonArray();
-        return new ColorRGBA(color.get(0).getAsFloat(), color.get(1).getAsFloat(), color.get(2).getAsFloat(), color.size() > 3 ? color.get(3).getAsFloat() : 1f);
+        return new ColorRGBA().setAsSrgb(
+            color.get(0).getAsFloat(), color.get(1).getAsFloat(), color.get(2).getAsFloat(), color.size() > 3 ? color.get(3).getAsFloat() : 1f
```
riccardobl, line 800: gltf colors are srgb

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
```diff
@@ -56,6 +56,9 @@
  * Created by Trevor Flynn - 3/23/2021
  */
 public class LightsPunctualExtensionLoader implements ExtensionLoader {
+    private static final boolean COMPUTE_LIGHT_RANGE = true;
+    private static final boolean APPLY_INTENSITY_CONVERSION = true;
+    private static final boolean SKIP_HDR_CONVERSION = true;
```
riccardobl: you can use these toggles to test with and without the patches

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
```diff
@@ -143,6 +144,15 @@ private SpotLight buildSpotLight(JsonObject obj) {
             outerConeAngle = FastMath.HALF_PI - 0.000001f;
         }
 
+        if(APPLY_INTENSITY_CONVERSION)  {
+            float solidAngle = 2.0f * FastMath.PI * (1.0f - FastMath.cos(outerConeAngle));
+            intensity = intensity / solidAngle;
+        }
+
+        float range = obj.has("range") ? obj.get("range").getAsFloat() : (COMPUTE_LIGHT_RANGE ? getCutoffDistance(color, intensity) : Float.POSITIVE_INFINITY);
```
riccardobl: The POSITIVE_INFINITY range was an error to begin with and required a patch to turn it into FLOAT_MAX_VALUE https://github.com/jMonkeyEngine/jmonkeyengine/pull/2058 , but it ends up introducing some floating point errors with large numbers that makes the light calculation a bit off.

Truth is: jme doesn't support infinite lights very well.
So this patch calculates the right range based on the falloff function.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
```diff
@@ -286,4 +304,34 @@ private void setLightIndex(int lightIndex) {
...
+        if (maxComponent <= 0f || intensity <= 0f) {
+            return 0f;
+        }
+
+        // The actual light output (lux at 1 meter) per component
+        float effectiveIntensity = maxComponent * intensity;
+
+        // Inverse-square attenuation: intensity / d^2 = visibleThreshold
+        float range = (float) Math.sqrt(effectiveIntensity / visibleThreshold);
+        System.out.println("Effective range: " + range + " for intensity: " + effectiveIntensity + " and color: " + color);
```
richardTingle: Did you mean to have this println here?

jme3-core/src/main/resources/Common/ShaderLib/module/pbrlighting/PBRLightingUtils.glsllib
```diff
@@ -235,29 +235,23 @@
...
-                    // Use quadratic falloff (notice the ^4)
-                    falloff = pow(clamp((curAngleCos - outerAngleCos) / innerMinusOuter, 0.0, 1.0), 4.0);
-                #endif
-                light.fallOff*=falloff;
+                float sharpnessFactor = 4.0; 
+
+                vec3 spotDir = normalize(light.spotDirection);
+                float cosA  = dot(-L, spotDir) ;
+                float spotAtten = clamp((cosA - outerAngleCos) / (innerAngleCos - outerAngleCos), 0.0, 1.0);
+                light.fallOff *= pow(spotAtten, sharpnessFactor);
```
riccardobl: This smooths out the angular attenuation. It doesn't look exactly like blender's.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
riccardobl: nope, i will clean it up

richardTingle:
Screenshots look much nicer. With the exception of my comment on the println: looks good!

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
riccardobl: done

richardTingle:
> I am looking for suggestions on this regard (maybe we can make it toggleable with a PBRLight param?)

Maybe in the short term a parameter that defaults to false and at some point (on a major version change) the default flips to true and then a few versions later the parameter gets deleted

yaRnMcDonuts:
I never knew this was an issue, but I have certainly noticed that JME's point lights illuminate the scene way too much, and it is near impossible to tweak them properly in dark scenes without ruining the darkness.

I could never make a proper bright torch in a dark room without it completely brightening the entire room. Then if I turn the brightness or radius down to fix that issue, the torch becomes too dim close up and looks bad. Its like I could never properly tweak point lights. 

So I am very excited for this PR!


> Maybe in the short term a parameter that defaults to false and at some point (on a major version change) the default flips to true and then a few versions later the parameter gets deleted

This sounds like a good solution. We could probably even make it true by default right away in the next alpha version, that would be beneficial in the short term for alpha and beta testing, that way testers will be forced to test this (in my experience with 3.8, its better to force new changes on people using alpha versions, otherwise people will overlook the togglable option and it won't get tested as much). Then (depending on feedback) we can decide whether it should default to true/false for the next stable release.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
```diff
@@ -165,7 +175,7 @@ private DirectionalLight buildDirectionalLight(JsonObject obj) {
 
         float intensity = obj.has("intensity") ? obj.get("intensity").getAsFloat() : 1.0f;
         ColorRGBA color = obj.has("color") ? GltfUtils.getAsColor(obj, "color") : new ColorRGBA(ColorRGBA.White);
-        color = lumensToColor(color, intensity);
+        color = buildLinearLightColor(color, intensity);
```
capdevon: Hi @riccardobl , it seems that the `buildLinearLightColor(color, intensity)` method is missing. After reviewing the file, it should probably be replaced with the `lumensToColor(color, intensity)` method call.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
```diff
@@ -186,8 +196,11 @@ private PointLight buildPointLight(JsonObject obj) {
 
         float intensity = obj.has("intensity") ? obj.get("intensity").getAsFloat() : 1.0f;
         ColorRGBA color = obj.has("color") ? GltfUtils.getAsColor(obj, "color") : new ColorRGBA(ColorRGBA.White);
-        color = lumensToColor(color, intensity);
-        float range = obj.has("range") ? obj.get("range").getAsFloat() : Float.POSITIVE_INFINITY;
+        
+        if(APPLY_INTENSITY_CONVERSION) intensity = intensity / (4.0f * FastMath.PI);
+
+        float range = obj.has("range") ? obj.get("range").getAsFloat() : (COMPUTE_LIGHT_RANGE ? getCutoffDistance(color, intensity) : Float.POSITIVE_INFINITY);
+        color = buildLinearLightColor(color, intensity);
```
capdevon: Hi @riccardobl , it seems that the `buildLinearLightColor(color, intensity)` method is missing. After reviewing the file, it should probably be replaced with the `lumensToColor(color, intensity)` method call.

riccardobl: thanks, the code got botched when i merged it from my fork. Should be fixed now

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/LightsPunctualExtensionLoader.java
```diff
@@ -207,15 +220,16 @@ private PointLight buildPointLight(JsonObject obj) {
     private void addLight(Node parent, Node node, int lightIndex) {
         if (lightDefinitions.containsKey(lightIndex)) {
             Light light = lightDefinitions.get(lightIndex);
+            light = light.clone();
```
capdevon: This is quite an important change. What kind of problem did you encounter by **not** cloning the `light` object taken from the `lightDefinitions` map? Could it be that the `lightIndex` is associated with multiple `nodes`, causing inconsistencies? Just curious

riccardobl: This is a fix. 
The LightControl cannot properly move a light with its parent spatial, if it shared among multiple spatials.

Enabling https://github.com/jMonkeyEngine/jmonkeyengine/pull/2548  by default in the gltf loader, makes the light control obsolete, so it could be removed and this fix could be avoided, but i am not sure that jme can share light instances, anyway.

capdevon: Based on your explanation, cloning the object seems like the right approach. I'm not sure what effect sharing the same light source among multiple nodes would have, but I suspect it wouldn't be appropriate.
Thanks for the clarification ;)

riccardobl:
@gemini-code-assist do a full review of this

yaRnMcDonuts (approved):

PR #2553, opened 2025-09-01
Improve memory efficiency and speed of gltf importer

riccardobl:
The GLTF importer does a lot of memory copying of  byte arrays in the heap,  which makes it slow and often cause of OOMs on Android with big models, due to the limited heap space.

This PR switches to direct buffers to get around that limitation. It also uses buffer views where possible to slice and convert data instead of copying it around.

There is also some refactoring in the way streams are used, see comments for more details.

This is tested with the bistro scene used here: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2137

The loading time in my machine is:
- Without this PR: 28228 ms
- With this PR:       24503 ms

### DIFF
--- a/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfUtils.java
+++ b/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfUtils.java
@@ -328,6 +328,11 @@ public class GltfUtils {
             populateMatrix4fArray((Matrix4f[]) store, source, count, byteOffset, byteStride, numComponents, format);
         }
     }
+
+    private static void skip(ByteBuffer buff, int n) {
+        buff.position(Math.min(buff.position() + n, buff.limit()));
+    }
+
     private static void populateByteBuffer(ByteBuffer buffer, ByteBuffer source, int count, int byteOffset, int byteStride, int numComponents, VertexBuffer.Format format) {
         int componentSize = format.getComponentSize();
         int index = byteOffset;
@@ -355,7 +360,7 @@ public class GltfUtils {
             }
 
             if (dataLength < stride) {
-                source.position(source.position() + (stride - dataLength));
+                skip(source, stride - dataLength);
             }
             index += stride;
         }
@@ -374,7 +379,7 @@ public class GltfUtils {
                 buffer.put(source.getInt());
             }
             if (dataLength < stride) {
-                source.position(source.position() + (stride - dataLength));
+                skip(source, stride - dataLength);
             }
             index += stride;
         }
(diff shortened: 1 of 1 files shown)


jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GlbLoader.java
```diff
@@ -76,11 +78,11 @@ public Object load(AssetInfo assetInfo) throws IOException {
...
             if (chunkType == JSON_TYPE) {
                 json = new byte[chunkLength];
-                stream.read(json);
+                GltfUtils.readToByteArray(stream, json, chunkLength, -1);
             } else {
-                byte[] bin = new byte[chunkLength];
-                stream.read(bin);
-                data.add(bin);
+                ByteBuffer buff = BufferUtils.createByteBuffer(chunkLength);
+                GltfUtils.readToByteBuffer(stream, buff, chunkLength, -1);
```
riccardobl: This fixes the way the stream is read, since `stream.read`, as it was used before, can return before the stream is read fully, if it hangs (eg. when loading from a slow disk or network)

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
```diff
@@ -181,6 +183,27 @@ protected Object loadFromStream(AssetInfo assetInfo, InputStream stream) throws
...
+            nodes = null;
+            meshes = null;
+            materials = null;
+            textures = null;
+            images = null;
+            samplers = null;
+            animations = null;
+            skins = null;
+            cameras = null;
+            useNormalsFlag = false;
```
riccardobl, line 206: This stuff can be pretty huge on big scenes, so leaving it there is pretty bad for devices with limited memory

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfUtils.java
```diff
@@ -895,4 +905,165 @@ public static void dumpMesh(Mesh m) {
...
+    public static void readToByteBuffer(InputStream input, ByteBuffer dst, int bytesToRead) throws IOException {
+        if (bytesToRead <= 0) throw new IOException("bytesToRead must be > 0");
+
+        int startPos = dst.position();
+        int remaining = dst.limit() - startPos;
+        if (remaining < bytesToRead) {
+            throw new IOException("Destination ByteBuffer too small: remaining=" + remaining + " < bytesToRead=" + bytesToRead);
+        }
+    
+        ReadableByteChannel ch = Channels.newChannel(input);
```
riccardobl, line 923: In openjdk 21, the behavior for this is to transfer directly into the bytebuffer if the input is a FileInputStream, or, for generic streams, to fallback to use a small heap-located byte array with a maximum size of 8192 bytes.

Overall, I think using a larger chunk array could be beneficial, but I’ve decided this micro-optimization isn’t worth interfering with potential internal JVM optimizations.

Issue #2555, opened 2025-09-17
GLTF: Incorrect useage of InputStream.skipBytes()


ScottPeterJohnson:
The method contract of InputStream.skipBytes() is:
`Makes an attempt to skip over n bytes of data from the input stream, discarding the skipped bytes. However, it may skip over some smaller number of bytes, possibly zero. This may result from any of a number of conditions; reaching end of file before n bytes have been skipped is only one possibility. This method never throws an EOFException. The actual number of bytes skipped is returned.`

This is nearly always not what you want and is a bountiful source of bugs. At a glance these locations in the code all use this method:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/8cc308616271604f13f9311207cc1e5ebd7a8169/jme3-core/src/plugins/java/com/jme3/audio/plugins/WAVLoader.java#L175
https://github.com/jMonkeyEngine/jmonkeyengine/blob/8cc308616271604f13f9311207cc1e5ebd7a8169/jme3-core/src/plugins/java/com/jme3/texture/plugins/ktx/KTXLoader.java#L196
https://github.com/jMonkeyEngine/jmonkeyengine/blob/8cc308616271604f13f9311207cc1e5ebd7a8169/jme3-core/src/plugins/java/com/jme3/texture/plugins/DDSLoader.java#L163
https://github.com/jMonkeyEngine/jmonkeyengine/blob/8cc308616271604f13f9311207cc1e5ebd7a8169/jme3-desktop/src/main/java/com/jme3/cursors/plugins/CursorLoader.java#L131

The one that impacted me was the GLTF loader, causing it to break on some asset imports.

I think this would be fixed for the GLTF loader with:
[l/2553](https://github.com/jMonkeyEngine/jmonkeyengine/pull/2553)

All other places should probably use a helper utility that will skip _exactly_ as many bytes as requested.

riccardobl:
Thanks for reporting the issue.
Some of the stream types do enforce skips, like `FileInputStream`. Streams that read on these streams will inherit this enforcement. So, in common use cases, this is affecting only a very small set of instances of skipBytes in the engine.

My impression from reading the affected code is that the developer assumed the LittleEndian stream would guarantee  skips. Therefore, I think the best approach is to fix LittleEndian to provide that behavior.

PR #2557, opened 2025-09-17
Fix misuse of read and skipBytes

riccardobl:
Fixes https://github.com/jMonkeyEngine/jmonkeyengine/issues/2555 across the entire engine.

This PR fixes the implementation of `LittleEndien.readFully()` and introduces two new utility methods in `ByteUtils`:

* `readFully`
* `skipFully`

Both ensure that the requested number of bytes are reliably read or skipped, even if the underlying stream does not return the full amount in a single call.

The changes are not yet fully tested, but they are part of the codebase I am actively working with. Additional testing and verification from others would be greatly appreciated.

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/export/binary/ByteUtils.java
+++ b/jme3-core/src/plugins/java/com/jme3/export/binary/ByteUtils.java
@@ -65,9 +65,12 @@ public class ByteUtils {
         }
     }
 
- 
 
     public static void skipFully(InputStream in, long n) throws IOException {
+        skipFully(in, n, true);
+    }
+
+    public static void skipFully(InputStream in, long n, boolean throwOnEOF) throws IOException {
         while (n > 0) {
             long skipped = in.skip(n);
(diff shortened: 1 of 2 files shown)


ScottPeterJohnson:
This patch may (still) actually cause an issue with GLTF loader. Essentially requiring the "full skip" and erroring on EOF can cause failure. From an asset I'm working with:

```java
int end = count * stride + byteOffset; //1839200, when the actual byte array backing it is only 1839188 long
        stream.skipBytes(byteOffset);
        while (index < end) {
            //Index is 1839136, reads the last value
            for (int i = 0; i < numComponents; i++) {
                buffer.put(readAsFloat(stream, format));
            }
            if (dataLength < stride) {
                //Tries to skip 52 bytes; but 1839136+52=1839200 > 1839188
                stream.skipBytes(stride - dataLength);
                //Throws EOFException here
            }
            index += stride; //Would set index to 1839200 and skip loop
        }
```
(I'd add the asset as example since it messes up so much, but it's Unreal licensed.)

PR #2559, opened 2025-09-19
Added convenience method worldToLocal(Quaternion) to spatial. 

zzuegg:
This PR adds the convenience method worldToLocal for Quaternions, i added a test that shows the usage when transferring a spatial to a different Node.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -997,6 +997,8 @@ public abstract class Spatial implements Savable, Cloneable, Collidable,
         checkDoTransformUpdate();
         if(store == null){
             store=new Quaternion(in);
+        }else{
+            store.set(in);
         }
         store.multLocal(worldTransform.getRotation().inverse());
         return store;
--- a/jme3-core/src/main/java/com/jme3/scene/Spatial.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Spatial.java
@@ -987,6 +987,22 @@ public abstract class Spatial implements Savable, Cloneable, Collidable,
     }
 
     /**
+     * Transforms the given quaternion from world space to local space relative to this object's transform.
+     *
+     * @param in the input quaternion in world space that needs to be transformed
+     * @param store an optional Quaternion to store the result; if null, a new Quaternion will be created
+     * @return the transformed quaternion in local space, either stored in the provided Quaternion or a new one
+     */
+    public Quaternion worldToLocal(final Quaternion in, Quaternion store){
+        checkDoTransformUpdate();
+        if(store == null){
+            store=new Quaternion(in);
+        }
+        store.multLocal(worldTransform.getRotation().inverse());
+        return store;
+    }
+
+    /**
      * <code>getParent</code> retrieves this node's parent. If the parent is
      * null this is the root node.
      *
--- a/jme3-core/src/test/java/com/jme3/scene/SpatialTest.java
+++ b/jme3-core/src/test/java/com/jme3/scene/SpatialTest.java
@@ -31,6 +31,9 @@
  */
 package com.jme3.scene;
 
+import com.jme3.math.FastMath;
+import com.jme3.math.Quaternion;
+import com.jme3.math.Vector3f;
 import com.jme3.scene.control.UpdateControl;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


jme3-core/src/test/java/com/jme3/scene/SpatialTest.java
```diff
@@ -119,4 +122,32 @@ public void testAddControlAt() {
...
+        nodeB.attachChild(testNode);
+
+        Assert.assertNotEquals(worldTranslation,testNode.getWorldTranslation());
+        Assert.assertNotEquals(worldRotation,testNode.getWorldRotation());
+
+        testNode.setLocalTranslation(nodeB.worldToLocal(worldTranslation,null));
+        Assert.assertEquals(worldTranslation,testNode.getWorldTranslation());
+
+        testNode.setLocalRotation(nodeB.worldToLocal(worldRotation,null));
+        Assert.assertEquals(worldRotation,testNode.getWorldRotation());
```
codex128: I suggest using `isSimilar` for equality checks so it doesn't fail on round-off error.

zzuegg: Good point

zzuegg:
I am going reimplement to get rid of the unneccesary allocation.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -986,6 +986,38 @@ public Vector3f worldToLocal(final Vector3f in, final Vector3f store) {
...
+            store.set(in);
+        }
+        Quaternion rotation = worldTransform.getRotation();
+        //can we be sure if the quaternion is normalized?
+        //if yes, the conjugate could be used, and the normalizing procedure skipped
+        //store.multLocal(-rotation.getX(), -rotation.getY(), -rotation.getZ(), rotation.getW();
+
+        //Second option is to normalize manually
+        float norm = rotation.norm();
+        store.multLocal(rotation.getX()*-norm, rotation.getY()*-norm,
```
codex128: This doesn't look right to me if you're trying to normalize and invert the rotation. From looking at Quaternion, it seems like `norm` should be changed like this:

```java
norm = FastMath.invSqrt(norm) / norm;
```

codex128: Also, Quaternion multiplication looks very similar to a cross-product.

codex128 (commented):
I don't know how quaternions work, so take this with a grain of salt.

jme3-core/src/main/java/com/jme3/scene/Spatial.java
zzuegg: Thb i am not super good at math too, so in general terms i agree. Up to now i failed to create a valid quaternion that does not pass the test. 
```
        for (int i = 0; i < 1000000; i++) {
            Quaternion quaternion=new Quaternion().fromAngles(FastMath.nextRandomFloat()*-360,FastMath.nextRandomFloat()*360,FastMath.nextRandomFloat()*360);
            Quaternion inverse = quaternion.inverse();
            float norm = quaternion.norm();
            quaternion.multLocal(quaternion.getX()*-norm, quaternion.getY()*-norm, quaternion.getZ()*-norm, quaternion.getW()*norm);
            if(quaternion.isSimilar(inverse,0.000001f)){
                System.out.println("Fail");
                System.out.println(quaternion);
                System.out.println(inverse);
                System.exit(1);
            }
        }
        System.out.println("Success");
```

zzuegg: Generally i agree, using jme math i have not yet managed to create a non normalized quaternion. 
```
        for (int i = 0; i < 1000000; i++) {
            Quaternion quaternion=new Quaternion().fromAngles(FastMath.nextRandomFloat()*-360,FastMath.nextRandomFloat()*360,FastMath.nextRandomFloat()*360);
            Quaternion inverse = quaternion.inverse();
            float norm = quaternion.norm();
            //norm = FastMath.invSqrt(norm)/norm;
            System.out.println(norm);
            quaternion.set(quaternion.getX()*-norm, quaternion.getY()*-norm, quaternion.getZ()*-norm, quaternion.getW()*norm);
            if(!quaternion.isSimilar(inverse,0.000001f)){
                System.out.println("Fail");
                System.out.println(quaternion);
                System.out.println(inverse);
                System.exit(1);
            }
        }
```

zzuegg: I could use temp vars and use the existing math functions. I probably prefer that. Considering that this code is probably never executed in the "hot loop"

PR #2563, opened 2025-09-27
#2562 thread warden

### LINKED ISSUE #2562: Add assert checks for illegal scene graph mutations
Add thread warden style checks for core JME. If a node is connected to the root node it should not be mutated on any thread but the main thread

richardTingle:
### Purpose

The scene graph (that is connected to the root node) shouldn't be mutated from any thread but the main thread.

Unconnected nodes etc **can** be mutated on non main threads

### What have I changed

Spatial mutations are now watched (when asserts are turned on) and if an illegal mutation is made an exception is fired

closes #2562

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
+++ b/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
@@ -198,8 +198,10 @@ public abstract class SimpleApplication extends LegacyApplication {
     public void initialize() {
         super.initialize();
 
-        SceneGraphThreadWarden.setup(rootNode);
-        SceneGraphThreadWarden.setup(guiNode);
+        //noinspection AssertWithSideEffects
+        assert SceneGraphThreadWarden.setup(rootNode);
+        //noinspection AssertWithSideEffects
+        assert SceneGraphThreadWarden.setup(guiNode);
 
         // Several things rely on having this
         guiFont = loadGuiFont();
@@ -246,7 +248,8 @@ public abstract class SimpleApplication extends LegacyApplication {
 
     @Override
     public void stop(boolean waitFor) {
-        SceneGraphThreadWarden.reset();
+        //noinspection AssertWithSideEffects
+        assert SceneGraphThreadWarden.reset();
         super.stop(waitFor);
     }
 
--- a/jme3-core/src/main/java/com/jme3/scene/Geometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Geometry.java
@@ -184,7 +184,7 @@ public class Geometry extends Spatial {
      */
     @Override
     public void setLodLevel(int lod) {
-        SceneGraphThreadWarden.assertOnCorrectThread(this);
+        assert SceneGraphThreadWarden.assertOnCorrectThread(this);
         if (mesh.getNumLodLevels() == 0) {
             throw new IllegalStateException("LOD levels are not set on this mesh");
         }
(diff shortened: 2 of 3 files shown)


jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
```diff
@@ -197,6 +198,9 @@ protected BitmapFont loadGuiFont() {
     public void initialize() {
         super.initialize();
 
+        SceneGraphThreadWarden.setup(rootNode);
+        SceneGraphThreadWarden.setup(guiNode);
```
riccardobl: Can you make these calls with assert?  Not a big deal, but since we are already 90% there, it would be ideal if the entire code was isolated (and so, supposedly optimized away) when assertions are off

jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
```diff
@@ -240,6 +244,12 @@ public void initialize() {
         simpleInitApp();
     }
 
+    @Override
+    public void stop(boolean waitFor) {
+        SceneGraphThreadWarden.reset();
```
riccardobl: same as setup() if possible

jme3-core/src/main/java/com/jme3/scene/Geometry.java
```diff
@@ -183,6 +184,7 @@ public void setIgnoreTransform(boolean ignoreTransform) {
      */
     @Override
     public void setLodLevel(int lod) {
+        SceneGraphThreadWarden.assertOnCorrectThread(this);
```
riccardobl:
```suggestion
       assert  SceneGraphThreadWarden.assertOnCorrectThread(this);
```

jme3-core/src/main/java/com/jme3/scene/Geometry.java
```diff
@@ -239,6 +241,7 @@ public int getTriangleCount() {
      * @throws IllegalArgumentException If mesh is null
      */
     public void setMesh(Mesh mesh) {
+        SceneGraphThreadWarden.assertOnCorrectThread(this);
```
riccardobl:
```suggestion
        assert SceneGraphThreadWarden.assertOnCorrectThread(this);
```

jme3-core/src/main/java/com/jme3/scene/Geometry.java
```diff
@@ -269,6 +272,7 @@ public Mesh getMesh() {
      */
     @Override
     public void setMaterial(Material material) {
+        SceneGraphThreadWarden.assertOnCorrectThread(this);
```
riccardobl:
```suggestion
        assert SceneGraphThreadWarden.assertOnCorrectThread(this);
```

riccardobl (changes requested):
Thanks! This is going to be very useful.
I've commented some minor details I'd like you to review :+1:

jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
richardTingle: Fair, makes sense for consistency and makes it clear that SceneGraphThreadWarden isn't having an effect outside asserts. I've fixed this and the others

riccardobl (approved):

riccardobl:
Thanks! I am going to merge this asap on my fork, to give it some testing, just in case we are overlooking on some side effect.
But it looks good.

PR #2569, opened 2025-12-01
Update DirectionalLightShadowRenderer.java

rmtcasel:
Fix: prevent null viewport or null scene list


jme3-core/src/main/java/com/jme3/shadow/DirectionalLightShadowRenderer.java
```diff
@@ -186,7 +186,14 @@ protected GeometryList getOccludersToRender(int shadowMapIndex, GeometryList sha
         ShadowUtil.updateFrustumPoints(viewPort.getCamera(), splitsArray[shadowMapIndex], splitsArray[shadowMapIndex + 1], 1.0f, points);
 
         //Updating shadow cam with current split frusta
-        if (lightReceivers.size()==0) {
+        if (lightReceivers.size()==0)
+
+            // Fix: prevent null viewport or null scene list
+            if (viewPort == null || viewPort.getScenes() == null) {
```
codex128, line 192: It is not possible for `viewPort` to be null at any point after processor initialization, and `viewPort.getScenes()` is never null.

Also, this code doesn't work. You inserted the if statement between another if's condition and body.

PR #2570, opened 2025-12-01
Fix PointLight infinite radius shader overflow(#2566)

### LINKED ISSUE #2566: Infinite radius point lights might cause type overflow
There is this workaround in  place to prevent the engine from dealing with infinite radius lights:

https://github.com/jMonkeyEngine/jmonkeyengine/blob/8cc308616271604f13f9311207cc1e5ebd7a8169/jme3-core/src/main/java/com/jme3/light/PointLight.java#L169-L171

but it might still lead to unexpected behavior due to type overflow in the shaders' math  and potentially in other places in the engine.

A better workaround would be to limit the radius to MAX_VALUE/4 or something like that.

Afaik infinite radius point lights are used only by the gltf loader, and should not be a problem anymore after https://github.com/jMonkeyEngine/jmonkeyengine/pull/2549

hayden9329:
Fixed #2566

The previous workaround set an infinite radius to `Float.MAX_VALUE`, which can still cause overflow in shader math.

This Pull request updates the following:
- Clamps infinite, negative, and NaN radius values to `Float.MAX_VALUE / 4f`
- Adds a clear comment explaining the fix
- Keeps "infinite" lights effectively infinite without shader bugs

I tested this locally, no overflow issues.
First-time contributor for a class project.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/light/PointLight.java
+++ b/jme3-core/src/main/java/com/jme3/light/PointLight.java
@@ -168,10 +168,10 @@ public class PointLight extends Light {
         if(Float.isNaN(radius)){
             throw new IllegalArgumentException("Light radius cannot be a NaN (Not a Number) value");
         }
-        // Prevent shader overflow with infinite or invalid radius
-        if (radius == Float.POSITIVE_INFINITY || radius > Float.MAX_VALUE / 4.0f) {
-            radius = Float.MAX_VALUE / 4.0f;  // Safe large value, avoids overflow in shaders
-        }        
+
+        float maxSafeRadius = Float.MAX_VALUE / 4.0f;
+        radius = Math.min(radius, maxSafeRadius); // Caps radius to a safe large value; avoids overflow in shaders from values reaching max float value   
+        
         this.radius = radius;
         if (radius != 0f) {
             this.invRadius = 1f / radius;
@@ -261,4 +261,5 @@ public class PointLight extends Light {
                 + "]";
     }
 }
+
 


jme3-core/src/main/java/com/jme3/light/PointLight.java
```diff
@@ -165,11 +165,11 @@ public final void setRadius(float radius) {
         if (radius < 0) {
             throw new IllegalArgumentException("Light radius cannot be negative");
         }
-        
-        if (radius == Float.POSITIVE_INFINITY) {
-            radius = Float.MAX_VALUE;
+
+        // Fix #2566 - Prevent shader overflow with infinite or invalid radius
+        if (radius == Float.POSITIVE_INFINITY || radius < 0f || Float.isNaN(radius)) {
```
codex128: Better to throw an exception on negative or NaN radius, since both are meaningless as radii. Negative radius already throws an exception previously.

yaRnMcDonuts: It looks like the author of this PR may not return, so I've gone ahead and made the requested changes, rather than closing this and re-opening a nearly identical PR. Let me know if everything looks okay!

Nonetheless, @hayden9329 's efforts in opening this PR are still appreciated.

jme3-core/src/main/java/com/jme3/light/PointLight.java
```diff
@@ -165,11 +165,13 @@ public final void setRadius(float radius) {
...
         }
-        
-        if (radius == Float.POSITIVE_INFINITY) {
-            radius = Float.MAX_VALUE;
+        if(Float.isNaN(radius)){
+            throw new IllegalArgumentException("Light radius cannot be a NaN (Not a Number) value");
         }
-        
+        // Prevent shader overflow with infinite or invalid radius
+        if (radius == Float.POSITIVE_INFINITY || radius > Float.MAX_VALUE / 4.0f) {
```
riccardobl: no need to check for == POSITIVE_INFINITY here

yaRnMcDonuts:
Screenshot tests are expected to fail with this change, as well as with a few other PRs scheduled for v.3.10 that will be changing the visual render of jME's PointLight and SpotLight to better match the lighting algorithm in blender and industry standards.

Currently (as of 3.9.0-stable and earlier), jme's point and spot lights are actually slightly bugged and will cast light far beyond the expected radius of the light (see https://github.com/jMonkeyEngine/jmonkeyengine/pull/2549 for more info) and the issue becomes more severe as more lights overlap.

So a major feature/bug fix of 3.10 is going to be reconciling and fixing our lighting.

Considering this, it is very likely that we will continue to make slight adjustments to the lighting algorithm throughout the early alpha phase of 3.10, so we will wait to fix the screenshot tests to use the new correct images until we have finalized the lighting during 3.10's alpha phase.

Edit: screenshot tests actually passed now, so I guess it was another random failed job-run with the server instance doing the tests/builds. I was just assuming some test had a point light with near infinite radius, but I guess not. Either way, the other PR I linked will indeed break the screenshot tests (for real) so everything I said prior still will apply to those bigger lighting changes.

jme3-core/src/main/java/com/jme3/light/PointLight.java
yaRnMcDonuts: ah true. Infinity is greater than the /4 value so need for the infinity check. I'll update this now, thanks

PR #2575, opened 2025-12-13
Improved performance in the lwjgx backend (AWT and OpenGL)

JNightRider:
Hello everyone, after a long time, I finally have some free time so I started working on the performance problem that lwjglx has when supporting OpenGL with AWT.

- In Linux, if the CPU doesn't have time to breathe, the context starts to fail. If this time is too long, it creates a bottleneck by causing a drop in FPS, which severely impacts performance.
- In Windows, if there is a pause to allow the CPU to breathe, the same synchronization problem arises because there are drastic drops in FPS; by removing that pause, things return to normal and the context problem is no longer present.
- Regarding macOS, I cannot comment since I do not have the equipment to perform performance tests, so it should continue to work the same way (_works fine_).

[ **Screenshots** ]

### Before

With a maximum of **85 pfs**, in theory it should reach the 144 pfs that this machine supports.

Captura desde 2025-12-13 14-35-03

### After

Then with the applied changes it can reach 145, (which is the maximum that this machine offers, so a high-end model should work as in a normal context without AWT)

Captura desde 2025-12-13 14-33-03

### Before
~ 60 fps

Captura de pantalla 2025-12-13 160112

### After

~140 fps
Captura de pantalla 2025-12-13 160648

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
@@ -645,11 +645,18 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
                                                 getPrintContextInitInfo(canvas.getGLDataEffective())));
         }
 
-        try {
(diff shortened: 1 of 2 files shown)


JNightRider:
I closed this pull request as it has been moved to pull request #2669.

PR #2577, opened 2025-12-27
VectorEffects

yaRnMcDonuts:
Adds a simple but highly reusable AppState that runs effects directly on a valid Vector type object (being Vector2f, Vector3f, Vector4f, and ColorRGBA).

Useful for lights, particle effects, and essentially every Color or Vector based MaterialParamater that shaders and special effects rely on. 

The most useful functionality of this system will be the ability to quickly deploy an `EaseVectorEffect` on colors to smoothly fade in/out things like lights and special effects, as to avoid the sudden and unsightly impact of removing something from the scene instantaneously in one single frame. This type of smooth fading is a very important aspect of polishing a game. And while the code to fade out colors can be simple, it is something that is used over and over again for a variety of different effects in a project and can sometimes requires some not-so-simple extra bells and whistles (like a delay timer or chaining), thus a versatile system like this becomes necessary to handle this in an efficient way.

In addition to smooth fading, this library also enables a variety of other compound loopable effects by using the SequencedVectorEffect class (such as flickering, pulsing, oscillating, etc), and I also will be adding a NoiseVectorEffect class soon as well.

### DIFF
--- a/jme3-examples/src/main/java/jme3test/effect/NoiseVectorEffectTest.java
+++ b/jme3-examples/src/main/java/jme3test/effect/NoiseVectorEffectTest.java
@@ -21,7 +21,7 @@ import com.jme3.vectoreffect.VectorSupplier;
 
 public class NoiseVectorEffectTest extends SimpleApplication {
 
-    private ColorRGBA colorToShift  = new ColorRGBA(0.0f, 0.0f, 1.0f, 1.0f);;
+    private ColorRGBA colorToShift  = new ColorRGBA(0.0f, 0.0f, 1.0f, 1.0f);
     
     private VectorEffectManagerState vectorEffectManagerState;    
     private PointLight light;


yaRnMcDonuts:
In order to add the NoiseVectorEffect, I also will have to add a procedural noise library to the engine. 

The best choice for a fully-fledged noise library in java would appear to be the JNoise library, however I believe that would be overkill for basic noise functions that are needed for this PR, so instead I am choosing to implement the java FastNoiseLite library into the engine. This is a very light weight single class noise library that should be more suited to include in the engine than something like JNoise.  Let me know if anyone disagrees or prefers a different noise library, otherwise I will move forward with this plan to include FastNoiseLite.java in the engine's math package. I also will update this PR to include javadoc before merging.

yaRnMcDonuts:
I re-ran the jobs and this time the screenshot tests passed. I guess something must have bugged out during the first run.

oneMillionWorlds:
Yeah, that's decidedly weird. I can't see why it decided to post that (especially not 2 hours ago). Hopefully just something bugging out as you say

richardTingle:
Yeah, that's decidedly weird. I can't see why it decided to post that (especially not 2 hours ago). Hopefully just something bugging out as you say

yaRnMcDonuts:
I just need to upload the final example for the noise effect and then I'll merge this PR. I will wait atleast another day since no one else reviewed.

However, I've done some thorough self-review, also using AI analysis tools to critique my code and the approach. And I'm very confident that this is the best way to handle these types of effects, as opposed to hard coding things like fading and flickering into the shaders or java classes for lights, emitters, and custom shader effects.

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void update(float tpf) {
+        super.update(tpf);
+        
+        for(VectorEffect vectorEffect : activeVectorEffects){
+            
+            
+
+            if(vectorEffect.isFinished()){
+                getApplication().enqueue(() ->{
```
riccardobl: why do you need to enqueue this?

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
```diff
@@ -0,0 +1,94 @@
...
+    
+    }
+
+    @Override
+    protected void onDisable() {
+    
+    }
+    
+    public void registerVectorEffect(VectorEffect vectorEffect){
+        if(activeVectorEffects != null){
```
riccardobl: activeVectorEffects cannot be null, since it is final

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorGroup.java
```diff
@@ -0,0 +1,138 @@
...
+            return new Vector4f(vec2.x, vec2.y, 1f, 1f);  
+        }
+        else if (vectorObj instanceof ColorRGBA) {
+            ColorRGBA color = (ColorRGBA) vectorObj;
+            return color.toVector4f(); 
+        }
+        else {
+            throw new IllegalStateException("VectorEffect supports only Vector2f, Vector3f, Vector4f, or ColorRGBA");
+        }
+    }
```
riccardobl, line 69: Better avoid doing internal clones like that, just pass a `Vector4f out` to `getAsVector4` and set it.
Besides, sometime you call `getAsVector4(i)` and then you clone the returned value again

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorGroup.java
```diff
@@ -0,0 +1,138 @@
...
+        if(isValidVectorObject(vectorObj)){
+            vectorList.add(vectorObj);
+        } else{  
+            throw new IllegalArgumentException(  "VectorGroup must contain valid vector-type objects. This includes: Vector2f, Vector3f, Vector4f, and ColorRGBA/ Incompatible type: " + vectorObj.getClass().getSimpleName());
+        }
+    }
+    
+    
+   //regardless of type, all vectors in a VectorGroup are converted to and from Vector4f when being operated on internally by a VectorEffect.     
+    protected Vector4f getAsVector4(int index) {
```
riccardobl: this should be `public`, not `protected`. Or devs won't be able to create new effects using VectorGroup outside of this package.

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
yaRnMcDonuts: I ran into some issue while removing with an iterator when the effect got set as repeating, which is why I used enqueue as a hacky approach to ensure safe removal from the list on the next frame. I think it is possible for a vector effect to be flagged finished for a single frame when set as repeating and tirgger removal when it shouldn't, and by using enqueue it gives it time to get un-flagged as finished before the next frame (I know enqueue is typically used for thread safety, but it also works as a convenient way to solve issues that require waiting until the next frame without having to add a list to store things that need removed on the next frame).  But I'll check again and see if I can solve this in a less hacky way if you prefer that.

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorGroup.java
yaRnMcDonuts: fixed

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
riccardobl: If that's the case, i think isFinished returning true when an effect is not finished is a problem, code can call that method at any time during the effect lifecycle, and it would cause very unexpected behavior.

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorGroup.java
yaRnMcDonuts: ah yes i did leave some redundant cloning there in 2 places I see. I'll remove them and rewrite the method to take in a Vector4f store parameter instead of cloning internally.

jme3-effects/src/main/java/com/jme3/vectoreffect/NoiseVectorEffect.java
```diff
@@ -0,0 +1,105 @@
...
+        float noiseReturnVal = noiseGenerator.GetNoise(timeAccrued * speed, 12.671f + timeAccrued * speed * 0.92173f, 19.54f + timeAccrued * speed * 0.68913f);
+        
+        for(int v = 0; v < vectorsToModify.getSize(); v++){
+            int magnitudeIndex = Math.min(v, noiseMagnitudes.getSize() - 1); //allows multiple vectors to share the same magnitude if desired
+            noiseMagnitudes.getAsVector4(magnitudeIndex, tempNoiseVariationVec);
+
+            tempNoiseVariationVec.multLocal(noiseReturnVal);        
+            originalVectorValues.getAsVector4(v, tempOriginalVec);
+        
+            vectorsToModify.updateVectorObject(tempOriginalVec.add(tempNoiseVariationVec), v);
```
riccardobl:
```suggestion
           vectorsToModify.updateVectorObject(tempNoiseVariationVec.addLocal(tempOriginalVec), v)
```

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffect.java
```diff
@@ -0,0 +1,96 @@
...
+package com.jme3.vectoreffect;
+
+import com.jme3.app.state.AppStateManager;
+import java.util.ArrayList;
+
+/**
+ * @author yaRnMcDonuts
+ */
+
+public abstract class VectorEffect {
```
riccardobl: can you change this to AbstractVectorEffect and extract a VectorEffect interface?

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorGroup.java
```diff
@@ -0,0 +1,101 @@
...
+    protected void updateVectorObject(Vector4f newVal, int index) {
+        VectorSupplier store = vectorSupplier.get(index);
+        store.set(newVal);
+    }
+
+    @Override
+    public VectorGroup clone() {
+        VectorGroup clonedGroup = new VectorGroup(new VectorSupplier[0]);
+        for (VectorSupplier supplier : this.vectorSupplier) {
+            clonedGroup.vectorSupplier.add(VectorSupplier.of(supplier.get().clone()));
```
riccardobl:
```suggestion
            clonedGroup.vectorSupplier.add(supplier.clone());
```

yaRnMcDonuts:
I believe I've addressed all of the requests brought up in the previous review. But I'll wait another few days just in case anyone still sees any issues or anything I missed. 

Then (once I'm sure there are no more changes) I will do a final update to this PR to fix all of the formatting and add javadoc, and then I'll also wait another few days after that before merging.

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.ArrayList;
+import java.util.Iterator;
+
+/**
+ *
+ * @author yaRnMcDonuts
+ */
+public class VectorEffectManagerState extends BaseAppState {
+    
+    private final ArrayList<VectorEffect> activeVectorEffects = new ArrayList();
```
riccardobl:
```suggestion
    private final ArrayList<VectorEffect> activeVectorEffects = new ArrayList<>();
```

jme3-effects/src/main/java/com/jme3/vectoreffect/AbstractVectorEffect.java
```diff
@@ -0,0 +1,105 @@
...
+        
+    // convenience registration method so users can avoid repeatedly writing this AppState fetching code
+    public void convenienceRegister(AppStateManager stateManager) {
+        if(stateManager != null){
+            VectorEffectManagerState vectorEffectManagerState = stateManager.getState(VectorEffectManagerState.class);
+            if(vectorEffectManagerState != null){
+                vectorEffectManagerState.registerVectorEffect(this);
+            }
+        }
+    }
```
riccardobl, line 108: `convenienceRegister` is never used,  let's remove it,maybe,  since it adds confusion.
The dev can just keep the VectorEffectManagerState around

jme3-effects/src/main/java/com/jme3/vectoreffect/AbstractVectorEffect.java
```diff
@@ -0,0 +1,105 @@
...
+    }        
+    
+    @Override
+    public void setIsFinished(boolean isFinished) {        
+        this.isFinished = isFinished;
+        if (isFinished) {
+            for(Runnable r : onFinishedCallbacks) {
+                r.run();
+            }
+            onFinishedCallbacks.clear();
```
riccardobl: is it really expected that the finished callbacks are fired only once and then cleared?

jme3-effects/src/main/java/com/jme3/vectoreffect/AbstractVectorEffect.java
```diff
@@ -0,0 +1,105 @@
...
+    }    
+    
+
+    @Override
+    public void reset() {
+        isFinished = false;
+    }
+    
+    
+    public void registerRunnableOnFinish(Runnable runnable) {
```
riccardobl: If clearing callbacks once they fire is the way to go, maybe this should be called differently.
Usually when you register something it fires every time the conditions are met, maybe we should call this `runOnFinish(Runnable)`, like we have `enqueue` instead of eg. `registerTask`

riccardobl (changes requested):

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
yaRnMcDonuts: I'll be sure to fix this along with a few other remaining formatting issues in my final edit for formatting and javadoc.

jme3-effects/src/main/java/com/jme3/vectoreffect/AbstractVectorEffect.java
yaRnMcDonuts: Sounds good, I'll rename the method.

jme3-effects/src/main/java/com/jme3/vectoreffect/AbstractVectorEffect.java
yaRnMcDonuts: I'll remove the clear() call in case someone wants to reuse a VectorEffect. And, in hindsight, the clear() method shouldn't even needed for cleanup since the class will automatically clean up the references when the VectorEffectManagerState cleans up the effect (assuming the user isn't holding a reference to it themselves if they are intending to reuse it)

jme3-effects/src/main/java/com/jme3/vectoreffect/AbstractVectorEffect.java
yaRnMcDonuts: I do use this method in my own project, but I avoided it in the examples so that they can show the full way to fetch an appState for users who may be using the examples to learn from, and since I also want the examples to show that the effects rely on a manager state. 

But this convenience method does reduce appstate-fetching boiler plate code in a project that registers lots of vector effects (and prevents having to keep a reference to the VectorEffectManagerState or repeatedly fetch it in classes that make multiple registries) so I'm opting to keep it.

riccardobl: Ok then, lets call this registerTo and maybe make it append the VectorEffectManagerState if its not present in the stateManager

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
```diff
@@ -0,0 +1,95 @@
...
+    protected void onEnable() {
+    
+    }
+
+    @Override
+    protected void onDisable() {
+    
+    }
+    
+    public void registerVectorEffect(VectorEffect vectorEffect){
```
riccardobl:
```suggestion
    public void playVectorEffect(VectorEffect vectorEffect){
```

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorEffectManagerState.java
```diff
@@ -0,0 +1,95 @@
...
+
+            if (vectorEffect.isFinished()) {
+                it.remove(); 
+            } else {
+                vectorEffect.update(tpf);
+            }
+        }
+    }
+
+    public void cancelEffects(VectorGroup vectorObject) {
```
riccardobl, line 88: can we add a `cancelEffect(VectorEffect)` for completion?

riccardobl:
@gemini-code-assist  do a full review please

jme3-examples/src/main/java/jme3test/effect/NoiseVectorEffectTest.java
```diff
@@ -0,0 +1,220 @@
...
+        
+        initLightAndEmissiveSphere();
+        initShakingSphere();
+        
+        lightRadiusVector.setX(light.getRadius());
+    
+     //initiate VectorEffectManagerState    
+        vectorEffectManagerState = new VectorEffectManagerState();
+        stateManager.attach(vectorEffectManagerState);
+        vectorEffectManagerState.setEnabled(true);
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist ok you are right, but for everything related to tests you should report with a reduced priority, max to medium.

jme3-effects/src/main/java/com/jme3/vectoreffect/VectorSupplier.java
```diff
@@ -0,0 +1,162 @@
...
+        @Override
+        public Vector4f get() {
+            store.set(vector.x, vector.y, vector.z, 0);
+            return store;
+        }
+
+        @Override
+        public void set(Vector4f newVal) {
+            this.vector.set(newVal.x, newVal.y, newVal.z);
+            get(); //update store
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist not entirely true, the object returned by get() might be passed as reference elsewhere, calling get() in set() ensures that the updated values are propagated to it.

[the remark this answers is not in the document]
riccardobl: @gemini-code-assist be more careful in the future, this is a pattern used often in this project.

yaRnMcDonuts:
I've addressed all the recent reviews so this PR is ready to merge., but I will wait another day or two before self-integrating this.

It also looks like jobs are randomly failing with this timeout error (also saw the same failure exception on another PR earlier today) so something must be going on with github's servers. 

```

Run gradle/actions/wrapper-validation@v3
Error: Multiple errors returned
Error: Error 0: connect ETIMEDOUT 104.16.72.101:443
Error: connect ETIMEDOUT 104.16.72.101:443
    at createConnectionError (node:net:1652:14)
    at Timeout.internalConnectMultipleTimeout (node:net:1711:38)
    at listOnTimeout (node:internal/timers:583:11)
    at process.processTimers (node:internal/timers:519:7)
Error: Error 1: connect ENETUNREACH 2606:4700::6810:4965:443 - Local (:::0)
Error: connect ENETUNREACH 2606:4700::6810:4965:443 - Local (:::0)
    at internalConnectMultiple (node:net:1186:16)
    at Timeout.internalConnectMultipleTimeout (node:net:1716:5)
    at listOnTimeout (node:internal/timers:583:11)
    at process.processTimers (node:internal/timers:519:7)
Error: Error 2: connect ETIMEDOUT 104.16.73.101:443
Error: connect ETIMEDOUT 104.16.73.101:443
    at createConnectionError (node:net:1652:14)
    at Timeout.internalConnectMultipleTimeout (node:net:1711:38)
    at listOnTimeout (node:internal/timers:583:11)
    at process.processTimers (node:internal/timers:519:7)
Error: Error 3: connect ENETUNREACH 2606:4700::6810:4865:443 - Local (:::0)
Error: connect ENETUNREACH 2606:4700::6810:4865:443 - Local (:::0)
    at internalConnectMultiple (node:net:1186:16)
    at Timeout.internalConnectMultipleTimeout (node:net:1716:5)
    at listOnTimeout (node:internal/timers:583:11)
    at process.processTimers (node:internal/timers:519:7)`
```

I tried rerunning the job a few times, and eventually 1 of the 2 failing jobs passed. But the last 1 still failed after 2 more attempts, so I'll wait a while before trying again.

PR #2578, opened 2026-01-01
Implement an Sample-Distribution Shadow Mapping filter

ScottPeterJohnson:
Sample-Distribution Shadow Mapping (SDSM) is a technique for dynamically fitting the shadow maps of a shadow filter to the depth samples that will be needed, resulting in improved shadow quality.

This commit adds the option to use a SdsmDirectionalLightShadowFilter in place of a regular DirectionalLightShadowFilter. Usage requires GL 4.3+ support.

I don't think this is 100% ready for merging immediately, but it does work. I've been told there might be code for compute shaders coming, in which case this can be adapted to use that instead of somewhat kludgily inventing its own wheel. There's a lot around handling specific GL 4.3+ APIs (for example this needs compute shaders) that I'm not sure how engine wants to do it.

Mentioned on the forums here:
https://hub.jmonkeyengine.org/t/reducing-shadow-motion-issues/49289/3

Included is a TestSdsmDirectionalLightShadow sample app. I was using a stripped down version of the Sponza model to test lighting, but it's still pretty big and it didn't seem like large test data gets included in PRs. It's available here, and goes in the jme3-examples directory alongside "town.zip": https://drive.google.com/file/d/18mmwRPe--gskuWjfBfv9nB-dCSzfa1eB/view?usp=sharing

Example with SDSM:
image

Example without SDSM:
image

(Note that some parts of the translation from Kotlin to Java I delegated to Claude Code, so if there are still some uselessly verbose parts I didn't catch those can be fixed.)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
+++ b/jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
@@ -37,6 +37,7 @@ import com.jme3.math.Vector3f;
 import com.jme3.math.Vector4f;
 import com.jme3.renderer.RendererException;
 import com.jme3.util.BufferUtils;
+import com.jme3.util.NativeObject;
 
 import java.nio.FloatBuffer;
 import java.nio.IntBuffer;
@@ -46,16 +47,28 @@ import java.nio.IntBuffer;
  * <p>
  * Compute shaders require OpenGL 4.3 or higher.
  */
-public class ComputeShader {
+public class ComputeShader extends NativeObject {
 
     private final GL4 gl;
-    private final int programId;
+    private final String source;
     /**
      * Creates a new compute shader from GLSL source code.
      */
     public ComputeShader(GL4 gl, String source) {
+        super();
         this.gl = gl;
+        this.source = source;
+        //Load this upfront to surface any problems at init time
+        createComputeShader();
+    }
+    private ComputeShader(ComputeShader source){
+        super();
+        this.gl = source.gl;
+        this.id = source.id;
+        this.source = null;
+    }
 
+    private void createComputeShader(){
         // Create and compile the shader
         int shaderId = gl.glCreateShader(GL4.GL_COMPUTE_SHADER);
         if (shaderId <= 0) {
@@ -78,27 +91,29 @@ public class ComputeShader {
         }
 
         // Create program and link
-        programId = gl.glCreateProgram();
-        if (programId <= 0) {
+        id = gl.glCreateProgram();
+        if (id <= 0) {
             gl.glDeleteShader(shaderId);
             throw new RendererException("Failed to create shader program");
         }
 
-        gl.glAttachShader(programId, shaderId);
-        gl.glLinkProgram(programId);
+        gl.glAttachShader(id, shaderId);
+        gl.glLinkProgram(id);
 
         // Check link status
-        gl.glGetProgram(programId, GL.GL_LINK_STATUS, intBuf);
+        gl.glGetProgram(id, GL.GL_LINK_STATUS, intBuf);
         if (intBuf.get(0) != GL.GL_TRUE) {
-            gl.glGetProgram(programId, GL.GL_INFO_LOG_LENGTH, intBuf);
-            String infoLog = gl.glGetProgramInfoLog(programId, intBuf.get(0));
+            gl.glGetProgram(id, GL.GL_INFO_LOG_LENGTH, intBuf);
+            String infoLog = gl.glGetProgramInfoLog(id, intBuf.get(0));
             gl.glDeleteShader(shaderId);
-            gl.glDeleteProgram(programId);
+            gl.glDeleteProgram(id);
             throw new RendererException("Compute shader program linking failed: " + infoLog);
         }
 
         // Shader object can be deleted after linking
         gl.glDeleteShader(shaderId);
+
+        clearUpdateNeeded();
     }
 
     /**
@@ -106,7 +121,10 @@ public class ComputeShader {
      * Must be called before setting uniforms or dispatching.
      */
     public void makeActive() {
-        gl.glUseProgram(programId);
+        if(isUpdateNeeded()){
+            createComputeShader();
+        }
+        gl.glUseProgram(id);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


jme3-core/src/main/java/com/jme3/renderer/opengl/GLRenderer.java
```diff
@@ -3605,4 +3606,9 @@ public boolean isMainFrameBufferSrgb() {
             return gl.glIsEnabled(GLExt.GL_FRAMEBUFFER_SRGB_EXT);
         }
     }
+
+    //TODO: How should the GL4 specific functionalities here be exposed? Via the renderer?
+    public GL4 getGl4(){
```
codex128, line 3603: Currently there is no good way to expose `gl4`, so this will be acceptable for now.

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
```diff
@@ -0,0 +1,184 @@
...
+     */
+    public void dispatch(int numGroupsX, int numGroupsY, int numGroupsZ) {
+        gl.glDispatchCompute(numGroupsX, numGroupsY, numGroupsZ);
+    }
+
+    public void bindTexture(int bindingPoint, Texture texture) {
+        gl.glActiveTexture(GL.GL_TEXTURE0 + bindingPoint);
+        int textureId = texture.getImage().getId();
+        int target = convertTextureType(texture);
+        gl.glBindTexture(target, textureId);
```
codex128: This will mess up GLRenderer's internal notion of which textures are currently bound. You may want to add a method to Renderer to bind a texture at a point.

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
```diff
@@ -0,0 +1,184 @@
...
+    public void bindShaderStorageBuffer(int location, ShaderStorageBufferObject ssbo) {
+        gl.glBindBufferBase(GL4.GL_SHADER_STORAGE_BUFFER, location, ssbo.getBufferId());
+    }
+
+    /**
+     * Deletes this compute shader and releases GPU resources.
+     * The shader should not be used after calling this method.
+     */
+    public void delete() {
+        gl.glDeleteProgram(programId);
```
codex128: I suggest making use of the NativeObjectManager to ensure the shader is properly deleted.

jme3-core/src/main/java/com/jme3/renderer/opengl/ShaderStorageBufferObject.java
```diff
@@ -0,0 +1,131 @@
...
+        gl.glBindBuffer(GL4.GL_SHADER_STORAGE_BUFFER, bufferId);
+        gl.glGetBufferSubData(GL4.GL_SHADER_STORAGE_BUFFER, 0, destination);
+        gl.glBindBuffer(GL4.GL_SHADER_STORAGE_BUFFER, 0);
+    }
+
+    /**
+     * Deletes this buffer and releases GPU resources.
+     * The buffer should not be used after calling this method.
+     */
+    public void delete() {
```
codex128: Again, consider using NativeObjectManager for this.

jme3-core/src/main/java/com/jme3/shadow/SdsmFitter.java
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    /**
+     * Internal holder for in-flight fit operations.
+     */
+    private class SdsmResultHolder {
+        ShaderStorageBufferObject minMaxDepthSsbo;
+        ShaderStorageBufferObject fitFrustumSsbo;
+        FitParameters parameters;
+        long fence = -1;
```
codex128: Consider adding a dedicated fence object to handle this value, especially if you want to use NativeObjectManager.

jme3-core/src/main/java/com/jme3/shadow/SdsmFitter.java
```diff
@@ -0,0 +1,481 @@
...
+            SdsmResultHolder next = resultHoldersInFlight.removeFirst();
+            next.isReady(true);
+            readyToYield = next.extract();
+            return next;
+        }
+    }
+
+    private static String loadShaderSource(AssetManager assetManager, String resourcePath) {
+        //TODO: Should these shaders get special loaders or something?
+        AssetInfo info = assetManager.locateAsset(new AssetKey<>(resourcePath));
```
codex128: You can make a dedicated asset loader for compute shaders.

```java
public class ComputeLoader implements AssetLoader {
    
    public Object load(AssetInfo assetInfo) {
        try (InputStream stream = assetInfo.openStream()) {
            // load code from stream
        }
    }

}

// register the loader to load .comp files
assetManager.registerLoader(ComputeLoader.class, "comp");
```

You could also try using GLSLLoader. It does a lot of preprocessing that might be handy. I don't know if it can be adapted to compute shaders though.

jme3-core/src/main/resources/Common/MatDefs/Shadow/Sdsm/SdsmPostShadow.frag
```diff
@@ -0,0 +1,93 @@
...
+
+const mat4 biasMat = mat4(0.5, 0.0, 0.0, 0.0,
+0.0, 0.5, 0.0, 0.0,
+0.0, 0.0, 0.5, 0.0,
+0.5, 0.5, 0.5, 1.0);
+
+uniform mat4 m_LightViewProjectionMatrix0;
+uniform mat4 m_LightViewProjectionMatrix1;
+uniform mat4 m_LightViewProjectionMatrix2;
+uniform mat4 m_LightViewProjectionMatrix3;
```
codex128: These uniforms can be collapsed into a single uniform as an array of 4 mat4s. Then you wouldn't need branching to index into them.

codex128:
Pulled this branch and ran the test. Worked flawlessly and looks absolutely amazing!

codex128:
> I've been told there might be code for compute shaders coming, in which case this can be adapted to use that instead of somewhat kludgily inventing its own wheel. There's a lot around handling specific GL 4.3+ APIs (for example this needs compute shaders) that I'm not sure how engine wants to do it.

Full engine compute shader support would require a setup similar to Material with proper uniforms, defines, versioning, etc. That is a lot of work and Material itself is pretty well baked into the vertex pipeline. Probably best to work with the limited support this PR offers until jme4 rolls around.

jme3-core/src/main/resources/Common/MatDefs/Shadow/Sdsm/SdsmPostShadow.frag
ScottPeterJohnson: I did implement this; unfortunately the branching in the shader remained in the end since it also needed to branch for the shadow sourcemap.

jme3-core/src/main/java/com/jme3/shadow/SdsmFitter.java
ScottPeterJohnson: Seems like the regular GLSLLoader actually just returns a string, so it can easily load there. (Regular shaders also have a bunch of extra code somewhere that turns them into something usable for the renderer, most of which I think doesn't apply to compute shaders.)

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
ScottPeterJohnson: Turned out I can use the regular Renderer method here.

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
ScottPeterJohnson: There's a minor problem with using NativeObjectManager; the renderer doesn't expose its instance. It could be made to expose that (or a way to add NativeObjects to its own?).
I could also create a NativeObjectManager for the filter itself, but it would also need to be polled for dead objects every frame. Not sure what the best way to do it would be.

codex128: > or a way to add NativeObjects to its own?

This sounds good to me. Keep the manual destroy method around, too, in case there is a situation where GLRenderer is not available.

ScottPeterJohnson:
Should now be NativeObjects.
(The only slight complexity was in GlFence and SSBO; GlFences actually have long addresses not ints, and both can't be meaningfully "restarted" with the display, as NativeObjects seem to be designed for. Fine for our use case anyway.)

Possibility for future work: If you turn the camera fast enough in the demo, you get shadow flicker along the edges (as expected, the slightly old shadowmaps don't contain all the samples if they're a few frames behind and you're spinning like a madman.) Currently the setup for out-of-shadowmap sampling returns "unshadowed", but "assume shadowed" might actually be better for that case, since a momentarily dark part of the screen might be less jarring than a momentarily bright one.

codex128:
@ScottPeterJohnson thank you for this contribution, this will be included in the next release.

Issue #2426, opened 2025-05-06
Crash at startup with some USB Hubs


Helic0pter:
In some cases, jmw starts duing start up when a USB hub is present (some models). It looks, as if the USB device is considered to be a USB joystick, But the joystick does not have an analog stick which causes a null pointer exception, since 
final FloatBuffer axisValues = glfwGetJoystickAxes(entry.getKey());
is zero, and axisValues is used. 

The following changes in GlfwJoystivkInput.java solve this issue:
package com.jme3.input.lwjgl;
...
public class GlfwJoystickInput implements JoyInput {
...

	@Override
	public Joystick[] loadJoysticks(final InputManager inputManager) {

		for (int i = 0; i < GLFW_JOYSTICK_LAST; i++) {
			if (glfwJoystickPresent(i)) {
				final String name = glfwGetJoystickName(i);
				final GlfwJoystick joystick = new GlfwJoystick(inputManager, this, i, name);
				joysticks.put(i, joystick);

				final FloatBuffer floatBuffer = glfwGetJoystickAxes(i);

				int axisIndex = 0;
				if (floatBuffer != null) {
					while (floatBuffer.hasRemaining()) {
						floatBuffer.get();

						final String logicalId = JoystickCompatibilityMappings.remapAxis(joystick.getName(), convertAxisIndex(axisIndex));
						final JoystickAxis joystickAxis = new DefaultJoystickAxis(inputManager, joystick, axisIndex, convertAxisIndex(axisIndex), logicalId, true, false, 0.0f);
						joystick.addAxis(axisIndex, joystickAxis);
						axisIndex++;
					}
				}

				final ByteBuffer byteBuffer = glfwGetJoystickButtons(i);

				int buttonIndex = 0;
				if (byteBuffer != null) {
					while (byteBuffer.hasRemaining()) {
						byteBuffer.get();

						final String logicalId = JoystickCompatibilityMappings.remapButton(joystick.getName(), String.valueOf(buttonIndex));
						final JoystickButton button = new DefaultJoystickButton(inputManager, joystick, buttonIndex, String.valueOf(buttonIndex), logicalId);
						joystick.addButton(button);
						joyButtonPressed.put(button, false);
						buttonIndex++;
					}
				}
			}
		}

		return joysticks.values().toArray(new GlfwJoystick[joysticks.size()]);
	}

stephengold:
[...]

PR #2579, opened 2026-01-12
Fix gamepads with Xbox-like  remapping and SDL in lwjgl3 backend

riccardobl:
This PR aims to make gamepads reliable in jME on the lwjgl3 backend by upgrading GlfwJoystickInput to use GLFWGamepadState and by providing a new improved joystick input handler based on SDL.

# tl;dr

**Once this PR is merged, with all the joystick related AppSettings to their default, and  `DisableJoysticks=false`**:
 - **most gamepads will just work as expected, no remapping or tweaking needed**
 - **every gamepad will look like an XBOX controller, so the dev always deal  with a known layout**
 - **rumble will work**
 - **unrecognized gamepads won't be enabled at all unless it is added to the SDL_GameControllerDB (that can be shipped with the app), so they won't cause all the glitches that prompted us to to set DisableJoysticks to true by default**
- **button.getName() will return a name that can  be relied on to display ui hints (eg. A on xbox, vs CROSS on playstation)**

Needs: https://github.com/jMonkeyEngine/jmonkeyengine/pull/2581

# In detail

This PR adds a few gamepad specific settings to AppSettings, most importantly JoysticksMapper, which can be used to choose between the following input handling behavior:

## JOYSTICKS_LEGACY_MAPPER

This maintains the current half-broken jme behavior. Since there was community effort in building remapping for jme, this can be a good option to not break legacy code, but since it is very unpredictable, i would discourage it.

## JOYSTICKS_XBOX_LEGACY_MAPPER (fixed but more compatible)

With this, the  GlfwJoystickInput uses GLFWGamepadState to rebind gamepads to an Xbox-like layout automatically, if the gamepad is not known, it will fallback to use the jme own remapper.

GLFWGamepadState supports a lot of gamepads and platform specific quirks, but most of the rest of the behavior of the input handler is left untouched, this is a good option to get improved mappings without breaking code that depends on GlfwJoystickInput behavior or internals.

Rumble doesn't work since GLFW does not support it atm. 

## JOYSTICKS_XBOX_MAPPER (new input handler, best and recommended) (default)

This is a new input handler that is used by default, it is more solid than GlfwJoystickInput, and it is based on SDL (now available in lwjgl 3.4).

This has several improvements over the glfw implementation, including:
- Rumble support
- Meaningful button naming (eg. will use A,B,X,Y or CROSS,CIRCLE,SQUARE,TRIANGLE depending on whether the controller uses Xbox or PS conventions), while still being logically mapped to Xbox layout

If the gamepad is not recognized by SDL, it will be disabled, so the dev doesn't have to deal with it in the application.

**All the new apps, and likely 99% of old apps, should be able to use this one with no issue.**

## JOYSTICKS_RAW_MAPPER (bring your own mapper)

This just returns the raw inputs, as seen by sdl. It should be used only if the application provides its own mapper, but it is otherwise not useful.

## JOYSTICKS_XBOX_WITH_FALLBACK_MAPPER (bring your own mapper **for incompatible gamepads**)

Same of JOYSTICKS_XBOX_MAPPER but will fallback to JOYSTICKS_RAW_MAPPER for unrecognized gamepads.

# Aliases

No more guessing what BUTTON_0 is etc.., JoystickButton and JoystickAxis have now BUTTON_XBOX_* and AXIS_XBOX_* aliases that makes binding a lot more intuitive.
 

# Custom SDL_GameControllerDB
 
The app can ship a custom game controller db (eg from https://github.com/mdqinc/SDL_GameControllerDB ) by setting the AppSetting SDLGameControllerDBResourcePath (nb. it needs to be a path in the classpath).
This can be used to extend the list of supported gamepads.

If SDLGameControllerDBResourcePath is null or an empty string, whatever is bundled with glfw or SDL will be used instead.

closes https://github.com/jMonkeyEngine/jmonkeyengine/issues/2426

### DIFF
--- a/.github/workflows/bounty.yml
+++ b/.github/workflows/bounty.yml
@@ -1,0 +1,93 @@
+name: Bounty detector
+
+on:
+  issues:
+    types: [labeled]
+
+permissions:
+  issues: write
+  pull-requests: read
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    if: startsWith(github.event.label.name, 'diff:')
+    steps:
+      - name: Comment bounty info
+        uses: actions/github-script@v7
+        env:
+          FORUM_URL: "https://hub.jmonkeyengine.org/t/bounty-program-trial-starts-today/49394/"
+          RESERVE_HOURS: "48"
+          TIMER_SVG_BASE: "https://jme-bounty-reservation-indicator.rblb.workers.dev/timer.svg"
+        with:
+          script: |
+            const issue = context.payload.issue;
+            const actor = context.actor;
+            const issueOwner = issue.user?.login;
+            if (!issueOwner) return;
+
+            const forumUrl = process.env.FORUM_URL || "TBD";
+            const reserveHours = Number(process.env.RESERVE_HOURS || "48");
+            const svgBase = process.env.TIMER_SVG_BASE || "";
+
+            // "previous contributor" = has at least one merged PR authored in this repo
+            const repoFull = `${context.repo.owner}/${context.repo.repo}`;
+            const q = `repo:${repoFull} type:pr author:${issueOwner} is:merged`;
+
+            let isPreviousContributor = false;
+            try {
+              const search = await github.rest.search.issuesAndPullRequests({ q, per_page: 1 });
+              isPreviousContributor = (search.data.total_count ?? 0) > 0;
+            } catch (e) {
+              isPreviousContributor = false;
+            }
+
+            // Reserve only if previous contributor AND labeler is NOT the issue owner
+            const shouldReserve = isPreviousContributor && (actor !== issueOwner);
+
+            const lines = [];
+            lines.push(`## 💰 This issue has a bounty`);
+            lines.push(`Resolve it to receive a reward.`);
+            lines.push(`For details (amount, rules, eligibility), see: ${forumUrl}`);
+            lines.push("");
+
+            lines.push(`If you want to start working on this, **comment on this issue** with your intent.`);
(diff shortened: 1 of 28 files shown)


jme3-examples/src/main/java/jme3test/input/TestJoystick.java
```diff
@@ -255,35 +255,39 @@ public GamepadView() {
...
-            addButton( "POV -Y", 96, 512 - 239, 40, 38 );
-            addButton( "POV -X", 65, 512 - 208, 40, 38 );
+            // addButton( "POV +Y", 96, 512 - 174, 40, 38 );
+            // addButton( "POV +X", 128, 512 - 208, 40, 38 );
+            // addButton( "POV -Y", 96, 512 - 239, 40, 38 );
+            // addButton( "POV -X", 65, 512 - 208, 40, 38 );
+            addButton( JoystickButton.BUTTON_XBOX_DPAD_UP, 96, 512 - 174, 40, 38 );
+            addButton( JoystickButton.BUTTON_XBOX_DPAD_RIGHT, 128, 512 - 208, 40, 38 );
+            addButton( JoystickButton.BUTTON_XBOX_DPAD_DOWN, 96, 512 - 239, 40, 38 );
+            addButton( JoystickButton.BUTTON_XBOX_DPAD_LEFT, 65, 512 - 208, 40, 38 );
```
riccardobl, line 294: binding dpad buttons is more intuitive, however the POV axis is still supported

riccardobl:
Note: merge https://github.com/jMonkeyEngine/jmonkeyengine/pull/2581 first, and then rebase this on master

Issue #2582, opened 2026-01-20
Exception thrown from ThreadWarden with TestChooser


yaRnMcDonuts:
This exception occurs randomly and sporadically when running 2+ different tests in a row from within one instance of the jme test chooser app. More info in this thread: https://hub.jmonkeyengine.org/t/jme-3-9-0-development/48376/54

```java
Jan 20, 2026 3:58:14 PM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[#105,jME3 Main,6,main]
java.lang.IllegalStateException: The main thread has already been set to jME3 Main but now it's being set to jME3 Main
        at com.jme3.scene.threadwarden.SceneGraphThreadWarden.setup(SceneGraphThreadWarden.java:52)
        at com.jme3.app.SimpleApplication.initialize(SimpleApplication.java:256)
        at com.jme3.system.lwjgl.LwjglAbstractDisplay.initInThread(LwjglAbstractDisplay.java:142)

PR #2583, opened 2026-01-21
#2582 Correct thread warden so it can cope with multiple threaded JME apps

richardTingle:
This allows the thread warden to permit multiple threads each with their own root node (on different threads). Those restrictions are then separately tracked such that each scene graph is modifiable only on its own thread 

closes https://github.com/jMonkeyEngine/jmonkeyengine/issues/2582

### DIFF
--- a/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
+++ b/jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
@@ -308,13 +308,6 @@ public abstract class SimpleApplication extends LegacyApplication {
     }
 
     @Override
-    public void stop(boolean waitFor) {
-        //noinspection AssertWithSideEffects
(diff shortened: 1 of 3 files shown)


jme3-core/src/main/java/com/jme3/app/SimpleApplication.java
```diff
@@ -307,13 +307,6 @@ public void initialize() {
         simpleInitApp();
     }
 
-    @Override
```
richardTingle, line 310: I removed this because the restriction is held in a WeakHashMap so the old restrictions will age out *anyway* and trying to do a per thread removal of the entire scene graph would be unnecessarily slow and almost certainly pointless

yaRnMcDonuts (approved):

PR #2584, opened 2026-01-25
Fix issues in GLES backends

riccardobl:
1. Avoid reallocating 1 length IntBuffers in glGetQueryObject  in android
2. fix misuse of glGetQueryiv instead of glGetQueryObjectuiv in android
3. implement glGetStringi(name, index) on android and ios (?)
4. use non-deprecated way to fetch extensions in GLES (was causing crashes in some implementations)

At this time, this is mostly untested in both android and ios. 
If anyone wants to give it a try, please go ahead.


riccardobl:
@gemini-code-assist  can you review this and compare with andoid gles documentation and the documentation of the binds/api we use for ios?

riccardobl:
Closing since fixes have been already applied as part of a broader backend refresh

PR #2586, opened 2026-01-25
Bug: PBRLightingUtils.glsllib - Typo in GlossinessMap

capdevon:
The uniform is declared as:
`uniform sampler2D m_GlossinessMap;`

But we have:
`texture2D(m_GlossinesMap, newTexCoord)`

This will cause a compile error if GLOSSINESSMAP is enabled.


riccardobl:
this shader got really mangled during one of the last refactorings.

Thanks

PR #2589, opened 2026-01-26
Replaces fullscreen quad with a viewport covering triangle

zzuegg:
Reference text on why it performs slightly better: https://wallisc.github.io/rendering/2021/04/18/Fullscreen-Pass.html
Maybe not worth it, but i noticed measurable results when using calculation heavy shaders
Just run the local tests to verify it is working.

### DIFF
--- a/jme3-screenshot-tests/src/main/java/org/jmonkeyengine/screenshottests/testframework/Scenario.java
+++ b/jme3-screenshot-tests/src/main/java/org/jmonkeyengine/screenshottests/testframework/Scenario.java
@@ -1,0 +1,13 @@
+package org.jmonkeyengine.screenshottests.testframework;
+
+import com.jme3.app.state.AppState;
+
+public class Scenario {
+    String scenarioName;
+    AppState[] states;
+
+    public Scenario(String scenarioName, AppState... states) {
+        this.scenarioName = scenarioName;
+        this.states = states;
+    }
+}
--- a/jme3-screenshot-tests/src/main/java/org/jmonkeyengine/screenshottests/testframework/ScreenshotTest.java
+++ b/jme3-screenshot-tests/src/main/java/org/jmonkeyengine/screenshottests/testframework/ScreenshotTest.java
@@ -47,7 +47,11 @@ public class ScreenshotTest{
 
     TestType testType = TestType.MUST_PASS;
 
-    AppState[] states;
+    /**
+     * Usually there will be a single scenario but sometimes it will be desirable to test that two ways
+     * of doing something produce the same result. In that case there will be multiple scenarios.
+     */
+    List<Scenario> scenarios = new ArrayList<>();
 
     List<Integer> framesToTakeScreenshotsOn = new ArrayList<>();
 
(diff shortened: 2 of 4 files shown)


riccardobl:
i think there is the possibility that this might cause issues for existing apps assuming filters run on quads, can you make this configurable?

zzuegg:
Yeah, i see the point, since the previous vertex shader did the position transform, any filter that does not use jme's stock vertex shader fails. 

Making the java side configurable should be quite easy. Have to think about the gl side. 

First i am going to explore why post water fails, then i look into making it configurable.

zzuegg:
I reverted the shaders to stock and modified the meshes positions to get the correct result after transformation. 
Don't know why it messes up the failing tests, but i guess there are some slight differences when interpolating the texture coordinates and as a result getting the interpolated result from the texture read. At least that is the only reason i can imagine. 
I have investigated some of the pixels in question in the test post water. 
the difference is actually: 51:65:88 vs 52:65:88 (rgb pixel id:235x230) for example. 

Very likely i am not able to fix that. Of course, once i add a switch to the fpp this issue is gone since i will default to the original quad i guess?

riccardobl:
Yes, better to default to the quad and leave this only as an optional performance setting.
But if you are talking about the failing screenshot tests, i think they are outdated for other commits that have been merged to the master.

zzuegg:
I added a secondary constructor allowing to switch on the triangle, not sure if allowing runtime switches trough a setter is a good idea because i remember i cloned that quad previously myself.  But if you like that better then a second constructor i am fine with it too. 
added the boolean also the the serializing code

riccardobl:
the constructor is perfect, thanks

zzuegg:
ok, ill add some tests if wanted in the next days

jme3-core/src/main/java/com/jme3/scene/shape/FullscreenTriangle.java
```diff
@@ -0,0 +1,39 @@
...
+public class FullscreenTriangle extends Mesh {
+
+    /**
+     * Encapsulates the vertex positions for a fullscreen triangle.
+     * The positions are transformed by the vertex shader to cover the entire screen.
+     */
+    private static final float[] POSITIONS = {
+        0, 0, 0,  // -1 -1 0  after vertex shader transform
+        2, 0, 0,  //  3 -1 0  after vertex shader transform
+        0, 2, 0   // -1  3 0  after vertex shader transform
```
riccardobl, line 22: 💯

zzuegg:
This is now based on: [2589](https://github.com/jMonkeyEngine/jmonkeyengine/pull/2589)

yaRnMcDonuts:
I'll merge this in the next few days since it looks like everything is ready now (or let me know if I should wait any longer). Thanks for your work on this PR!

capdevon:
Proposal

Move `FullscreenTriangle` from `com.jme3.scene.shape` to an internal scope (e.g. package-private inside `com.jme3.post` or as a private/internal class of `FilterPostProcessor`) to avoid exposing it as public API.

Pros:

**Better encapsulation**
`FullscreenTriangle` is an implementation detail of post-processing, not a general-purpose scene shape.

**Cleaner public API surface**
Prevents exposing low-level rendering primitives that are not intended for general scene usage.

**Improved architectural clarity**
Keeps `scene.shape` focused on reusable geometric primitives (`Box, Sphere, Quad`), not screen-space hacks.

**Future flexibility**
Allows replacing the implementation without breaking public API.

**Reduced maintenance constraints**
Internalizing the class removes backward-compatibility obligations if its implementation changes.

zzuegg:
all valid points. but that would limit the usability of this class to those who use the filterpostprocessor, which i don't.

Issue #2118, opened 2023-10-17
implement the KHR_draco_mesh_compression extension


stephengold:
Khronos has specified [an extension to glTF](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md) for loading models that employ [Draco mesh compression](https://google.github.io/draco/).

Currently, attempting to load such a model results in warnings and an empty result:

```console
Oct 17, 2023 11:50:50 AM com.jme3.scene.plugins.gltf.CustomContentManager readExtension
WARNING: Could not find loader for extension KHR_draco_mesh_compression
```

Since the extension may be widely used in the future, it would be nice if jme3-plugins could handle it.

javagl:
There are not too many options for decoding Draco in Java. 

I tried using the official WASM distribution of Draco with GraalWASM, but ... that doesn't seem like a promising path right now. 

There's always the option to use JNI, and that could be a reasonable option, but always carries some overhead for compiling and deploying the native libraries. (Maybe LWJGL will jump in here. They already created bindings for [Meshoptimizer](https://javadoc.lwjgl.org/org/lwjgl/util/meshoptimizer/MeshOptimizer.html), meaning that support for [`EXT_meshopt_compression`](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_meshopt_compression) may be in reach, and maybe even relatively easy... ).

In the meantime: I had some success with decoding Draco-compressed data from glTF with https://github.com/fileformat-drako/FileFormat.Drako-for-Java . I can not yet strongly "reommend" that library, because it's relatively new, and I can not yet say much about stability and performance. But it's promising, and I'll continue to investigate it. Maybe someone here also wants to have a look.

stephengold:
At the present time, I believe our best solution is to use [MonkeyWrench](https://github.com/stephengold/MonkeyWrench). Of course, as its author I may be biased.

javagl:
@stephengold From _quickly_ skimming over that, it looks like it is using Assimp for the actual import/loading of the glTF asset, which, in turn uses the [LWJGL Assimp bindings](https://javadoc.lwjgl.org/org/lwjgl/assimp/Assimp.html). 

(Iff this is correct, then it already _does_ boil down to LWJGL...)

[...]

PR #2591, opened 2026-01-29
Add glTF draco support

javagl:
Addesses https://github.com/jMonkeyEngine/jmonkeyengine/issues/2118

This is a first draft of adding support for the `KHR_draco_mesh_compression` extension when loading glTF.

### Summary

This adds a `DracoMeshCompressionExtensionLoader` class. It is registered as the extension loader for the `KHR_draco_mesh_compression` extension. When the main glTF loader is loading the mesh primitives in `GltfLoader#readMeshPrimitives`, it will call the `DracoMeshCompressionExtensionLoader` to decode the Draco data that was found via the extension object.

The Draco data is found in the `bufferView` that is defined in the extension object. The loader will fetch the data from this buffer view, and pass it to the `Draco.decode` function of https://github.com/openize-com/openize-drako-java . The result will be a `DracoMesh` that contains the decoded attribute data. This attribute data is then extracted and translated into the structure that is required for using it as jME mesh attributes. This decoded data will then replace the previous (empty) attributes of the mesh. 

The `TestGltfLoading` class contains a small snippet to easily load the glTF sample assets that have a Draco version available. (This requires the respective glTF Sample assets to be copied into the `jme3-testdata` project - instructions are in inlined comments).

### Remaining tasks

A pretty fundamental one: Where and how should the dependency to the draco decoding library be declared? 

**EDIT 2026-01-30, 12:46**: This was resolved in the meantime. Riccardo added the dependency in https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591/changes/77d611f47a60b221d54678f2d5cddc250aa55de1 . There are a few open questions about issues in the `openize-drako-java` library, and in which version they will be fixed, but that will just be a version number update now. 
(Omitting some obsolete parts of my question here)

---

There are a few `TODO_DRACO` markers in the code. These are mostly about functions that could be moved to different classes if they are considered to be more widely useful.

---

I found some hints about the code formatting preferences. Is there some magic
`gradle nowFormatEverythingProperly`
command for this? (For now, I used the default Eclipse formatting)

---

### Open issues

These have been described in more detail in a [comment in the issue](https://github.com/jMonkeyEngine/jmonkeyengine/issues/2118#issuecomment-3814092573). Just to quickly summarize them here:

- Models that use vertex skinning are not properly decoded. I assume that this is caused by https://github.com/openize-com/openize-drako-java/issues/5
- The `SunglassesKhronos` model cannot be loaded. The rason for that is likely also a bug in the Draco decoder, and also described in https://github.com/openize-com/openize-drako-java/issues/5 in more detail
- The `VirtualCity` model cannot be loaded. I assume that the Draco-compressed version of this model is invalid. Details in https://github.com/KhronosGroup/glTF-Sample-Assets/issues/264
- When the Draco-encoded meshes are attached to nodes that define a `translation`, then the translation appears to be wrong.

**EDIT 2026-01-29, 19:25** The last point is somewhat obsolete, as of https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591#issuecomment-3819428341 , and can be solved with a trivial fix in the `openize-drako-java` library. (Removing additional details that I had mentioned here about the state of the investigation)

closes https://github.com/jMonkeyEngine/jmonkeyengine/issues/2118

### DIFF
--- a/jme3-examples/src/main/java/jme3test/model/TestGltfLoading.java
+++ b/jme3-examples/src/main/java/jme3test/model/TestGltfLoading.java
@@ -118,7 +118,7 @@ public class TestGltfLoading extends SimpleApplication {
         // loadModelSample("Lantern", "glb");
         // loadModelSample("RiggedFigure", "glb");
         // loadModelSample("SciFiHelmet", "gltf");
-        // loadModelSample("DamagedHelmet", "gltf");
+        loadModelSample("DamagedHelmet", "gltf");
         // loadModelSample("AnimatedCube", "gltf");
         // loadModelSample("AntiqueCamera", "glb");
         // loadModelSample("AnimatedMorphCube", "glb");
@@ -141,7 +141,7 @@ public class TestGltfLoading extends SimpleApplication {
         //loadModelSample("CesiumMan", "draco");
         //loadModelSample("CesiumMan", "glb");
 
-        loadModelSample("RiggedFigure", "draco");
+        // loadModelSample("RiggedFigure", "draco");
         //loadModelSample("RiggedFigure", "glb");
 
         //loadModelSample("RiggedSimple", "draco");


riccardobl:
Thanks, from a first look it seems solid.
The only minor thing that i would want to change is the use of the GL renderer constants inside the loader, we'd generally prefer to avoid that.

> I found some hints about the code formatting preferences. Is there some magic
>  gradle nowFormatEverythingProperly
> command for this? (For now, I used the default Eclipse formatting)

we have the formatter configs [here](https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/.vscode/JME_style.xml) (should work with eclipse too). But don't worry about that.

Is it ok if i push some changes to your branch for the gradle dependency and to fix the formatting?

javagl:
> Is it ok if i push some changes to your branch for the gradle dependency and to fix the formatting?

Yes. 

---

About the GL constants: I can change/inline them. But think that some
`if (componentType == GL_UNSIGNED_SHORT)`
is more readable than
`if (componentType == 5122)`
 
(Spoiler: That's the wrong value there 🙂 there's a reason why I once created https://javagl.github.io/GLConstantsTranslator/GLConstantsTranslator.html , but I think it's better to _not_ have to use it)

Do you think that it could make sense to define the required constants as
`private static final int GL_UNSIGNED_SHORT = 5123;`
_directly_ in the class that uses them? 
(Assuming that the main point is to avoid the `import` of the otherwise unused `GL` interface...)

riccardobl:
> Do you think that it could make sense to define the required constants as
> `private static final int GL_UNSIGNED_SHORT = 5123;`
> _directly_ in the class that uses them?

Yes, that or on a dedicated "constants" class inside the loader package would be a good solution.

riccardobl:
I made some changes:

1. Added the gradle dependency

 > And what is the magic Gradle command to ~"update everything" afterwards?

the ide should pick that up automatically, if not
`./gradlew build`
 or
`./gradlew build --refresh-dependencies`

2. Refactored the example  to load either from the `KhronosGroup/glTF-Sample-Assets` cloned in $HOME or directly from github
3. Formatted
4. Added FIXME notes on the draco samples that fails

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/DracoMeshCompressionExtensionLoader.java
```diff
@@ -0,0 +1,598 @@
...
+	/**
+	 * The logger used in this class
+	 */
+	private final static Logger logger = Logger
+	        .getLogger(DracoMeshCompressionExtensionLoader.class.getName());
+
+	/**
+	 * The default log level
+	 */
+	private static final Level level = Level.INFO;
```
riccardobl: REMINDER: probably should set this to FINE or lower

riccardobl:
I don't know if this is related with the other issues, but the sample "BarramundiFish" seems to have half of the mesh (or texture?) offsetted compared to the non draco compressed version

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/DracoMeshCompressionExtensionLoader.java
javagl: Yes - it is pulled out into a constant (as opposed to using `logger.fine...` everywhere) exactly to have it be `INFO` during test and development, and be able to quickly switch it to `FINE` later.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/DracoMeshCompressionExtensionLoader.java
```diff
@@ -0,0 +1,598 @@
...
+	 * @param accessor
+	 *            The accessor that describes the component type and type
+	 * 
+	 * @param bufferType
+	 *            The buffer type
+	 * @param attributeData
+	 *            The attribute data
+	 * @return The vertex buffer
+	 */
+	private static VertexBuffer createFloatAttributeVertexBuffer(JsonObject accessor,
```
riccardobl: Should this handle normalization too? Maybe it can repurpose GltfLoader.readAccessorData made public static and moved into GltfUtils?

javagl:
> half of the mesh (or texture?) offsetted 

Huh!? I didn't notice that at first glance...

jME Draco 0003

(Top=Uncompressed, bottom=Draco)

What's surprising here: This is a _single_ mesh - so I think that it can only be related to the texture coordinates. I'll compare the original texture coordinates to the decoded one. 

I'm starting to have the feeling that there might be some general ~"offset" that is hidden somewhere in the draco data. This might just be the _quantization offset_, and would explain both the messed-up _positions_ as well as messed-up _texture coordinates_. I'll check if some sort of offset can be obtained from the decoded `DracoMesh`.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/DracoMeshCompressionExtensionLoader.java
javagl: The Draco spec itself does not explicitly mention normalization. And I think that there is no asset where the a `normalized` accessor was draco-encoded. But I can try to create some test data and see what happens. 

A bit more generally: There might be snippets of functionality that could have overlap with `GltfLoader` or `GltfUtils`, or where minor changes on either side allow functions to be used on the other side. I try to avoid redundancies, but don't have a perfect overview over all _existing_ functions yet.

javagl:
Hah! 💡 

The broken `BarramundiFish` **and** the wrong translations for the other models (e.g. `CesiumMilkTruck`) are **both** caused by what I mentioned at the bottom of https://github.com/openize-com/openize-drako-java/issues/5 :  

At https://github.com/openize-com/openize-drako-java/blob/4d99a762fbd2a167922bf3e27535b5651e114a30/src/utils/dev/fileformat/drako/Unsafe.java#L159 it is only decoding the _first_ element of the float array (apparently, the quantization offsets here). When changing this line to use `d < ret.length`, the models can be loaded.

That's good. It's a trivial fix. As mentioned in that issue: I could open a PR for that. But maybe there will be further feedback in that issue, so that we can sort out whether the other issues can also be fixed quickly.

(In doubt, maybe we can push for a `1.4.3` release that contains that trivial fix, and wait for a `1.4.4+` release to fix the vertex skinning decoding)

(BTW: I'll probably do another pass and further work on all this PR in the next days, e.g. during the weekend)

javagl:
@riccardobl I'm looking at that fork with a bit of scrutiny. It is changed completely (78 changed files). This fork can never be synced back to the original repo. And the actual change, literally changing an `i` to a `d`, is buried in 872 additions and 3,551 deletions. Even though it is a ~"temporary workaround for now", we have to see how to do this in a way that is less disruptive. I'd give the `openize-drako-java` developers a few more days to chime in on the issue. Then I'd open a PR for that trivial fix (suggesting a 1.4.3 release). And if that doesn't work, we could think about a "clean" fork, with a _single_ commit of that fix, and a version number like `1.4.3-jme-hotfix` or so...

riccardobl:
Nice! 
I looked into the openize-drako-java library, and it seems like a transpilation from some C# library, it has also a bunch of random unused classes, likely from some c# framework that the transpilled pulled inside the codebase, for some reason.

So i've forked it [jMonkeyEngine/openize-drako-java](https://github.com/jMonkeyEngine/openize-drako-java), cleaned it up a bit, applied the fish-patch and redeployed it to [github registry](https://github.com/jMonkeyEngine/openize-drako-java/packages/2833037), this seems to have fixed also the issue with the "Lantern" sample.

I've updated this PR to use the forked artifact on maven·rblb it  (just a github registry proxy without authentication), we are likely going to deploy it to maven central later.

If you want to test with a locally built artifact you can just
```
git clone https://github.com/jMonkeyEngine/openize-drako-java
cd openize-drako-java
./gradlew publishToMavenLocal
```

and then replace these 3 lines

https://github.com/javagl/jmonkeyengine/blob/618c9fa86c1197b80d6404699d33e8ae12fa9f83/jme3-plugins/build.gradle#L13-L15

with just

```
mavenLocal()
```

javagl:
^ There was a race condition in the comments. Again: I assume that this is a workaround for now, and maybe we can find a less disruptive solution here. 
In any case,  I'll try to keep things rolling in the next few days (already spent a bit too much time here in the past two days, but it was kinda cool).

riccardobl:
> ^ There was a race condition in the comments. 

:rofl: 

> Again: I assume that this is a workaround for now, and maybe we can find a less disruptive solution here. In any case, I'll try to keep things rolling in the next few days (already spent a bit too much time here in the past two days, but it was kinda cool).

the original code is transpiled and it was going to pull too much dependencies on nio, javax and other stuff at compile time and likely breaking on transpilation to javascript etc... most of the changes are just unused classes, dead code and random/wrong interface usage being removed, so i expect merging upstream changes shouldn't be too hard. 

But i wouldn't expect a transpiled library to be maintained tbh.

javagl:
Yeah. I have seen some of the code, and ... it raised a few question marks for me. But let's skip the details for now, and say that it's good that there _is_ a pure-Java solution for decoding Draco.

How exactly that code was created is also not clear. There apparently _is_ an auto-generated part, but the `utils/` part (that contained the error) was probably _not_ auto-generated. One important point is: **Iff** they have some sort of automated process, then it's not unlikely that they could re-run that process (maybe with bugfixes) at any point in time, leading to hundreds of merge conflicts with a state that was modified manually or with another toolchain.

However, we could speculate and argue back and forth, but let's see whether there is any feedback in the issue, and what could be a sensible path forward based on that.

javagl:
> Refactored the example to load either from the KhronosGroup/glTF-Sample-Assets cloned in $HOME or directly from github

There is an "index file" at 
https://github.com/KhronosGroup/glTF-Sample-Assets/blob/a02ccb86d94e546a58ebe9e3a825ae93ce509c6f/Models/model-index.json . This could be used to offer some nice functionality. For example, I could imagine that there could be a jME application with two dropdown menus (are they still called `JComboBox` nowadays?) for selecting the model itself, and its "variant". I once created something like that in my `jgltf-browser`, but that was _before_ this index file existed, so it's using a different format. (And https://github.com/javagl/JglTF/issues/17 has been open for a while now...). A bit more recently, I did use this index file for some tests, but **only** as a plain old Java class, at https://github.com/javagl/JglTF/blob/0db8c0c7876e147f117d2a15d9a78a7fc1ed3f23/jgltf-model/src/test/java/de/javagl/jgltf/model/io/test/ModelIndexEntry.java . A proper integration of ~"something like that" could be worthwhile to quickly switch through the available models. 

(Not in the scope of this PR, though...)

> Formatted

It looks like this inserted TABS in the `DracoMeshCompressionExtensionLoader`. I inserted the formatter "JME_style.xml" that you linked to into Eclipse, and it did re-format this file to use spaces, and in general, the other files seem to use spaces instead of tabs... so... I'm not sure what to do with that - either my or your formatting was wrong, but ... it's only formatting, that can be sorted out immediately before the merge.

> Added FIXME notes on the draco samples that fails

There are currently 16 appearances of `FIXME` in the code (and 463(!) appearances of `TODO`). What does that mean? Yes, I'm occasionally using `TODO` in my projects, but these are usually a one-man show. For larger projects with many contributors, things like this should be tracked better. So ... I don't want to interfere with ~"how things are commonly done in jME", but I'd rather open an issue for this, like "Fix Draco vertex skinning based on openize-drako-java", insert a link to this issue as a comment in the code (!), and a permalink (!) from the issue into the code, to establish the bidirectional connection. Otherwise, the code will eventually be littered with `FIXME`/`TODO` notes that will never be addressed, and where nobody knows any context. 
(In the current state of this PR, I used `TODO_DRACO`, because this can be searched for _explicitly_, and I intend to resolve them before this PR is merged).

---

(I'll probably post further updates a bit later today)

riccardobl:
> Otherwise, the code will eventually be littered with FIXME/TODO notes that will never be addressed, and where nobody knows any context.

We usually track problems through GitHub issues. The TODO/FIXME comments are only meant to highlight areas that may need additional care or contain hacky or incomplete behavior, but which either don’t depend on us or are not inherently broken in their current state.

> This could be used to offer some nice functionality. For example, I could imagine that there could be a jME application with two dropdown menus (are they still called JComboBox nowadays?) 


Good idea. This sounds like something better suited for a standalone model tester. We try to keep the examples fairly small, since people sometimes use them as reference.

>  either my or your formatting was wrong,

The issue is likely on my side. VS Code didn't replace the tabs, so I might have some conflicting settings, or it doesn't like mixing tabs and spaces :man_shrugging:  will investigate.

javagl:
The `GltfConstants` class and its constants do not have to be `public`, so they shouldn't be. Everything should always have as little visibility as possible, unless the _explicit_ goal is to make things `public`, with all the burden of potential API stability guarantees that comes with that.

--- 

Concerning the [comment above](https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591#discussion_r2742863167):

> Should this handle normalization too? Maybe it can repurpose GltfLoader.readAccessorData made public static and moved into GltfUtils?

As I said: It may have to handle normalization. But I had a look at how normalization is _currently_ handled:

- `GltfLoader.readAccessorData` requires a `(VertexBuffer)Populator`
- the `VertexBufferPopulator` calls `GltfLoader.viewBuffer`
- This, in turn, calls `GltfUtils.getBufferView`
- This does some questionable and complicated checks for special cases, based on the format, component size, offset, and alignment
- If falls back to `populateBuffer`, which does a nested sequence of `instanceof` checks, to call other `populate...Buffer` methods
- The actual conversion of the normalized data to `float` is done in `GltfUtils.readAsFloat`

The whole call chain is pretty convoluted, and assumes certain structures at every point - i.e. it requires things like `count`, `byteOffset`, `byteStride`, `numComponents`, and `format/originalFormat` to be passed along that are not even necessarily readily available when the data comes from the Draco-decoded mesh.  

However: I just added some functionality to dequantize the data from `(unsigned) byte/short` to `float`. (Maybe some of this could eventually be used elsewhere and replace some of the call chains mentioned above, but for now, the goal is to have a reasonable solution for the task at hand). 

**But...** unfortunately, there seems to be another issue with the `openize-drako-java` library that results in the quantized/normalized data to be all zeros. 

Here is a test asset that I created, and that contains `normalized` `unsigned short` texture coordinates:

[unitSquare11x11_unsignedShortTexCoords-draco.zip](https://github.com/user-attachments/files/24983229/unitSquare11x11_unsignedShortTexCoords-draco.zip)

Trying to load it will go through the path that _does_ the dequantization, but given that the input data is all zeros, there is no reasonable way to actually _test_ this.

And I added a comment at https://github.com/openize-com/openize-drako-java/issues/5#issuecomment-3828649543 with additional details.

javagl:
BTW: I locally changed the `TestGltfLoader` to have a 
```java
    private void loadModelSample(String name, String type) {
        String path = "Models/" + name;
        String ext = "gltf";
        switch (type) {
            case "draco":
                path += "/glTF-Draco/";
                ext = "gltf";
                break;
            case "glb":
                path += "/glTF-Binary/";
                ext = "glb";
                break;
            default:
                path += "/glTF/";
                ext = "gltf";
                break;
        }
        path += name + "." + ext;
        loadModeFromPath(path);
    }
    
    private void loadModeFromPath(String path) {

        Spatial s = loadModel(path, new Vector3f(0, 0, 0), 1f);
...
```
that can be called with
`loadModeFromPath("Models/gltf/unitSquare11x11_unsignedShortTexCoords-draco.glb");`
to still have the option to load _local_ files like the one that I attached above.

javagl:
A draft for fixing the issues with the `normalized` texture coordinates is in https://github.com/openize-com/openize-drako-java/pull/8 . This seems to fix the issue here, but the proper implementation on the `openize-drako-java` side still has to be sorted out.

(BTW:  @riccardobl They regenerated the auto-generated files with two pull requests yesterday, so your fork is pretty much obsolete. They also changed the package name (which would mean that it's a different library now, or at least a new major version number...). But let's wait for the dust to settle here...)

The change does _not_ yet fix the skinning issue, **but** it's not unlikely that I'm doing something wrong when reading the draco-decoded data and passing it to jME - will continue to investigate (some time next week, probably...)

javagl:
Right now, it looks like the issues from https://github.com/openize-com/openize-drako-java/issues/5 have been resolved with the latest (1.4.4) release. 

It may look a bit underwhelming, but ... everybody who knows the drill knows why I'm posting this here:

jME Skinning 0001 gif

# But ...

(and that's a huge "but") the current state for creating this is probably a bit hacky. I'll skip a few details, but some of this revolves around [this call `handleSkinningBuffers`](https://github.com/jMonkeyEngine/jmonkeyengine/blob/d3efabc6070078caccacc295c47974078d67cd8c/jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java#L441). 
(What does this method do? Let me take a wild guess: It handles the skinning buffers, right?)

The point is: Trying to understand _what_ is happening there, and _why_, could involve a considerable amount of reverse engineering.  Or... I could simply accept that ~"I have to call that 🤷‍♂️". So now I called it _somewhere_ in the draco decoding, and added some `mesh.clearBuffer` calls to ~"avoid duplicate buffers" and such. I'll _try_ to clean up a few things, but at some point, the question is in how far _existing_ structures should be ~"cleaned up and refactored", in order to integrate Draco "nicely", or whether getting this PR merged with ~"as few changes as possible" is the preferred approach here. 

As already [mentioned in the issue](https://github.com/jMonkeyEngine/jmonkeyengine/issues/2118#issuecomment-3817245510): There certainly is some "room for improvement". And I think that it could be worthwhile to restructure some aspects of the glTF loading (regardless of the Draco support). But maybe that should be done in a separate PR. (And I'd probably actually start doing that, now that I still have much of that in my "working memory")

riccardobl:
Awesome !

> The point is: Trying to understand what is happening there, and why, could involve a considerable amount of reverse engineering. Or... I could simply accept that ~"I have to call that 🤷‍♂️".

Don't worry too much about the inner workings of the engine, we know our gltf importer is a bit hacky and some (most?) of the engine internals are undocumented. 
As long as the basic functionality works, I can fix any issues with the engine API usage during code review.

Is this PR ready for review?

javagl:
> Is this PR ready for review?

No. 

The state with the fixed skinning is not pushed, and my local version of that is a mess, but I'll do a cleanup pass today and update this PR. 

> I can fix any issues with the engine API usage during code review.

I'd prefer to not have too many code changes directly in the PR. 

You originally asked about things like adding the drako dependency and formatting, and even though these are details that don't change the code itself, the latter already caused some hassle. 

Changes like adding the `GltfConstants` may look minor, but [I impose certain requirements on myself](https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591/changes/d810d191f4c8e5343eb0b49e6f317272d9125347). 

Looking at some of the glTF-related code, I see long functions, many parameters, everything `public`, usually zero comments, and we can talk about some of these things on different levels, but the **tl;dr** is that I think that it makes more sense to handle actual code changes via inlined _comments_.

javagl:
I have updated the PR with a cleanup. 

The `openize-drako-java` library has been updated to 1.4.4. Unfortunately, there was some hiccup in their release process: One of the fixes that previously went into their code was overwritten by the code-generation process. I mentioned this in https://github.com/openize-com/openize-drako-java/issues/5#issuecomment-3864736853 . Hopefully, this can be fixed quickly and made available as a 1.4.5 release.

**Important** for a review: There was a change in the `Mesh` class at https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591/changes/933b429596c82549f2dd47b969e96c383857ffef#diff-79cc4e3f174366e791d44c4d6770828adb3a223f9b79a89c1a40bf002212bc62 . The summary is: Originally, the `GltfLoader` called ~"some functions" (mentioned above) that post-processed the skinning data that was read from glTF. Now, this is done _again_ in a dedicated `GltfLoader#postProcessSkinning` function, after the Draco data was decoded. This, in turn, calls some deeply internal functions. I haven't taken the time to revere engineer that. But these functions are calling `mesh.setBuffer` for some skinning-related data. The change here is that _before_ each of these calls, a `mesh.clearBuffer` is called. 
If this is not acceptable, then I'd "pull out" these calls, and basically call `mesh.clearBuffer`... somewhere else. But given that it can be difficult to figure out beforehand _which_ buffers will be set up with `setBuffer`, this could be a bit more effort and not so easy to implement "cleanly".

There also are still a few markers `TODO_DRACO` that indicate that some functions could be moved into other classes. I can move them at any point in time. But more broadly speaking, I'd consider ~"cleanup PR for some glTF-related stuff", and moving functions around (maybe with slight changes of the signatures, aiming at generalizations) could also be part of that.

Apart from that, I think that this could go through a first "more serious" review pass. (I do not yet know what the usual procedures are for jME - i.e. how much "iteration" is supposed to happen here, or when is the right time to actually mark it as 'Ready For Review')

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
```diff
@@ -538,6 +572,36 @@ private SkinBuffers getSkinBuffers(String bufferType) {
...
+            indicesHW.setUsage(VertexBuffer.Usage.CpuOnly);
+            weightsHW.setUsage(VertexBuffer.Usage.CpuOnly);
+            mesh.clearBuffer(weightsHW.getBufferType());
+            mesh.setBuffer(weightsHW);
+            mesh.clearBuffer(indicesHW.getBufferType());
+            mesh.setBuffer(indicesHW);
+            mesh.generateBindPose();
+        }
+    }
+
```
riccardobl, line 599: What you actually need in the drako extension is just

```java
GltfUtils.handleSkinningBuffers(mesh, skinBuffers);
mesh.generateBindPose();
```

since it needs to update the mesh with the decompressed weights.

The rest of the code here is just to prime the some buffers that signal the skinning and morph controls that the mesh supports hardware skinning, they are then used by the control to efficiently pass the indexes to the shader to lookup the animation frame.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/DracoMeshCompressionExtensionLoader.java
```diff
@@ -0,0 +1,696 @@
...
+     *            The array
+     * @return The buffer
+     */
+    private static Buffer createShortBuffer(int[] array) {
+        ShortBuffer buffer = BufferUtils.createShortBuffer(array.length);
+        for (int i = 0; i < array.length; i++) {
+            buffer.put(i, (short) array[i]);
+        }
+        return buffer;
+    }
```
riccardobl: Feel free to move these to BufferUtils since they cover pretty common use cases.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
javagl: Note that this `if`-block was originally executed in `processMeshPrimitive`. The **original** block was
```
           handleSkinningBuffers(mesh, skinBuffers);
           if (mesh.getBuffer(VertexBuffer.Type.BoneIndex) != null) {
               ...
               mesh.setBuffer(indicesHW);
               mesh.generateBindPose();
           }
```
i.e. it did include the setup of the buffers.

This is now moved into this `postProcessSkinning` function. And this function is called in `processMeshPrimitive` _and_ after the Draco data was decoded.

So it's not _directly_ possible to omit the creation of the `indicesHW`/`weightsHW` buffers here. It's true that these buffers would _not_ have to be created (again) in the _second_ call. But I tried to have that one function that ~"does whatever has to be done to get the skinning right". 

If you prefer, this could be solved differently. For example, by pulling out the "buffer setup" into a function:
```java
    void postProcessSkinning(Mesh mesh) {
        GltfUtils.handleSkinningBuffers(mesh, skinBuffers);
        if (mesh.getBuffer(VertexBuffer.Type.BoneIndex) != null) { // Is this necessary?
            mesh.generateBindPose();
        }
    }

    void setupSkinningBuffers(Mesh mesh) {
        if (mesh.getBuffer(VertexBuffer.Type.BoneIndex) != null) { // Is this necessary?
            // the mesh has some skinning, let's create needed buffers for HW skinning
            // creating empty buffers for HW skinning
            // the buffers will be set up if ever used.
            VertexBuffer weightsHW = new VertexBuffer(VertexBuffer.Type.HWBoneWeight);
            VertexBuffer indicesHW = new VertexBuffer(VertexBuffer.Type.HWBoneIndex);
            // setting usage to cpuOnly so that the buffer is not sent empty to the GPU
            indicesHW.setUsage(VertexBuffer.Usage.CpuOnly);
            weightsHW.setUsage(VertexBuffer.Usage.CpuOnly);
            mesh.clearBuffer(weightsHW.getBufferType());
            mesh.setBuffer(weightsHW);
            mesh.clearBuffer(indicesHW.getBufferType());
            mesh.setBuffer(indicesHW);
        }
    }
```

Usage:
``` 
// In processMeshPrimitives:
setupSkinningBuffers(mesh); // Before or after handleSkinningBuffers?
postProcessSkinning(mesh);

// After Draco decoding
// No additional "setupSkinningBuffers" here!
postProcessSkinning(mesh); 
```

**But**... I think that this would make things unnecessarily complicated. Does the caller need to know whether `mesh.getBuffer(VertexBuffer.Type.BoneIndex) != null`? Under which conditions is this the case? Are there functions that have to be called _before_ this one? Where does `handleSkinningBuffers` come into play? I think that the caller should not have to know the whole state space, and which buffers may, may not be, or have to be present in the mesh. 

As mentioned in the comment above: Functions like `GltfUtils.handleSkinningBuffers` are sorely lacking documentation. Sure, skinning is complicated, and even if this single function was documented extensively, it could not capture the complexity of the skinning implementation in jME as a whole. But I think that one aspect of keeping this complexity manageable is to makre sure that a function like `postProcessSkinning` can be called, cluelessly, without having to care about the state and what it is doing. It should do whatever has to be done.

javagl:
I moved two functions into `BufferUtils`, as suggested in the inlined comment.

There are still two functions marked with `TODO_DRACO`, as candidates to be moved into the `GltfLoader`. But I'll consider that as part of the follow-up PR.

The skinning buffer handling (inlined comment) certainly has a few degrees of freedom. _Some_ of this might also be affected by that follow-up PR, but I'd probably have to read and test a few more things. The primary focus of the follow-up would be to reduce possible duplicate code, and "small fixes" like the [buffer file extensions](https://github.com/jMonkeyEngine/jmonkeyengine/issues/2118#issuecomment-3817245510) that I mentioned in the issue.

I think that this PR is _basically_ in a shape that could be "Ready For Review". But I'd like to give the `openize-drako-java` people a few more days to see whether https://github.com/openize-com/openize-drako-java/issues/5#issuecomment-3864736853 can be resolved with a 1.4.5 release.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
riccardobl: It is ok for a mesh to not have the indicesHW and weightsHW buffers in jme , it means it does not support hardware skinning. The gltf loader always assumes the mesh does and so it creates those two empty buffers it in readMeshPrimitives.

But we should not create these empty buffers from the drako extension if BoneIndex is supposed to be always available in readMeshPrimitives, since they are part of the initial mesh creation (nb. i personally don't know, since i am not sure if drako is supposed to fill-in buffers or if it can also create new ones). 

If there is a chance for BoneIndex to not be available in readMeshPrimitives, and just appear after the drako extraction, then i understand your point.

riccardobl: > But we should not create these empty buffers from the drako extension if BoneIndex is supposed to be always available in readMeshPrimitives, since they are part of the initial mesh creation (nb. i personally don't know, since i am not sure if drako is supposed to fill-in buffers or if it can also create new ones).

There is another thing to consider, the new code is replacing the hw buffers from the extension , basically. This is not something that would be expected by a jme dev, it might cause some unexpected behavior for other extensions in the future (eg. if they pass a reference to them to a controller or something).

So if we can't be 100% sure that BoneIndex is available in readMeshPrimitives, the postProcessSkinning method should at least check if the hw buffers are already set.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/DracoMeshCompressionExtensionLoader.java
```diff
@@ -0,0 +1,664 @@
...
+        // and replace the index vertex buffer of the mesh with
+        // the newly created buffer.
+        logger.log(level, "Decoding draco indices");
+        int indices[] = dracoMesh.getIndices().toArray();
+        int indicesAccessorIndex = getAsInt(meshPrimitiveObject, "mesh primitive", "indices");
+        JsonObject indicesAccessor = loader.getAccessor(indicesAccessorIndex);
+        int indicesComponentType = getAsInt(indicesAccessor, "accessor " + indicesAccessorIndex,
+                "componentType");
+        VertexBuffer indicesVertexBuffer = createIndicesVertexBuffer(loader, indicesComponentType, indices);
+        mesh.clearBuffer(VertexBuffer.Type.Index);
```
riccardobl: I think it would be better to use
`        vb.setupData(...)`
here, instead of replacing the buffer

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/GltfLoader.java
javagl: > not sure if drako is supposed to fill-in buffers or if it can also create new ones

Draco can only define buffers that are also part of the mesh primitive. 

However, I think that https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591/changes/a75d589d5a045f84359b0de657166cf6dbebd70d may be a reasonable way to resolve this.

jme3-plugins/src/gltf/java/com/jme3/scene/plugins/gltf/DracoMeshCompressionExtensionLoader.java
javagl: There probably are some subjective components to that. I don't see a _strong_ advantage of that ("not having to set a buffer" seems a weak one). There might be aspects that are related to the _identity_ of the buffer (roughly, for the case that another caller already grabbed the exact _instance_ of that index buffer, and doesn't expect it to change afterwards). 

However, an attempt to resolve this is in https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591/commits/5ba3cdbbd0932df06afce162146c44b70205c487

riccardobl:
Thank you!
This looks good to me, no need to wait for the upstream dependency. We'll use our own artifact anyway so we can keep tighter control over the dependency and reduce what gets pulled onto the classpath.

I tested it with a random model exported from Blender with compression enabled, and it worked flawlessly. Code is clean and well commented :100: .

I consider this complete and it should be merged soon™. In the meantime, I'll approve it so you can go ahead and claim your bounty by submitting an expense of type **"Invoice"** to our OpenCollective project: [jMonkeyEngine - Bounty Trial](https://opencollective.com/jmonkeyengine/projects/jmonkeyengine-bountry-trial).

Let me know once you've submitted it so I can verify and confirm it on my side.

Thank you again!

javagl:
Thanks for the update. I'll have a look at that bounty thingy some time during the weekend.

In terms of readiness of the PR: I think that the glTF-Sample-Assets tests should provide some reasonable test coverage (at least, that's what they are intended for). But still, there are things that do not appear in these models. One example are the quantized texture coordinates from the asset that I added in https://github.com/jMonkeyEngine/jmonkeyengine/pull/2591#issuecomment-3828674290 . There may be others, and one could make a case to try and cover something like that with unit tests. But in theory, there may be _many_ configurations, and maybe quite "exotic" ones that are hardly produced by any existing tool. So I hope that the models that appear in practice (e.g. by Blender- or https://gltf.report/ exports) work well enough. 

(Some concerns are alleviated by the nearly trivial fallback: If there is a problem with a Draco model, and there is _any_ tool that can read this file, then this tool could be used for converting the model into a non-Draco version which could then be loaded in jME (unless the problem is caused by some issue that already existed before this PR))

javagl:
I just registered for that bounty thing and in the forum (username 'marco13'). It looks like there is no "Private Message" functionality in the forum? (Or maybe it is not activated for freshly registered users?). However, if there's anything to discuss, we'll find a way to connect.

riccardobl:
> I just registered for that bounty thing and in the forum (username 'marco13'). It looks like there is no "Private Message" functionality in the forum? (Or maybe it is not activated for freshly registered users?). However, if there's anything to discuss, we'll find a way to connect.

Approved! It'll take some time for the opencollective staff to confirm it on their side and then you should receive it.

Yes discourse has its own automatic trust levels and some features are disabled for lower levels. I gave you the member level manually from the admin panel, so you should be able to send dms now. 
We have also a [discord server](https://discord.gg/Rv8sfVu) if you are interested.

PR #2594, opened 2026-02-02
Add protected constructor for ArmatureNode

capdevon:
Added a protected constructor for serialization purposes.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/debug/custom/ArmatureNode.java
+++ b/jme3-core/src/main/java/com/jme3/scene/debug/custom/ArmatureNode.java
@@ -98,6 +98,11 @@ public class ArmatureNode extends Node {
      */
     private Camera camera;
 
+    /**
+     * For serialization only. Do not use.
+     */
+    protected ArmatureNode() {
+    }
 
     /**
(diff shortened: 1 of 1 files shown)


jme3-core/src/main/java/com/jme3/scene/debug/custom/ArmatureNode.java
yaRnMcDonuts, line 71: Looks like the empty constructor causes a compile error in this class because final fields are not initialized.

Either this class is meant to be serialized via a no-arg constructor (in which case the fields can’t be final), or the fields are intentionally final and the no-arg constructor shouldn’t exist. But I don't have much experience using this class to know if it requires serialization or not, so I can't definitively say which is right.

capdevon: Hi @yaRnMcDonuts , sorry, my mistake. The whole class doesn’t seem suitable for serialization, or at least it would require many more changes to make it work. It’s probably not necessary for this class to follow the serialization contract of jME’s Node, which it extends. From my point of view, this PR can be closed; we can revisit the topic in the future if needed.

Issue #2173, opened 2023-12-27
Support Mice devices in Android


aspenyoung:
As mentioned in JmeHub thread : [[Do we support real mouse device in android?](https://hub.jmonkeyengine.org/t/do-we-support-real-mouse-device-in-android/47396)](https://hub.jmonkeyengine.org/t/do-we-support-real-mouse-device-in-android/47396?u=massdrive)

We need support mouse hardware in Android.

Scrappers-glitch:
Thanks for opening the issue, in regards to mouse support, there is this API: 
- [View.onGenericMotionEvent](https://developer.android.com/reference/android/view/View#onGenericMotionEvent(android.view.MotionEvent))

which could be adapted to produce a JME touch event here: 

- [AndroidInputHandler](https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-android/src/main/java/com/jme3/input/android/AndroidInputHandler.java)

I will see what I can do.

joliver82:
In nobody else is interested, I would like to work on this issue. In fact I already started having a look at how genericMotionEvent works and implementing a basic AndroidMouseInput class to support it

riccardobl:
> In nobody else is interested, I would like to work on this issue. In fact I already started having a look at how genericMotionEvent works and implementing a basic AndroidMouseInput class to support it

This has been open since 2023, so feel free to go ahead 👍

pavly-gerges:
> In nobody else is interested, I would like to work on this issue. In fact I already started having a look at how genericMotionEvent works and implementing a basic AndroidMouseInput class to support it

Go ahead! I could help with reviewing and testing the PR.

joliver82:
Just published a PR for this. I've tested it on three devices:

- Samsung A16 (Android 16) -> Everything worked successfully
- Xiaomi Redmi note 8 pro (Android 11) -> Not able to fully capture mouse (feature included in API 26)  not working on this device, everything else worked OK
- Xiaomi 11Lite 5G NE (Android 14) -> Everything worked successfully

I don't have any more devices I can test with. Any test and feedback is appreciated ;)

PR #2597, opened 2026-02-06
Feature - Mouse devices on android

joliver82:
This pull request adds mouse device support to android. (Issue https://github.com/jMonkeyEngine/jmonkeyengine/issues/2173 )

In addition to specific android code, I had to add a new method "setMouseGrab" to both InputManager and MouseInput (and classes implemeting it).

This changed was done because when Android changes to grab/capture mode while dragging all events after mouse down are ignored till mouse up is received. FlyByCamera and ChaseCamera changes visibility while dragging if set dragToRotate=true. So changing grab state in setVisibility method avoids being able to move the camera in this mode. Also I wanted to be able to properly capture the mouse if the game requires it like jme3 works on desktop when running setVisibility(false)

### DIFF
--- a/jme3-android/src/main/java/com/jme3/input/android/AndroidInputHandler14.java
+++ b/jme3-android/src/main/java/com/jme3/input/android/AndroidInputHandler14.java
@@ -104,7 +104,7 @@ public class AndroidInputHandler14 extends AndroidInputHandler implements View.O
         boolean isMouse = ((source & InputDevice.SOURCE_MOUSE) == InputDevice.SOURCE_MOUSE);
         if (isMouse && mouseInput != null) {
             // send the event to the mouse processor
-            consumed = ((AndroidMouseInput14)mouseInput).onHover(event);
+            consumed = consumed || ((AndroidMouseInput14)mouseInput).onHover(event);
         }
 
         return consumed;
@@ -151,7 +151,7 @@ public class AndroidInputHandler14 extends AndroidInputHandler implements View.O
         boolean isMouse = ((event.getSource() & InputDevice.SOURCE_MOUSE) == InputDevice.SOURCE_MOUSE);
         if (isMouse && mouseInput != null) {
             // send the event to the mouse processor
-            consumed |= ((AndroidMouseInput14)mouseInput).onGenericMotion(event);
+            consumed = consumed || ((AndroidMouseInput14)mouseInput).onGenericMotion(event);
         }
 
         return consumed;


riccardobl:
Looks solid, but i am not sure about the new setMouseGrab method, usually in jme we consider setCursorVisible(false) = mouse grab, isn't there a way to make this behavior work in the new android input handler?

joliver82:
Thanks for reviewing.

I know jme3 uses setVisibility as grab method but in Android it's not possible to do it this way unless we accept that drag to rotate functionality won't work when using a mouse.

As I wrote in the first comment, if changed grab state after a mouse down event, the following events (move events) are not sent by the OS till mouse up (checked in all my devices and the behavior is the same).

Notice that I updated all existing mouse handlers to include the set grab method as wrapper for setVisibility(!grab) and also updated all classes using mouse handler setVisibility to use it or setGrab (only three classes)

I don't see much issue in adding this method, in fact I feel it's a more meaningful name. Also I don't know if there're lots of users using setVisibility directly or just through current existing jme3 classes (like FlyByCamera).

riccardobl:
My goal would be to make this behave exactly like in desktop without any api change.
I did some debugging and in my device calling `requestPointerCapture()` normally hides the mouse pointer automatically, however i think the problem in this implementation is that `ACTION_CANCEL` are handled like `RELEASE`, this causes a button up event to be triggered when android cancels the click action in favor of a drag action, and in turn it breaks the chasecamera use case and shows the cursor again

riccardobl:
I've pushed the changes that fixed it for me, without the setMouseGrab, here on this branch. 
Could you please review it and test on your devices? Thanks

(feel free to revert the commit or to force push, if it is conflicting with your local branch)

joliver82:
I've checked your change in my devices and "drag to rotate" works perfectly. When I was testing this before adding the setGrab methods, maybe I didn't notice the ACTION_CANCEL events... :/ Moreover, I was not receiving the ACTION_MOVE events. Maybe those were not sent because I was returning true when I received the ACTION_CANCEL?

Thanks for the fix ;)

I've added a small commit removing the log in each mouse event. I added it when checking it worked as expected and forgot to remove it before creating the PR

jme3-android/src/main/java/com/jme3/input/android/AndroidInputHandler14.java
```diff
@@ -91,6 +101,12 @@ public boolean onHover(View view, MotionEvent event) {
             consumed = ((AndroidTouchInput14)touchInput).onHover(event);
         }
 
+        boolean isMouse = ((source & InputDevice.SOURCE_MOUSE) == InputDevice.SOURCE_MOUSE);
+        if (isMouse && mouseInput != null) {
+            // send the event to the mouse processor
+            consumed = ((AndroidMouseInput14)mouseInput).onHover(event);
```
riccardobl: shouldn't this be

```suggestion
            consumed = consumed || ((AndroidMouseInput14)mouseInput).onHover(event);
```

?

riccardobl (changes requested):
Nice!
I think there is another issue: the callbacks that are shared between touch and mouse, always fire the touch events, even if the source is the mouse. 

I think we should keep the two devices separated, we already have a setting to simulate mouse events on touch, you could add a setting that allows to simulate touch events on mouse(if you think it is needed), but there should be a way to treat the devices as entirely separated.

jme3-android/src/main/java/com/jme3/input/android/AndroidInputHandler14.java
joliver82: Not sure I understand your comment correctly... Do you mean that for example on the AndroidInputHandler14.OnTouch method both (touch screen and mouse) events are processed? One or the other will be processed differently depending on the source, there's a check for that. Also both events are received in this callback. It's the same case for joystick and mouse in the onGenericMotion method, those events are both received there.

If you mean other thing, please could you rephrase?

About the suggested change, I've accepted it. Thanks ;) Honestly I don't it'll make any difference because the event will come either from touch screen or mouse, so when reaching this point 100% times should be the default value false.

riccardobl: > AndroidInputHandler14.OnTouch method both (touch screen and mouse) events are processed?

Right, i didn't notice the check on the parent class

joliver82:
Thanks for taking the time to review this PR and the improvements ;) I'm glad it'll be merged soon.

I'll check the bounty page later today and maybe get back to you in the hub if I don't find the options to submit the expense. I'm new to this platform

PR #2612, opened 2026-02-12
Introduce cachedLocalMat and update cloning logic

capdevon:
Added a cachedLocalMat field and updated cloneFields method.


yaRnMcDonuts (approved):

riccardobl:
what is the goal here? This matrix is used only for storage in `getTransformMatrix`, there is no need to clone it

PR #2623, opened 2026-02-21
fix regression that disabled custom fpp color format in some platforms

riccardobl:
Fixes the issue reported [here](https://hub.jmonkeyengine.org/t/cartoonedgefilter-makes-the-entire-background-black/49419/9)

### DIFF
--- a/jme3-core/src/main/java/com/jme3/post/FilterPostProcessor.java
+++ b/jme3-core/src/main/java/com/jme3/post/FilterPostProcessor.java
@@ -206,14 +206,16 @@ public class FilterPostProcessor implements SceneProcessor, Savable {
         }
 
         // Determine optimal framebuffer format based on renderer capabilities
-        if (fbFormat == null) fbFormat = Format.RGB111110F;
-        if (!renderer.getCaps().contains(Caps.PackedFloatTexture)) {
-            if (renderer.getCaps().contains(Caps.FloatColorBufferRGB)) {
-                fbFormat = Format.RGB16F;
-            } else if (renderer.getCaps().contains(Caps.FloatColorBufferRGBA)) {
-                fbFormat = Format.RGBA16F;
-            } else {
-                fbFormat = Format.RGB8;
+        if (fbFormat == null) {
+            fbFormat = Format.RGB111110F;
+            if (!renderer.getCaps().contains(Caps.PackedFloatTexture)) {
+                if (renderer.getCaps().contains(Caps.FloatColorBufferRGB)) {
+                    fbFormat = Format.RGB16F;
+                } else if (renderer.getCaps().contains(Caps.FloatColorBufferRGBA)) {
+                    fbFormat = Format.RGBA16F;
+                } else {
+                    fbFormat = Format.RGB8;
+                }
             }
         }
 


jme3-core/src/main/java/com/jme3/post/FilterPostProcessor.java
```diff
@@ -206,6 +206,7 @@ public void initialize(RenderManager rm, ViewPort vp) {
         }
 
         // Determine optimal framebuffer format based on renderer capabilities
+        if (fbFormat == null) fbFormat = Format.RGB111110F;
         if (!renderer.getCaps().contains(Caps.PackedFloatTexture)) {
```
codex128: Shouldn't it be this instead?

```java
if (fbFormat == null) {
    fbFormat = Format.RGB111110F;
    if (!renderer.getCaps().contains(Caps.PackedFloatTexture)) {
        ...
    }
}
```

capdevon:
It might also be worth adding a validation step to ensure that the framebuffer format selected by the user is actually supported by the renderer.

Given this addition, we may also want to consider whether `fbFormat` should be serialized, so that the chosen format can be preserved across saves and restores.

```java
        if (fbFormat == null) {
            // Determine optimal framebuffer format based on renderer capabilities
            fbFormat = getDefaultFrameBufferFormat(renderer);
        } else {
            // User requested a specific format → verify GPU support
            if (!isFormatSupported(fbFormat, renderer)) {
                Format fallback = getDefaultFrameBufferFormat(renderer);
                logger.warning("Requested framebuffer format " + fbFormat +
                        " not supported. Falling back to " + fallback);
                fbFormat = fallback;
            }
        }
```

```java
    private Format getDefaultFrameBufferFormat(Renderer renderer) {
        Collection<Caps> caps = renderer.getCaps();
        if (caps.contains(Caps.PackedFloatTexture)) {
            return Format.RGB111110F;
        } else if (caps.contains(Caps.FloatColorBufferRGB)) {
            return Format.RGB16F;
        } else if (caps.contains(Caps.FloatColorBufferRGBA)) {
            return Format.RGBA16F;
        } else {
            return Format.RGB8;
        }
    }

    private boolean isFormatSupported(Format format, Renderer renderer) {
        Collection<Caps> caps = renderer.getCaps();
        switch (format) {
            case RGB111110F:
                return caps.contains(Caps.PackedFloatTexture);
            case RGB16F:
                return caps.contains(Caps.FloatColorBufferRGB);
            case RGBA16F:
                return caps.contains(Caps.FloatColorBufferRGBA);
            case RGB8:
                return true;
            default:
                return false;
        }
    }
```

@riccardobl

jme3-core/src/main/java/com/jme3/post/FilterPostProcessor.java
riccardobl: yes, you are right. Fixed

riccardobl:
Yes, for serialization. However, I  suggest that we avoid verifying whether the format is supported when set manually.
If a developer is setting this low-level value, it likely means they have a specific need for a particular format. Changing it, even with a warning, could make debugging more difficult. The developer can perform the necessary validation on their side if needed.

PR #2629, opened 2026-02-22
Removed  references to removed jme3-vr project

aegroto:
I had some issues building jME in local, then noted there were dangling references to the removed jme3-vr project.


riccardobl (approved):
As you've noted we are getting rid of the deprecated jme3-vr module for the next release, but it seems there were still some left overs. Thanks!

PR #2630, opened 2026-02-24
Fix depth in SDSM shadows

### LINKED ISSUE #2620: Setting 'setNumSamples' on FilterPostProcessor causes SdsmDirectionalLightShadowFilter to crash.
Test case

```
package test;

import com.jme3.light.DirectionalLight;
import com.jme3.post.FilterPostProcessor;
import com.jme3.renderer.queue.RenderQueue;
import com.jme3.shadow.SdsmDirectionalLightShadowFilter;

public class TestShadowCrash2 extends SimpleApplication {
    
    @Override
    public void simpleInitApp() {
        Geometry box = new Geometry("", new Box(0.5f, 0.5f, 0.5f));
        box.setMaterial(new Material(getAssetManager(), "Common/MatDefs/Light/Lighting.j3md"));
        box.setShadowMode(RenderQueue.ShadowMode.CastAndReceive);
        rootNode.attachChild(box);
        
        Geometry floor = new Geometry("", new Box(0.5f, 0.5f, 0.5f));
        floor.setMaterial(new Material(getAssetManager(), "Common/MatDefs/Light/Lighting.j3md"));
        floor.setShadowMode(RenderQueue.ShadowMode.CastAndReceive);
        floor.setLocalTranslation(0, -1, 0);
        floor.setLocalScale(5, 1, 5);
        rootNode.attachChild(floor);
        
        DirectionalLight light = new DirectionalLight(new Vector3f(-1, -1, -1));
        rootNode.addLight(light);
        
        fpp.setNumSamples(2);
        viewPort.addProcessor(fpp);
        
        SdsmDirectionalLightShadowFilter shadowFilter = new SdsmDirectionalLightShadowFilter(assetManager, 512, 2);
        shadowFilter.setLight(light);
        fpp.addFilter(shadowFilter);
    }

    @Override
        super.simpleUpdate(tpf);
    }
    
        TestShadowCrash2 app = new TestShadowCrash2();
        app.start();
    }
}
```

Error log

```
Feb 21, 2026 12:20:50 AM com.jme3.app.SimpleApplication start
INFO: AppSettings not set, creating default settings.
Feb 21, 2026 12:20:50 AM com.jme3.system.JmeSystemDelegate lambda$new$1
WARNING: JmeDialogsFactory implementation not found.
Feb 21, 2026 12:20:50 AM com.jme3.system.JmeDesktopSystem initialize
INFO: Running on jMonkeyEngine 3.10.0-alpha1
 * Branch: HEAD
 * Git Hash: d3efabc
 * Build Date: 2026-02-09
Feb 21, 2026 12:20:51 AM com.jme3.system.lwjgl.LwjglContext printContextInitInfo
INFO: LWJGL 3.4.0+20 context running on thread jME3 Main
 * Graphics Adapter: GLFW 3.5.0 Win32 WGL Null EGL OSMesa VisualC DLL
Feb 21, 2026 12:20:51 AM com.jme3.renderer.opengl.GLRenderer loadCapabilitiesCommon
INFO: OpenGL Renderer Information
 * Vendor: NVIDIA Corporation
 * Renderer: NVIDIA GeForce RTX 3060/PCIe/SSE2
 * OpenGL Version: 3.2.0 NVIDIA 560.94
 * GLSL Version: 1.50 NVIDIA via Cg compiler
 * Profile: Core
Feb 21, 2026 12:20:51 AM com.jme3.audio.openal.ALAudioRenderer printAudioRendererInfo
INFO: Audio Renderer Information
 * Device: OpenAL Soft
 * Vendor: OpenAL Community
 * Renderer: OpenAL Soft
 * Version: 1.1 ALSOFT 1.25.0
 * Supported channels: 64
 * ALC extensions: ALC_ENUMERATE_ALL_EXT ALC_ENUMERATION_EXT ALC_EXT_CAPTURE ALC_EXT_debug ALC_EXT_DEDICATED ALC_EXT_direct_context ALC_EXT_disconnect ALC_EXT_EFX ALC_EXT_thread_local_context ALC_SOFT_device_clock ALC_SOFT_HRTF ALC_SOFT_loopback ALC_SOFT_loopback_bformat ALC_SOFT_output_limiter ALC_SOFT_output_mode ALC_SOFT_pause_device ALC_SOFT_reopen_device ALC_SOFT_system_events
 * AL extensions: AL_EXT_ALAW AL_EXT_BFORMAT AL_EXT_debug AL_EXT_direct_context AL_EXT_DOUBLE AL_EXT_EXPONENT_DISTANCE AL_EXT_FLOAT32 AL_EXT_IMA4 AL_EXT_LINEAR_DISTANCE AL_EXT_MCFORMATS AL_EXT_MULAW AL_EXT_MULAW_BFORMAT AL_EXT_MULAW_MCFORMATS AL_EXT_OFFSET AL_EXT_source_distance_model AL_EXT_SOURCE_RADIUS AL_EXT_STATIC_BUFFER AL_EXT_STEREO_ANGLES AL_LOKI_quadriphonic AL_SOFT_bformat_ex AL_SOFT_bformat_hoa AL_SOFT_block_alignment AL_SOFT_buffer_length_query AL_SOFT_callback_buffer AL_SOFTX_convolution_effect AL_SOFT_deferred_updates AL_SOFT_direct_channels AL_SOFT_direct_channels_remix AL_SOFT_effect_target AL_SOFT_events AL_SOFT_gain_clamp_ex AL_SOFTX_hold_on_disconnect AL_SOFT_loop_points AL_SOFTX_map_buffer AL_SOFT_MSADPCM AL_SOFT_source_latency AL_SOFT_source_length AL_SOFTX_source_panning AL_SOFT_source_resampler AL_SOFT_source_spatialize AL_SOFT_source_start_delay AL_SOFT_UHJ AL_SOFT_UHJ_ex
Feb 21, 2026 12:20:51 AM com.jme3.audio.openal.ALAudioRenderer initEfx
INFO: Audio effect extension version: 1.0
Feb 21, 2026 12:20:51 AM com.jme3.audio.openal.ALAudioRenderer initEfx
INFO: Audio max auxiliary sends: 2
Feb 21, 2026 12:20:51 AM com.jme3.app.LegacyApplication handleError
SEVERE: Uncaught exception thrown in Thread[#33,jME3 Main,5,main]
java.lang.IllegalArgumentException: Material parameter is not defined: NumSamplesDepth
	at com.jme3.material.Material.checkSetParam(Material.java:515)
	at com.jme3.material.Material.setParam(Material.java:531)
	at com.jme3.material.Material.setInt(Material.java:726)
	at com.jme3.post.FilterPostProcessor.renderFilterChain(FilterPostProcessor.java:372)
	at com.jme3.post.FilterPostProcessor.postFrame(FilterPostProcessor.java:428)
	at com.jme3.renderer.pipeline.ForwardPipeline.pipelineRender(ForwardPipeline.java:127)
	at com.jme3.renderer.RenderManager.renderViewPort(RenderManager.java:1303)
	at com.jme3.renderer.RenderManager.render(RenderManager.java:1345)
	at com.jme3.app.SimpleApplication.update(SimpleApplication.java:361)
	at com.jme3.system.lwjgl.LwjglWindow.runLoop(LwjglWindow.java:735)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:825)
	at java.base/java.lang.Thread.run(Thread.java:1583)

Feb 21, 2026 12:20:51 AM com.jme3.system.JmeSystemDelegate lambda$new$0
WARNING: JmeDialogsFactory implementation not found.
Uncaught exception thrown in Thread[#33,jME3 Main,5,main]
IllegalArgumentException: Material parameter is not defined: NumSamplesDepth
Feb 21, 2026 12:20:51 AM com.jme3.audio.openal.ALAudioRenderer destroyOpenAL
INFO: OpenAL context destroyed.

Exception: java.lang.AssertionError thrown from the UncaughtExceptionHandler in thread "jME3 Main"
```
Test environment： win11, java21, jMonkeyEngine 3.10.0-alpha1

jeannekamikaze:
This fixes #2620, though I am sure not in the cleanest way. Please provide feedback or feel free to throw this whole PR away.

#2620 shows a sample app for the SDSM filter that also uses multisampling. SDSM was not written with multisampling in mind. This change makes SDSM work with and without it.

As I explained in the issue, the problem is that the depth texture that the SDSM uses becomes multisampled, so `texture2D` becomes `texture2DMS`, and the texture cannot be sampled but only fetched. Moreover, the logic in the `FilterPostProcessor` to handle multisampling mostly deals with materials, which the two compute shaders in SDSM are agnostic of.

The change is to have `SdsmFitter` check whether the input depth texture is multisampled and compile the two shaders accordingly. This introduces some CPU overhead but is the simplest solution without having to track and maintain additional state.

The crux of the change is in `MultiSample.glsllib`. I have added additional overloads there to work with/out multisampled textures.

I think there are still three problems not addressed here, though:

- The existing `getDepth(in sampler2DMS tex, in vec2 texC)` is kind of cooked. It calls `textureFetch()`, which is doing an average of all MS samples. This is rarely what you want for depth. Specifically, at an edge between geometry and background, it'll average values that range between "foreground" and "background" and give something in between. Rather, for depth, you typically want either a max or a min of the samples, depending on the algorithm. Here's a random post about this: https://wickedengine.net/2016/11/how-to-resolve-an-msaa-depthbuffer/ For SDSM in particular, I went with max. I also did not want to change the existing `getDepth()` because that might break a whole bunch of other shaders that I have not tested.
- Going with max with SDSM may defeat the purpose of multisampling a bit. Like I mentioned in #2620, for some shadow algos it's often better to run the algo on each sample and then average the results. This typically gives a smoother shadow. However, that would also increase the cost of the shader by `m_NumDepthSamples`. The max should be just as bad as in the no-multisampling case, but not worse. I'm also not entirely familiar with SDSM anyway.
- I'm not really sure it should be up to `SdsmFitter` and individual compute shader-based post-filters to deal with multisampling. I think this would ideally be abstracted away, either by having the `FilterPostProcessor` manage compute shaders better, or by resolving the MS textures before the filters are called. But this would require a larger change and more discussion.

**Testing**

I have tested the sample app in the issue with and without the line `fpp.setNumSamples(2)`.

Screenshot From 2026-02-23 19-57-17

### DIFF
--- a/jme3-core/src/main/resources/Common/MatDefs/Shadow/Sdsm/SdsmPostShadow.frag
+++ b/jme3-core/src/main/resources/Common/MatDefs/Shadow/Sdsm/SdsmPostShadow.frag
@@ -44,8 +44,8 @@ float determineShadow(int index, vec4 worldPos){
 }
 
 void main() {
-    float depth = getColor(m_DepthTexture,texCoord).r;
-    vec4 color  = getDepth(m_Texture,texCoord);
+    float depth = getDepth(m_DepthTexture,texCoord).r;
+    vec4 color  = getColor(m_Texture,texCoord);
 
     //Discard shadow computation on the sky
     if(depth == 1.0){


jme3-core/src/main/java/com/jme3/shadow/SdsmFitter.java
```diff
@@ -313,15 +319,38 @@ void cleanup() {
...
+                fitFrustumsShaderMs = buildShader(fitSource, true);
+            }
+            depthReduceShader = depthReduceShaderMs;
+            fitFrustumsShader = fitFrustumsShaderMs;
 
-        depthReduceShader = new ComputeShader(gl, reduceSource);
-        renderer.registerNativeObject(depthReduceShader);
-        fitFrustumsShader = new ComputeShader(gl, fitSource);
-        renderer.registerNativeObject(fitFrustumsShader);
+        } else if (depthTexture.getImage().getMultiSamples() == 1) {
```
codex128: else if can be simplified to just an else

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
```diff
@@ -50,25 +50,53 @@
...
-        //Load this upfront to surface any problems at init time
+        // Load this upfront to surface any problems at init time
         createComputeShader();
     }
-    private ComputeShader(ComputeShader source){
+
+    /**
+     * Creates a new compute shader from GLSL source code and a set of defines.
+     */
+    public ComputeShader(GL4 gl, String source, String[][] defines) {
```
codex128, line 73: Document how each define pair should be formatted, or use a "ComputeDefine" class.

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
```diff
@@ -50,25 +50,53 @@
...
+    private ComputeShader(ComputeShader source) {
         super();
         this.gl = source.gl;
         this.id = source.id;
         this.source = null;
     }
 
-    private void createComputeShader(){
+    private String addDefines(String source, String[][] defines) {
+        String[] sourceLines = (String[])source.split("\\r?\\n");
```
codex128: There should be a faster way to do this besides splitting the entire source code by line. Maybe limit the number of splits that can be made?

jme3-core/src/main/java/com/jme3/shadow/SdsmFitter.java
```diff
@@ -58,10 +58,16 @@ public class SdsmFitter {
...
 
-    private final ComputeShader depthReduceShader;
-    private final ComputeShader fitFrustumsShader;
+    private ComputeShader depthReduceShader;
+    private ComputeShader fitFrustumsShader;
+
+    private ComputeShader depthReduceShaderNoMs;
+    private ComputeShader fitFrustumsShaderNoMs;
+    private ComputeShader depthReduceShaderMs;
+    private ComputeShader fitFrustumsShaderMs;
```
codex128: Perhaps use an inner class that would manage each pair of shaders?

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
```diff
@@ -50,25 +50,53 @@
...
-    private ComputeShader(ComputeShader source){
+
+    /**
+     * Creates a new compute shader from GLSL source code and a set of defines.
+     */
+    public ComputeShader(GL4 gl, String source, String[][] defines) {
+        super();
+        this.gl = gl;
+        this.source = addDefines(source, defines);
+        System.out.println(this.source);
```
codex128: Delete debugging

codex128:
Unrelated to your changes, but I found a bug while testing this PR. #2631

jeannekamikaze:
Hello, thank you. Will address your comments soon. I am a Java noob and admittedly should have done a better self-review.

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
jeannekamikaze: Went for the docstring. I tried using `record` but apparently the project uses Java 8? Defining a class just for this seemed too much. And [Pair](https://docs.oracle.com/javase//9/docs/api/javafx/util/Pair.html) apparently requires `javafx.util`; I don't know if that's available on Android or all the targets you support?

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
jeannekamikaze: Yep, thanks. Single split now, then merging the two parts together.

jme3-core/src/main/resources/Common/ShaderLib/MultiSample.glsllib
```diff
@@ -40,25 +40,57 @@ vec4 getColorSingle(in sampler2DMS tex, in vec2 texC){
...
+    return textureSize(tex);
+}
+#endif // Multisampling
 
 vec4 fetchTextureSample(in sampler2D tex,in vec2 texC,in int sampleId){
+#if __VERSION__>=430
+    return texture(tex,texC);
+#else
     return texture2D(tex,texC);
+#endif
```
riccardobl: this should not be handled here. GLSLCompat already redefined texture2D to texture

jme3-core/src/main/resources/Common/ShaderLib/MultiSample.glsllib
```diff
@@ -40,25 +40,57 @@ vec4 getColorSingle(in sampler2DMS tex, in vec2 texC){
...
 }
 
 vec4 getColor(in sampler2D tex, in vec2 texC){
+#if __VERSION__>=430
+    return texture(tex,texC);
+#else
```
riccardobl: same here: GLSLCompat already handles this

jme3-core/src/main/resources/Common/ShaderLib/MultiSample.glsllib
```diff
@@ -40,25 +40,57 @@ vec4 getColorSingle(in sampler2DMS tex, in vec2 texC){
...
     return texture2D(tex,texC);
+#endif
 }
 
 vec4 getColorSingle(in sampler2D tex, in vec2 texC){
-    return texture2D(tex, texC);
+#if __VERSION__>=430
+    return texture(tex,texC);
```
riccardobl: same here: GLSLCompat already handles this

jme3-core/src/main/resources/Common/ShaderLib/MultiSample.glsllib
```diff
@@ -40,25 +40,57 @@ vec4 getColorSingle(in sampler2DMS tex, in vec2 texC){
...
+#else
+    return texture2D(tex,texC);
+#endif
 }
 
-vec4 getDepth(in sampler2D tex,in vec2 texC){
-    return texture2D(tex,texC);
+vec4 getDepth(in sampler2D tex, in vec2 texC){
+    ivec2 iTexC = ivec2(texC * vec2(textureSize(tex, 0)));
+    return texelFetch(tex, iTexC, 0);
```
riccardobl: can you revert this function to the original behavior and add a getDepthTexel for the new behavior?

jme3-core/src/main/resources/Common/ShaderLib/MultiSample.glsllib
jeannekamikaze: Unfortunately that does not work on my end:

```
com.jme3.renderer.RendererException: Compute shader compilation failed: 0:66(9): error: no function with name 'texture2D'
0:66(2): error: could not implicitly convert return value to vec4, in function `fetchTextureSample'
```

It sounds like I might be using a core profile, but I don't know why anything would be different on my end. I have not changed any of the project settings. Ubuntu 25.10, OpenJDK 21, Netbeans 29.

jme3-core/src/main/resources/Common/ShaderLib/MultiSample.glsllib
jeannekamikaze: Thanks, good catch. I actually don't need this change anymore, so reverted to the texture sample.
I had to add that version check per my other comment, though. Need to figure that one next.

jme3-core/src/main/resources/Common/ShaderLib/MultiSample.glsllib
riccardobl: It is likely because of this

https://github.com/jMonkeyEngine/jmonkeyengine/blob/a8b89f88144c3251dd8e476bd45944e051ab4fcc/jme3-core/src/main/resources/Common/ShaderLib/GLSLCompat.glsllib#L2

This, and other similar ifdef in this file should be changed to
`#if defined(FRAGMENT_SHADER} || defined(COMPUTE_SHADER)`

and COMPUTE_SHADER needs to be always appended to the compute shader defines

jeannekamikaze: Ok, thanks. I just had to import the compat file from the two SDSM compute shaders; the `texture()` defs are not under `FRAGMENT_SHADER` or any other guard.

riccardobl:
@gemini-code-assist  please review

jme3-core/src/main/resources/Common/MatDefs/Shadow/Sdsm/SdsmPostShadow.frag
```diff
@@ -43,8 +44,8 @@ float determineShadow(int index, vec4 worldPos){
 }
 
 void main() {
-    float depth = texture2D(m_DepthTexture,texCoord).r;
-    vec4 color = texture2D(m_Texture,texCoord);
+    float depth = getColor(m_DepthTexture,texCoord).r;
+    vec4 color  = getDepth(m_Texture,texCoord);
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist please remember in this project to always submit code suggestions when possible.
Submit a code suggestion now

jme3-core/src/main/java/com/jme3/renderer/opengl/ComputeShader.java
```diff
@@ -50,25 +50,55 @@
...
+    private String addDefines(String source, String[][] defines) {
+        // The #version pragma must appear before anything else. Insert the
+        // defines after it.
+        String[] sourceLines = (String[])source.split("\\r?\\n", 2);
+        StringBuilder builder = new StringBuilder();
+        builder.append(sourceLines[0] + "\n");
+        for (String[] pair : defines) {
+            builder.append("#define " + pair[0] + " " + pair[1] + "\n");
+        }
+        builder.append(sourceLines[1] + "\n");
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist you really expect a shader code to have zero new lines?

PR #2639, opened 2026-03-05
Hardened memory allocator for jme

riccardobl:
**tl;dr jME fully hardened against memory corruption with better memory management, no config needed, just add `jme3-safeallocator`.**

This PR introduces a hardened native memory allocator for Windows, Linux, macOS, and Android.

It relies on [saferalloc](https://github.com/NostrGameEngine/saferalloc), which wraps [mimalloc](https://github.com/microsoft/mimalloc) on desktop and acts as a pass-through on Android, where the system allocator is already hardened. For jme, this provides a single cross-platform allocator, removing platform-specific implementations and mitigating common heap vulnerabilities.  If hardened memory is passed to a buggy native library, exploiting memory corruption vulnerabilities becomes significantly more difficult.

It also adds a long-missing feature: native memory pressure tracking. Previously, large native allocations referenced by  ByteBuffer objects would not create enough Java heap pressure to trigger GC (java sees just a ByteBuffer that is basically an empty object), delaying native memory release. The new allocator tracks total native allocations and, when a configurable adaptive budget is exceeded, attempts to trigger the GC. The budget automatically grows or shrinks based on runtime behavior to adapt to the app memory needs.

A new submodule, `jme3-safeallocator`, enables this functionality. When included, the lwjgl3 and Android backends use SaferAllocator instead of their default allocators. It is also injected into lwjgl3 memory management, so all lwjgl3 allocations are hardened. It is bound to BufferUtils, which is used throughout jme, and is available for your own native code as well.

Important: memory allocated by SaferAllocator must also be freed by it. Always use BufferUtils.destroyDirectBuffer and never call free() directly from the JNI side.

All configuration is optional and works out of the box.

I have added the module to the examples, so you can clone this branch and try it out. It would be very helpful if someone could test it on a large existing project to evaluate whether there is any noticeable performance impact.

## Summary by CodeRabbit

* **New Features**
  * Optional safer memory allocator with native-backed allocation, phantom-reference cleanup, and adaptive soft-budgeting that coordinates GC and maintenance cycles.
  * Automatic integration into Android and LWJGL3 runtimes with transparent fallback when the safer allocator isn’t present.

* **Tests**
  * Added unit tests validating adaptive budgeting behavior and GC-trigger scenarios.

* **Chores**
  * Added safer-allocator project, native platform dependencies, and build inclusion/config updates.

### DIFF
--- a/jme3-saferallocator/src/main/java/com/jme3/util/SaferAllocMemoryGuard.java
+++ b/jme3-saferallocator/src/main/java/com/jme3/util/SaferAllocMemoryGuard.java
@@ -6,6 +6,7 @@ import java.util.function.LongSupplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.ngengine.saferalloc.SaferAlloc;
 import org.ngengine.saferalloc.SaferAllocNative;
 
 public class SaferAllocMemoryGuard {
@@ -127,7 +128,7 @@ public class SaferAllocMemoryGuard {
     private static final AtomicLong lastAdaptUpdate = new AtomicLong(0L);
     private static final AtomicLong highPressureCount = new AtomicLong(0L);
     private static final AtomicLong lowPressureCount = new AtomicLong(0L);
-    private static volatile LongSupplier allocatedBytesSupplier = SaferAllocNative::currentAllocatedBytes;
+    private static volatile LongSupplier allocatedBytesSupplier = SaferAlloc::currentAllocatedBytes;
     private static volatile LongSupplier nowSupplier = System::currentTimeMillis;
     private static volatile Runnable gcInvoker = System::gc;
 
--- a/jme3-saferallocator/src/main/java/com/jme3/util/SaferBufferAllocator.java
+++ b/jme3-saferallocator/src/main/java/com/jme3/util/SaferBufferAllocator.java
@@ -23,6 +23,7 @@ public final class SaferBufferAllocator implements BufferAllocator {
     static {
         reaperThread.setDaemon(true);
         reaperThread.start();
+        SaferAlloc.ensureLoaded();
     }
 
     public SaferBufferAllocator() {


jme3-saferallocator/src/main/java/com/jme3/util/SaferAllocMemoryGuard.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    private static void requestGC(long now) {
+        lastGCRun.updateAndGet(last -> {
+            if (now - last >= gcIntervalMillis) {
+                if(LOGGER.isLoggable(Level.FINER)){
+                    LOGGER.log(Level.FINER, "!!! Requesting GC...");
+                }
+                gcInvoker.run();
+                gcInvoker.run();
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist submit a code suggestion

jme3-saferallocator/src/main/java/com/jme3/util/SaferAllocMemoryGuard.java
```diff
@@ -0,0 +1,356 @@
...
+
+    private static String human(long bytes) {
+        if (bytes >= 1024L * 1024L * 1024L) {
+            return String.format(Locale.ROOT, "%.2f GB", bytes / (1024d * 1024d * 1024d));
+        }
+        if (bytes >= 1024L * 1024L) {
+            return String.format(Locale.ROOT, "%.2f MB", bytes / (1024d * 1024d));
+        }
+        return bytes + " B";
+    }
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist submit suggestion

jme3-saferallocator/src/main/java/com/jme3/util/SaferAllocMemoryGuard.java
```diff
@@ -0,0 +1,364 @@
...
+        if (now - lastGCRun.get() >= gcIntervalMillis) {
+            if (LOGGER.isLoggable(Level.FINER)) {
+                LOGGER.log(Level.FINER, "!!! Requesting GC...");
+            }
+
+            // Calling gc() twice is a common heuristic to increase the likelihood of a full
+            // garbage collection cycle, which is important for timely release of native memory.
+            gcInvoker.run();
+            gcInvoker.run();
+
```
richardTingle, line 237: Do we want that on a soft limit hit? (I'm thinking about GC stalls) Wouldn't we want to hint at wanting a GC and then be more insistent at a higher limit

riccardobl: I think this is fine here, because the soft limit will eventually stabilize to the application memory requirements and the gc call frequency is capped, so you will likely see this only right after you have a very big allocation (eg. you load a map) and then it'll stabilize.

I think having another limit would make the code too complex, but if we see that this is causing performance problems, we can call gc only once, it is probably enough.

richardTingle: Fair, lets give it a try and see what happens

PR #2641, opened 2026-03-11
update the Gradle build tool to v8.14.4

stephengold:
Currently Gradle v8.8 is specified, which supports building on (and with) Java 1.8 through 22.

This change causes Gradle v8.14.4 to be used instead. This should allow the Engine to be built on (and with) Java 23 and 24.

In order to support Java 25+ build environments, we'd need to upgrade to Gradle v9, but unfortunately the latest production release of JMonkeyEngine SDK (3.8.0) is based on NetBeans 25, which doesn't work right with Gradle v9. Also, the current build scripts are incompatible with Gradle v9 because they use some deprecated Gradle features (such as auto-loading test framework implementation dependencies).


stephengold:
I believe this is ready to be integrated into the "master" branch.

@yarnmcdonuts if you would like me to perform the merge, let me know.

PR #2642, opened 2026-03-16
Fix memory leaks caused by reused arrays of Spatials

parysto:
If (multiple) Geometries with Textures are being attached to and detached from the Sceengraph at runtime (in `update(...)`), the Texture memory counter increases continuously and does not reach 0 again. This happens especially after `Spatial#getWorldTranslation()` has been called.
Even after detaching the Geometries' parent node, cleaning the asset-cache and calling `System.gc()`, it does not go back to 0.

After some debugging, I found out that this is caused by how ref-variables are used to improve performance:

 * `Spatial#checkDoTransformUpdate()` uses a `Spatial[]` from `TempVars`, but does not clean up the references.
 * `ListSort#sort(...)` performs a merge-sort if there are more than 128 entries. The merge sort stores a reference to a temporary array of sorted entries in a class field, but does not clean up the references set in this array after sorting.

This means that parts of the detached sub-Sceengraph survive garbage collection. The Texture memory counter just happens to be the most visible indicator...

### DIFF
--- a/jme3-core/src/main/java/com/jme3/collision/bih/BIHNode.java
+++ b/jme3-core/src/main/java/com/jme3/collision/bih/BIHNode.java
@@ -140,7 +140,7 @@ public final class BIHNode implements Savable {
         private final BIHNode node;
         private final float min, max;
 
-        public BIHStackData(BIHNode node, float min, float max) {
+        BIHStackData(BIHNode node, float min, float max) {
             this.node = node;
             this.min = min;
             this.max = max;
--- a/jme3-core/src/main/java/com/jme3/util/ListSort.java
+++ b/jme3-core/src/main/java/com/jme3/util/ListSort.java
@@ -73,7 +73,7 @@ public class ListSort<T> {
      */
     private static final int MIN_SIZE = 128;
     private T[] array;
-    T[] tmpArray;
+    private T[] tmpArray;
     private Comparator<T> comparator;
     
     /**
--- a/jme3-core/src/test/java/com/jme3/util/ListSortTest.java
+++ b/jme3-core/src/test/java/com/jme3/util/ListSortTest.java
@@ -31,6 +31,7 @@
  */
 package com.jme3.util;
 
+import java.lang.reflect.Field;
 import java.util.Arrays;
 import java.util.Comparator;
 import static org.hamcrest.CoreMatchers.everyItem;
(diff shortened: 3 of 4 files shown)


jme3-core/src/main/java/com/jme3/scene/Spatial.java
```diff
@@ -706,15 +706,21 @@ void checkDoTransformUpdate() {
...
+                // (Spatials added to the stack might get detached from their parent
+                // after this method has been called -> they and their OGL object will not 
+                // get garbage collected when subsequent calls to checkDoTransformUpdate() 
+                // require only smaller stacks)
+                stack[j] = null;
                 //rootNode.worldTransform.set(rootNode.localTransform);
                 //rootNode.worldTransform.combineWithParent(rootNode.parent.worldTransform);
                 //rootNode.refreshFlags &= ~RF_TRANSFORM;
                 rootNode.updateWorldTransforms();
             }
```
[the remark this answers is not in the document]
riccardobl: @gemini-code-assist what are you saying? The whileloop assigns ` stack[i] = rootNode;` , am i missing something?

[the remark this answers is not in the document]
riccardobl: I know how loops work @gemini-code-assist , but in our case we break it here

```

                if ((hisParent.refreshFlags & RF_TRANSFORM) == 0) {
                    break;
                }
```

before i++

riccardobl:
Good find! Since this is likely a bug that can potentially present itself everytime that TempVars field is used without proper cleanup, maybe we should fix it right into the TempVars.release method?

https://github.com/jMonkeyEngine/jmonkeyengine/blob/f02c4e671690d2a456bed5ade9b62971af13cdbe/jme3-core/src/main/java/com/jme3/util/TempVars.java#L126-L142

we can call Arrays.fill from there, thoughts?

parysto:
Yes, calling `Arrays.fill(...)` in TempVars would yield the same result and would probably be a lot cleaner, but it would always iterate over the stack even if it wasn't used by the releasing code. I think this causes a little performance impact we might want to avoid.

capdevon:
Could `TempVars.collisionResults` have a similar issue? I haven’t looked into it in detail, but at first glance it seems the list is cleared before each use and then not fully reset afterward.

If that’s the case, it might make sense to address it in a separate PR to avoid slowing down this one.

riccardobl:
i think we should fix all those in the tempvars release(). Those small extra loops are worth the peace of mind of having an engine-wide fix, also for user code that might use this.

It won't hurt performances significantly and we can always revert the change if we noticed that when profiling.

riccardobl:
We should also call clear on bihStack, since it is technically unbound and could still use references that might themselves be used to hold references to other objects

parysto:
Thanks to both of you for the feedback!

I've also checked the other two temp-vars you mentioned, and you're right, they get never cleared afterwards.

What do the others think about always clearing all these three temp-vars in `release()` (especially @yaRnMcDonuts, since he has already approved the PR)? I'd update the PR, if you are fine with these changes, too.

(I'll add a unit test to `ListSort` in the meantime.
Edit: Added unit test and updated `ListSort` to only clean references that have actually been set by the merge-sort).

yaRnMcDonuts:
> What do the others think about always clearing all these three temp-vars in `release()` (especially @yaRnMcDonuts, since he has already approved the PR)? I'd update the PR, if you are fine with these changes, too.


Yes that sounds good, I'd bet that there's more code in the engine that has this same bug if it managed to slip through to the Spatial class where you discovered it, so there's a good chance putting the fix into the release() method may fix the same issue in other yet to be discovered places as well.

jme3-core/src/main/java/com/jme3/collision/bih/BIHNode.java
```diff
@@ -140,7 +140,7 @@ public static final class BIHStackData {
         private final BIHNode node;
         private final float min, max;
 
-        BIHStackData(BIHNode node, float min, float max) {
+        public BIHStackData(BIHNode node, float min, float max) {
```
riccardobl: lets not make this public just for unit tests.
You can create a small helper inside the test folder under the `com.jme3.collision.bih` package:
```java
public class BIHTestHelper{
    public static BIHStackData newBIHStackData(BIHNode node, float min ,float max){ 
         return new BIHStackData(node,min,max);
    }
}

```

or use reflection

jme3-core/src/main/java/com/jme3/util/ListSort.java
```diff
@@ -71,7 +73,7 @@ public class ListSort<T> {
      */
     private static final int MIN_SIZE = 128;
     private T[] array;
-    private T[] tmpArray;
+    T[] tmpArray;
```
riccardobl: don't change access modifier just for unit tests, use reflection for this

jme3-core/src/main/java/com/jme3/collision/bih/BIHNode.java
parysto: Do you know why the constructor was set to package-private instead of private? If setting it to private (now or in the future) makes sense, I'd prefer reflection.

riccardobl: > Do you know why the constructor was set to package-private instead of private?

I don't know for sure, but it is likely that it was used by some code that was refactored and it was never changed.

> If setting it to private (now or in the future) makes sense, I'd prefer reflection.

Yes,  we might do a  cleanup of all access modifiers in a future major release.

So, yes reflection is likely better here, long term.

riccardobl (approved):
Thank you!

PR #2656, opened 2026-03-22
PBR Megafix

riccardobl:
Over the years we refactored and fixed PBR a lot, this introduced a bit of compounded  weirdness, also, due to the fact that our pbr was never really completed or validated.

So, with the help of gpt-5.4 I baked this megafix, that solves pretty much everything (i think?)

## List of changes

* Specular IBL bake/runtime consistency

  * Make bake-time roughness progression match runtime `sqrt(roughness)` mip sampling in `IBLGLEnvBaker.java` and `IBLHybridEnvBakerLight.java`.
  * Keep mip chains on captured environment maps by switching the capture cubemap to trilinear minification in `GenericEnvBaker.java`.
 *  ~~Use mipmaps when sampling the baked EnvMap~~ had to revert this since the code that generates the mipmaps was **crashing** the gallium software renderer in screenshot tests, afaik it should be correct and it does work on nvidia :shrug:  might be some driver quirk
  * Remove the hardcoded workaround HDR luminance clamp during specular prefiltering.

* Shared GGX parameterization cleanup

  * Standardize the bake/runtime convention in `Math.glsl`:

    * `roughness` = perceptual roughness
    * `alpha = roughness * roughness`
    * `alpha2 = alpha * alpha`
  * Make `ImportanceSampleGGX()` and `GeometrySmith()` consistently consume `alpha`.
  * Add explicit GGX PDF helpers used by the prefilter kernel.

* Runtime PBR shading fixes

  * Normalize Lambert diffuse and combine it with Fresnel-based energy splitting in `PBR.glsllib`.
  * Guard the GGX `D` term against tiny denominators.
  * Separate diffuse AO from indirect specular occlusion by introducing `computeSpecularAO()` and removing the old uniform AO multiply over all indirect lighting.
  * Fix BRDF LUT coordinate usage in the legacy LUT helper while keeping the default runtime path on `ApproximateSpecularIBLPolynomial()`.

* PBR surface / material 

  * Clamp metallic and roughness after texture/scalar composition in `PBRLightingUtils.glsllib`.
  * Prevent bad tangent data from producing invalid TBNs; fall back to geometry normal / plain texcoords when tangent-space data is degenerate.
  * Keep basic baked light maps out of specular/F0 terms.
  * Improve spec-gloss handling by clamping roughness and reducing diffuse energy based on the specular color.
  * Add guards to multi-probe blending to avoid unstable divisions and degenerate weights.

* Add `TestPBRValidation.java` with:

  * white furnace
  * roughness sweep
  * metallic sweep
  * HDR hotspot
  * dielectric vs metal comparison

* Tangent generator bug fixes

  * Fix missing trailing-run sorts in `buildNeighborsFast()` in `MikktspaceTangentGenerator.java`.
  * Fix quad-path `TriInfo` initialization.
  * Fix the slow shared-vertex fallback so it actually stores the matched welded index.
  * Reject unsupported `Mesh.Mode.Patch` instead of pretending tangent generation supports it.


riccardobl:
before
image

after
image

before
image


after
image


before
image


after
image

yaRnMcDonuts (approved):

PR #2686, opened 2026-04-07
Update GitHub Actions and enable Dependabot

8Keep:
Updating current github actions and enabling dependabot for github actions from now on.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,10 +9,14 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 30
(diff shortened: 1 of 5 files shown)


riccardobl (approved):
Thanks!
I took the liberty of making a few changes to this PR:

* downgraded some GitHub Actions that were released very recently to slightly older versions
* added a 30 days cooldown to Dependabot

The goal is to avoid tracking the very latest releases immediately, giving some time for potential supply-chain attacks to surface before we upgrade.

8Keep:
Thanks @riccardobl, those changes sound good to me too. I don't think I can merge, let me know when you can do that.

PR #2696, opened 2026-04-07
Upgrade to Gradle 9.4.1

8Keep:
https://github.com/jMonkeyEngine/jmonkeyengine/issues/2695

@stephengold mentioned in https://github.com/jMonkeyEngine/jmonkeyengine/pull/2641 that the sdk would block gradle 9, but I see that it is already on netbeans 28 and gradle 9 now. 

So it's not a problem for this now, but do we still block jmonkey core on the sdk? Wondering if there is an agreed on policy.


8Keep:
Oh, it looks like we need to be running jdk 17+ for gradle 9. We need to first update some other dependencies... Trying to untangle the order of operations of all this.

@riccardobl we will need your PR to upgrade android build tools.

stephengold:
The [latest stable release of the JME SDK](https://github.com/jMonkeyEngine/sdk/releases/tag/v3.8.0-stable-sdk1) is "3.8.0-stable-sdk1", which was based on Netbeans 25. Netbeans didn't really support Gradle v9 prior to [Netbeans 27](https://github.com/apache/netbeans/releases/tag/27).

I think the first step in untangling is to decide whether the JME SDK is still important. If it is, then bring the SDK up to date and publish a stable release.

tonihele:
> The [latest stable release of the JME SDK](https://github.com/jMonkeyEngine/sdk/releases/tag/v3.8.0-stable-sdk1) is "3.8.0-stable-sdk1", which was based on Netbeans 25. Netbeans didn't really support Gradle v9 prior to [Netbeans 27](https://github.com/apache/netbeans/releases/tag/27).
> 
> I think the first step in untangling is to decide whether the JME SDK is still important. If it is, then bring the SDK up to date and publish a stable release.

There is the pre-release that is on Netbeans 28. While that needs a re-release to be marked as done, it is in the near future.

But in general I don't think you should consider the SDK as a blocker in this regard. For the end user the SDK will satisfy every point regardless the Gradle version used in jME itself. Only thing a SDK user would struggle with is building the latest jME version itself within the SDK but I doubt there are many people doing this.

All the SDK features work regardless of jME Gradle version.

8Keep:
I think we'll have to drop jdk 11 to upgrade.

8Keep:
Screenshot tests fail, but I checked the report and they are all minuscule changes. I am not sure why these tests are flaky, I will try to investigate later. 

This PR seems ready to go, dropping jdk 11 support

Issue #2666, opened 2026-03-28
Deprecate fbx asset loader


riccardobl:
The FBX loader was a great addition when it was first introduced. At the time, the 3D ecosystem was different and there was no truly portable format.

However, it is now falling behind due to the complexity of maintaining it and the availability of more standardized alternatives such as glTF.

For these reasons, we should deprecate the FBX loader and mark it for removal.

mondogo24:
I am seeing there is a lot of FBX related files and mentions. Most of them in their corresponding directory:

https://github.com/jMonkeyEngine/jmonkeyengine/tree/ae1715fc3beeafeab1642653412ad8eeb8c4f025/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx

With deprecate the FBX loader, you mean deprecate **only** the [FbxLoader.java](https://github.com/jMonkeyEngine/jmonkeyengine/blob/ae1715fc3beeafeab1642653412ad8eeb8c4f025/jme3-plugins/src/fbx/java/com/jme3/scene/plugins/fbx/FbxLoader.java) or the entire support?

riccardobl:
I mean deprecate the entire support, and if there are some tests or examples that depend on FBX they should be ported over to something else (ie. GLTF, j3o or obj), unless they are specific to showcasing FBX, in that case they should be marked for removal too.

Do you need any  of those FBX files your project? This is just an idea, not a final decision, if there is demand we can keep it

mondogo24:
No, I was thinking about taking care of the deprecate marking  and make the pull request. So I was asking to be sure about what to do. Thanks for the clarification!

8Keep:
The fxb loader loads into the old animation system. I would say that the removal of the old animation system is blocked on ether deprecation then removal of the fbx system or modernization of the fbx system.

8Keep:
Doing this in https://github.com/jMonkeyEngine/jmonkeyengine/pull/2715

PR #2715, opened 2026-04-19
Deprecate FBX loader

8Keep:
https://github.com/jMonkeyEngine/jmonkeyengine/issues/2666

* Also remove it from general.cfg, so that assetmanager does not know how to load fbx files by default. If users want fbx support, they should add it to assetmanager manually for their app.


riccardobl:
Thanks! 

From recent forum posts, it seems like some people might be interested in preserving and fixindg the fbx loader ( eg. @capdevon  ), so i am merging this for now, but if there is any demand, instead of removing the fbx loader in the future, we can move it to its own dedicated repo with a maintainer that is capable of testing and improving it.

PR #2720, opened 2026-04-24
Modify VS Code settings for Java configuration

riccardobl:
Some changes to vscode json:
- avoid unexpected formatting
- increase memory limits
- prevent build from being stuck in a loop due to updateBuildConfiguration:automatic in some versions of vscode's java ext


riccardobl:
@gemini-code-assist review again

PR #2750, opened 2026-05-09
Target Java 11 bytecode

8Keep:
Android gradle plugin has support for desugaring based on which min sdk you target. Is it time to bump this up to 11, along with other modernization we are doing with jme 3.10?


riccardobl:
The current ios backend is compatible only with java8, so for now we are stuck on that.
I am working to a replacement

riccardobl:
New ios backend is merged, so we are not stuck on java 8 anymore. This can be rebased on the new master branch

codex128:
What is the status of this PR? I'd like to use Cleaner from java 9, but if we're still targeting java 8 I'll go with something else.

PR #2753, opened 2026-05-09
Remove very old deprecated code from com.jme3.input package

8Keep:
This stuff has been deprecated for a very long time. I think we should remove it now.


riccardobl:
I'd probably keep this around for now, unless it is problematic

8Keep:
> I'd probably keep this around for now, unless it is problematic

How long? It's been deprecated for a long time.

riccardobl:
I'd say until it becomes a cost maintaining it

8Keep:
I disagree, it's valuable to tidy up apis after 10 years of being deprecated... There is likely zero cost to removing it. People read code far more than it's written, and people are having to read dead code is a cost. There is already a cost for it just existing.

PR #2755, opened 2026-05-09
Remove isJava8Compatible() checks from Gradle files

8Keep:
Since the project now requires JDK 8+, the JavaVersion.current().isJava8Compatible() checks are no longer needed.


8Keep:
Addressed comments

PR #2764, opened 2026-05-10
Secure native library temp extraction

riccardobl:
### Motivation
- Prevent local privilege escalation where an attacker can pre-create predictable extraction paths under `java.io.tmpdir` and cause a victim process to load attacker-supplied native libraries. 
- Replace the previous shared `jme3/natives_<hash>` cache behavior with a per-process private extraction directory to remove the race/planting window. 

### Description
- Replace creation of a predictable `jme3/natives_<hash>` directory under `java.io.tmpdir` with a freshly-created private directory returned by a new `createPrivateTempNativesDirectory(...)` helper. 
- The helper uses `Files.createTempDirectory(...)` with an owner-only POSIX permission request (`rwx------`) and falls back to the non-POSIX overload when attributes are unsupported. 
- Update imports and Javadoc to reflect the new private-temp behavior instead of a reusable shared temp subdirectory. 

### Testing
- Ran the Java compilation for the desktop module with `./gradlew :jme3-desktop:compileJava`, which completed successfully. 
- Ran `git diff --check` to validate the working tree for whitespace/patch issues and it reported no remaining problems.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ff51477330832eb3856290fcd30457)

### DIFF
--- a/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
+++ b/jme3-desktop/src/main/java/com/jme3/system/NativeLibraryLoader.java
@@ -36,8 +36,11 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.FileAttribute;
+import java.nio.file.attribute.PosixFilePermissions;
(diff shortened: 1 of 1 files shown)


riccardobl:
PSA: this was generated by codex that can be pretty dramatic. It is a good thing to fix but it's nowhere near the vulnerability it's made out to be. 
Games are not run as root normally.

PR #2770, opened 2026-05-12
Fix Android prebuilt native extraction

8Keep:
I messed up PR #2700 which caused the android examples to fail finding the native openal lib.

The native snapshot zip stores files under an `android/` top-level directory. The extraction task tried to strip that first path segment with `RelativePath`, but the task produced only empty ABI directories. As a result, Android example APKs had no packaged `.so` files and crashed at runtime when audio initialized.

Before the fix, `:jme3-android-examples:installAndroidExamples` built an APK with `mergeDebugNativeLibs NO-SOURCE`, and launching `jme3test.helloworld.HelloLoop` on an emulator failed with `UnsatisfiedLinkError: library "libopenalsoftjme.so" not found`.

before:
before-native-missing

after:
after-native-packaged

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,8 +12,6 @@ buildscript {
     }
 }
 
-import org.gradle.api.file.RelativePath
-
 // Set the license for IDEs that understand this
 ext.license = file("$rootDir/source-file-header-template.txt")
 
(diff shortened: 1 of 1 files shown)


8Keep:
this is some evidence that getting the android examples working in ci would be extremely valuable

riccardobl:
Agreed, @richardTingle is working on android screenshot tests

PR #2774, opened 2026-05-15
 Restore deprecated Android Harness and Surface for backward compatibility

### LINKED ISSUE #2765: Bring back a deprecated AndroidHarness to make it easier to move to 3.10
In https://github.com/jMonkeyEngine/jmonkeyengine/pull/2727 AndroidHarness was removed entirely. We should bring back AndroidHarness, based on top of the new fragment code, to allow people to migrate to 3.10 easier, and later fix up their app to not use the deprecated code path on their own time.

Risk: the new migration friendly deprecated AndroidHarness won't be exactly the same as the old code path, though the API should be, so we will need testing from users with real apps.

8Keep:
Fixes https://github.com/jMonkeyEngine/jmonkeyengine/issues/2765

Add a new AndroidHarness activity that emulates the old one, just wrapping a AndroidHarnessFragment.

Also changes the gradle build so the androidx stubs are treated as compile time only. A real jmonkey android application should provide its own androidx dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -179,6 +179,7 @@ tasks.register('mergedJavadoc', Javadoc) {
 
     options.overview = file("javadoc-overview.html")
     source = mergedJavadocSubprojects.collect { project(it).sourceSets.main.allJava }
+    source += fileTree("jme3-android/src/androidx-stubs/java")
     classpath = files(mergedJavadocSubprojects.collect { project(it).sourceSets.main.compileClasspath })
 }
 
--- a/jme3-android/build.gradle
+++ b/jme3-android/build.gradle
@@ -18,8 +18,6 @@ dependencies {
 }
 
 compileJava {
-    // The Android-Native Project requires the jni headers to be generated, so we do that here
-    options.compilerArgs += ["-h", "${project.rootDir}/jme3-android-native/src/native/headers"]
     options.sourcepath = files("src/androidx-stubs/java")
     options.compilerArgs += ["-implicit:none"]
 }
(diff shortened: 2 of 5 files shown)


8Keep:
Need some testing with some users existing android apps

riccardobl:
@copilot can you rebase and fix this?

riccardobl:
@copilot rebase

stephengold:
I tried to test this PR by checking out commit 016311c, doing a local install, and building the [BasicGameOnDroid](https://github.com/stephengold/BasicGame-on-Droid) app with local JME libraries in Android Studio Panda 4 Patch 1. The build succeeded, but when I ran the app on a Medium Phone API 35 emulator, the familiar blue cube did not appear. Instead I got an `AssetLoadException`:

NoLoader

The same app runs fine when built with 3.9.0-stable libraries.

8Keep:
@riccardobl Pretty sure this is related to https://github.com/jMonkeyEngine/jmonkeyengine/commit/2cd8aeb85dc026b72cd44838e79beb31f7ed4737

Let me start a new issue

riccardobl:
@stephengold  the new png loader  is in jme3-plugins  with the others, since it is not android specific anymore.
You need to uncomment this line https://github.com/stephengold/BasicGame-on-Droid/blob/44a84b40eadcc334c09e74fb8329ba61fd4bb2be/game/build.gradle#L16

8Keep:
Moving to another issue for this https://github.com/jMonkeyEngine/jmonkeyengine/issues/2830

If jme3-android requires plugins, then it should probably depend on it

stephengold:
> You need to uncomment this line

With line 16 un-commented, the app runs fine on the emulator.

riccardobl:
@nickidebruyn could you please look into this PR and see if the harness has the features your apps need?

PR #2776, opened 2026-05-15
Remove more deprecated android code and unify openal settings

riccardobl:
This PR removes the old untested android surface based path and unifies the LWJGL and android OpenAL AppSetting to simply AppSetting.OPENAL, then removes the createSettings in android fragment, that is now useless.

The goal is to let the developer use the same SimpleApplication creation pattern used in the desktop backend


richardTingle (approved):

pavly-gerges:
Why removing the surface view API?

riccardobl:
My understanding is that the fragment api + GLSurfaceView is the way to go with modern android, and I rewrote our fragment harness to fix all the issues it had, so it should be a pretty stable option now.
Is there something the surface view api without a fragment is useful for? If so I can figure out a way to bring it back

pavly-gerges:
The primary purpose of the surfaceview is to enable the seamless integration of a jMonkeyEngine Application into the already established Android Applications including Androidx Applications (not necessarily games) which were not supported by the Android hardness and Android Fragment harness at the time the surfaceview was written.

pavly-gerges:
I proposed a plan to include a CompatHarness fragment and activity that support Androidx and uses the JmeSurfaceView with very minimal code; as all the code and binding among the jMonkeyEngine application and android application lifecycles are handled within the surfaceview component...

pavly-gerges:
Other uses for the surfaceview is to enable using Android Native Views by attaching them directly to the parent android view.

PR #2778, opened 2026-05-16
Add JMH benchmark harness

8Keep:
Adds a dedicated `jme3-benchmark` module with initial JMH coverage for bounding volumes, LightList mutations/sorting, and GeometryList clearing.

The benchmark module is included only when an explicit benchmark task is requested, so normal builds, distribution archives, and publication configuration do not include it. Nothing runs it in CI because machine/load variation would make performance gating unreliable.

List available benchmarks:

```sh
./gradlew :jme3-benchmark:jmh -PjmhArgs='-l'
```

Run selected JMH options:

```sh
./gradlew :jme3-benchmark:jmh -PjmhArgs='<JMH options>'
```

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,6 +4,7 @@
 
 checkstyle = "13.3.0"
 jacoco = "0.8.12"
+jmh = "1.37"
 lwjgl3 = "3.4.1"
 angle = "2026-05-09"
 libjglios = "0.4"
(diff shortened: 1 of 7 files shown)


yaRnMcDonuts (approved):

riccardobl (changes requested):
I think we should move all benchmark to a dedicated jme3-benchmark module, so that it is easier to separate them from the actual engine source code.

8Keep:
Should be good to go now. @riccardobl there's something going on with graal in libjglios in the ci recently. I assume it's unrelated to this pr?

riccardobl:
The only problem left here is that the jme3-benchmarks module is going to be included in the release zip and published to maven central.
I suggest to include it in settings.gradle only when the jmh task is launched, similar to what i did for ios examples: 
https://github.com/jMonkeyEngine/jmonkeyengine/blob/032a493230f083908c1c9f930c41dc97c1e753e2/settings.gradle#L29-L36

8Keep:
Sorry for the delay. Fixed

PR #2784, opened 2026-05-17
Improve Lightlist performance

8Keep:
Stacked on #2778; the first commit is the JMH harness.

## Summary

- avoid copying and clearing unused retained `LightList` backing capacity
- use bulk copy paths where `LightList.update()` has no filter
- keep the retained sort buffer sized to active lights rather than backing-array capacity
- remove unused `distToOwner` state
- add focused tests for removal ordering, update ordering/filtering, and sorting after retained capacity

## Benchmarking

The harness includes mutation and retained-capacity sort benchmarks. The original unfiltered `updateFromLocalAndParent` number is not presented as a production claim: `Spatial.updateWorldLightList()` uses a global-light filter. `updateWithGlobalLightFilter` covers that representative path; compact local measurements were promising at medium/large list sizes but need longer controlled runs before a PR performance claim.

## Testing

```sh
./gradlew :jme3-core:test --tests com.jme3.light.LightListTest --tests com.jme3.light.LightSortTest :jme3-benchmark:jmh -PjmhArgs='-l'
```

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,6 +4,7 @@
 
 checkstyle = "13.3.0"
 jacoco = "0.8.12"
+jmh = "1.37"
 lwjgl3 = "3.4.1"
 angle = "2026-05-09"
 libjglios = "0.4"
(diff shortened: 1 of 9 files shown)


yaRnMcDonuts (approved):

riccardobl (changes requested):
Benchmarks should be moved into their own jme3-benchmark module,  so that we don't have to deal with them in the normal codebase.

Other than that and the `distToOwner` comment by gemini, everything looks good.

8Keep:
Should be solid now

8Keep:
@riccardobl could you take another look?

Issue #2609, opened 2026-02-11
BinaryExporter fails when serializing multiple UserData objects containing lists


riccardobl:
I was playing around with @copilot and it says it found this issue:  https://github.com/jMonkeyEngine/jmonkeyengine/pull/2607#issuecomment-3887504117

Reporting here for tracking, but i haven't validated it yet.

-----

When a Spatial has multiple UserData entries containing lists (e.g., "int-list", "string-list", "float-list"), the BinaryExporter/BinaryImporter fails to deserialize correctly. The first list in the userData HashMap deserializes successfully, but subsequent lists return null and cause an `ArrayIndexOutOfBoundsException` during deserialization.

**Steps to reproduce:**
1. Set `testLists=true` in `jme3-plugins/src/test/java/com/jme3/export/JmeExporterTest.java` (line 142)
2. Run `./gradlew :jme3-plugins:test --tests "com.jme3.export.JmeExporterTest"`
3. Observe that BinaryImporter (Node 1) returns null for int-list and string-list, while float-list works (being first in iteration order)

**Error:**
```
java.lang.ArrayIndexOutOfBoundsException: Index -3 out of bounds for length 4
    at com.jme3.export.binary.ByteUtils.rightAlignBytes(ByteUtils.java:554)
    at com.jme3.export.binary.BinaryInputCapsule.readInt(BinaryInputCapsule.java:834)
    at com.jme3.export.binary.BinaryInputCapsule.setContent(BinaryInputCapsule.java:154)
```

**Notes:**
- XMLExporter/XMLImporter works correctly with all lists
- Only affects BinaryExporter when multiple UserData objects with lists exist
- Appears to be a byte stream alignment issue in the binary format

PR #2785, opened 2026-05-18
Fix 2609 binary userdata lists

Dingyi189-eng:


xinzhoudev (commented):
Motivation
The BinaryExporter in jMonkeyEngine currently fails to correctly deserialize Savable objects that contain multiple UserData lists. When a scene object carries more than one such list, the binary import process misreads the stream, causing subsequent data to be parsed from the wrong offset. This leads to corrupted deserialization results and potential crashes, as reported in issue #2609.
The root cause lies in how BinaryInputCapsule handles the index pointer after reading certain values. The existing logic did not account for cases where multiple consecutive list structures are present, causing the read cursor to fall out of sync with the actual byte stream layout.

Summary
This PR makes two changes to address the issue:
1. BinaryInputCapsule.java — Index correction on NULL_OBJECT / DEFAULT_OBJECT sentinel values
When reading a long value that matches the NULL_OBJECT (-1) or DEFAULT_OBJECT (-2) sentinel constants defined in BinaryOutputCapsule, the index is now rewound by 4 bytes. This compensates for the fact that BinaryOutputCapsule.deflate() encodes these special int-sized markers in a compressed form, so the reader must not advance the full 8-byte stride in such cases.

⚠️ Note: A concern has been raised (by Gemini Code Assist) that this correction may itself introduce stream corruption when a genuine long value of -1L or -2L is written, since deflate() does not apply the same single-byte compression to 8-byte longs. This edge case should be validated before merge.

2. JmeExporterTest.java — Enable list serialization tests
The testLists flag in the exporter consistency test suite is changed from false to true, ensuring that list serialization and deserialization (including multiple UserData lists) is now covered by automated tests, preventing regression.

jme3-core/src/plugins/java/com/jme3/export/binary/BinaryInputCapsule.java
```diff
@@ -959,6 +959,10 @@ protected long readLong(byte[] content) throws IOException {
         }
         bytes = ByteUtils.rightAlignBytes(bytes, 8);
         long value = ByteUtils.convertLongFromBytes(bytes);
+        if (value == BinaryOutputCapsule.NULL_OBJECT
+                || value == BinaryOutputCapsule.DEFAULT_OBJECT) {
+            index -= 4;
+        }
```
JNightRider, line 965: I tested these changes without this code snippet; the test runs without issues, so I suspect it's not related to the problem.

jme3-core/src/main/java/com/jme3/scene/UserData.java
```diff
@@ -138,7 +156,16 @@ public static byte getObjectType(Object type) {
     public void write(JmeExporter ex) throws IOException {
         OutputCapsule oc = ex.getCapsule(this);
         oc.write(type, "type", (byte) 0);
-        
+
+        if (uniqueId == null) {
+            synchronized (this) {
+                if (uniqueId == null) {
```
JNightRider, line 162: I don't understand the double validation; isn't one validation enough?

JNightRider (commented):
I believe this PR contains the same changes as this [PR - #2610](https://github.com/jMonkeyEngine/jmonkeyengine/pull/2610), only this one slightly modifies the format of the prefix proposed by the previous pull request.

I'm not so sure about the other one, @riccardobl  was there any reason to close that [PR - #2610](https://github.com/jMonkeyEngine/jmonkeyengine/pull/2610)?

riccardobl:
> I'm not so sure about the other one, @riccardobl was there any reason to close that [PR - #2610](https://github.com/jMonkeyEngine/jmonkeyengine/pull/2610)?

It was more complex than what I had imagined and I didn't trust the robots to make changes that were backward compatible in this specific area, so I wanted to resolve this manually sometimes in the future (has low priority for me atm).

This PR also looks AI generated, so not something I would merge, jme's UserData are kinda messed up and have a lot of side effects

PR #2799, opened 2026-05-23
Enable Javadoc doclint errors

8Keep:
## Summary
- fail Javadoc tasks on doclint errors by removing the global doclint suppression and failOnError override
- remove module-local doclint suppressions for terrain, lwjgl3, and saferallocator
- fix the doclint errors exposed in core Javadocs and exclude generated libjglios native-image glue from iOS examples Javadocs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -175,7 +175,8 @@ tasks.register('mergedJavadoc', Javadoc) {
     destinationDir = file("dist/javadoc")
 
     options.encoding = 'UTF-8'
-    options.addStringOption('Xdoclint:none', '-quiet')
+    options.addBooleanOption('Werror', true)
+    options.addBooleanOption('Xdoclint:all,-missing', true)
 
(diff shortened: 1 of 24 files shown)


8Keep:
Rebased + now ready for review

8Keep:
This one should be good @riccardobl but I believe there's an issue with the build recently with your ios changes

riccardobl:
Thanks!

riccardobl:
Reverted this for now, because it was causing the build to fail due to javadoc errors https://github.com/jMonkeyEngine/jmonkeyengine/actions/runs/26560877797/job/78243326387

Feel free to open a new pr for this. 
Would it be possible to make the doclint errors enable only when passing a -PenableJavadocError flag, and then make a dedicated CI task to check the doc, that fails with a notice without stopping the build? It would be less frustrating for contributors i think.

PR #2802, opened 2026-05-24
Add KeyNamesTest to test key code to name mapping

ziyaom2-stack:
Add KeyNamesTest to test key code to name mapping


riccardobl:
Thanks for taking the time to look into jMonkeyEngine.

In this case, I don't think we need a unit test. Since `KeyNames` is essentially a static mapping table, testing it would mostly mean duplicating the same mapping in the test code, which adds maintenance cost without giving us much additional confidence in correctness.

PR #2804, opened 2026-05-25
Add Matrix4f affine inverse round-trip test

boni-teppanyaki:
## Summary

Adds an initial `Matrix4fTest` for `com.jme3.math.Matrix4f`, which previously had no dedicated test suite.

The test builds a typical affine transform using translation, scale, and rotation via `setTransform`, inverts it with `invert()`, multiplies the transform by its inverse with `mult()`, and verifies the result is approximately `Matrix4f.IDENTITY`.

## Motivation

`Matrix4f` is used throughout jMonkeyEngine for world/view/projection matrices, camera math, skinning, shadows, and collision transforms. However, it currently has limited direct test coverage. This test introduces a basic mathematical property of affine transforms to exercise several important methods together, and provides a starting point for future, more comprehensive `Matrix4f` tests.

### DIFF
--- a/jme3-core/src/test/java/com/jme3/math/Matrix4fTest.java
+++ b/jme3-core/src/test/java/com/jme3/math/Matrix4fTest.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2009-2021 jMonkeyEngine
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ *
+ * * Neither the name of 'jMonkeyEngine' nor the names of its contributors
+ *   may be used to endorse or promote products derived from this software
+ *   without specific prior written permission.
+ *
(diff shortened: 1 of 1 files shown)


jme3-core/src/test/java/com/jme3/math/Matrix4fTest.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright (c) 2009-2021 jMonkeyEngine
```
jaime-jmebot: This file is new in this PR, so the copyright range `2009-2021` is inaccurate. The repository header template (`source-file-header-template.txt`) emits `2009-<current year>`, and other recently added tests in this package do the same (e.g. `QuaternionTest.java` uses a current-year header).

```suggestion
 * Copyright (c) 2026 jMonkeyEngine
```

jaime-jmebot (commented):
Reviewed the test against `Matrix4f`/`Matrix3f` sources.

- The assertions are sound: `isSimilar(Matrix4f, float)` performs an absolute per-element comparison, and checking both `transform * inverse` and `inverse * transform` covers non-commutativity. The earlier suggestion in the gemini thread is already present in cb78df9.
- `Matrix3f.fromAngleAxis` normalizes via `axis.normalize()` (non-mutating), so passing the shared `Vector3f.UNIT_Y` constant is safe.
- Only issue found: the header of this newly added file carries a stale year range (see inline comment).
- Scope note, not a defect: the test only asserts the inverse round-trip property; it does not check that `setTransform` reproduced the requested translation/scale/rotation, nor does it cover `invertLocal()`, which duplicates the same math.

PR #2821, opened 2026-05-27
Preserve buffer positions during binary export

### LINKED ISSUE #2312: BinaryExporter resets buffer positions to 0 when writing them.

ittaigolde:
## Summary
- Preserve original NIO buffer positions when BinaryOutputCapsule writes FloatBuffer, IntBuffer, ByteBuffer, and ShortBuffer values
- Add a binary exporter regression test that verifies positions are unchanged and serialized contents still round-trip

Fixes #2312.

## Testing
- ./gradlew.bat :jme3-core:test --tests com.jme3.export.binary.BinaryOutputCapsuleBufferPositionTest --no-daemon
- ./gradlew.bat :jme3-core:test --no-daemon

### DIFF
--- a/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryOutputCapsule.java
+++ b/jme3-core/src/plugins/java/com/jme3/export/binary/BinaryOutputCapsule.java
@@ -913,16 +913,10 @@ final class BinaryOutputCapsule implements OutputCapsule {
             write(NULL_OBJECT);
             return;
         }
-        int position = value.position();
-        try {
-            value.rewind();
-            int length = value.limit();
-            write(length);
-            for (int x = 0; x < length; x++) {
-                writeForBuffer(value.get());
-            }
-        } finally {
-            value.position(position);
+        int length = value.limit();
+        write(length);
+        for (int x = 0; x < length; x++) {
+            writeForBuffer(value.get(x));
         }
     }
 
@@ -933,17 +927,11 @@ final class BinaryOutputCapsule implements OutputCapsule {
             write(NULL_OBJECT);
             return;
         }
-        int position = value.position();
-        try {
-            value.rewind();
-            int length = value.limit();
-            write(length);
+        int length = value.limit();
+        write(length);
 
-            for (int x = 0; x < length; x++) {
-                writeForBuffer(value.get());
-            }
-        } finally {
-            value.position(position);
+        for (int x = 0; x < length; x++) {
+            writeForBuffer(value.get(x));
         }
     }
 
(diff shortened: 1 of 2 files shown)


ittaigolde:
Could a maintainer please approve/run the gated GitHub Actions workflow for this fork PR? If the screenshot comment workflow fails, it appears to be the same infrastructure issue: `Build jMonkeyEngine` is gated for the fork PR, so the `Run Screenshot Tests` check is not created for the monitor workflow to wait on.

riccardobl:
Thank you! Nice find and patch :100:

PR #2822, opened 2026-05-27
Tolerate gated screenshot test checks

ittaigolde:
## Summary

- keep the screenshot comment workflow from failing when the screenshot check was never created
- report the missing check as a notice, which can happen when the pull request build workflow is waiting for maintainer approval

## Reasoning

Fork pull requests can leave the `Build jMonkeyEngine` workflow in `action_required` before any jobs are created. In that case the `Run Screenshot Tests` check never exists, but the `pull_request_target` comment workflow still runs and currently fails while waiting for that check.

This keeps real screenshot failures working the same way, while avoiding a misleading failed check when there was no screenshot test run to inspect.

### DIFF
--- a/.github/workflows/screenshot-test-comment.yml
+++ b/.github/workflows/screenshot-test-comment.yml
@@ -24,6 +24,8 @@ jobs:
         run: sleep 120
 
       - name: Wait for Screenshot Tests to complete
+        id: wait-for-screenshot-tests
+        continue-on-error: true
         uses: lewagon/wait-on-check-action@9312864dfbc9fd208e9c0417843430751c042800 # v1.7.0
         with:
           ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


riccardobl:
I think this will make the screenshot test fail silently if it breaks for whatever reason, what do you think @richardTingle ?

Maybe we need to check for an outcome or something  to make sure it is failing due to missing approval.

richardTingle:
Not posting the comment isn't a huge deal. The pipeline would still fail, it just wouldn't produce a comment explaining why.

But if this is going to happen regularly now pipelines require approval it might be necessary to revisit the auto comment (it may just not be possible anymore).

Really the comment was important when the tests were new to explain them, now they're embedded it isn't quite so essential

riccardobl:
> Not posting the comment isn't a huge deal. The pipeline would still fail, it just wouldn't produce a comment explaining why.
> 
> But if this is going to happen regularly now pipelines require approval it might be necessary to revisit the auto comment (it may just not be possible anymore).
> 
> Really the comment was important when the tests were new to explain them, now they're embedded it isn't quite so essential

makes sense, i think it would be enough to generate an annotation with the same text using https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message

PR #2824, opened 2026-05-27
Don't force non-srgb capable framebuffer in desktop backend

### LINKED ISSUE #2823: "Couldn't find matching GLX visual" during LWJGL v3 display initialization
I tested JMonkeyEngine v3.10.0-alpha5 by plugging it into my [VhacdTuner](https://github.com/stephengold/VhacdTuner) app. Something fails during display initialization:

```console
May 27, 2026 12:18:29 PM com.jme3.app.LegacyApplication handleError
java.lang.RuntimeException: Failed to create SDL window: Couldn't find matching GLX visual
	at com.jme3.system.lwjgl.LwjglWindow.createContext(LwjglWindow.java:338)
	at com.jme3.system.lwjgl.LwjglWindow.initInThread(LwjglWindow.java:693)
	at com.jme3.system.lwjgl.LwjglWindow.run(LwjglWindow.java:1105)
	at java.base/java.lang.Thread.run(Thread.java:1583)

May 27, 2026 12:18:30 PM com.jme3.system.lwjgl.LwjglWindow run
SEVERE: Display initialization failed. Cannot continue.
```

This is with jme3-lwjgl3 on Linux. With jme3-lwjgl (that is, LWJGL v2) the app runs normally.

The same app ran normally with v3.10.0-alpha3, so I think this is a recent regression.

riccardobl:
As reported here: https://github.com/jMonkeyEngine/jmonkeyengine/issues/2823

When using OpenGL (instead of the default ANGLE_GLES) in the new desktop renderer backend, the window creation can fail.

This surfaced two issues:
- ~~The OpenGL path uses GLX instead of EGL~~
- the backend requests specifically a non-srgb capable framebuffer, that might limit the compatibility with some platforms

This PR fixes both. 
_(edit: EGL is not forced on X11, since gemini says some X11 platforms might not be compatible)_

closes https://github.com/jMonkeyEngine/jmonkeyengine/issues/2823


stephengold:
I verified this fix by using commit 4f26497 to run `TestBrickTower` using LWJGL-OpenGL3.

PR #2835, opened 2026-05-29
Memory allocators overhaul

### LINKED ISSUE #1990: AndroidNativeBufferAllocator memory leak

riccardobl:
This PR improves memory allocators in every backend

- add memory management to android and ios allocators: closes https://github.com/jMonkeyEngine/jmonkeyengine/issues/1990
- update to saferalloc 0.0.10 with iOS support and 16kb pages for android
- default to saferalloc based allocators in ios and android when `jme3-saferallocator` dependency is present
- rework LWJGL3 allocator to be fully threadsafe 
- deprecate `ConcurrentLWJGLBufferAllocator` and turn it to an alias of `LWJGLBufferAllocator` (since it is threadsafe now)
- deprecate `ReflectionAllocator`  that is not supported in modern java nor guaranteed to work in non hostpot jvms
- fix some issues with saferalloc memory pressure estimation
- use saferalloc in screenshot tests


riccardobl:
/gemini review again, i did not apply the AtomicIntegerFieldUpdater suggestion because it is a micro-optimization not worth the complexity

PR #2845, opened 2026-06-05
Fix #2354: Cache and reuse VertexBuffers in MorphControl

### LINKED ISSUE #2354: Cache and reuse VertexBuffers in MorphTarget
When the set of active morph targets changes often it keeps reuploading data to the GPU due to:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/8ba6574ef6c2cba463e2f03a87263c4074466e10/jme3-core/src/main/java/com/jme3/anim/MorphControl.java#L313
Because `MorphTarget` holds `FloatBuffer`s instead of `VertexBuffer`s that could simply be bound instead of overwritten:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/scene/mesh/MorphTarget.java#L13

Ojas2095:
Fixes #2354 by caching VertexBuffers in MorphControl to avoid re-uploading buffer data to the GPU every frame.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/anim/MorphControl.java
+++ b/jme3-core/src/main/java/com/jme3/anim/MorphControl.java
@@ -79,6 +79,8 @@ public class MorphControl extends AbstractControl implements Savable {
     private float[] tmpNormArray;
     private float[] tmpTanArray;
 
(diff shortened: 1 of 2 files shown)


riccardobl:
I won't even waste time reviewing this, since it is obvious slop.

PR #2867, opened 2026-06-10
Add low-level hook for SDL3 input events

### LINKED ISSUE #2864: Add low-level hook for SDL3 input event
Feature request from [the forum](https://hub.jmonkeyengine.org/t/jme-3-10-0-beta1/49603/38?u=riccardoblb)

Ivan1pl:
Resolves #2864

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -1123,7 +1123,11 @@ public abstract class LwjglWindow extends LwjglContext implements Runnable {
 
     private void dispatchSDLEvent(SDL_Event event) {
         for (SdlEventListener listener : sdlEventListeners.getArray()) {
-            listener.onSDLEvent(event);
+            try {
+                listener.onSDLEvent(event);
+            } catch (Exception e) {
+                LOGGER.log(Level.SEVERE, "Error handling SDL event", e);
+            }
         }
         if (keyInput instanceof SdlEventListener) {
             ((SdlEventListener) keyInput).onSDLEvent(event);
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
@@ -37,6 +37,7 @@ import com.jme3.input.JoyInput;
 import com.jme3.input.KeyInput;
 import com.jme3.input.MouseInput;
 import com.jme3.input.TouchInput;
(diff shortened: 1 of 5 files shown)


jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
```diff
@@ -255,6 +257,14 @@ public void removeWindowSizeListener(WindowSizeListener listener) {
         windowSizeListeners.remove(listener);
     }
 
+    public void registerSdlEventListener(SdlEventListener listener) {
```
capdevon, line 260: I wonder if `registerSdlEventListener(...)` could be renamed to `addSdlEventListener(...)` for consistency with `removeSdlEventListener(...)`.

Using complementary method pairs (`add`/`remove`, `get`/`set`, etc.) is a well-established Java API convention and tends to make APIs more intuitive. Since the method simply adds the listener to an internal collection, `addSdlEventListener()` also seems to better reflect what it actually does.

capdevon (changes requested):

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java
Ivan1pl: True. I went for consistency with existing API for `WindowSizeListener`: `registerWindowSizeListener` and `removeWindowSizeListener`. With that in mind, should I rename this to `addSdlEventListener`, or keep it as it is?

capdevon: You're right, and I owe you an apology for assuming your suggestion was inconsistent with the existing API. Your reply made it clear that you were actually following the current `WindowSizeListener` naming (`registerWindowSizeListener` / `removeWindowSizeListener`).

At this point, I see two possible approaches:

1. **Adopt the `add/remove` convention for new APIs.** In that case, this PR could introduce `addSdlEventListener()` / `removeSdlEventListener()`, and a follow-up PR could deprecate `registerWindowSizeListener()` in favor of `addWindowSizeListener()`, while keeping backward compatibility for the time being.

2. **Keep consistency with the existing API.** That would mean naming the new methods `registerSdlEventListener()` / `removeSdlEventListener()`, matching the current `WindowSizeListener` API even though the `register/remove` pairing is less consistent with common Java naming conventions.

That said, this is only my opinion. The final decision on API direction should rest with the engine maintainers.
@riccardobl

riccardobl: I think this PR is fine, i wouldn't worry too much about the method names here. These are basically exposed internal APIs.

Issue #2919, opened 2026-07-29
NullPointer on Geometry.setMorphState(String, float)


DieterDePaepe:
(Created with help from AI)

`Geometry.setMorphState(String, float)` throws NPE on a geometry whose morph state was never set

## Summary

`Geometry` allocates its `morphState` array lazily. `setMorphState(float[])` and
`getMorphState()` both allocate it on demand, but the two name-based overloads 
`setMorphState(String, float)` and `getMorphState(String)` index the field
directly. Calling either of them before one of the array-based methods throws:

```
java.lang.NullPointerException: Cannot store to float array because "this.morphState" is null
        at com.jme3.scene.Geometry.setMorphState(Geometry.java:632)
```

## Version

Reproduced on **3.8.1-stable**, **3.9.0-stable** and **3.10.0-beta1** (the newest
release on Maven Central), and the code is unchanged on current `master`
(`65e413232`).

## Reproducer

No `Application` or render loop needed this is pure scene-graph state:

```java
import com.jme3.scene.Geometry;
import com.jme3.scene.Mesh;
import com.jme3.scene.mesh.MorphTarget;

public class Repro {
    public static void main(String[] args) {
        Mesh mesh = new Mesh();
        mesh.addMorphTarget(new MorphTarget("morph"));
        Geometry geometry = new Geometry("body", mesh);

        System.out.println("morph index = " + mesh.getMorphIndex("morph")); // 0
        geometry.setMorphState("morph", 1f);                                // NPE
    }
}
```

`getMorphState(String)` fails the same way, at `return morphState[index];`.

## Cause

In `jme3-core/src/main/java/com/jme3/scene/Geometry.java`, the array-based setter
allocates:

```java
public void setMorphState(float[] state) {
    if (mesh == null || mesh.getMorphTargets().length == 0) {
        return;
    }
    int nbMorphTargets = mesh.getMorphTargets().length;
    if (morphState == null) {
        morphState = new float[nbMorphTargets];   // <-- allocated here
    }
    System.arraycopy(state, 0, morphState, 0, morphState.length);
    this.dirtyMorph = true;
}
```

â€¦while the name-based one immediately below it does not:

```java
public void setMorphState(String morphTarget, float state) {
    int index = mesh.getMorphIndex(morphTarget);
    if (index >= 0) {
        morphState[index] = state;   // <-- NPE if nothing allocated it yet
        this.dirtyMorph = true;
    }
}
```

[...]

PR #2920, opened 2026-07-29
Fix NPE in Geometry's morph state (#2919)

DieterDePaepe:
Geometry allocates its morphState array lazily. setMorphState(float[]) and getMorphState() both create it on demand, but the name-based overloads setMorphState(String, float) and getMorphState(String) indexed the field directly, so calling either one first threw NullPointerException:

    java.lang.NullPointerException: Cannot store to float array because
    "this.morphState" is null

Nothing else populates the array. MorphControl only reads it, and the control itself is attached by GltfLoader only when the asset carries an animation channel targeting "weights". An application driving morph targets from its own code -- expression blending, a slider, anything procedural -- therefore has an unallocated array precisely when it reaches for the name-based setter, which is the natural one to use since it is the only API that resolves a morph by name.

Route both overloads through getMorphState(), which already allocates on demand, so the by-name and by-array paths behave alike. The lookup guard is unchanged: an unknown name still writes nothing and reads -1.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/scene/Geometry.java
+++ b/jme3-core/src/main/java/com/jme3/scene/Geometry.java
@@ -629,7 +629,7 @@ public class Geometry extends Spatial {
     public void setMorphState(String morphTarget, float state) {
         int index = mesh.getMorphIndex(morphTarget);
         if (index >= 0) {
-            morphState[index] = state;
+            getMorphState()[index] = state;
             this.dirtyMorph = true;
         }
     }
@@ -677,7 +677,7 @@ public class Geometry extends Spatial {
         if (index < 0) {
             return -1;
         } else {
-            return morphState[index];
+            return getMorphState()[index];
         }
     }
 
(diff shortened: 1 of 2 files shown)


jaime-jmebot (approved):
Reviewed the two changed files plus the surrounding call sites (`Mesh`, `MorphControl`).

- `setMorphState(String, float)` and `getMorphState(String)` now both reach the array through the allocating `getMorphState()`, so the NPE is gone for callers that hit the by-name API first.
- `Mesh.getMorphIndex()` returns -1 for an unknown name, so the documented contract ("no state set" / "-1 returned") is unchanged; `MorphControl` only reads the array and is unaffected by the added allocation.
- The new tests are genuine regression tests: both the set-by-name and get-by-name cases invoke a name-based accessor on a Geometry whose `morphState` is still null, so they fail on master and pass with this change; the unknown-name case covers the guard.
- Non-blocking, no change requested: `setMorphState(String, float)` still NPEs when `mesh` is null, whereas `setMorphState(float[])` returns early. The getters behave the same way as before, so this is pre-existing asymmetry rather than something this PR needs to resolve — a follow-up would be reasonable if full parity is wanted.

PR #2927, opened 2026-08-04
fix LWJGL3 Canvas Frame Rate

riccardobl:


jme3-lwjgl3/src/main/java/com/jme3/system/lwjglx/Win32GLPlatform.java
```diff
@@ -42,6 +53,47 @@
  */
 final class Win32GLPlatform extends PlatformWin32GLCanvas implements LwjglxGLPlatform {
 
+    @Override
+    public long create(Canvas canvas, GLData data, GLData effective) throws AWTException {
+        effective.swapInterval = null;
```
JNightRider, line 58: Windows already implements the function for the `swapInterval` call, so this is an unnecessary duplication ([see](https://github.com/jMonkeyEngine/lwjgl3-awt/blob/960ef4b64b8677c69a290d46d6f7af8b1932f438/src/org/lwjgl/opengl/awt/PlatformWin32GLCanvas.java#L280)).

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -575,18 +590,22 @@ public void run() {
...
                 if (renderable.get() && canvas.hasContext() && canvas.isValid()) {
                     try {
                         if (allowSwapBuffers && autoFlush) {
                             // calls swap buffers | lock, etc.
-                            try {
-                                canvas.lock();
-                                canvas.swapBuffers();
-                            } finally {
-                                canvas.unlock();
+                            if (canvas.lock()) {
```
JNightRider, line 602: I think this could be a problem, since if the `lock()` function fails and throws an exception, the application could crash, given that the `unlock()` function would never be called.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -305,8 +320,8 @@ public void doDisposeCanvas() {
         /**
          * This is where you actually draw on the canvas (framebuffer).
          */
-        public void swapBuffers() {
-            platformCanvas.swapBuffers();
+        public boolean swapBuffers() {
```
JNightRider, line 323: I think this is unnecessary...

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -272,12 +285,14 @@ public GLData getGLDataEffective() {
          * To start drawing on the AWT surface, the AWT threads must be locked to
          * avoid conflicts when drawing on the canvas.
          */
-        public void lock() {
+        public boolean lock() {
```
JNightRider, line 288: I think this is unnecessary...

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -113,6 +114,18 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
     /** Logger class. */
     private static final Logger LOGGER = Logger.getLogger(LwjglCanvas.class.getName());
 
+    static boolean swapBuffersAndCheckNativeVsync(boolean nativeVsyncEnabled,
```
JNightRider, line 117: I really don't think this static method is necessary, since it is only used once and could be integrated directly into the code block where it is required.

JNightRider (changes requested):
I think this PR needs a bit more work and cleanup... I haven't tested it, as I can't reproduce the error.

PR #2935, opened 2026-08-16
fix: patch for the main thread (awt | lwjgl3)

JNightRider:
This PR fixes execution on macOS on the main thread; the `invokeLater()` method is used to run the loop. This applies only to macOS, as it uses the [main thread](https://github.com/jMonkeyEngine/jmonkeyengine/blob/c40043a3b58be3af3f6b99e109d7a4a42ba77204/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglWindow.java#L700) for rendering; other platforms are not affected.

### DIFF
--- a/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
+++ b/jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
@@ -340,7 +340,7 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
                 hasNativePeer.set(false);
                 reinitcontext.set(true);
 
-                 while (reinitcontext.get()) {
+                while (reinitcontext.get() && parallel.get()) {
                     try {
                         lock.wait();
                     } catch (InterruptedException ex) {
@@ -348,8 +348,6 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
                         return;
                     }
                 }
-
-                 reinitcontext.set(false);
             }
 
             // GL context is dead at this point
@@ -410,6 +408,13 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
 
     /** Notify if there is a change in canvas dimensions. */
     private final AtomicBoolean needResize = new AtomicBoolean(false);
+
+    /**
+     * Flag indicating whether a custom thread is used to separate GL rendering
+     * from the EDT; its value is false if the main thread is used via the
+     * {@code SwingUtilities.invokeLater() } function.
+     */
+    private final AtomicBoolean parallel = new AtomicBoolean(false);
 
     /**
      * Flag that uses the context to check if it is initialized or not, this prevents
@@ -526,7 +531,6 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
             if (needResize.getAndSet(false)) {
                 settings.setResolution(framebufferWidth, framebufferHeight);
                 listener.reshape(framebufferWidth, framebufferHeight, framebufferWidth, framebufferHeight);
-                listener.reshape(framebufferWidth, framebufferHeight);
             }
 
             synchronized (lock) {
@@ -588,9 +592,6 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
                             } finally {
                                 canvas.unlock();
                             }
-
-                            // Sync the display on some systems.
-                            Toolkit.getDefaultToolkit().sync();
                         }
                     } catch (Throwable ex) {
                         listener.handleError("Error while swapping buffers", ex);
@@ -607,6 +608,17 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
             if (needClose.get()) {
                 break;
             }
+
+            if (! parallel.get()) {
+                // Sync the display on some systems.
+                Toolkit.getDefaultToolkit().sync();
+                break;
+            }
+        }
+
+        if (!parallel.get() && !needClose.get()) {
+            SwingUtilities.invokeLater(() -> run());
+            return;
         }
 
         deinitInThread();
@@ -669,6 +681,12 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
         if (this.contextFlag.get()) {
             return;
         }
+        /*
+         * Note that JME does not run on a thread parallel to the AWT EDT; 
+         * this applies only to macOS.
+         */
+        this.parallel.set(Platform.get() != Platform.MACOSX);
+
         // create context
         super.create(waitFor);
         this.contextFlag.set(true);
@@ -720,8 +738,8 @@ public class LwjglCanvas extends LwjglWindow implements JmeCanvasContext, Runnab
 
         RENDER_CONFIGS.computeIfAbsent(settings.getRenderer(), (t) -> {
             return (data) -> {
-                data.majorVersion = 2;
-                data.minorVersion = 0;
+                data.majorVersion = 3;
+                data.minorVersion = 2;
             };
         }).accept(glData);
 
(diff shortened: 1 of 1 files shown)


jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -526,7 +531,6 @@ public void run() {
             if (needResize.getAndSet(false)) {
                 settings.setResolution(framebufferWidth, framebufferHeight);
                 listener.reshape(framebufferWidth, framebufferHeight, framebufferWidth, framebufferHeight);
```
jaime-jmebot, line 533: This removes the call to `SystemListener.reshape(int, int)`. That overload is deprecated but explicitly "kept only for backward compatibility", and `LwjglWindow.updateSizes()` still invokes both overloads. After this change an AWT-canvas app whose `SystemListener`/`Application` overrides only the 2-arg method silently stops receiving resize notifications when the canvas is resized, while the same app on the `Display` path still gets them. This is also outside the scope of the main-thread fix. Keep both calls unless the removal is deliberate and documented:

```suggestion
                settings.setResolution(framebufferWidth, framebufferHeight);
                listener.reshape(framebufferWidth, framebufferHeight, framebufferWidth, framebufferHeight);
                listener.reshape(framebufferWidth, framebufferHeight);
```

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -607,6 +608,17 @@ public void run() {
             if (needClose.get()) {
                 break;
             }
+
+            if (! parallel.get()) {
+                // Sync the display on some systems.
+                Toolkit.getDefaultToolkit().sync();
+                break;
+            }
```
jaime-jmebot, line 616: `Toolkit.getDefaultToolkit().sync()` was moved out of the swap block, so it now runs only when `parallel` is false (macOS). On Windows/Linux the AWT canvas no longer syncs the toolkit after `canvas.swapBuffers()`, even though the PR description states that other platforms are not affected. If the sync was only ever needed on macOS, say so in the commit message; otherwise keep the call in the swap block (after `canvas.unlock()`), where it was, and keep this branch free of it.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
```diff
@@ -720,8 +738,8 @@ protected void createContext(AppSettings settings) {
 
         RENDER_CONFIGS.computeIfAbsent(settings.getRenderer(), (t) -> {
             return (data) -> {
-                data.majorVersion = 2;
-                data.minorVersion = 0;
+                data.majorVersion = 3;
+                data.minorVersion = 2;
```
jaime-jmebot, line 742: The fallback renderer configuration changed from GL 2.0 to GL 3.2 here. It is unrelated to the main-thread fix (it only triggers for renderer strings missing from `RENDER_CONFIGS`) and it changes which context is requested on every platform. Bundling it makes the diff harder to reason about and to bisect; please split it into its own PR or state why the canvas path needs 3.2 as the fallback.

jaime-jmebot (commented):
Reviewed `jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java` against base `c40043a` (the only source file that differs from the base).

- The core change is coherent: on macOS the loop is executed via `SwingUtilities.invokeLater()` (`parallel == false`) so it runs on the AWT/main thread, `removeNotify()` no longer blocks the EDT waiting for the render thread, and `deinitInThread()` is still reached when `needClose` is set.
- Three items need attention: a removed deprecated `reshape(int,int)` callback, `Toolkit.getDefaultToolkit().sync()` dropped for the non-macOS (parallel) path, and an unrelated GL fallback version change — see inline comments.
- `LwjglWindow`, `AwtPanelsContext` and `LwjglContext` are unchanged, so the stated macOS-only scope otherwise holds.

jaime-jmebot (commented):
No new state since the previous review on this revision: the review request in the timeline (`2026-09-17T16:52Z`) precedes my submitted review, the head content of `LwjglCanvas.java` is unchanged (the 2-arg `listener.reshape(...)` is still missing in `run()` while `LwjglWindow.updateSizes()` still calls both overloads, `Toolkit.getDefaultToolkit().sync()` still sits in the `!parallel` branch outside the swap block, and the fallback render config still requests 3.2), and none of my three inline threads has an author reply. Nothing new to add; re-posting the same findings would be duplicate noise.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
JNightRider: When I started working on this PR, the new `reshape()` method called the old method (causing a duplicate call) but it seems this has been removed (in this PR #2881).

Was there really any difference between having the new method call the old one and making the call explicitly whenever necessary?

_~working on the changes~_

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
JNightRider: Even though it is a minor change that doesn't actually affect anything (since the minimum supported version is already set to that value) should I create a separate PR for it?

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
jaime-jmebot: Good question — and yes, there is one real difference: the two mechanisms replace each other, they don't stack.

- **Default delegates:** `SystemListener.reshape(int,int,int,int)` forwards to the deprecated 2-arg method, so a context only needs to call the 4-arg one.
- **Both defaults empty (current state):** each context has to call both overloads, like `LwjglWindow.updateSizes()` does.

Mixing them is exactly what produced the duplicate you saw: a delegating default *plus* an explicit 2-arg call means a listener that only overrides the old `reshape(int,int)` gets notified twice. What you have now (both defaults empty in `SystemListener`, both calls in `LwjglCanvas`) matches the Display path and gives each listener exactly one live callback, so it's the safer arrangement. The delegating variant would be tidier (single call site everywhere), but it moves the compatibility shim into `jme3-core`'s public interface and changes what every backend does, so keeping it local in the context is the lower-risk option.

Thanks for pushing the 2-arg call back in — that closes the item I raised. Tiny nit: `LwjglWindow.updateSizes()` calls the 4-arg first and then the 2-arg, while the canvas does the opposite; matching that order would keep both paths identical for listeners that override both overloads.

jme3-lwjgl3/src/main/java/com/jme3/system/lwjgl/LwjglCanvas.java
jaime-jmebot: Fair point. That fallback only fires for renderer strings that aren't in `RENDER_CONFIGS` (the deprecated `LWJGL_OPENGL2`/`30`/`31`, or a custom string), and the canvas only ever registers 3.2+ configs, so requesting 3.2 there is at least consistent with what this context supports.

So it's optional rather than required: if you'd rather keep this PR focused on the main-thread fix, moving it to its own PR is nice-to-have; otherwise a one-line note in the PR description explaining why the fallback moved to 3.2 is enough for me. Not a blocker either way.

jaime-jmebot (commented):
Replied in both existing review threads:

- **reshape (id 4054707594):** explained why the two approaches are mutually exclusive (a delegating 4-arg default + an explicit 2-arg call causes the double notification), confirmed the current head — both defaults empty in `SystemListener` and both calls in `LwjglCanvas` — is the consistent, local-to-context fix, and noted the minor call-ordering difference vs `LwjglWindow.updateSizes()`.
- **GL fallback (id 4054718801):** reassured that the 3.2 fallback is consistent with the canvas's registered 3.2+ configs and is optional to split out; a one-line note in the PR description would suffice.

No new findings to add as inline comments and the head commit is unchanged, so no further review submission was made.

jaime-jmebot (commented):
No reply is warranted here: the triggering comment (id 4054718801, the question about splitting the GL fallback change into its own PR) is already answered in that same review thread by my reply (id 4058640197), and the PR head is unchanged since — the `reshape` 2-arg call is back in `run()`, the toolkit `sync()` still sits in the `!parallel` branch, and the fallback render config still requests 3.2. Re-posting the same points would just add duplicate noise.

(Note: the `react_to_trigger` tool referenced in my operating rules is not present in this session's tool set, so no reaction could be recorded.)

PR #2940, opened 2026-08-22
FlyByCamera:  prevent field-of-view from zooming past 179 degrees (issue #2939)

### LINKED ISSUE #2939: FlyCam shouldn't zoom out past 180-degree field-of-view
For perspective cameras, there is a check in `FlyByCamera.zoomCamera()` to prevent the field-of-view angle from going negative or "too close" to zero, which is considered an invalid state: 

https://github.com/jMonkeyEngine/jmonkeyengine/blob/c40043a3b58be3af3f6b99e109d7a4a42ba77204/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java#L447-L453

However, there's nothing to prevent the angle from exceeding 180 degrees.  This puts the camera in an invalid state where, for instance, `getFrustumTop() < cam.getFrustumBottom()`.

For reasonable values of `zoomSpeed` (such as 20) it's easy to zoom out a perspective camera until the field-of-view exceeds 180 degrees.

To prevent this, the check in `zoomCamera()` should be changed to something like:

```java
if (newFov > 0.01f && newFov < 179f) {
```

Testing suggests this is not a new issue. It was present in JME v3.2.4-stable.

stephengold:
This adds a runtime check to `FlyByCamera` so that "zooming out" a perspective camera cannot cause its field-of-view to exceed 179 degrees of arc.

### DIFF
--- a/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
+++ b/jme3-core/src/main/java/com/jme3/input/FlyByCamera.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2025 jMonkeyEngine
+ * Copyright (c) 2009-2026 jMonkeyEngine
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


stephengold:
I believe this PR is ready for integration.

Issue #2948, opened 2026-08-31
Android virtual joystick disappeared due to virtual keyboard events


terranprog:
I saw this issue happen once.

jME 3.10.0-beta2

To reproduce:

1. Set virtual joystick to AppSettings.VIRTUAL_JOYSTICK_AUTO
2. Run application on an Android phone with touchscreen and no physical keyboard -> virtual joystick will appear

I used the app for many hours. One time, the virtual joystick just disappeared and wouldn't come back. I had to restart the app to get it back.
In VIRTUAL_JOYSTICK_AUTO, the joystick should disappear if a keyboard event is detected. This should not have been possible as my device does not have a keyboard.

A.I. theorised that android can send virtual keyboard events even if there is no physical keyboard. And that the joystick handler should differentiate between physical keyboard events and virtual keyboard events.

riccardobl:
Fixed by https://github.com/jMonkeyEngine/jmonkeyengine/pull/2949 . Thanks!

PR #2949, opened 2026-08-31
Prevent virtual keyboard events from hiding the virtual joystick

terranprog:
Fix for #2948 

This PR adds extra checks to make sure only a real physical keyboard can cause the virtual joystick to disappear when in VIRTUAL_JOYSTICK_AUTO mode.

Disclaimer: I asked A.I. to diagnose this observed issue and it suggested the following code fix. I have run my app with this code change and have not seen the issue reoccur. I did not write the fix myself but it makes sense as far as I can tell.


riccardobl:
I added a commit to this PR.
This was basically the correct patch, it only needed some tweaks to be compatible with android < 16.

PR #2950, opened 2026-09-03
Fix misuse of constants

riccardobl:
This PR fixes several cases of constants misuse that could end up being mutated, and one case where the code used the wrong constant.

Since passing references is often intentional in jme, I made sure to clone objects only when passing a reference was already causing the internal state to break.

### DIFF
--- a/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
+++ b/jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
@@ -902,7 +902,7 @@ public class TerrainPatch extends Geometry {
         oc.write(totalSize, "totalSize", 16);
         oc.write(quadrant, "quadrant", (short)0);
         oc.write(stepScale, "stepScale", Vector3f.UNIT_XYZ);
-        oc.write(offset, "offset", Vector3f.UNIT_XYZ);
+        oc.write(offset, "offset", Vector2f.ZERO);
         oc.write(offsetAmount, "offsetAmount", 0);
         //oc.write(lodCalculator, "lodCalculator", null);
         //oc.write(lodCalculatorFactory, "lodCalculatorFactory", null);
@@ -920,7 +920,7 @@ public class TerrainPatch extends Geometry {
         totalSize = ic.readInt("totalSize", 16);
         quadrant = ic.readShort("quadrant", (short)0);
         stepScale = (Vector3f) ic.readSavable("stepScale", Vector3f.UNIT_XYZ);
-        offset = (Vector2f) ic.readSavable("offset", Vector3f.UNIT_XYZ);
+        offset = (Vector2f) ic.readSavable("offset", Vector2f.ZERO);
         offsetAmount = ic.readFloat("offsetAmount", 0);
         //lodCalculator = (LodCalculator) ic.readSavable("lodCalculator", new DistanceLodCalculator());
         //lodCalculator.setTerrainPatch(this);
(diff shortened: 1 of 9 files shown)


jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
```diff
@@ -920,7 +920,7 @@ public void read(JmeImporter im) throws IOException {
         totalSize = ic.readInt("totalSize", 16);
         quadrant = ic.readShort("quadrant", (short)0);
         stepScale = (Vector3f) ic.readSavable("stepScale", Vector3f.UNIT_XYZ);
-        offset = (Vector2f) ic.readSavable("offset", Vector3f.UNIT_XYZ);
+        offset = (Vector2f) ic.readSavable("offset", Vector2f.ZERO);
```
JNightRider, line 923: Won't this case represent a radical change? After all, the value used to be (1, 1, 1), whereas now it is (0, 0, 0)

riccardobl: This was likely a bug, because the writer defaults to ZERO

riccardobl:
@jaime-jmebot review

riccardobl:
@jaime-jmebot review

jme3-terrain/src/main/java/com/jme3/terrain/geomipmap/TerrainPatch.java
jaime-jmebot: Confirming @riccardobl's reasoning from what the code does today.

In `TerrainPatch`, `write()` passes `Vector3f.UNIT_XYZ` as the default for `stepScale` and `Vector2f.ZERO` as the default for `offset`, and `read()` now mirrors both. The default argument exists so that a capsule can skip the field when the value equals it, so the omitted case is, by definition, "offset == writer's default" — under the previous reader default (a 1-filled vector) such assets came back with a wrong offset, which is the bug being fixed here. `Vector2f.ZERO` is also the offset the terrain tree starts from (`TerrainQuad`'s constructors pass `new Vector2f()`), so (0,0) is the intended default rather than a radical change.

The only case that would still be impacted is an asset written by a much older build whose `write()` default was (1,1): that field would have been omitted and would now load as (0,0). I see no sign of such a writer default in the current code, so I'd only double-check the history of that `oc.write(offset, ...)` default before merging — otherwise this line is a fix, not a regression.

jaime-jmebot (commented):
Reviewed the serialization-default change discussed in the inline thread, plus the "copy shared constants into mutable fields" pattern.

- `TerrainPatch.read()` now mirrors `write()`: `stepScale` defaults to `Vector3f.UNIT_XYZ` and `offset` to `Vector2f.ZERO`, matching the defaults passed to `oc.write(...)`. `Vector2f.ZERO` is also the offset the terrain tree starts from (`TerrainQuad` constructors pass `new Vector2f()`), so the reader default is now the correct constant rather than a behaviour change. Details in the inline thread.
- The only asset class that could still be affected is one saved by a build whose `write()` default for `offset` was a 1-filled vector: such a field would have been omitted and would now be read as (0,0). I found no evidence of that in the current code; a quick look at that line's history is enough to close it out.
- For the mutation-type fixes, the pattern used is the right one: any field that is later modified must not hold a shared constant (e.g. `TerrainQuad.lastScale` is initialised from `Vector3f.UNIT_XYZ.clone()` and later updated via `lastScale.set(...)`, so it must be a distinct instance). Passing constants read-only, as is still done in a few places, remains fine and needs no change.

Nothing here requires a code change from my side.

PR #2951, opened 2026-09-04
Android: Allow virtual joystick and touch events at the same time

terranprog:
Improve handling of android touch events when virtual joystick is being used. 
Previously it was not possible to move using the joystick AND rotate the camera using the touch screen at the same time. You could do one or the other at any one time. If you were using the joystick, touch events would be ignored.
This change fixes that so the user can move forward/back/left/right AND rotate the camera at the same time; just like using a mouse and keyboard.
Disclaimer: this is done using Claude code. I don't have the expertise to understand the changes and what they do. This code should be reviewed by a human who understands this part of the engine.
I found it a big improvement to the UX of my android application so I think it should be seriously considered for inclusion.

### DIFF
--- a/jme3-android/src/main/java/com/jme3/input/android/AndroidInputHandler.java
+++ b/jme3-android/src/main/java/com/jme3/input/android/AndroidInputHandler.java
@@ -63,6 +63,11 @@ public class AndroidInputHandler implements View.OnTouchListener,
     protected AndroidTouchInput touchInput;
     protected AndroidJoyInput joyInput;
     protected MouseInput mouseInput;
+    /**
+     * Pointer ids owned by the on-screen virtual joystick for the duration of the MotionEvent
+     * being dispatched - see {@link #isPointerCapturedByJoystick(int)}.
+     */
+    private long joystickPointerMask;
 
     public AndroidInputHandler() {
         touchInput = new AndroidTouchInput(this);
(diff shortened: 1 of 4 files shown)


jaime-jmebot (commented):
Reviewed the head snapshot (`eb5309c`). Since the PR body says the mechanism was written without full understanding of this part of the engine, here is what the change actually does, then the concerns.

## What the change does

* `VirtualJoystick` exposes `getCapturedPointerMask()`: a `long` bitmask of the Android pointer ids it currently owns (stick captures *and* button captures), built from the `captures` map under `inputLock`, with ids ≥ 64 dropped.
* `AndroidJoyInput.getCapturedPointerMask()` / `onTouch()` delegate to that joystick.
* `AndroidInputHandler.onTouch()` now reads the mask **before** and **after** handing the event to the joystick and unions the two (`capturedBefore | getCapturedPointerMask()`). That union is what makes the transition events behave: on `ACTION_DOWN` the pointer is only in the "after" set (it is being captured now), on `ACTION_UP` only in the "before" set (it is being released now). Either way it stays hidden from touch handling.
* `AndroidTouchInput` then skips exactly those pointers in `DOWN`/`MOVE`, counts only *reported* pointers for the one-finger mouse-emulation guard, and drops an `UP` whose pointer it never reported.
* The gesture/scale detectors are skipped whenever the event contains a captured pointer.

Per-pointer filtering instead of dropping the whole `MotionEvent` is the right granularity, and the before/after union is a neat way to cover both ends of a capture. Two things I think are worth fixing, plus two nits.

## Concerns

**1. Gesture detectors are starved mid-gesture, with no `ACTION_CANCEL` (medium).**
`if (numPointers == event.getPointerCount())` withholds *every* event from `GestureDetector`/`ScaleGestureDetector` as soon as a captured pointer exists — including the `UP`/`CANCEL` of a finger the detectors were already tracking — and nothing ever tells them the gesture ended. Concretely: finger 1 goes `DOWN` (detector posts its long-press message and starts tracking), finger 2 then presses the stick; the `MOVE`/`UP` of finger 1 are never delivered, so `mStillDown` stays set and the posted long-press can still fire even though the finger has moved away or lifted. A fling or pinch that started before the stick finger landed can likewise never complete. Suggest delivering a synthetic `ACTION_CANCEL` (and/or a detector reset) to both detectors when the "contains a captured pointer" state changes, so their state cannot outlive the gesture.

**2. Captures dropped mid-gesture produce `MOVE`/`UP` with no matching `DOWN` (edge case).**
`updateVisuals()` calls `releaseAllLocked(...)` on a GUI resize (e.g. orientation change) and `setEnabled(false)` from the auto-visibility path also releases, both while a finger may still be physically down. From that moment `isJoystickPointer()` no longer matches, so the `MOVE` branch creates a `lastPositions` entry for that pointer (first move has delta 0, so nothing is emitted; the next move is) and the app then sees `MOVE…UP` for a pointer that was never reported `DOWN`. The `lastPositions.remove(pointerId) == null` guard only covers the case where no `MOVE` was reported in between. Tracking a small set of "hidden/never reported" pointer ids (cleared on `DOWN` and `UP`) would close this.

**3. Nit:** `AndroidInputHandler.isPointerCapturedByJoystick()` is public but reads a field whose value is only valid while the nested `touchInput.onTouch()` call is on the stack. It works today; it is a lifetime trap for a later caller. Passing the mask explicitly down the call chain (or wrapping it in a dispatch-scoped object) would remove the trap.

**4. Verification.** The capture/release logic is plain Java (capture on `DOWN`, release on `UP`/`CANCEL`/`setEnabled(false)`, ids ≥ 64 ignored) and could be unit-tested headlessly, but I don't see such a test in the checkout, and none of the CI jobs here exercise Android input. Before merge this wants a manual on-device pass: stick held while dragging a second finger (the case this PR fixes), stick released, orientation change / GUI resize while touching, and pinch/fling while the stick is visible.

Given the disclaimer in the PR body, I'd also want someone who works on Android input regularly to confirm behaviour on a real device rather than relying on the code reading.

PR #2962, opened 2026-09-17
Absolute value operations on vector classes

AfonsoDGBatista:
Added abs and absLocal methods on Vector2f, Vector3f and Vector4f classes.


jme3-core/src/main/java/com/jme3/math/Vector3f.java
```diff
@@ -828,6 +828,29 @@ public Vector3f normalizeLocal() {
...
     }
 
+    /**
+     * Computes the absolute value of each component and returns the result as a
+     * new instance. The current instance is unaffected.
+     *
+     * @return a new Vector3f with non-negative components
+     */
+    public Vector3f abs() {
+        return new Vector3f(Math.abs(x), Math.abs(y),Math.abs(z));
```
jaime-jmebot, line 838: Missing space after the comma between `Math.abs(y)` and `Math.abs(z)`. The rest of this file (and the repo's documented style, Google style per CONTRIBUTING.md) uses a single space after argument commas.

```suggestion
        return new Vector3f(Math.abs(x), Math.abs(y), Math.abs(z));
```

jme3-core/src/main/java/com/jme3/math/Vector4f.java
```diff
@@ -782,6 +782,30 @@ public Vector4f normalizeLocal() {
...
     }
 
+    /**
+     * Computes the absolute value of each component and returns the result as a
+     * new instance. The current instance is unaffected.
+     *
+     * @return a new Vector4f with non-negative components
+     */
+    public Vector4f abs() {
+        return new Vector4f(Math.abs(x), Math.abs(y),Math.abs(z),Math.abs(w));
```
jaime-jmebot, line 792: Same formatting issue as in `Vector3f`: missing space after the comma before `Math.abs(z)`.

```suggestion
        return new Vector4f(Math.abs(x), Math.abs(y), Math.abs(z), Math.abs(w));
```

jme3-core/src/main/java/com/jme3/math/Vector3f.java
```diff
@@ -828,6 +828,29 @@ public Vector3f normalizeLocal() {
...
+     * current instance.
+     *
+     * @return the (modified) current instance (for chaining)
+     */
+    public Vector3f absLocal() {
+        x = Math.abs(x);
+        y = Math.abs(y);
+        z = Math.abs(z);
+        return this;
+    }
```
jaime-jmebot, line 852: The six new public methods have no test coverage. `jme3-core/src/test/java/com/jme3/math/Vector3fTest.java` exercises essentially every other public method of `Vector3f` (including `maxLocal`, `minLocal`, `negate`), so `abs()`/`absLocal()` should get matching tests there. Worth asserting the edge cases the existing suite cares about: negative components, `-0.0f` (→ `+0.0f`), and NaN propagation. `Vector2f`/`Vector4f` have no test classes, so covering the `Vector3f` variants is the minimum.

jaime-jmebot (changes requested):
- The added logic is correct: `abs()`/`absLocal()` follow the engine's `...Local` naming, and `Math.abs` maps `-0.0f` to `+0.0f` and propagates NaN consistently with the rest of `com.jme3.math`.
- Two formatting violations (missing space after a comma) in `Vector3f.abs()` and `Vector4f.abs()`; committable one-line fixes are posted inline.
- No tests for the six new methods. `Vector3fTest` is the established home for per-method vector tests and should be extended to cover the abs variants, including `-0.0f` and NaN inputs.
- Per CONTRIBUTING.md, non-trivial additions are expected to start with a forum thread; link the discussion in the PR description if one exists.

JNightRider (commented):
It appears that `java.lang.Math` is being used. Since [`FastMath`](https://github.com/jMonkeyEngine/jmonkeyengine/blob/81dc88be22f9aaee1b84c51483e332dacde85193/jme3-core/src/main/java/com/jme3/math/FastMath.java#L592) has this same method, shouldn't we take advantage of the `FastMath` class instead of the `Math` class?

yaRnMcDonuts:
> Since [`FastMath`](https://github.com/jMonkeyEngine/jmonkeyengine/blob/81dc88be22f9aaee1b84c51483e332dacde85193/jme3-core/src/main/java/com/jme3/math/FastMath.java#L592) has this same method, shouldn't we take advantage of the `FastMath` class instead of the `Math` class?

We could do that for consistency's sake, but otherwise I don't think there's any significant advantages to using one or the other, so I'd be fine with it either way.

jaime-jmebot (changes requested):
Thanks for the update — I re-checked the current head (`68f12f3`).

- ✅ The two comma-spacing issues from the previous round are fixed in `Vector3f.abs()` and `Vector4f.abs()`.
- Narrower surface than before: `Vector2f`, `Vector3f`, and `Vector4f` now expose a consistent `abs()` / `absLocal()` pair, and `absLocal()` correctly returns `this` for chaining.
- ⏳ Still open: tests. `Vector3fTest` covers essentially every other public method of `Vector3f`, so the `abs` pair is the one gap; the existing inline note lists the cases worth asserting.
- On the `FastMath` vs `java.lang.Math` question: `Math.abs` is actually the safer choice here. `FastMath.abs(-0.0f)` returns `-0.0f`, whereas `Math.abs` gives `+0.0f` — and `Vector3f.equals` distinguishes the two, so `Math.abs` matches the "non-negative components" javadoc. Performance-wise the two are equivalent, so no change needed either way.

JNightRider:
> We could do that for consistency's sake

It might be better to maintain consistency.

---
We can apply the changes in another PR to maintain consistency (in the future).

